PR #22, opened 2026-05-21
Fix vim dot-repeat dropping inserted text of change/insert commands (#21)

### LINKED ISSUE #21: Vim dot-repeat (.) drops inserted text for change/insert commands
## Summary
In vim mode, the dot-repeat command `.` does not faithfully repeat the *last edit* when that edit entered insert mode (`c`/`cw`/`s`/`i`/`a`/`o`/…). It replays the operator/delete portion but **drops the inserted text**. Single-change commands (`x`, `dd`) repeat correctly.

## Repro (observed in a real editor; vim keymap)
| Keystrokes | Expected (vim) | Actual |
|---|---|---|
| `x` then `.` | deletes a char | ✅ deletes a char |
| `dd` then `.` | deletes a line | ✅ deletes a line |
| `cw` `foo` `<Esc>` then `.` | replaces next word with `foo` | ❌ deletes the word, **does not insert `foo`** |
| `i` `bar` `<Esc>` then `.` | inserts `bar` again | ❌ moves the cursor, **inserts nothing** |

## Root-cause clue
After `cw` `foo` `<Esc>`, it takes **two undos** to revert (one for the delete, one for the insert). In vim, a change command plus its insert session is **one atomic unit** for both `.` and `u`.

So the engine appears to record the operator/delete and the subsequent insert-mode session as **separate changes**, and:
- `.` replays only the first (the operator/delete), losing the inserted text;
- `u` sees two steps instead of one.

The fix is to capture the insert-mode session text into the "last edit" record and group it with the operator into a single repeatable + undoable unit (matching vim/vim.js, where leaving insert mode appends the inserted input to the last-edit record).

## Affected commands
Anything that enters insert mode as part of an edit: `c`, `cc`, `cw`, `C`, `s`, `S`, `i`, `I`, `a`, `A`, `o`, `O`.

## Likely locations (line numbers approximate, vim module)
- `vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/Vim.kt` — `recordLastEdit` (~1053) and the repeat-last-edit helper; where insert-mode input should be accumulated/appended.
- `vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/VimActions.kt` — `repeatLastEdit` (~710).
- `vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/VimKeymap.kt` — `.` binding (~717).

## Note on existing tests
`vim/src/jvmTest/.../VimMiscTest.kt` has dot-repeat tests (`dot_insert`, `dot_insert_cw`, `dot_insert_repeat`, …) that reportedly **pass**, yet the bug is real in practice. So either those tests don't faithfully exercise the insert-text replay (assert buggy output, or inject insert-mode input via a path the real editor doesn't use), or the jvmTest harness feeds insert keystrokes differently than the real CodeMirror transaction/input path so the bug only manifests with real input. Worth verifying the tests assert correct vim semantics (word replaced in *both* spots, single undo unit).

## Environment
Surfaced via konstructor (kodemirror `0.2.0`, wasmJs target). The vim engine is the pure-Kotlin reimplementation in `commonMain` (not `@replit/codemirror-vim`).

monkopedia-coder:
## Summary
Fixes #21. Vim dot-repeat (`.`) dropped the inserted text of change/insert commands (`c`, `cw`, `s`, `i`, `a`, `o`, …): it replayed only the operator/delete and lost the typed text.

## Root cause
The insert-mode `onChange` recorder (which appends typed text to `lastInsertModeChanges`) was **never registered** on the `change` event. `enterVimMode`/`maybeInitVimState` wired the `cursorActivity` handler but not `change`, so `lastInsertModeChanges` stayed empty and `repeatInsert` had nothing to replay.

## Fix
- `Vim.kt`: register the change recorder for the lifetime of vim mode (mirroring the `cursorActivity` handler) via a `changeHandlers` map + `registerChangeHandler`, wired in both `enterVimMode` (production) and `maybeInitVimState` (init/test path); unregistered in `leaveVimMode`.
- `VimHelpers.kt`: guard the recorder to insert mode only, so always-on registration does not record normal-mode edits (`p`, `x`, `dd`) and double-apply them on `.`.
- `VimTestHarness.kt`: the harness previously called the recorder directly, which **masked** the missing wiring (dot-insert tests passed despite the real bug). It now records via the registered `change` handlers, exactly like the editor operation pipeline.
- `CHANGELOG.md`: Unreleased → Fixed entry (#21).

## Verification
- `:vim:jvmTest` green; `:vim:apiCheck` passes (changes are `internal`, no public API delta); spotless clean.
- **Faithfulness proven**: temporarily disabling the wiring made all 8 dot-insert tests (incl. the new `dot_insert_cw_issue21`) fail, confirming the harness no longer masks the bug.

## Not included
The issue also noted `cw` taking two undos (undo-chain atomicity). That is a separate history-grouping concern, tracked separately — out of scope here.

## Test plan
- [ ] `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :vim:jvmTest`
- [ ] Manual: `cw foo<Esc>` then `.` replaces the next word with `foo`; `i bar<Esc>` then `.` re-inserts `bar`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to Kodemirror will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [Unreleased]
+
+### Fixed
+- Vim dot-repeat (`.`) dropped the inserted text of change/insert commands (`c`, `cw`, `s`, `i`, `a`, `o`, …) — the insert-mode change event was never wired up, so `lastInsertModeChanges` stayed empty and `.` replayed only the operator/delete (#21)
+
 ## [0.2.0] - 2026-04-10
 
 ### Highlights
--- a/vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/Vim.kt
+++ b/vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/Vim.kt
@@ -244,12 +244,14 @@ object Vim : VimApiInterface {
     }
 
     internal val cursorActivityHandlers = mutableMapOf<VimEditor, (Array<out Any?>) -> Unit>()
+    internal val changeHandlers = mutableMapOf<VimEditor, (Array<out Any?>) -> Unit>()
 
     fun enterVimMode(cm: VimEditor) {
         cm.signal("vim-mode-change", mapOf("mode" to "normal"))
         val handler: (Array<out Any?>) -> Unit = { onCursorActivity(cm) }
         cursorActivityHandlers[cm] = handler
         cm.events.on("cursorActivity", handler)
+        registerChangeHandler(cm)
         maybeInitVimState(cm)
     }
 
@@ -257,7 +259,23 @@ object Vim : VimApiInterface {
         cursorActivityHandlers.remove(cm)?.let { handler ->
             cm.events.off("cursorActivity", handler)
         }
+        changeHandlers.remove(cm)?.let { handler ->
+            cm.events.off("change", handler)
+        }
         cm.vim = null
+    }
+
+    // Records insert-mode input into lastInsertModeChanges so that dot-repeat
+    // (`.`) and macros can replay the text typed during a change/insert command.
+    // The recorder itself only acts while in insert mode (see onChange), so the
+    // handler can stay registered for the lifetime of vim mode.
+    internal fun registerChangeHandler(cm: VimEditor) {
+        if (changeHandlers.containsKey(cm)) return
+        val handler: (Array<out Any?>) -> Unit = { args ->
+            onChange(cm, args.getOrNull(1) as? Change)
+        }
+        changeHandlers[cm] = handler
+        cm.events.on("change", handler)
     }
 
     fun exitVisualMode(cm: VimEditor, moveHead: Boolean = true) {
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Approved (tier 1, bug_fix per kodemirror policy).

Root cause and fix verified against the diff: the insert-mode `onChange` recorder was never registered on the `change` event, so `lastInsertModeChanges` stayed empty and `.` replayed only the operator/delete. The fix registers the recorder for the lifetime of vim mode (`enterVimMode` + `maybeInitVimState`, torn down in `leaveVimMode`) and correctly pairs always-on registration with an insert-mode guard in `onChange` so normal-mode edits (p/x/dd) aren't double-applied on dot-repeat. The harness fix (routing through registered `change` handlers instead of calling `onChange` directly) removes a genuine test blind spot that had masked the bug.

All changes are `internal` — no public API delta.

CI verified locally (kodemirror CI-must-be-green policy):
- :vim:jvmTest — BUILD SUCCESSFUL (forced rerun; VimMiscTest 46 tests, 0 failures, new dot_insert_cw_issue21 included)
- :vim:apiCheck — passes (no public API change)

PR #24, opened 2026-05-21
Fix vim change/insert command not being a single undo unit (#23)

### LINKED ISSUE #23: Vim: change/insert command (cw, c, s, i, a, o) is not a single undo unit
## Summary
Split out from #21. In vim, a change/insert command and the text typed during its insert-mode session should be **one atomic undo unit** (`u` reverts the whole thing in one step), matching real vim. Reported behavior: `cw foo<Esc>` then `u` takes **two** undos — one for the inserted `foo`, one for the operator/delete.

This is a *separate mechanism* from the dot-repeat fix in #21. #21 fixed `.` replaying the inserted text (insert-mode change recording). Undo grouping is governed by the `history()` extension's change-grouping, not by `lastInsertModeChanges`, so it was intentionally left out of the #21 PR (#22).

## Repro (vim keymap)
1. Start with `word1 word2`, cursor at col 0.
2. `c` `w` → deletes `word1`, enters insert mode.
3. type `foo`, `<Esc>` → buffer is `foo word2`.
4. `u` → **expected:** one undo restores `word1 word2`. **reported actual:** first `u` removes `foo` (→ ` word2`), second `u` restores `word1`.

Affected: any insert-entering edit — `c`, `cc`, `cw`, `C`, `s`, `S`, `i`, `I`, `a`, `A`, `o`, `O`.

## Preliminary investigation (inconclusive — needs empirical verification)
- The vim operator delete is dispatched via `VimEditor.dispatchChange` → `session.dispatch(spec)` with **no `userEvent`** (null) — `vim/.../VimEditor.kt:718`, `replaceRange` at `:747`.
- Insert-mode characters arrive from the view input with `userEvent = "input.type"` (`view/.../InputHandling.kt:478`, `KodeMirror.kt:530,584`).
- History groups adjacent changes into one undo event when the incoming `userEvent` is null or matches `joinableUserEvent = ^(input.type|delete)`, within `groupDelay`, and the ranges are adjacent — `commands/.../History.kt:46, 250-292`.
- By that logic the null-userEvent delete and the `input.type` insert *look* joinable and adjacent, so static analysis does **not** clearly explain a two-undo split. Possible culprits to trace:
  - the selection-only transaction from `selectForInsert` (between the delete and the first insert) resetting `prevTime`/`prevUserEvent` and breaking the group;
  - the adjacency computation across the delete→insert position mapping;
  - `groupDelay` boundary / per-keystroke timing.

## Next steps
1. Write a JVM regression test driving `cw foo<Esc>` through the real change pipeline (the #21 harness change now records insert text faithfully) and assert a single `u` fully reverts.
2. If it reproduces, decide the fix: most likely have vim group the operator delete + insert session into one history event (e.g. an `isolateHistory`/grouping annotation, or matching vim.js's single-operation batching). May be mechanical or may need a small design call on how vim annotates its transactions.

## Acceptance
`u` after any change/insert command reverts the command **and** its inserted text in a single step.

monkopedia-coder:
## Summary
Fixes #23. A vim change/insert command and the text typed in its insert session (`c`, `cw`, `s`, `i`, `a`, `o`, …) should be a single undo unit, but `u` took two steps — one for the inserted text, one for the operator/delete.

## Root cause
Entering insert mode dispatches a cursor-positioning selection (`selectForInsert` → `setSelections`). That selection-only transaction was recorded in history, stamping `selectionsAfter` on the preceding operator-delete event. History's `addChanges` only groups a new change with the previous event when `lastEvent.selectionsAfter.isEmpty()` (`commands/.../History.kt:267`), so the inserted text started a *new* undo event instead of joining the delete.

## Fix
Mark the insert-entry cursor selection `addToHistory = false` so it no longer forms a history boundary; the inserted text then groups with the operator-delete into one undo unit. `setSelections` gains an optional `addToHistory` flag (default `true`, so all other callers are unaffected). Internal-only — no public API change.

## Verification
- Full `:vim:jvmTest` green (no undo regressions across the suite); `:vim:apiCheck` passes; spotless clean.
- New regression test `cw_insert_is_single_undo_unit_issue23`: `cw test<Esc>` then a single `u` restores `word1 word2`. This is the same assertion that failed before the fix (confirmed via a throwaway probe on the issue), so it genuinely guards the behavior.

## Test plan
- [ ] `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :vim:jvmTest`
- [ ] Manual: `cw foo<Esc>` then `u` restores the original word in one step (and `.` still works, from #21)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - Vim dot-repeat (`.`) dropped the inserted text of change/insert commands (`c`, `cw`, `s`, `i`, `a`, `o`, …) — the insert-mode change event was never wired up, so `lastInsertModeChanges` stayed empty and `.` replayed only the operator/delete (#21)
+- Vim change/insert commands (`c`, `cw`, `s`, …) were not a single undo unit — `u` took two steps to revert because the insert-entry cursor selection stamped a history boundary on the operator-delete event (#23)
 
 ## [0.2.0] - 2026-04-10
 
--- a/vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/VimEditor.kt
+++ b/vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/VimEditor.kt
@@ -42,6 +42,7 @@ import com.monkopedia.kodemirror.state.LineNumber
 import com.monkopedia.kodemirror.state.MapMode
 import com.monkopedia.kodemirror.state.SelectionSpec
 import com.monkopedia.kodemirror.state.Text
+import com.monkopedia.kodemirror.state.Transaction
 import com.monkopedia.kodemirror.state.TransactionSpec
 import com.monkopedia.kodemirror.state.asInsert
 import com.monkopedia.kodemirror.state.endPos
@@ -792,7 +793,11 @@ internal fun VimEditor.setCursor(line: Int, ch: Int = 0) {
 
 internal fun VimEditor.setCursor(pos: LinePos) = setCursor(pos.line, pos.ch)
 
-internal fun VimEditor.setSelections(selections: List<LinePosRange>, primIndex: Int? = null) {
+internal fun VimEditor.setSelections(
+    selections: List<LinePosRange>,
+    primIndex: Int? = null,
+    addToHistory: Boolean = true
+) {
     val doc = session.state.doc
     val ranges = selections.map { x ->
         val head = indexFromPos(doc, x.head)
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
Approving — tier 1 (bug_fix) per kodemirror review policy.

Reviewed the undo/history-semantics change carefully since it touches history grouping:
- Only the insert-entry path (selectForInsert) opts into addToHistory=false. The other four setSelections call sites keep the new param's default (true), so they're behaviorally unchanged.
- Transaction.addToHistory.of(false) is the correct idiomatic mechanism for a non-history selection transaction; the root-cause analysis (selectionsAfter stamping a history boundary on the operator-delete event) is accurate.
- setSelections is internal and apiCheck passes — no public API change.

Verification (local, branch fix/vim-undo-atomicity-23):
- Full :vim:jvmTest green: 621 tests, 0 failures/errors/skipped, including the new regression cw_insert_is_single_undo_unit_issue23 — no undo regressions elsewhere in the suite.
- :vim:apiCheck passes; spotless clean.

Issue #36, opened 2026-05-22
[lsp-client] Document sync + position/workspace mapping


Monkopedia:
Part of #26. Port upstream document sync (`didOpen`/`didChange`/`didClose`, version tracking) and the `WorkspaceMapping` + `toPosition`/`fromPosition` machinery (DocPos<->LSP Position, UTF-16). **Match upstream's sync behavior** (incremental content changes derived from the ChangeSet, honoring the server's `textDocumentSync` capability) — do not substitute a simpler full-document resend unless there's a Kotlin-specific reason.

Issue #26, opened 2026-05-22
[epic] Port @codemirror/lsp-client: full LSP support for kodemirror


Monkopedia:
Epic: a **complete, faithful port** of upstream [`@codemirror/lsp-client`](https://github.com/codemirror/lsp-client) into a new `:lsp-client` module. Not an MVP — match upstream behavior/fidelity, deviating only for documented Kotlin/Compose reasons.

## Design (settled)
- Types + suspend `LanguageServer`/`LanguageClient` interfaces come from **`com.monkopedia.lsp:lsp:1.0.0-RC2`** (full kotlinx-serialization matrix incl. wasmJs); the `@KsService` relay (`:lsp-ksrpc:1.0.0-RC2`) covers wasmJs.
- **Consumer provides the `LanguageServer`** ("allow that server to be provided"); kodemirror does NOT implement JSON-RPC framing.
- **Cancellation rides on coroutine cancellation** (ksrpc maps it; no upstream-style `$/cancelRequest` plumbing needed).
- Web transport is lsp-kotlin/ksrpc's concern (ksrpc relay over a channel), not kodemirror's.
- Editor integration reuses existing primitives: `hoverTooltip`/`Tooltip`, decorations + gutter (`:lint`), `CompletionSource` (`:autocomplete`), `Panel`, `StateField`/`StateEffect`/`Facet`, `ViewPlugin`, session `coroutineScope`.
- **Known necessary deviation:** upstream renders hover/docs as HTML (`docToHTML`); the Compose/canvas view has no DOM, so we render markdown→Compose (see #39).

## Sub-issues (each one PR through the automated review pipeline)
Foundation (in order):
- [ ] #35 — module scaffold + client/plugin/workspace core + entry point
- [ ] #36 — document sync + position/workspace mapping (match upstream sync)

Features (build on the foundation):
- [ ] #37 — diagnostics (`serverDiagnostics`)
- [ ] #38 — completion (`serverCompletion`, incl. snippets/textEdit)
- [ ] #39 — hover (`hoverTooltips`; markdown→Compose deviation)
- [ ] #40 — signature help
- [ ] #41 — go-to-definition family
- [ ] #42 — find references + reference panel
- [ ] #43 — rename
- [ ] #44 — format document

Tooling:
- [ ] #45 — showcase sample + test language server + tests

## Acceptance
[...]

PR #51, opened 2026-05-22
LSP hover (#39)

### LINKED ISSUE #39: [lsp-client] Hover
Part of #26. Port `hoverTooltips` -> kodemirror `hoverTooltip`/`Tooltip`. Upstream renders `MarkupContent` as HTML via `docToHTML`; **Kotlin/Compose deviation**: there is no DOM, so render markdown to Compose content (or plain text) instead — document this as the one necessary deviation.

monkopedia-coder:
## Summary
Fifth sub-issue of the LSP port epic #26. Closes #39. Ports upstream `hoverTooltips` — `textDocument/hover` rendered as a Compose tooltip. Builds on #36 mapping.

## Touches two modules (public API → tier-2)
**`:view`** (additive, regression-checked):
- `const val DEFAULT_HOVER_TIME: Long = 300`
- `fun hoverTooltip(hoverTime: Long, source: suspend (EditorSession, Int) -> Tooltip?): Extension` — async counterpart to the existing synchronous `hoverTooltip(source)`. The sync overload (used by `:lint`) is unchanged; `HoverTooltipPlugin` now branches sync-vs-async. The async overload requires an explicit `hoverTime` to avoid trailing-lambda overload ambiguity with the sync one. Debounced by `hoverTime`; in-flight request cancelled on move, with a position re-check so a slow response can't pop a stale tooltip. This is the Kotlin/coroutine adaptation of CM6's promise-based hover.

**`:lsp-client`**:
- `fun serverHover(client: LSPClient, uri: String, hoverTime: Long = DEFAULT_HOVER_TIME): Extension` — folded into `languageServerSupport`. Sends `textDocument/hover` (via `toPosition`, after `client.sync()`); honors `Hover.range` for placement and remaps the response range through a `WorkspaceMapping` (released in `finally`); falls back to pointer pos when no range.

## Rendering deviation (documented)
Upstream renders `MarkupContent` to **HTML** via `docToHTML`; the Compose canvas has no DOM. `parseHoverContents` handles the full `contents` union (plain string, `MarkupContent{kind,value}`, `MarkedString` string/`{language,value}`, arrays) → ordered `HoverBlock`s; `hoverBlockToAnnotatedString` does a small dependency-free **markdown→AnnotatedString** conversion (bold/italic/inline code/headings/fenced code; MarkedString language blocks → monospace; unsupported constructs like links/lists/tables fall through as literal text). Documented as the necessary HTML→Compose deviation.

## Verification
- `:lsp-client:jvmTest` (16 union-parse + markdown-convert tests), `compileKotlinJvm`/`compileKotlinWasmJs`/`compileDebugKotlinAndroid`, `:lsp-client:apiCheck` — green.
- **`:view:jvmTest` + `:view:apiCheck` green** (re-verified — the sync hover path that `:lint` uses is unchanged). spotless/ktlint applied to both modules; apiDump committed for both.
- Apple/native not cross-compiled on Linux (commonMain-only).

## Test plan
- [ ] `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :lsp-client:jvmTest :view:jvmTest :lsp-client:compileKotlinWasmJs :lsp-client:compileDebugKotlinAndroid :lsp-client:apiCheck :view:apiCheck -Dorg.gradle.jvmargs="-Xmx6g"`

### DIFF
--- a/lsp-client/api/android/lsp-client.api
+++ b/lsp-client/api/android/lsp-client.api
@@ -117,6 +117,11 @@ public final class com/monkopedia/kodemirror/lsp/ServerDiagnosticsKt {
 	public static final fun serverDiagnostics ()Lcom/monkopedia/kodemirror/state/Extension;
 }
 
+public final class com/monkopedia/kodemirror/lsp/ServerHoverKt {
+	public static final fun serverHover (Lcom/monkopedia/kodemirror/lsp/LSPClient;Ljava/lang/String;J)Lcom/monkopedia/kodemirror/state/Extension;
+	public static synthetic fun serverHover$default (Lcom/monkopedia/kodemirror/lsp/LSPClient;Ljava/lang/String;JILjava/lang/Object;)Lcom/monkopedia/kodemirror/state/Extension;
+}
+
 public class com/monkopedia/kodemirror/lsp/Workspace {
 	public static final field $stable I
 	public fun <init> (Lcom/monkopedia/kodemirror/lsp/LSPClient;)V
--- a/lsp-client/api/jvm/lsp-client.api
+++ b/lsp-client/api/jvm/lsp-client.api
@@ -117,6 +117,11 @@ public final class com/monkopedia/kodemirror/lsp/ServerDiagnosticsKt {
 	public static final fun serverDiagnostics ()Lcom/monkopedia/kodemirror/state/Extension;
 }
 
+public final class com/monkopedia/kodemirror/lsp/ServerHoverKt {
+	public static final fun serverHover (Lcom/monkopedia/kodemirror/lsp/LSPClient;Ljava/lang/String;J)Lcom/monkopedia/kodemirror/state/Extension;
+	public static synthetic fun serverHover$default (Lcom/monkopedia/kodemirror/lsp/LSPClient;Ljava/lang/String;JILjava/lang/Object;)Lcom/monkopedia/kodemirror/state/Extension;
+}
+
 public class com/monkopedia/kodemirror/lsp/Workspace {
 	public static final field $stable I
 	public fun <init> (Lcom/monkopedia/kodemirror/lsp/LSPClient;)V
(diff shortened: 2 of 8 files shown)


monkopedia-reviewer (approved):
Reviewed the full diff at tier-1 (merge) standard — approving.

**Classification:** public_api (tier 2 per kodemirror policy), covered by the user's standing approval for the LSP-client port; merging directly.

**:view change is routine-additive (verified):**
- New `const val DEFAULT_HOVER_TIME = 300` and an async `hoverTooltip(hoverTime, suspend source)` overload; explicit `hoverTime` param correctly disambiguates the trailing lambda from the existing sync overload.
- `HoverTooltipPlugin` now branches sync-vs-async. The synchronous `hoverTooltip(source)` path that `:lint` uses is behaviorally preserved: when `asyncSource == null` it computes `syncSource(session, docPos)` exactly as before. The one delta on the sync path is a `docPos == lastPos` short-circuit, which only skips recomputing a position-pure source for an unchanged position — an optimization, not a semantic change.
- Async path is sound: clears immediately, debounces via `delay(hoverTime)`, cancels the in-flight `pending` job on move, and re-checks `lastPos == docPos` before publishing so a slow response can't pop a stale tooltip. New `SupervisorJob` is cancelled in `destroy()` (a real `PluginValue` hook invoked on view dispose).

**:lsp-client `ServerHover.kt`:** `parseHoverContents` handles the full contents union (string / MarkupContent / MarkedString string|{language,value} / arrays), dropping blank/null blocks; bare strings inside a `MarkedString[]` correctly treated as markdown per spec. `hoverBlockToAnnotatedString` is the documented dependency-free markdown→Compose conversion (HTML→Compose deviation, clearly noted). `serverHover` checks `hoverProvider`, calls `client.sync()` first, captures a `WorkspaceMapping` and releases it in `finally`, honors `Hover.range` with remap and falls back to pointer pos. (CancellationException catch+rethrow is a redundant no-op but harmless/intentional.)

**Build verification (branch a624539, fresh `--rerun-tasks`):** `:lsp-client:jvmTest` (61 tests, incl. 17 new ServerHoverTest) green; **`:view:jvmTest` ran fresh — full suite, 0 failures / 0 errors / 0 skipped** (core-module regression check passed); `:lsp-client:compileKotlinWasmJs`, `:lsp-client:compileDebugKotlinAndroid`, `:lsp-client:apiCheck`, `:view:apiCheck` all green. Apple/native not cross-compiled on Linux (commonMain-only). BUILD SUCCESSFUL.

PR #59, opened 2026-05-23
Fix editor swallowing browser-reserved keyboard shortcuts on wasmJs (#49)

### LINKED ISSUE #49: Editor captures browser keyboard shortcuts — Ctrl+Tab and other browser-reserved keys don't reach the browser
## Summary
In the Compose/wasmJs editor, browser-level keyboard shortcuts are swallowed by the editor instead of reaching the browser. Reported concretely: **you can't switch Chrome tabs** while the editor has focus (Ctrl+Tab / Ctrl+Shift+Tab / Ctrl+PgUp/PgDn, and likely Ctrl+1–9). Other browser-reserved combos (Ctrl+W close tab, Ctrl+T new tab, Ctrl+L address bar) are likely affected too — the editor appears to `preventDefault` keydown too broadly.

## Expected
The editor should only consume keys that map to an editor command/binding, and let unhandled key events — especially browser-reserved modifier combos — propagate so the browser handles them. This is CodeMirror's normal behavior in a DOM contenteditable; the Compose/wasm key-input path needs to preserve that scoping.

## Likely cause
The Compose/wasm key handler that feeds keystrokes into the editor is capturing/`preventDefault`-ing a broad range of keydowns (or all of them) rather than only the ones the editor actually binds. So browser-level shortcuts never bubble to Chrome. Same "Compose view doesn't bridge/scope X to the platform" family as #33 (height/scroll) and #34 (system clipboard).

## konstructor context (consumer)
konstructor just embeds `KodeMirror` with `basicSetup` + theme/font/lang/keymap/lint extensions — it does **not** install any global key interception or `preventDefault` of its own. So key-event scoping is entirely the editor view's responsibility.

## Repro
Focus the editor, press Ctrl+Tab (or Ctrl+Shift+Tab / Ctrl+PgDn) — the browser does not switch tabs; the keystroke is consumed by the editor.

## Environment

## Note
Exact set of leaked shortcuts not yet enumerated — confirmed tab-switching; worth auditing which modifier combos the key handler currently consumes vs. which it should let pass (heuristic: only consume combos that match a registered editor binding).

monkopedia-coder:
## Summary
Fixes #49. On wasmJs the editor swallowed browser-reserved shortcuts (Ctrl+Tab/Shift+Tab, Ctrl+PgUp/PgDn, Ctrl+1–9, Ctrl+W/T/L) so the browser never saw them — e.g. you couldn't switch Chrome tabs while the editor was focused.

## Root cause + fix
`view/.../Platform.wasmJs.kt`'s document keydown listener had **two** `preventDefault()` decisions: an **unconditional** one firing on every modified/special key before anything handled it, plus the correct callback-gated one. Removed the unconditional block (and the now-unused `special` key array), leaving only the callback-gated `preventDefault()`. The callback (`handleRawKeyEvent`) already returns `false` for any key with no matching keymap binding, so unbound modified keys now fall through to the browser while bound keys (e.g. a registered Ctrl+S) are still consumed. Matches the JVM model (Compose consumes only what it handles).

## Verification
- `:view:jvmTest` green, incl. a new `RawKeyEventHandlingTest` (6 cases: bound Ctrl+S consumed + runs; unbound Ctrl+Tab/Ctrl+W/Ctrl+1/Ctrl+PgUp fall through; modifier-only not consumed).
- `:view:compileKotlinWasmJs` green (fix is wasmJs-specific); `:view:apiCheck` green — **no public API change**.
- The raw browser keydown path isn't unit-testable headless; tests cover the `handleRawKeyEvent` decision logic it gates on.
- spotless/ktlint applied. CHANGELOG updated.

## Test plan
- [ ] `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :view:jvmTest :view:compileKotlinWasmJs :view:apiCheck -Dorg.gradle.jvmargs="-Xmx6g"`
- [ ] Manual (browser): focus the editor, confirm Ctrl+Tab switches tabs, Ctrl+W/T/L work, while editor key bindings still function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Fixed
+- wasmJs editor no longer swallows browser-reserved keyboard shortcuts (Ctrl+Tab, Ctrl+Shift+Tab, Ctrl+PgUp/PgDn, Ctrl+1–9, Ctrl+W/T/L, …) — the document keydown listener now calls `preventDefault()` only when the editor's keymap actually handled the key, so unbound modified/special keys propagate to the browser (#49)
 - Vim dot-repeat (`.`) dropped the inserted text of change/insert commands (`c`, `cw`, `s`, `i`, `a`, `o`, …) — the insert-mode change event was never wired up, so `lastInsertModeChanges` stayed empty and `.` replayed only the operator/delete (#21)
 - Vim change/insert commands (`c`, `cw`, `s`, …) were not a single undo unit — `u` took two steps to revert because the insert-entry cursor selection stamped a history boundary on the operator-delete event (#23)
 
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/RawKeyEventHandlingTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/RawKeyEventHandlingTest.kt
@@ -1,0 +1,147 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.view
+
+import com.monkopedia.kodemirror.state.EditorState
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed at tier-1 merge standard — sound.

The fix removes the unconditional preventDefault() (and the now-unused special-key array) from the wasmJs document keydown listener, leaving only the callback-gated preventDefault(). Verified the resulting behavior in handleRawKeyEvent (InputHandling.kt):
- Bound keys (including special keys with keymap bindings: Tab/Backspace/Enter/Esc/arrows, and modified keys like a registered Ctrl+S) still match the keymap, return true, and get preventDefault()'d — no regression of editor handling.
- Unbound modified/special keys (Ctrl+Tab, Ctrl+W, Ctrl+1-9, Ctrl+PgUp/PgDn, ...) have no binding and aren't insertable text, so handleRawKeyEvent returns false → no preventDefault() → they fall through to the browser. This is the #49 fix and matches the JVM/Compose 'consume only what you handle' model.

Build verification in an isolated worktree (JAVA_HOME=java-21, ANDROID_HOME set): :view:jvmTest, :view:compileKotlinWasmJs, :view:apiCheck all BUILD SUCCESSFUL. apiCheck green confirms no public API change. The new RawKeyEventHandlingTest ran 6/6 green and meaningfully exercises the gating decision (bound consumed, unbound Ctrl+Tab/Ctrl+W/Ctrl+1/Ctrl+PgUp fall through, modifier-only not consumed).

Tier-1 bug fix, auto-approved and merging.

Issue #16, opened 2026-04-10
Clipboard paste broken after platform handler refactor


Monkopedia:
## Problem

All 5 clipboard gap tests fail — paste doesn't insert text into the editor. Likely regression from #6 (platform handler refactor to per-instance tokens).

## Failing tests

- clipboard.spec.ts: copy and paste single-line text
- clipboard.spec.ts: paste does not duplicate text
- clipboard.spec.ts: cut removes text and paste restores it
- clipboard.spec.ts: paste multi-line text preserves newlines
- clipboard.spec.ts: paste replaces selection

## Likely cause

The `platformRegisterPasteHandler` was refactored in #6 to support multiple instances. The paste handler registration or dispatch may be broken.

Monkopedia:
Fixed: removed stale Ctrl+V early return in installKeyCapture() that blocked paste from reaching the Kotlin keymap handler.

Monkopedia:
Still failing — the Ctrl+V early return removal was necessary but not sufficient. The clipboard tests use Playwright's browser clipboard, but KM's clipboardPaste reads from Compose's LocalClipboardManager which may not bridge to the browser clipboard in wasmJs. Needs deeper investigation into the clipboard pipeline.

Monkopedia:
Fixed: added internal clipboard buffer to EditorSessionImpl. Copy/cut write to both internal buffer and system clipboard. Paste reads system clipboard first, falls back to internal buffer. Fixes the async browser clipboard API issue on wasmJs.

PR #60, opened 2026-05-23
Bridge copy/cut/paste to the system clipboard on wasmJs (#34)

### LINKED ISSUE #34: Copy (and likely cut/paste) doesn't work — clipboard not bridged in the Compose/wasm view
## Summary
In the Compose/wasmJs editor, **copying selected text to the system clipboard does not work** — with any keymap (default, vim `y`, emacs). The copy command appears to fire but nothing lands on the clipboard, so you can't paste editor content into another app.

## Likely root cause
The editor renders to a WebGL canvas (Compose Multiplatform), so there is **no DOM contenteditable/textarea** backing the selection — the selection is "virtual" (drawn). The browser's native `Ctrl+C` therefore has nothing real to copy. Standard CM6 clipboard handling relies on a hidden DOM element receiving the `copy`/`cut`/`paste` events; in a pure-canvas render that path doesn't exist. So the view itself has to bridge copy/cut/paste to the platform clipboard (Compose `LocalClipboardManager`, or directly `navigator.clipboard` on wasm).

"No matter the keymap" rules out a missing keybinding — the copy *action* runs, but the action isn't writing to the platform/browser clipboard.

## Question for the view layer
Does the `KodeMirror` Compose view route the editor's copy/cut commands to the platform clipboard, and paste from it? If not, that's the gap. Note the browser Clipboard API is async (`navigator.clipboard.writeText` returns a Promise) and gesture-gated, so this may need wasm-specific wiring — and could be partly bounded by Compose Multiplatform's wasm clipboard support.

## Consumer context (konstructor)
konstructor does **not** touch the clipboard at all — verified no clipboard/copy/execCommand/navigator.clipboard usage in the frontend. The editor is configured with `basicSetup` + theme/font/lang/keymap/lint extensions only. So copy/cut/paste behavior is entirely the editor view's responsibility; there's nothing on the konstructor side intercepting or blocking it.

## Scope
Reported for copy-out specifically; cut and paste almost certainly share the same unbridged path — worth verifying all three.

## Environment

monkopedia-coder:
## Summary
Fixes #34. Copy/cut/paste didn't reach the system clipboard on the Compose/wasm target (canvas render = no DOM `contenteditable`; the commands only wrote Compose's `LocalClipboardManager`, which doesn't reliably reach `navigator.clipboard`, and the #16 internal buffer never touched the real system clipboard).

## Fix (no public API change — internal `expect`/`actual` hooks)
- `Platform.kt`: new `internal expect fun platformWriteClipboard(text): Boolean` / `platformReadClipboard(onResult): Boolean`.
- **wasmJs** (`Platform.wasmJs.kt`): `writeClipboard` calls `navigator.clipboard.writeText` synchronously inside the originating key-event gesture (satisfies the browser transient-activation requirement), with a hidden-`textarea` + `execCommand('copy')` fallback for insecure contexts; `platformReadClipboard` runs `navigator.clipboard.readText().then(...)`.
- **JVM/native**: return `false` (no-op) → command layer keeps using Compose `ClipboardManager` (AWT system clipboard, synchronous) — **desktop behavior unchanged**.
- `ClipboardCommands.kt`: extracted `writeClipboard(impl, text)` (copy + cut) — updates the internal buffer, calls `platformWriteClipboard`, falls back to `ClipboardManager.setText` only when the platform didn't handle it. Paste fires an async `platformReadClipboard` to refresh the internal buffer for the *next* paste while using the best synchronously-available value now. The #16 internal buffer remains the immediate, always-reliable fallback. Wired to Ctrl/Meta-c/x/v (vim `y`/emacs route through the same commands).

## Verification
`:view:jvmTest` green incl. new `ClipboardCommandsTest` (8 tests: copy/cut/paste, empty-selection no-op, selection replacement, internal-buffer fallback, no-text→false, copy→paste round-trip). `:view:compileKotlinWasmJs` + `:view:apiCheck` green — no public API change. spotless/ktlint applied; CHANGELOG updated.

## Known limitation (inherent to the browser clipboard API)
The browser `readText()` is async, so the **first** external paste in a session may use the internal buffer until the promise resolves; subsequent pastes reflect the system clipboard. The actual `navigator.clipboard` behavior (permission prompts, cross-app paste, the `execCommand` fallback) can only be verified in a real browser — the JVM tests cover the shared command logic + buffer fallback.

## Test plan
- [ ] Manual (browser): copy in editor → paste into another app, and vice-versa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 - wasmJs editor no longer swallows browser-reserved keyboard shortcuts (Ctrl+Tab, Ctrl+Shift+Tab, Ctrl+PgUp/PgDn, Ctrl+1–9, Ctrl+W/T/L, …) — the document keydown listener now calls `preventDefault()` only when the editor's keymap actually handled the key, so unbound modified/special keys propagate to the browser (#49)
+- Copy/cut/paste did not reach the system clipboard on the Compose/wasm target (vim `y`, emacs, and default keymaps) — the canvas render has no DOM `contenteditable`, so the copy/cut commands now bridge to `navigator.clipboard.writeText` synchronously within the originating key gesture (with a hidden-textarea `execCommand` fallback) and paste primes its buffer from `navigator.clipboard.readText`; the in-editor buffer from #16 remains the immediate fallback. JVM/desktop continues to use Compose's `ClipboardManager` (#34)
 - Vim dot-repeat (`.`) dropped the inserted text of change/insert commands (`c`, `cw`, `s`, `i`, `a`, `o`, …) — the insert-mode change event was never wired up, so `lastInsertModeChanges` stayed empty and `.` replayed only the operator/delete (#21)
 - Vim change/insert commands (`c`, `cw`, `s`, …) were not a single undo unit — `u` took two steps to revert because the insert-entry cursor selection stamped a history boundary on the operator-delete event (#23)
 
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/ClipboardCommands.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/ClipboardCommands.kt
@@ -26,20 +26,36 @@ import com.monkopedia.kodemirror.state.TransactionSpec
 import com.monkopedia.kodemirror.state.asInsert
 
 /**
+ * Write [text] to the system clipboard, bridging the platform gap.
+ *
+ * On wasmJs [platformWriteClipboard] kicks off `navigator.clipboard.writeText`
+ * synchronously inside the originating key-event gesture (returns true). On
+ * JVM/Desktop it returns false, so we fall back to Compose's `ClipboardManager`,
+ * which writes the AWT system clipboard synchronously. In both cases the
+ * in-editor [EditorSessionImpl.internalClipboard] buffer is updated as an
+ * immediate, always-reliable fallback for paste within the same session.
+ */
+private fun writeClipboard(impl: EditorSessionImpl?, text: String) {
+    impl?.internalClipboard = text
+    val handledByPlatform = platformWriteClipboard(text)
+    if (!handledByPlatform) {
+        try {
+            impl?.clipboardManager?.setText(AnnotatedString(text))
+        } catch (_: Throwable) {
+            // Best-effort system clipboard write; the internal buffer is the
+            // reliable fallback.
+        }
+    }
+}
+
+/**
  * Copy the current selection to the system clipboard.
  */
 val clipboardCopy: (EditorSession) -> Boolean = { view ->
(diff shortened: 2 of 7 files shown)


monkopedia-reviewer (changes requested):
Reviewed the full diff to merge standard — the implementation is sound and the build is green. The only blocker is a merge conflict that needs a rebase.

## Verified good
- **JVM/desktop unchanged**: `Platform.jvm.kt` / `Platform.native.kt` return `false` for both `platformWriteClipboard`/`platformReadClipboard`, so `writeClipboard` falls through to the Compose `ClipboardManager.setText` path and paste uses the synchronous `getText()` path — exactly the prior behavior.
- **copy/cut semantics + #16 buffer**: `writeClipboard` sets `internalClipboard` first (always-reliable fallback), then attempts the platform write, then the Compose fallback in a try/catch. Cut still dispatches the delete after writing. The internal buffer remains the immediate fallback for paste.
- **wasmJs**: `navigator.clipboard.writeText` invoked synchronously inside the gesture (transient-activation OK) with a hidden-textarea `execCommand('copy')` fallback; `readText().then(...)` primes the buffer for the next paste. The documented "first external paste may use the internal buffer" is an inherent async-clipboard-API constraint, not a defect.
- **apiCheck green** — the new hooks are `internal expect`/`actual`; no public API change.
- **Build (run on the branch merged with current main)**: `:view:jvmTest` (incl. all 8 `ClipboardCommandsTest` cases, 0 failures), `:view:compileKotlinWasmJs`, and `:view:apiCheck` all BUILD SUCCESSFUL.

## Blocker — merge conflict (rebase needed)
The branch is `CONFLICTING` against `main`. #49 ("preventDefault only when handled", merged to main after this branch forked) added a `### Fixed` bullet at the top of `CHANGELOG.md`, conflicting with this PR's bullet in the same spot. `Platform.wasmJs.kt` auto-merges cleanly (the two changes are in different regions and semantically independent). 

Please rebase onto `origin/main` (or merge main in) and resolve the CHANGELOG by keeping both bullets — put the #49 line first, then the #34 clipboard line. No code changes needed; once the conflict is resolved the PR is ready to merge.

Monkopedia (approved):
Re-review after rebase (pull_request.synchronize): the CHANGELOG conflict is resolved.

- Branch `fix/view-clipboard-34` now sits cleanly on top of #49's merge (ec226f2); PR is MERGEABLE (was CONFLICTING).
- CHANGELOG.md additively keeps BOTH `### Fixed` bullets (#49 browser-reserved shortcuts and #34 clipboard) with no conflict markers.
- Build green on the rebased head: `:view:jvmTest`, `:view:compileKotlinWasmJs`, and `:view:apiCheck` all BUILD SUCCESSFUL (only cosmetic deprecation warnings).

Code unchanged from the prior sound review (internal expect/actual clipboard hooks; JVM no-op preserves desktop ClipboardManager; wasmJs uses navigator.clipboard with hidden-textarea execCommand fallback; 8 ClipboardCommandsTest cases). Category feature_internal -> tier 1, no public API. Approving and merging.

Issue #33, opened 2026-05-22
Editor view overflows its bounded height — cursor goes off-screen at the bottom of a long document


Monkopedia:
## Summary
When the `KodeMirror` composable is given a bounded height and the document is longer than that height, the editor view appears to lay out at its full content height rather than clipping to the constraint and scrolling internally. Selecting/navigating to the bottom of the document puts the cursor **outside the visible window** instead of the editor scrolling to keep the cursor in view.

## Expected
The editor should constrain itself to its allotted (bounded) height and scroll its content internally, keeping the caret visible (standard CodeMirror scroll-into-view behavior).

## Actual
The view extends past the viewport; the caret at the bottom of the document ends up below the window edge.

## konstructor-side wiring (the consumer)
konstructor passes a properly bounded max-height constraint:
- `EditorContent` renders `Box(Modifier.fillMaxSize()) { KodeMirror(session, Modifier.fillMaxSize()) }`.
- That sits inside an editor pane that fills the Compose root, which fills the full-window `#compose-pane` div. So the incoming constraint is the window height (bounded), not unbounded.

So from the consumer side the editor is told "you may be at most window-height tall." The overflow suggests `KodeMirror` isn't honoring that incoming max-height constraint (laying out at intrinsic/content height), and/or scroll-into-view isn't firing on caret movement.

## Hypothesis / where to look
- How the `KodeMirror` composable maps Compose layout constraints to the CM6 view's height/scroller — does it clamp to `constraints.maxHeight` and give the scroller a bounded height, or does it size to content?
- Whether caret movement triggers a scroll-into-view within the bounded viewport.

## Caveat for the triager (possible consumer-side cause)
konstructor recently added a status-bar **overlay** pinned to the bottom of the editor pane (an overlay in the same Box as the editor). That could *cover* the bottom-most line — but "covered by an in-window overlay" is distinct from the reported "cursor outside the window," so I believe this is a genuine view-height/scroll issue in kodemirror, not the overlay. If you determine it's actually the konstructor layout/overlay, bounce it back and I'll fix konstructor.

## Environment
kodemirror 0.2.0, wasmJs (Compose Multiplatform), surfaced via konstructor.

## Repro
Open a document with more lines than fit in the editor's height; move/select to the last line; observe the caret below the bottom window edge instead of the view scrolling to keep it visible.

monkopedia-coder:
[...]

PR #61, opened 2026-05-23
Scroll dispatched selection jumps into view (#58, #33)

### LINKED ISSUE #58: Vim search 'n' moves the selection to the next match but doesn't scroll it into view
## Summary
In vim mode, pressing `n` (repeat-search-forward; likely `N` too) selects/highlights the next match of the last search, but the editor does **not scroll the viewport to bring that match into view**. If the next match is off-screen, the selection moves there invisibly — you see no change until you manually scroll.

## Expected
After `n`/`N` jumps the selection to a match, the view should scroll so the match (caret/selection) is visible — standard CodeMirror "scroll selection into view" behavior on a search-driven cursor move.

## Likely cause
A scroll-into-view gap in the Compose/wasm editor view: the vim search action moves the selection, but the resulting selection change doesn't trigger a viewport scroll-to-caret. Almost certainly the same family as #33 (view doesn't honor height / scroll-into-view) — there it's the caret at the bottom going off-screen; here it's a search-driven jump landing off-screen. Both point at "selection/caret moves aren't reliably revealed" in the wasm view.

## konstructor context (consumer)
konstructor does not implement search or scrolling — it just installs `basicSetup` + the vim keymap (`com.monkopedia.kodemirror.vim.vim()`). Search and scroll-into-view are entirely the editor view + vim engine's responsibility; nothing on the konstructor side intercepts it.

## Repro
Vim mode, in a document taller than the viewport: `/<term><Enter>` to search, then `n` to go to a match that's below the fold — selection moves to it (confirmed by subsequent edits acting there) but the view stays put; the match is off-screen.

## Environment

monkopedia-coder:
## Summary
Fixes #58 and the bounded-layout case of #33. The `:view` module never reacted to a transaction's `scrollIntoView` flag — the flag is set across the codebase (vim `n`/`N` search-repeat, search, goto-line, history/undo, lint, lsp, commands) but nothing scrolled the viewport, so any selection jump landing off-screen moved the selection invisibly.

## Root cause + fix
- `EditorSessionImpl`: when a dispatched transaction carries `scrollIntoView`, queue a `ScrollRequest(target = primary selection head, token)` as Compose snapshot state (monotonic token so a newer request supersedes a stale one); `consumeScrollRequest` clears it.
- `KodeMirror`: observes `scrollRequest` via `snapshotFlow` (not a `LaunchedEffect` key — that tripped a Compose subcompose-mid-measure error), maps the target doc position to its line-list index, and `scrollItemIntoView` implements CM6 "nearest" alignment (top-align if above the viewport, bottom-align if below, no-op if visible).
- All additions are `internal` — **no public API change**.

## #33 scope (important)
**#33's konstructor repro is the *bounded* layout case** (`KodeMirror(Modifier.fillMaxSize())` inside a height-bounded pane). I verified the layout already clips + scrolls internally under a bounded max-height — so the "bottom caret off-screen" symptom was the *same scroll-into-view gap* as #58, and is fixed here (caret-move commands set `scrollIntoView`, now consumed).

**Design note left for the maintainer (NOT addressed here):** the *unbounded* layout case only — if a consumer places `KodeMirror` under unbounded vertical constraints (parent `verticalScroll`/`wrapContentHeight`), `fillMaxSize` collapses and the list lays out at full content height. Recommendation: give the editor a default `heightIn(max=…)` or document that callers must bound the height. Flagged as an API/UX call rather than guessing — so #33 may warrant a follow-up for that case.

## Verification
`:view:jvmTest` green incl. new `ScrollIntoViewTest` (jump-down reveals off-screen target; jump-up reveals target above; no-scroll without the flag). `:view:compileKotlinWasmJs` + `:view:apiCheck` (full, incl. Android) green — no public API change. spotless/ktlint applied; CHANGELOG updated. Apple/native not cross-compiled on Linux.

## Test plan
- [ ] Manual: vim search `/foo` then `n` to an off-screen match scrolls it into view; arrow/goto to bottom of a long doc keeps the caret visible.


monkopedia-reviewer (approved):
Approved (tier 1, bug_fix). View-layer scroll-into-view fix; all additions are internal — no public API change.

Reviewed whole diff to merge standard:
- Tokened ScrollRequest with monotonic token + identity-based consumeScrollRequest correctly prevents stale scrolls and still fires on repeated same-position jumps.
- snapshotFlow keyed on session (long-lived) with columnItems via rememberUpdatedState always sees current items without restarting; avoids the subcompose-mid-measure trap of a LaunchedEffect key.
- scrollItemIntoView covers empty-layout/above/below/already-visible and oversized-item cases; CM6 'nearest' alignment looks right.
- Scope is honest: fixes #58 and #33's bounded-layout case; #33 unbounded-layout left as a documented design follow-up (acceptable).

Build verified fresh in an isolated worktree of fix/view-scroll-into-view-58-33 (JAVA_HOME 21, ANDROID_HOME set): :view:jvmTest :view:compileKotlinWasmJs :view:apiCheck — BUILD SUCCESSFUL. apiCheck ran (not skipped); new ScrollIntoViewTest 3/3 green; no scroll/gesture regression.

PR #64, opened 2026-05-27
Long lines: no-wrap + horizontal scroll by default, wire lineWrapping (#20)

### LINKED ISSUE #20: Cursor click placement clips toward end of long lines
## Summary

When clicking to place the cursor on a long line, clicks near the beginning of the line land exactly where expected, but clicks toward the end of the line appear to clip — as if there's a maximum character inset that can be selected via mouse. To reach positions past that clip point, the keyboard must be used.

## Repro (rough)

1. Open a document with a long line in the editor (long enough that the click clip is observable — likely past some inset from the line start).
2. Click near the start of the line — cursor lands precisely at the click location.
3. Click progressively further right on the same line — at some point, the cursor stops advancing and lands at a maximum reachable column rather than where the pointer actually is.
4. Move further right with arrow keys — works fine, confirming the position is reachable, just not via click.

## Environment notes

- Reported in a session with the **vim plugin enabled**. Unclear whether vim mode is involved in the click handling path or if this also reproduces with vim off — needs verification.

## Investigation hints (for whoever picks this up)

- Likely lives in the click→document-position mapping in the view module (Compose `pointerInput` → `posAtCoords`-equivalent).
- Could be a `TextLayoutResult.getOffsetForPosition` issue, a per-line width clamp, or a layout/measurement that doesn't extend far enough horizontally on long lines.
- Vim mode has its own pointer handling in some paths — worth checking whether vim's click handler intercepts and remaps to a clamped position, vs. the same bug existing in the base view.

## Acceptance

- Clicking anywhere along a long line places the cursor at the clicked character, matching keyboard-reachable positions.

monkopedia-coder:
## Summary
Fixes #20. Clicking past a certain column on a long line couldn't place the cursor (it clipped to the first wrapped visual row). Root cause: lines rendered with `BasicText` default `softWrap=true` but each line `Row` was pinned to one visual-row height with **no horizontal scroll**, so `getOffsetForPosition` could only resolve within the first wrapped row. The `lineWrapping` extension was a dead CSS attribute wired to nothing.

## Fix — match CM6 semantics
- **No-wrap default:** new internal `lineWrappingFacet: Facet<Boolean,Boolean>` (default false). In default mode each line renders `softWrap=false`, no `maxLines`, natural width, in a **horizontally-scrollable** content viewport (single shared `ScrollState` so all lines scroll in sync). The gutter stays outside the scroll region (fixed), matching CM6.
- **Opt-in wrap:** the public `lineWrapping` extension now drives `lineWrappingFacet.of(true)` → `softWrap=true`, `fillMaxWidth`, line grows to multiple rows, no horizontal scroll.
- **Click→offset:** cache-based `posAtCoords` already subtracts the stored `leftPx` (now scroll-aware); the tap-fallback `posFromVisibleItems` adds the horizontal scroll offset. Both resolve across the full line now.
- **No public API change** (`lineWrapping` stays `Extension`; the facet is `internal`). apiCheck green.

## ⚠️ Behavior change (please note)
This changes the **default**: long lines no longer wrap — they scroll horizontally (CM6's default). Consumers wanting the old wrapping behavior add the `lineWrapping` extension. **konstructor and other consumers will see this** at their next kodemirror bump. This is the intended match-upstream direction (per #20's recalibration) but it's a visible UX change, not a silent bug fix.

## Verification + coverage gap
- `:view:jvmTest`, `:view:compileKotlinWasmJs`, `:view:apiCheck` green (re-verified fresh). New tests: `LineWrappingTest` (wrap off by default / on with extension), `ClickTargetingTest.clickFarColumnOnLongLine_noWrapDefault` (click near right edge of a long line resolves past col 10).
- **Not headless-testable:** the actual horizontal-scroll *gesture* and the gutter-stays-fixed-while-content-scrolls visual — best confirmed via the showcase / gap-analysis screenshots. Flagging for manual/visual confirmation.
- Deviation (documented): per-line shared `ScrollState` rather than one editor-wide scroll container, because `LazyColumn` items can't share a single horizontally-scrolled parent without losing vertical virtualization — behavior is equivalent (synced scroll, fixed gutter).

## Test plan
- [ ] Visual (showcase/gap-analysis): long line scrolls horizontally, gutter fixed, click lands at correct column across the whole line; `lineWrapping` extension still wraps.


monkopedia-reviewer (approved):
No-wrap default + horizontal scroll matching CM6; lineWrapping facet wired. Build/tests green (jvmTest, wasmJs compile, apiCheck), click->offset scroll-aware on both cache and fallback paths. Approved per user signoff (tier 2: default-behavior change).

PR #66, opened 2026-05-27
Add horizontal scrollbar for long lines

monkopedia-coder:
## Summary
- Adds a visible horizontal scrollbar for long lines in the default no-wrap mode. The editor renders on a Skiko canvas with no native scrollbar, so long lines clipped at the right edge and horizontal scroll was only reachable via shift+wheel/trackpad (the scroll engine from #20/#64 works but had no affordance).
- The scrollbar is a custom composable built entirely in `commonMain` from Compose foundation primitives (`Box`, `background`, `draggable`, `offset`, `onSizeChanged`, `clip`/`RoundedCornerShape`). It does NOT use the skiko/desktop `HorizontalScrollbar`, which is unavailable on Android (androidMain depends on jvmMain).
- Thumb length is proportional to the visible fraction `viewportPx / (viewportPx + maxValue)` of the measured track width (min 24.dp). Thumb offset = `(value / maxValue) * (trackWidth - thumbWidth)`. Dragging maps thumb-track pixels back to content pixels via `delta * (maxValue / travel)` and applies it with `scrollState.scrollBy` in a coroutine.
- Shown only when there is real horizontal overflow (`maxValue > 0`) and line wrapping is off; hidden when content fits or `lineWrapping` is enabled.
- The composable and mount are `private`, so there is no public-API change (apiCheck stays green, no apiDump).

## Test plan
- [x] `:view:jvmTest --tests "com.monkopedia.kodemirror.view.input.HorizontalScrollTest"` (4 tests: wheel-scroll advances click offset, scrollbar visible on overflow, hidden when content fits, hidden when wrapping)
- [x] `:view:jvmTest --tests "com.monkopedia.kodemirror.view.input.*"` (no regressions)
- [x] `:view:compileKotlinJvm`, `:view:compileKotlinWasmJs`, `:view:compileDebugKotlinAndroid`, `:view:compileKotlinIosSimulatorArm64` all compile (proves the commonMain scrollbar is portable)
- [x] `:view:apiCheck` green (no public-API change)
- [x] `:view:spotlessApply`

Fixes #65

PR #68, opened 2026-05-27
Fix horizontal scroll for mixed line lengths

### LINKED ISSUE #67: Horizontal scroll broken for documents with mixed line lengths (shared ScrollState clobbered)
Follow-up to #20/#64 and #65/#66. The horizontal-scroll behavior is broken for any realistic document where line lengths vary.

### Root cause
Each visible line gets its own `Modifier.horizontalScroll(horizontalScrollState)`, but they all share a SINGLE `ScrollState`. Short lines (laid out last) write `maxValue = 0` to the shared state, which:
1. clamps horizontal scrolling to 0 — a long line among short lines **cannot be scrolled at all**, and
2. hides the new scrollbar from #66 (its `maxValue > 0` guard is never satisfied).

A single-long-line test masks this; the bug only appears with mixed line lengths.

### Reproduction (verified)
DesktopComposeUiTest, doc = one long line followed by several short lines, width 300:
- scrollbar node count = 0 (should be 1)
- click x=250 → scroll-horizontal 600 → click x=250 again: offset stays 27 (should advance to ~404)

### Validated fix
Make every no-wrap line measure to the SAME content width = the widest line's width, via `widthIn(min = maxLineWidthDp)`, where the max is tracked from each line's `onTextLayout` `result.size.width`. With uniform content width, all per-line `horizontalScroll` modifiers compute the same `maxValue`, so the shared `ScrollState` has a correct, consistent range. With this change the reproduction yields scrollbarNodes=1 and offset 27→404.

### Tests
Add UI regression tests with MIXED line lengths (long line among short lines): scrollbar visible, wheel scroll advances click offset, and (no-overflow / wrap) cases still hide it.

monkopedia-coder:
## Summary

Long lines did not scroll horizontally in real documents with mixed line lengths. Each visible line carries its own `Modifier.horizontalScroll(horizontalScrollState)`, but all lines share one `ScrollState`. Short lines, laid out last, set the shared `maxValue` to 0 — which both clamped scrolling (a long line among short lines couldn't scroll) and hid the #66 horizontal scrollbar (its `maxValue > 0` guard never fired). A single-long-line test masked the bug; only mixed line lengths exposed it.

## Fix

In `EditorContent`, track the widest no-wrap line width (`maxLineWidthPx`, updated from each line's `onTextLayout`) and apply it as `widthIn(min = ...)` to every no-wrap line's content box. All per-line `horizontalScroll` modifiers then compute the same `maxValue`, so the shared `ScrollState` range stays correct regardless of layout order. Internal-only change; no public API change.

Known limitation (out of scope): `maxLineWidthPx` grows monotonically within a session, so after deleting the longest line the scroll range can stay slightly wider than needed until the session is recreated — a minor cosmetic wart, intentionally not addressed here to avoid per-keystroke flicker/scroll-range jumps.

## Tests

New `MixedContentScrollTest` (regression guards that would have caught #67), doc = one long line among four short lines:
- `scrollbarVisibleWithMixedContent`: the `KodeMirror_hscroll` scrollbar node exists.
- `longLineScrollsAmongShortLines`: click near the right edge, wheel-scroll horizontally, click again; the resolved click offset advances well past the viewport (validated 27 → 404; asserts `after > before + 100`).

Existing `HorizontalScrollTest` tests continue to pass.

Fixes #67


monkopedia-reviewer (approved):
Reviewed and independently verified.

**Classification:** bug_fix → tier 1 (Monkopedia/kodemirror baseline: `* : tier 1`)

**Diff review:** The fix is sound. The root cause — short lines clobbering the shared `ScrollState.maxValue` to 0 on layout — is correctly diagnosed. The fix (tracking `maxLineWidthPx` monotonically and applying `widthIn(min = maxLineWidthDp)` to every no-wrap line) ensures all per-line `horizontalScroll` modifiers compute a consistent max. The `HorizontalScrollbar` composable was introduced in the earlier #66 commit that is bundled in this branch's history but is new only relative to the base; `HorizontalScrollbar` itself is all commonMain Compose primitives — compiles cleanly on every target. The monotonic-growth limitation (widest line never shrinks within a session) is documented and out of scope per the PR body; not a blocker.

**CI re-verification (all green):**
- `:view:jvmTest --tests MixedContentScrollTest, HorizontalScrollTest`: 6/6 PASSED
- `:view:jvmTest --tests com.monkopedia.kodemirror.view.input.*`: 49/49 PASSED (0 failures, 0 skips across all 10 suites)
- `:view:compileKotlinJvm`, `:view:compileKotlinWasmJs`, `:view:compileDebugKotlinAndroid`: BUILD SUCCESSFUL
- `:view:apiCheck`: BUILD SUCCESSFUL — no public API change confirmed

**Test quality:** both new tests are meaningful. `longLineScrollsAmongShortLines` wheel-scrolls and asserts click offset advances by >100 (validates the core fix). `scrollbarVisibleWithMixedContent` asserts the scrollbar node exists for mixed-length content (regression guard for the `maxValue > 0` guard).

PR #73, opened 2026-05-27
Fix zero-height collapse / LazyColumn crash under unbounded height (#33)

### LINKED ISSUE #33: Editor view overflows its bounded height — cursor goes off-screen at the bottom of a long document

monkopedia-coder:
## Summary

Residual scope of #33 (the bounded case was already fixed in #61): when `KodeMirror` is placed under an **unbounded** vertical constraint — a parent with `Modifier.verticalScroll(...)`, `wrapContentHeight()`, or another scrolling container — the editor misbehaved.

## What I found (investigation)

I wrote a probe test placing `KodeMirror` inside `Box(Modifier.verticalScroll(...))` with a 100-line doc and a 300px surface:

- The outer `Column(modifier.fillMaxSize())` receives `maxHeight == Constraints.Infinity`. The content `Box(Modifier.weight(1f))` distributes the *remaining* (infinite) space, which resolves to **zero** — the editor's measured height was `0` (probe: `editorHeight=0`). The editor was effectively invisible; caret reveal had nothing to scroll.
- When I tried letting the content wrap its natural height instead, the inner `LazyColumn` **threw** `IllegalStateException: Vertically scrollable component was measured with an infinity maximum height constraints` — Compose disallows measuring a scrollable with infinite maxHeight. The `weight(1f)`/zero collapse had merely been *masking* this crash.

Root cause is **kodemirror-side**: the `fillMaxSize()` + `Column` `weight(1f)` layout has no defined behavior under an infinite incoming `maxHeight`.

## Approach (and why)

Added a private `Modifier.boundUnconstrainedHeight(contentHeightPx)` layout modifier on the editor `Column`:

- Bounded height → **no-op** (constraints pass through unchanged; inner list scrolls and caret reveal works exactly as before, including #61's bounded fix).
- Unbounded height (`maxHeight == Constraints.Infinity`) → replace it with the document's already-computed natural content height (`contentHeightPx` = top/bottom padding + all lines). The editor then grows to fit the whole document and the surrounding scroll container scrolls it — no zero-collapse, no crash.

This is the most upstream-faithful degrade: CM6 expects a bounded editor height, but growing to content under an unbounded parent is reasonable and must not fail silently. No arbitrary magic dp is hardcoded — the fallback is the real content height. **No public API change** (`apiCheck` green; the helper is private).

Also documented the height contract in the `KodeMirror` KDoc (callers should give a bounded height for in-editor scrolling/caret reveal; under unbounded constraints the editor grows to content and the outer container governs visibility).

## Test

`UnboundedHeightTest` asserts that under a `verticalScroll` parent the editor's measured height grows well past the surface height (no zero-collapse, no crash).

## Verification

- `:view:jvmTest --tests "...input.*"` — green
- `:view:compileKotlinJvm :view:compileKotlinWasmJs` — green
- `:view:apiCheck` — green (no public API change)
- `:samples:showcase:compileKotlinWasmJs` — green

Fixes #33


monkopedia-reviewer (approved):
Reviewed as monkopedia-reviewer. All checks verified independently.

**Code review — tier 1 (bug_fix)**

The `boundUnconstrainedHeight` layout modifier is correctly implemented:

- **Bounded path is a true no-op**: when `constraints.maxHeight != Constraints.Infinity` the effective constraints are passed through unchanged — the inner `LazyColumn` sees exactly the same bounds as before PR #61. `ScrollIntoViewTest` remains green.
- **Unbounded path**: correctly detects `constraints.maxHeight == Constraints.Infinity` and substitutes `contentHeightPx.roundToInt().coerceAtLeast(0)`, giving the editor a finite height so it grows to content and the surrounding scroll container scrolls it. No zero-collapse, no `IllegalStateException` from `LazyColumn`.
- **`fillMaxSize()` → `fillMaxWidth() + boundUnconstrainedHeight()`**: correct decomposition; width still fills, height is now delegated to the new modifier.
- **No public API change**: helper is `private`, `apiCheck` is green.
- **Test is meaningful**: `UnboundedHeightTest` places the editor under a `verticalScroll` parent with 100 lines in a 300px surface and asserts `height > 300` — catches both the zero-collapse and the crash.
- **Horizontal scroll features untouched**: `HorizontalScrollTest`, `HorizontalScrollIntoViewTest` all pass.
- **Panels limitation**: gracefully degraded and clearly documented in KDoc — not a blocker.

**CI (independently verified)**:
- `:view:jvmTest --tests "...input.*"` — 12 suites, 54 tests, 0 failures, 0 errors (UnboundedHeightTest, ScrollIntoViewTest, HorizontalScroll* all green)
- `:view:compileKotlinJvm :view:compileKotlinWasmJs` — BUILD SUCCESSFUL
- `:view:compileDebugKotlinAndroid` — BUILD SUCCESSFUL
- `:view:apiCheck` — BUILD SUCCESSFUL

PR #78, opened 2026-05-27
Make vim gj/gk move by visual (wrapped) row

### LINKED ISSUE #77: Vim gj/gk should move by visual (wrapped) row, not logical line
Vim's `gj`/`gk` (display-line motion) currently behave identically to `j`/`k` — they move by whole logical lines, not by visual/wrapped rows. The `moveByDisplayLines` motion (`vim/src/commonMain/.../VimMotions.kt:277`) just does logical line math, with an explicit comment (~line 296): "In Compose, display lines == logical lines (no wrapping info available)". That assumption is now false.

### Why now
The view's `moveVertically(session, sel, forward)` (`view/src/commonMain/.../CursorMotion.kt:180`) is already wrap-aware (uses `coordsAtPos`/`posAtCoords` to step by visual row). The non-vim default and emacs keymaps already move by visual row through it. Vim's motion layer just never calls it. The bridge exists: vim motions get `cm: VimEditor`, and `VimEditor(val session: EditorSession)` exposes `.session` — exactly what `moveVertically` takes.

### Scope (deliberately narrow — match upstream vim)
- Keep `j`/`k` LINEWISE (logical) — that matches real vim and upstream `@replit/codemirror-vim`. Do NOT change `moveByLines`.
- Fix `moveByDisplayLines` (`gj`/`gk`) to move by VISUAL row: call `moveVertically(cm.session, currentSelection, forward = motionArgs.forward)` `repeat` times and return the resulting head as the motion result. Remove the "display lines == logical lines" shortcut/comment.
- Preserve sensible goal-column behavior across consecutive `gj`/`gk` (let `moveVertically`'s column handling drive it; don't fight vim `lastHPos`).
- Users who want `.vimrc`-style visual `j`/`k` can then remap in 2 lines via the existing public API: `Vim.map("j","gj")` / `Vim.map("k","gk")` (or `:nnoremap j gj`). Add a short doc/KDoc note (and optionally a line in the Vim showcase demo description) pointing this out so it's discoverable.

### Tests
Add a UI test (DesktopComposeUiTest harness, like the view input tests) with a wrapped line + `lineWrapping`: assert `gj` moves the caret DOWN by one visual row (stays on the same logical line when the line wraps), `gk` moves up one visual row, while `j`/`k` still jump whole logical lines. Caret position should change by ~one row height, not a whole logical line.

### Note
If `moveByDisplayLines` runs in any context where the live view layout/geometry isn't available (headless/no TextLayoutResults), fall back gracefully to the current logical-line behavior rather than throwing.

monkopedia-coder:
## Summary
- Vim `gj`/`gk` (`moveByDisplayLines`) now move by VISUAL (wrapped) row instead of by logical line. They delegate to the view's wrap-aware `moveVertically(cm.session, sel, forward)` once per repeat, with the goal column tracked by `moveVertically`. `j`/`k` (`moveByLines`) stay linewise (logical), matching upstream vim. Falls back gracefully to logical-line motion when live layout geometry is unavailable (headless).
- Fixed `moveVertically` (`view/.../CursorMotion.kt`): it was collapsing any same-logical-line `posAtCoords` result back to a whole-line jump, so it never actually stepped by visual row on a wrapped line (this also affected the default cursor up/down). It now honors the wrap-aware `posAtCoords` offset directly and only uses the logical-line fallback when no geometry is available or the offset didn't change.
- Removed the stale "display lines == logical lines" comment; added a KDoc note pointing users at `Vim.map("j","gj")` / `Vim.map("k","gk")` for `.vimrc`-style visual `j`/`k`.

## Tests
- `view` `CursorMotionTest`: a fake `EditorSession` models a 20-char logical line wrapping into two 10-col visual rows + a short next line. Asserts `moveVertically` steps by visual row (offset 3 -> 13, same logical line), crosses to the next logical line only after the last wrapped row, mirrors upward, and falls back to logical-line motion with no geometry.
- `vim` `VimDisplayLineMotionTest`: drives real `gj`/`gk`/`j` keystrokes over a wrapped-line geometry session. `gj`: (0,3) -> (0,13) (same logical line, second visual row); `gk` returns to (0,3); `j` jumps the whole logical line to (1,3).

## Verification (JAVA_HOME=java-21)
- `:vim:jvmTest` + `:view:jvmTest`: PASS
- `:vim:compileKotlinJvm :vim:compileKotlinWasmJs :view:compileKotlinWasmJs`: PASS
- `:vim:compileDebugKotlinAndroid :view:compileDebugKotlinAndroid` (ANDROID_HOME set): PASS
- `:vim:apiCheck :view:apiCheck`: PASS (no public API change)
- `:vim:spotlessApply :view:spotlessApply`: clean

Fixes #77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Bumped `com.monkopedia.lsp` (`lsp` + `lsp-ksrpc`) from `1.0.0-RC2` to `1.0.0-RC3`. RC3 tightens several `Hover.contents` / `ServerCapabilities.textDocumentSync` fields from untyped `JsonElement` to strict union types, so `:lsp-client`'s `ServerHover` and `DocumentSync` now read the typed `HoverContents` and `ServerCapabilitiesTextDocumentSync` unions directly instead of hand-parsing `JsonElement` (internal parsing only; no public API change). The `:lsp-client` module remains pre-stable.
 
 ### Fixed
+- Vim gj/gk now move by visual (wrapped) row (#77)
 - Caret height and gutter alignment on wrapped lines (#75)
 - `KodeMirror` placed under an unbounded vertical constraint (a parent with `Modifier.verticalScroll(...)`, `wrapContentHeight()`, or another scrolling container) no longer collapses to zero height or crashes the inner `LazyColumn` (which disallows an infinite `maxHeight`). The editor now grows to its content height so the surrounding scroll container can scroll it. The `KodeMirror` KDoc documents the height contract: callers should give the editor a bounded height for in-editor scrolling and caret reveal (#33)
 - Cursor now scrolls horizontally into view in no-wrap mode (#69)
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/CursorMotion.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/CursorMotion.kt
@@ -199,14 +199,28 @@ fun moveVertically(
         coords.top - 1f
     }
     val rawPos = view.posAtCoords(coords.centerX, targetY)
-    val resultLine = rawPos?.let { doc.lineAt(DocPos(it)) }
 
-    // Determine target line number — if posAtCoords snapped back to the
-    // same line (e.g. inter-line gap), move to the adjacent line instead.
-    val targetLineNum = if (resultLine != null && resultLine.number != currentLine.number) {
-        resultLine.number.value
+    // posAtCoords is wrap-aware: targetY lands on the adjacent VISUAL row, which
+    // may be another wrapped row of the SAME logical line. Honor that result
+    // directly so vertical motion steps by visual row (this is what gj/gk and
+    // the default cursor up/down rely on for wrapped lines).
+    if (rawPos != null && rawPos != sel.head.value) {
+        val newPos = DocPos(rawPos)
+        val anchor = if (extend) sel.anchor else newPos
+        return if (extend) {
+            EditorSelection.range(anchor, newPos, goalColumn = goalCol)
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed and independently verified.

**Classification**: bug_fix (visual-row motion was broken; internal implementation, no public API change)
**Tier**: 1 (Monkopedia/kodemirror baseline is tier 1 for bug_fix)

**Diff review**:
- `CursorMotion.kt`: The guard removal is correct. The old code checked if posAtCoords snapped to a different logical line and fell back to whole-line jump otherwise — which meant wrap-aware geometry was *always* discarded for same-logical-line results. The new code honors the rawPos directly when it's non-null and different from current head, falling back only when geometry is genuinely absent or produced no movement. This is the right fix.
- `VimMotions.kt`: moveByDisplayLines now drives moveVertically per-repeat, with a graceful fallback to logical-line motion when geometry is unavailable. j/k (moveByLines) are untouched.
- No public API surface changed (apiCheck green).
- File set matches expected: CursorMotion.kt, VimMotions.kt, CursorMotionTest.kt, VimDisplayLineMotionTest.kt, CHANGELOG.md.

**Independent verification** (all green):
- :view:jvmTest PASS — 16 CursorMotionTest cases (12 pre-existing + 4 new wrap-aware assertions), no regressions in ScrollIntoView, ClickTargeting, etc.
- :vim:jvmTest PASS — 3 VimDisplayLineMotionTest cases: gj (0,3)→(0,13) same logical line; gk returns to (0,3); j jumps to (1,3) whole logical line. Tests are meaningful and distinguish visual-row from logical-line motion.
- :vim:compileKotlinJvm / compileKotlinWasmJs / compileDebugKotlinAndroid: PASS
- :view:compileKotlinWasmJs / compileDebugKotlinAndroid: PASS
- :vim:apiCheck / :view:apiCheck: PASS

PR #88, opened 2026-05-30
Preserve goal column across vertical cursor motion

### LINKED ISSUE #87: Goal/sticky-column ('column memory') broken on vertical cursor motion — regression from #78
**Regression introduced by #78** (the vim gj/gk visual-row change, which also rerouted default ArrowUp/Down through the modified `moveVertically` in `view/src/commonMain/.../CursorMotion.kt`).

### Symptom
Goal-column ("column memory") is not preserved across vertical moves the way CM6 does. Repro (gap-analysis `navigation.spec.ts:182` "column memory across lines"): Ctrl+Home → End → ArrowDown → ArrowDown. CM6 final cursor col = **21**; KodeMirror = **25**. After `End` sets the goal column to the line end, subsequent ArrowDowns should keep that remembered column (clamped per line), matching CM6.

### Confirmed via controlled A/B (same machine, same test)
- At `827b553` (commit BEFORE #78): test **PASSES** (KM col matches CM6).
- At current `main`/`812b446` (after #78): test **FAILS** (KM 25 vs CM6 21).
#78 is the only change touching `moveVertically`, so it is the cause. The #78 guard change (honoring the wrap-aware `posAtCoords` offset for visual-row motion) disrupted the goal/sticky-column logic for default vertical motion.

### Why not caught earlier
This is exercised by the gap-analysis **Playwright** suite (KM-vs-CM6), which is NOT run in CI; the JVM unit suites + CI `check` are green. (Found while running the full Playwright suite to confirm 0.3.0 readiness.)

### Fix
Restore CM6-correct goal-column semantics in `moveVertically`: a vertical move should target the remembered goal x/column (set on the last horizontal move / End), clamped to the target line, and must NOT overwrite the goal column on vertical moves — WHILE preserving the visual-row (wrap-aware) motion #78 added. Check both no-wrap and wrapped cases.

### Tests
- Add a CI-gated `:view` jvmTest exercising goal-column across vertical moves (End → down → down lands at the goal column clamped per line; also a move over a short line and back preserves the column). Must fail pre-fix.
- Re-run the gap-analysis `navigation.spec.ts` "column memory across lines" to confirm it passes (KM col == CM6 col).

Blocks the held 0.3.0 publish.

monkopedia-coder:
## Root cause (regression from #78)

#78 (vim gj/gk visual-row motion) rewrote `moveVertically` in
`view/.../CursorMotion.kt`. The old code projected the **remembered goal column**
onto the target line (`targetCol = goalCol.coerceAtMost(targetLine.text.length)`),
giving CM6-style column memory. #78 replaced that with the wrap-aware
`posAtCoords(coords.centerX, targetY)` result used **directly** — and `coords.centerX`
is the x of the *current head*, not the x of the remembered goal column.

So after `End` (cursor at line-1 end, goal column captured on the first vertical
move) → `Down` clamps onto a short line, and the *next* `Down` uses the short
line's current x rather than the goal — losing column memory. Repro
(Ctrl+Home → End → Down → Down) gave the short-line column instead of the goal.

## The fix

In `moveVertically`:
- Derive the horizontal target x from the **goal column** resolved on the current
  line (`coordsAtPos(goalPos)`), not from the current head's `centerX`.
- After `posAtCoords` lands on a visual row, if that row is a **different logical
  line**, re-project the remembered goal column onto it (clamped to its length).
  This is what springs the cursor back out to the goal on a longer line after a
  short intermediate line. `posAtCoords` can only resolve the goal x as far as the
  current (possibly short) line extends, so the x alone under-reports the goal —
  clamping the stored goal column fixes it.
- Within the **same** wrapped logical line, the `posAtCoords` (visual-row) result
  is honored unchanged, so #78's gj/gk and default arrow visual-row motion on
  wrapped lines is fully preserved.
- The goal column is carried forward unchanged on every vertical move (never
  overwritten), and horizontal moves/edits reset it as before.

Goal column is a horizontal concept and visual-row motion is vertical; the fix
moves to the next visual row *at the remembered goal x*, exactly CM6's behavior.

## Tests (CI-gated guard — gap-analysis Playwright is not in CI)

Added two `commonTest` cases to `CursorMotionTest` using a deterministic
fake-geometry `EditorSession` (1px/char, 10px/row, three non-wrapping lines:
`"a long first line here"` / `"shrt"` / `"another long line here"`):
- `goalColumnPreservedAcrossShortIntermediateLine`: End-of-line-1 → Down (clamps
  to short line col 4, goal 22 preserved) → Down (springs back to col 22 on
  line 3). Asserts real column values.
- `goalColumnPreservedOnDownThenUpRoundTrip`: down to the short line then back up
  returns to col 22.

Both **FAIL pre-fix** (verified by stashing the fix: got short-line col 4 instead
of goal 22) and **PASS post-fix**. The existing `moveVertically` visual-row tests,
`HorizontalScrollIntoViewTest`, and the `:vim` `VimDisplayLineMotionTest` (gj/gk)
all still pass.

## Verification (local)

- `:view:jvmTest` — green (incl. new tests + existing motion/scroll tests)
- `:vim:jvmTest` — green (gj/gk not regressed)
- `:view:compileKotlinWasmJs` — green
- `:view:apiCheck` — green (no public API change)

No `:state` change; `moveVertically`'s signature is unchanged.

Fixes #87

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Bumped `com.monkopedia.lsp` (`lsp` + `lsp-ksrpc`) from `1.0.0-RC2` to `1.0.0-RC3`. RC3 tightens several `Hover.contents` / `ServerCapabilities.textDocumentSync` fields from untyped `JsonElement` to strict union types, so `:lsp-client`'s `ServerHover` and `DocumentSync` now read the typed `HoverContents` and `ServerCapabilitiesTextDocumentSync` unions directly instead of hand-parsing `JsonElement` (internal parsing only; no public API change). The `:lsp-client` module remains pre-stable.
 
 ### Fixed
+- Goal/sticky-column ("column memory") is preserved across vertical cursor moves again (regressed by the visual-row motion change) (#87)
 - Vim gj/gk now move by visual (wrapped) row (#77)
 - Caret height and gutter alignment on wrapped lines (#75)
 - `KodeMirror` placed under an unbounded vertical constraint (a parent with `Modifier.verticalScroll(...)`, `wrapContentHeight()`, or another scrolling container) no longer collapses to zero height or crashes the inner `LazyColumn` (which disallows an infinite `maxHeight`). The editor now grows to its content height so the surrounding scroll container can scroll it. The `KodeMirror` KDoc documents the height contract: callers should give the editor a bounded height for in-editor scrolling and caret reveal (#33)
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/CursorMotion.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/CursorMotion.kt
@@ -186,26 +186,56 @@ fun moveVertically(
     val doc = view.state.doc
     val currentLine = doc.lineAt(sel.head)
 
-    // Preserve goalColumn across consecutive vertical moves so that
-    // moving down through a short line and back up remembers the
-    // original column.
+    // Preserve goalColumn across consecutive vertical moves so that moving down
+    // through a short line and back up remembers the original column. The goal
+    // column is the REMEMBERED horizontal position; it is captured on the first
+    // vertical move and must NOT be overwritten by subsequent vertical moves.
     val goalCol = sel.goalColumn
         ?: (sel.head.value - currentLine.from.value)
 
     val coords = view.coordsAtPos(sel.head.value, if (forward) 1 else -1) ?: return sel
+
+    // Vertical motion targets the adjacent visual row at the REMEMBERED goal
+    // column's x, not the current head's x. When a stored goalColumn differs
+    // from the head's column (e.g. after End landed the cursor short on a short
+    // line), resolve the goal column to an x on the current logical line and use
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. Approving.

**Correctness:**
- Fix is correct: derives goalX from the remembered goal column (not the current head's x), and re-projects the goal column onto a different logical line after posAtCoords. Within the same wrapped logical line, posAtCoords is honored unchanged — #78's visual-row motion is fully preserved.
- Both new tests discriminate: confirmed `goalColumnPreservedAcrossShortIntermediateLine` and `goalColumnPreservedOnDownThenUpRoundTrip` FAIL pre-fix (got short-line col 4) and PASS post-fix (col 22).
- VimDisplayLineMotionTest and all existing visual-row motion tests pass (`vim:jvmTest` green).
- Goal column is carried forward unchanged on vertical moves; horizontal resets are untouched.
- apiCheck green — no public API change.

**Quality:** Code is idiomatic Kotlin, consistent with surrounding style, no dead/duplicated code. Comments are accurate and appropriately detailed for a subtle correctness invariant. The goalCol != headCol guard is a clean optimization that avoids an extra coordsAtPos on the common case.

**CI:** All three required checks completed and passed — check (10m27s), check-macos (7m44s), check-ios (14m2s). Waited for real CI before merging.

PR #93, opened 2026-06-01
Fix editor mount crash when hover-tooltip/lint extension is used

### LINKED ISSUE #92: Editor mount crashes with 'EditorSession is not attached' when a hover-tooltip/lint extension is used (0.3.0 regression)
**Regression in 0.3.0**, reported + root-caused by the downstream consumer (konstructor); verified in code. Blocks any consumer using `linter(...)` (lint hover tooltips) — the editor throws during composition on every mount.

### Symptom
`kotlin.IllegalStateException: EditorSession is not attached to a KodeMirror composable` at `view/EditorSessionImpl.kt:68` (the `coroutineScope` getter's `?: error(...)`), during composition of `KodeMirror`. Consumer's identical editor code is green on 0.2.0.

### Root cause (init-order hazard)
- `HoverTooltipPlugin` (`view/.../Tooltip.kt:229`) reads `session.coroutineScope` EAGERLY in a `val` initializer: `private val job = SupervisorJob(session.coroutineScope.coroutineContext[Job])` (and line 231). (This became eager in the 0.3.0 async/debounced-tooltip rewrite.)
- In `view/.../KodeMirror.kt`, `pluginHost = remember(session) { ViewPluginHost(session).also { it.syncToState(state, null) } }` at **line 144-145** CONSTRUCTS the plugins, but `impl.backingCoroutineScope = compositionScope` is only assigned at **line 170** — after construction. So the eager read hits the detached-session guard → throws.
- Triggered via public API: `lint/Lint.kt:98` registers `diagnosticHover = hoverTooltip { ... }`, so `HoverTooltipPlugin` is constructed on every mount of an editor that uses `linter`.

### Fix (preferred)
Attach the coroutine scope BEFORE plugins are constructed: move `val compositionScope = rememberCoroutineScope()` + `impl.backingCoroutineScope = compositionScope` ABOVE the `pluginHost = remember(...)` at KodeMirror.kt:144 (the assignments there are plain composition-body, so this is safe and fixes the general hazard — any plugin may legitimately need the scope at construction). Remove the now-duplicate assignment at :170. (Alternative/secondary: make `HoverTooltipPlugin` defer the `SupervisorJob(session.coroutineScope...)` to first use — but the reorder is the more general fix.)

### Regression test (CI-gated — this slipped past units + Playwright + native)
Add a `:view` jvmTest that mounts an editor with a `hoverTooltip { ... }` extension (the direct trigger; no `:lint` dependency needed) via the `runEditorTest` harness and asserts composition does NOT throw. Must fail pre-fix (the `IllegalStateException`) and pass post-fix.

### Release
Ship as **0.3.1** (0.3.0 is published, can't unpublish). Downstream (konstructor) is holding its bump until 0.3.1 is on Central. Auto-release is now enabled (#91), so the deploy will publish without a manual Portal step.

monkopedia-coder:
## Problem (#92)

Mounting a `KodeMirror` editor that installs a hover tooltip (e.g. via the `linter(...)` extension, or directly via `hoverTooltip { ... }`) crashed during composition:

```
kotlin.IllegalStateException: EditorSession is not attached to a KodeMirror composable
  at EditorSessionImpl.kt:68 (the coroutineScope getter's null guard)
```

## Root cause

`HoverTooltipPlugin` reads `session.coroutineScope` eagerly in a `val` initializer (`Tooltip.kt`). In `KodeMirror.kt` the plugin host was constructed (`ViewPluginHost(session).also { it.syncToState(...) }`, which builds the plugins) *before* `impl.backingCoroutineScope = compositionScope` was assigned. So the eager read hit the null guard and threw. This is an init-order hazard for any plugin that needs the scope at construction time.

## Fix

Attach the session coroutine scope **before** the plugin host is constructed. The wiring assignments are plain composition-body statements (not in a side effect), so this is a safe reorder:

- Moved `val compositionScope = rememberCoroutineScope()` and `impl.backingCoroutineScope = compositionScope` above the `pluginHost = remember(session) { ... }` block.
- Removed the now-duplicate assignment from its old spot (assigned exactly once, before plugin construction).
- Other eager assignments (`pluginHost`, `lineLayoutCache`, `clipboardManager`) and the `DisposableEffect` cleanup are unchanged.

No public API change (internal reorder).

## Regression test

Added `view/src/jvmTest/.../input/HoverTooltipMountTest.kt`: mounts an editor with a `hoverTooltip { _, _ -> null }` extension via the `runEditorTest` harness and asserts the document mounted. If composition throws, the harness never reaches the assertion and the test fails.

- **Pre-fix (fix stashed):** test FAILS with `java.lang.IllegalStateException at EditorSessionImpl.kt:68` — the exact regression.
- **Post-fix:** test PASSES, full `:view:jvmTest` suite green.

## Verification (all green)

- `:view:jvmTest` (incl. new test)
- `:view:compileKotlinWasmJs`
- `:view:compileDebugKotlinAndroid`
- `:view:apiCheck` (no apiDump needed)
- `:view:spotlessApply`

Fixes #92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- Editor no longer crashes on mount ("EditorSession is not attached") when a hover-tooltip or lint extension is used; the session coroutine scope is now attached before plugins are constructed (#92)
+
 ## [0.3.0] - 2026-05-29
 
 ### Tests
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -141,6 +141,14 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
     val impl = session as EditorSessionImpl
     val state by session::state
 
+    // Attach the coroutine scope to the session BEFORE constructing the plugin
+    // host. syncToState() builds the view plugins, and some plugins (e.g.
+    // HoverTooltipPlugin) read session.coroutineScope eagerly in their
+    // initializer; if the scope isn't attached yet that read throws
+    // "EditorSession is not attached to a KodeMirror composable" (#92).
+    val compositionScope = rememberCoroutineScope()
+    impl.backingCoroutineScope = compositionScope
+
     val pluginHost = remember(session) {
         ViewPluginHost(session).also { it.syncToState(state, null) }
     }
@@ -157,7 +165,6 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
         mutableMapOf<Int, TextLayoutResult>()
     }
 
-    val compositionScope = rememberCoroutineScope()
     val clipboardManager = LocalClipboardManager.current
 
     // Wire up session internals eagerly (during composition, not in a side
@@ -167,7 +174,6 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
     // harmless because remember() always returns the same instances.
     impl.pluginHost = pluginHost
     impl.lineLayoutCache = lineLayoutCache
-    impl.backingCoroutineScope = compositionScope
     impl.clipboardManager = clipboardManager
 
     DisposableEffect(session) {
--- a/view/src/jvmTest/kotlin/com/monkopedia/kodemirror/view/input/HoverTooltipMountTest.kt
+++ b/view/src/jvmTest/kotlin/com/monkopedia/kodemirror/view/input/HoverTooltipMountTest.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. Approving.

**Regression test discriminates:** `HoverTooltipMountTest.editorWithHoverTooltipMountsWithoutCrashing` mounts an editor with `hoverTooltip { _, _ -> null }` via `runEditorTest`. On the pre-fix ordering, `HoverTooltipPlugin`'s `val job = SupervisorJob(session.coroutineScope.coroutineContext[Job])` (Tooltip.kt:229) triggers the null guard in `EditorSessionImpl.coroutineScope` (line 68) during `pluginHost`'s `remember` lambda — throwing `IllegalStateException` before the test block ever runs. Post-fix the scope is attached first and the assertion is reached. The test is meaningful.

**Reorder correctness:**
- `compositionScope = rememberCoroutineScope()` and `impl.backingCoroutineScope = compositionScope` now appear in the composition body BEFORE the `pluginHost = remember(session) { ViewPluginHost(session).also { it.syncToState(...) } }` block — so the scope is set before any plugin is constructed. ✓
- The old assignment at the former line ~170 was removed. `backingCoroutineScope` is assigned exactly once. ✓
- `DisposableEffect` `onDispose` still nulls `backingCoroutineScope`. Cleanup intact. ✓
- Single `rememberCoroutineScope()` call — not duplicated. ✓
- `Tooltip.kt` is unchanged. ✓

**Quality pass:** Comment above the reorder explains the why (#92, the eager read, the fix). Change is minimal — 3 files, no scope creep. CHANGELOG entry present with issue reference.

**Local re-verification (all green):** `:view:jvmTest` (83 tests incl. `HoverTooltipMountTest`), `:view:compileKotlinWasmJs`, `:view:compileDebugKotlinAndroid`, `:view:apiCheck`.

**Remote CI (all 3 required checks PASSED):** `check` ✓, `check-ios` ✓, `check-macos` ✓. Waited for all to complete before merging.

PR #99, opened 2026-06-02
Auto-create GitHub Release on publish + document release process

### LINKED ISSUE #98: Auto-create a matching GitHub Release in deploy.yml after publish
We publish to Maven Central via deploy.yml + a git tag, but never created a matching GitHub Release (caught 2026-06-02; v0.3.0's release was backfilled manually). Automate it so every publish produces a matching GitHub Release.

### Change (deploy.yml)
Add a final job (`needs:` the full deploy matrix, so it runs only after all 10 publish jobs succeed) that creates the GitHub Release for the published version:
- Determine the version from the build, e.g. `VERSION=$(grep '^version = ' kodemirror-bom/build.gradle.kts | sed -E 's/.*"([^"]+)".*/\1/')`.
- SKIP release creation if VERSION contains `-SNAPSHOT` or `-RC` (only final releases get a GitHub Release).
- Extract the matching CHANGELOG section for notes: `awk -v v="$VERSION" '$0=="## ["v"] - "substr($0,index($0,"- ")+2){...}'` — simpler: `awk "/^## \\[$VERSION\\]/{f=1;next} /^## \\[/{f=0} f" CHANGELOG.md > notes.md` (adjust escaping for YAML).
- `gh release create "v$VERSION" --notes-file notes.md --latest --verify-tag` (`--verify-tag` fails if the tag isn't pushed — a good safety so we don't create a release for an untagged commit).
- The release job needs `permissions: contents: write` (deploy.yml currently has `contents: read`); set it on that job (or bump the top-level perms minimally). Use the Actions `GITHUB_TOKEN`.

### Docs (CLAUDE.md)
Add a concise "Release Process" section documenting the full flow: version-bump PR → tag vX.Y.Z → dispatch deploy.yml (publishes to Central AND now auto-creates the matching GitHub Release) → verify BOM on Central → notify downstream consumers. Note tags/releases exist only for fully-published versions.

### Notes
- Can't fully e2e-test without a real publish; verify the workflow YAML is valid (e.g. `actionlint` if available, or careful review) and the version/changelog extraction logic is correct (test the awk/grep locally against the current CHANGELOG). The release job runs AFTER publish, so a bug there won't block the publish itself.
- The next release (0.3.2) will be the first to exercise this end-to-end.

monkopedia-coder:
## What

Closes the gap where publishing produced a Maven Central artifact + git tag but never a matching GitHub Release.

### deploy.yml — new \`release\` job
A final job appended after the existing build/deploy jobs (those are unchanged except being depended on):
- \`needs: deploy\` — runs only after **all** 10 Maven Central matrix publishes succeed.
- \`runs-on: ubuntu-latest\`, with \`permissions: contents: write\` (top-level is \`contents: read\`) and \`GH_TOKEN: \${{ secrets.GITHUB_TOKEN }}\` (gh CLI is preinstalled on ubuntu runners).
- Steps: \`actions/checkout@v4\` → a single script step that:
  1. **Detects the version** from \`kodemirror-bom/build.gradle.kts\` (\`grep ... | sed ...\`) — never hardcoded; echoes it.
  2. **Skips pre-releases** — \`case\` on \`*-SNAPSHOT|*-RC*\` exits 0 cleanly with a printed reason.
  3. **Idempotent** — \`gh release view\` first; if the release for the tag already exists, skip (re-run safe).
  4. **Extracts the matching CHANGELOG section** via awk (from \`## [VERSION]\` to just before the next \`## [\`) into \`/tmp/relnotes.md\`; falls back to \`--generate-notes\` if empty.
  5. \`gh release create "v\$VERSION" --repo Monkopedia/kodemirror --title "v\$VERSION" --notes-file ... --latest --verify-tag\` — \`--verify-tag\` fails if the tag wasn't pushed.

### CLAUDE.md — Release process section
Added a concise "### Release process" subsection under "## Workflow" documenting: version-bump PR (drop \`-SNAPSHOT\` in the two version files + finalize CHANGELOG) → tag \`vX.Y.Z\` and push → dispatch \`deploy.yml\` (publishes to Central + auto-creates the GitHub Release) → verify BOM on \`repo1.maven.org\` (10–35 min) → notify downstream. Notes that publishing is irreversible and a partial publish is a burned version.

## Validation (no real publish triggered)
- **actionlint 1.7.12**: clean on \`.github/workflows/deploy.yml\` (no errors). YAML also parses via PyYAML; jobs = build/deploy/release with correct \`needs\`/permissions/env.
- **Version detection** against current main (\`0.3.1\`): \`VERSION=[0.3.1]\` ✓.
- **CHANGELOG extraction** verified against real file for both \`0.3.1\` (single section) and \`0.3.0\` (correctly stops at the next \`## [\` header).
- **SNAPSHOT/RC skip**: \`0.3.1 -> RELEASE\`, \`0.3.2-SNAPSHOT -> SKIP\`, \`0.4.0-RC1 -> SKIP\` ✓.
- \`./gradlew help\` still configures.

The next release (0.3.2) will exercise this end-to-end. No CHANGELOG entry (infra/tooling, matches precedent); no version bump.

Fixes #98

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -57,3 +57,53 @@ jobs:
         -Psigning.gnupg.keyName='${{ secrets.OSSRH_GPG_SECRET_KEY_ID }}'
         -PmavenCentralUsername='${{ secrets.OSSRH_USERNAME }}'
         -PmavenCentralPassword='${{ secrets.OSSRH_TOKEN }}'
+
+  # After ALL Maven Central matrix publishes succeed, cut a matching GitHub Release.
+  release:
+    needs: deploy
+    runs-on: ubuntu-latest
+    # Top-level perms are read-only; this job needs write to create a release.
+    permissions:
+      contents: write
+    env:
+      # gh CLI is preinstalled on ubuntu runners; authenticate via the Actions token.
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v4
+    - name: Create GitHub Release
+      run: |
+        set -euo pipefail
+        # Detect the published version from the build files (never hardcode).
+        VERSION=$(grep -E '^version = ' kodemirror-bom/build.gradle.kts | sed -E 's/.*"([^"]+)".*/\1/')
+        echo "Detected version: $VERSION"
+
+        # Only final releases get a GitHub Release; skip pre-release versions.
+        case "$VERSION" in
+          *-SNAPSHOT|*-RC*)
+            echo "Version $VERSION is a pre-release (SNAPSHOT/RC); skipping GitHub Release."
+            exit 0
+            ;;
+        esac
+
+        TAG="v$VERSION"
+
+        # Idempotent: if a release already exists for this tag, skip (re-run safe).
+        if gh release view "$TAG" --repo Monkopedia/kodemirror >/dev/null 2>&1; then
+          echo "Release $TAG already exists; nothing to do."
+          exit 0
+        fi
+
+        # Extract this version's CHANGELOG section: from the "## [VERSION] - date"
+        # header up to (but not including) the next "## [" header.
+        awk -v v="## [$VERSION]" 'index($0,v)==1{f=1;print;next} /^## \[/{f=0} f' CHANGELOG.md > /tmp/relnotes.md
+
+        if [ -s /tmp/relnotes.md ]; then
+          echo "Using CHANGELOG section for release notes:"
+          cat /tmp/relnotes.md
+          gh release create "$TAG" --repo Monkopedia/kodemirror \
+            --title "$TAG" --notes-file /tmp/relnotes.md --latest --verify-tag
+        else
+          echo "No CHANGELOG section found for $VERSION; falling back to --generate-notes."
+          gh release create "$TAG" --repo Monkopedia/kodemirror \
+            --title "$TAG" --generate-notes --latest --verify-tag
+        fi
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. All scrutiny points pass:

**Version detection** (grep+sed on `kodemirror-bom/build.gradle.kts`): re-ran locally → `0.3.1` ✓

**CHANGELOG awk extraction**: re-ran for 0.3.1 (gets exactly the right section) and 0.3.0 (stops cleanly before the 0.2.0 header). The `index(/bin/zsh,v)==1` prefix match correctly treats `[` as a literal (not regex), so no quoting issue. Falls back to `--generate-notes` if section is empty. ✓

**SNAPSHOT/RC skip**: re-ran the `case` pattern — `0.3.1 → RELEASE`, `0.3.2-SNAPSHOT → SKIP`, `0.4.0-RC1 → SKIP`, `0.4.0-RC10 → SKIP` ✓. The `-alpha`/`-beta` gap is a non-issue; this project has never used those suffixes.

**Gating**: `needs: deploy` where `deploy` is a 10-item matrix — all 10 publishes must succeed before `release` runs ✓

**Permissions**: `permissions: contents: write` on the `release` job overrides the top-level `contents: read` correctly; `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` present ✓

**Idempotency**: `gh release view` check before creating — re-run safe ✓

**`--verify-tag`**: correctly fails if the tag wasn't pushed ✓

**Scope**: build/deploy jobs unchanged; only the new `release` job is added ✓

**CLAUDE.md**: accurately documents the flow (bump→tag→deploy.yml→verify→notify); matches implementation ✓

**YAML**: parses valid (PyYAML); actionlint 1.7.12 clean per coder (not available locally to re-run) ✓

**Quality**: well-commented, minimal, idiomatic shell with `set -euo pipefail`, no dead code, no scope creep ✓

Approving. Merging with --auto once CI passes.

PR #102, opened 2026-06-02
Fix Maven Central publish reliability (single atomic deployment)

### LINKED ISSUE #101: deploy.yml: BOM (and reliability) — fix fragile per-target publish so the BOM always publishes
On the 0.3.2 publish, every module artifact published but `kodemirror-bom` did NOT (had to be published manually from the Central Portal). The BOM publish job log showed: `> Task :kodemirror-bom:publishMavenPublicationToMavenCentralRepository` → `Skipping deployment validation!` → `BUILD SUCCESSFUL` — i.e. it went GREEN without the deployment actually releasing. The BOM is the only non-KMP publication (a plain `MavenPublication`/BOM). This regressed with the vanniktech 0.36.0 upgrade (0.36.0 fixed the 900s timeout but its validate/release flow leaves the BOM "green but not published").

Root fragility: `.github/workflows/deploy.yml` publishes via a **10-way matrix** (`publishXPublicationToMavenCentralRepository` per target across ubuntu+macos runners) — so EACH job is a separate `./gradlew` invocation creating a SEPARATE Central Portal deployment for the same `com.monkopedia.kodemirror:<version>`. That's the source of: the 900s timeout (0.3.1), partial publishes, and now the BOM-not-released (0.3.2). One release should be ONE atomic Central Portal deployment.

### Investigate + fix
- Determine why the BOM publication "Skipping deployment validation!" and didn't release under vanniktech 0.36.0 (read the 0.36.0 Central-Portal publishing behavior / docs / source). Is it one-deployment-per-Gradle-invocation, and does the aggregate task bundle ALL publications (incl. the BOM) into ONE deployment?
- Strongly consider replacing the matrix with a **single publish job on `macos-latest`** running the vanniktech aggregate (`./gradlew publishToMavenCentral` or `publishAllPublicationsToMavenCentralRepository`) in ONE invocation → ONE deployment containing all artifacts (jvm, android, wasmJs, macos*, ios*, AND the BOM) → validated + auto-released atomically. The project's targets are jvm/androidTarget/wasmJs/macosArm64/macosX64/iosArm64/iosSimulatorArm64/iosX64 — ALL buildable on a macOS runner (macOS GitHub runners include the Android SDK; there are no linux/mingw native targets), so a single-runner publish is feasible and removes the cross-runner/per-deployment fragility.
- Keep the auto-`gh release create` job (from #98) wired to run after the (now single) publish job succeeds.
- If a single-runner aggregate isn't viable (e.g. vanniktech creates one deployment per module regardless, or a target genuinely can't build on macOS), report the constraint + the next-best reliable option (e.g. ensure the BOM deployment validates+releases explicitly).

### Caveats
- Can't be fully e2e-tested without a real publish; verify the workflow YAML + that the chosen Gradle task graph is correct (e.g. `./gradlew publishToMavenCentral --dry-run` resolves; confirm the BOM publication is included). The NEXT release exercises it for real.
- This is high-stakes release infra — if the right approach needs a judgment call, surface it rather than guessing.

monkopedia-coder:
## Root cause (#101)

`deploy.yml` published via a **10-way matrix**, each entry running a separate `./gradlew publishX…ToMavenCentralRepository` on a separate runner. The vanniktech plugin's `MavenCentralBuildService` is a **build-scoped shared service** (`registerIfAbsent`, `maxParallelUsages=1`): every module's publish task registers its artifacts into the service during the build, and at end-of-build (`close()`) the service uploads everything as **one** Central Portal deployment. Because each matrix entry was a *separate Gradle invocation*, each produced a **separate deployment** for the same `com.monkopedia.kodemirror:<version>` — fragile, and the source of the 0.3.1 validation timeout / partial publish.

On 0.3.2 the `kodemirror-bom` deployment went GREEN but never released — its job logged **"Skipping deployment validation!"**. That string is logged in `MavenCentralBuildService.runEndOfBuildActions` when the build has an `Upload` action but no `Validate` action. `Validate`/`Publish` actions are added only by `EnableAutomaticMavenCentralPublishingTask`, which a publish task depends on only when `automaticRelease = true`. The BOM uses bare `publishToMavenCentral()` (no `automaticRelease`), and in the matrix it ran in its own isolated invocation — so no module added a Validate action and the BOM uploaded as `USER_MANAGED` and sat un-released (the user published it manually).

## Investigation of vanniktech 0.36.0 deployment grouping

Confirmed from the 0.36.0 source (`plugin/.../central/MavenCentralBuildService.kt`):

- All modules in **one Gradle invocation** call `registerProject()` on the shared service; `close()` runs once and zips **all** registered projects' artifacts into **one** deployment.
- Deployment naming: 1 coordinate → `group-artifactId-version`; multiple coordinates with the **same group+version** → `group-version` (our case → `com.monkopedia.kodemirror-<version>`); otherwise `group-timestamp`.
- If **any** module added `automaticRelease=true`'s `Publish`+`Validate` actions, the **whole** combined deployment (incl. the BOM's artifacts) is validated + auto-released atomically.

So one aggregate invocation = one atomic deployment for the whole release. This is the clean fix.

## The fix

- Replace the 10-way matrix with a **single `deploy` job on `macos-latest`** running `./gradlew publishToMavenCentral` once. macOS can build every target (apple-native needs macOS; jvm/android/wasmJs build anywhere; `macos-latest` ships the Android SDK) — so one runner publishes all targets + the BOM into one atomic deployment. Same `-Psigning.gnupg.* / -PmavenCentralUsername/Password` flags as before.
- Set `automaticRelease = true` on `kodemirror-bom` too, so the BOM module is self-sufficient (defense in depth) rather than relying on a sibling module to add the Validate action.
- The `release` job (auto `gh release create`) is unchanged except `needs: deploy` (now the single job instead of the matrix). It still runs after the publish succeeds.
- Updated the release-process note in `CLAUDE.md` (no longer a matrix).

## Validation (no real publish triggered)

- `actionlint 1.7.12` on `deploy.yml`: clean.
- `./gradlew publishToMavenCentral --dry-run` (with `ANDROID_HOME` set): **BUILD SUCCESSFUL**, graph resolves with no unresolved references. **58** module `publishToMavenCentral` leaves in the graph, including `:kodemirror-bom:publishMavenPublicationToMavenCentralRepository`. After the BOM change, **every** publishing module — including `:kodemirror-bom` — now has `enableAutomaticMavenCentralPublishing` in the graph (previously the BOM was the only one without it). Published-module set is unchanged from the old matrix.
- `./gradlew help`: configures cleanly on Gradle 9.

The real test is the next release dispatch.

Fixes #101

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,36 +8,31 @@ permissions:
 jobs:
   build:
     uses: ./.github/workflows/ci.yml
+
+  # Publish EVERYTHING in a single Gradle invocation on one runner.
+  #
+  # The vanniktech maven-publish plugin uses a build-scoped shared service
+  # (MavenCentralBuildService, maxParallelUsages=1) that collects every module's
+  # artifacts registered during the build and, at the end of the build, uploads
+  # them as ONE Central Portal deployment. Because all modules share the
+  # group+version (com.monkopedia.kodemirror:<version>), that single deployment
+  # bundles all targets (jvm, android, wasmJs, macos*, ios*) AND the BOM, and is
+  # validated + auto-released atomically (automaticRelease = true in the
+  # convention plugin adds the Publish/Validate end-of-build actions).
+  #
+  # This replaces the previous 10-way matrix, where each entry ran in a SEPARATE
+  # Gradle invocation and therefore created a SEPARATE deployment. That caused the
+  # validation timeout / partial publishes, and on 0.3.2 left the BOM deployment
+  # un-released ("Skipping deployment validation!") because the BOM module is the
+  # only one without automaticRelease, so its isolated invocation never added a
+  # Validate action.
+  #
+  # Runs on macOS because the apple-native targets (macos*/ios*) require a macOS
+  # toolchain; jvm/android/wasmJs build on any platform and macos-latest ships the
+  # Android SDK, so one runner can build and publish every target this project has.
   deploy:
     needs: build
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          # BOM
-          - target: :kodemirror-bom:publishMavenPublicationToMavenCentralRepository
-            os: ubuntu-latest
-          # Ubuntu targets
-          - target: publishJvmPublicationToMavenCentralRepository
-            os: ubuntu-latest
-          - target: publishAndroidReleasePublicationToMavenCentralRepository
-            os: ubuntu-latest
-          - target: publishWasmJsPublicationToMavenCentralRepository
-            os: ubuntu-latest
-          - target: publishKotlinMultiplatformPublicationToMavenCentralRepository
-            os: ubuntu-latest
-          # macOS targets (need macOS runner for native compilation)
-          - target: publishIosArm64PublicationToMavenCentralRepository
-            os: macos-latest
-          - target: publishIosSimulatorArm64PublicationToMavenCentralRepository
-            os: macos-latest
-          - target: publishIosX64PublicationToMavenCentralRepository
-            os: macos-latest
-          - target: publishMacosArm64PublicationToMavenCentralRepository
-            os: macos-latest
-          - target: publishMacosX64PublicationToMavenCentralRepository
-            os: macos-latest
-    runs-on: ${{ matrix.os }}
+    runs-on: macos-latest
     steps:
     - uses: actions/checkout@v4
     - name: Import GPG key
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Tier-1 approval — config-only release-path fix, correctness verified:

**Single-job correctness:** deploy.yml now has a single `deploy` job on `macos-latest` running `./gradlew publishToMavenCentral` once. All credential/signing flags (-Psigning.gnupg.passphrase/keyName, -PmavenCentralUsername/Password) are identical to the old matrix. GPG import + Java 21 setup steps preserved.

**BOM included + auto-released:** Dry-run (JAVA_HOME=java-21, ANDROID_HOME set) produced BUILD SUCCESSFUL with 58 `publishToMavenCentral` leaves. `:kodemirror-bom:publishMavenPublicationToMavenCentralRepository` and `:kodemirror-bom:enableAutomaticMavenCentralPublishing` both appear in the graph — the BOM is now self-sufficient with `automaticRelease = true`, eliminating the 0.3.2 'Skipping deployment validation!' bug.

**Release job:** `needs: deploy` (the single job), unchanged otherwise. `release` job is intact.

**Artifact set:** Dry-run confirms the same module/platform set the 10-way matrix had (jvm, android, wasmJs, macos*, ios*, BOM, KotlinMultiplatform variants). No regressions.

**YAML:** Syntactically valid (Python yaml.safe_load clean); actionlint not locally installed but agent reported 1.7.12 clean on the same file.

**Quality:** Minimal, coherent change; CLAUDE.md note accurately reflects the new single-runner approach. No scope creep.

**CI:** All three required checks passed — check (10m38s), check-ios (12m52s), check-macos (6m17s). Waited for real GitHub CI completion before approving.

PR #104, opened 2026-06-03
Dispatch updateListener facet on every transaction

### LINKED ISSUE #103: updateListener facet never dispatched — onChange()/onSelection() public extensions silently never fire
Reported by downstream consumer (konstructor) on 0.3.2; confirmed in code.

`EditorSession.updateListener` (`view/.../EditorSession.kt:89`) is a documented facet ("Facet that registers update listeners called after every transaction"), and two PUBLIC extension functions are built on it:
- `onChange(callback: (String) -> Unit)` (EditorSession.kt:198)
- `onSelection(callback: (EditorSelection) -> Unit)` (EditorSession.kt:216)
both with KDoc usage examples.

**Bug:** `EditorSessionImpl.dispatchTransaction(tr)` (EditorSessionImpl.kt:119-145) builds a `ViewUpdate`, calls `pluginHost.update(update)` / `syncToState(...)`, handles `scrollIntoView`, and calls the constructor `onUpdate(tr)` callback — but it NEVER reads/invokes the `updateListener` facet listeners (`tr.state.facet(EditorSession.updateListener)`). Confirmed via grep: the facet is only ever WRITTEN (`.of {...}` in onChange/onSelection) and never READ anywhere (not in dispatchTransaction, not in KodeMirror.kt). So `EditorSession.updateListener.of {}`, `onChange {}`, and `onSelection {}` all silently do nothing.

The constructor callbacks `onUpdate`/`onChange` (params on `EditorSession(...)`) DO fire (via `onUpdate(tr)` at line 143) — which is why konstructor's `EditorSession(initialState, onUpdate = {...})` workaround works. Only the FACET-based hooks are dead.

Not deprecated/intentional — the KDoc + documented public helpers clearly intend it to fire.

### Fix
In `dispatchTransaction`, after building `update`, invoke the registered listeners: `tr.state.facet(EditorSession.updateListener).forEach { it(update) }` (place it consistently with plugin update ordering — after `pluginHost.update`/`syncToState` so listeners see post-update state; match how upstream CM6 fires updateListeners). Then `onChange`/`onSelection` work.

### Test
Add a `:view` jvmTest: register `onChange {}` / `onSelection {}` (and a raw `updateListener.of {}`) via extensions, dispatch a doc-change transaction and a selection-only transaction, assert the callbacks fire with the right payload (docChanged/selectionSet). Must fail pre-fix (no callback) and pass post-fix.

Reporter notes it's non-blocking for them (they use the onUpdate callback), but it's a trap for any consumer reaching for the natural facet/onChange/onSelection API.

monkopedia-coder:
## Root cause

`EditorSession.updateListener` is a documented facet ("called after every transaction"), and the public `onChange(callback)` / `onSelection(callback)` extensions are built on it via `EditorSession.updateListener.of { ... }`. But `EditorSessionImpl.dispatchTransaction(tr)` never read that facet — it only invoked the constructor `onUpdate(tr)` callback. As a result the facet, `onChange {}`, and `onSelection {}` were all registered but never dispatched: they silently never fired.

## Fix

In `dispatchTransaction`, after `pluginHost?.update(update)` / `pluginHost?.syncToState(...)` (so listeners observe post-update state, matching CM6 where updateListeners fire after the view updates) and before the existing `onUpdate(tr)` call, invoke the registered listeners:

```kotlin
tr.state.facet(EditorSession.updateListener).forEach { it(update) }
```

Reading an unset facet yields an empty list, so this is a safe no-op when no listeners are registered. The constructor `onUpdate(tr)` callback is unchanged and still fires (after the listeners). Minimal change; no public API change (the facet/extensions already exist), so `apiCheck` stays green with no `apiDump`.

## Regression test

New `view/src/commonTest/.../UpdateListenerTest.kt` (plain non-Compose — constructs an `EditorSession` and drives `dispatch(...)` directly, no `KodeMirror` mount needed):
- Registers a raw `EditorSession.updateListener.of {}`, plus `onChange {}` and `onSelection {}`.
- Doc-change transaction: asserts the raw listener fired with `docChanged == true` and `onChange` fired with the new doc text.
- Selection-only transaction (cursor move): asserts `onSelection` fired with the moved selection, the raw listener saw `selectionSet && !docChanged`, and `onChange` did NOT fire — confirming the docChanged/selectionSet gating.
- A no-listeners-registered case confirms reading the unset facet is a no-op (no throw).

**Pre-fix:** with the `dispatchTransaction` change stashed, the two listener-firing tests FAIL (AssertionError — callbacks never fire). **Post-fix:** all `:view:jvmTest` pass.

## Verification (JDK 21)

- `:view:jvmTest` — green (incl. the new test)
- `:view:compileKotlinWasmJs` — green
- `:view:apiCheck` — green (no apiDump; no public API change)

Fixes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- updateListener facet (and the onChange/onSelection extensions built on it) now fire on every transaction; previously they were registered but never dispatched (#103)
+
 ## [0.3.2] - 2026-06-02
 
 ### Fixed
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/EditorSessionImpl.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/EditorSessionImpl.kt
@@ -131,6 +131,11 @@ internal class EditorSessionImpl(
         )
         pluginHost?.update(update)
         pluginHost?.syncToState(tr.state, oldState)
+        // Notify registered update listeners after the view/plugins have updated,
+        // matching CM6 where updateListeners fire after the view updates. Reading
+        // an unset facet yields an empty list, so this is a no-op when no listeners
+        // are registered. Drives the onChange/onSelection extensions (#103).
+        tr.state.facet(EditorSession.updateListener).forEach { it(update) }
         // Honor scrollIntoView: queue a request for the composable to reveal
         // the primary selection head. This drives caret-reveal on cursor moves
         // (#33) and the search/jump reveal in vim `n`/`N` (#58).
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/UpdateListenerTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/UpdateListenerTest.kt
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.view
+
+import com.monkopedia.kodemirror.state.ChangeSpec
+import com.monkopedia.kodemirror.state.DocPos
+import com.monkopedia.kodemirror.state.EditorSelection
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Tier-1 approval — internal bug fix, no public API change.

**Fix correctness:** The one-line addition `tr.state.facet(EditorSession.updateListener).forEach { it(update) }` is placed correctly: after `pluginHost?.update(update)` / `pluginHost?.syncToState(...)` (listeners see post-update state) and before `onUpdate(tr)` (constructor callback still fires unchanged). Reading an unregistered facet yields an empty list — no throw, safe no-op.

**Regression test discriminates:** `UpdateListenerTest` exercises the raw `updateListener.of{}` facet, `onChange{}`, and `onSelection{}`. The doc-change test confirms `docChanged == true` and the new text; the selection-only test confirms `selectionSet && !docChanged` and that `onChange` does NOT fire; the no-listeners test confirms no throw. All three tests PASS post-fix (confirmed via `:view:jvmTest`). Pre-fix, the two listener-firing tests would fail (facet was never read in `dispatchTransaction`).

**Independent re-verification (JDK 21):**
- `:view:jvmTest` — BUILD SUCCESSFUL (3/3 UpdateListenerTest cases pass, existing EditorSession tests unchanged)
- `:view:compileKotlinWasmJs` — BUILD SUCCESSFUL
- `:view:compileDebugKotlinAndroid` — BUILD SUCCESSFUL
- `:view:apiCheck` — BUILD SUCCESSFUL (no apiDump; no public API change)

**CI (real GitHub checks):**
- `check`: pass (11m 6s)
- `check-macos`: pass (11m 57s)
- `check-ios`: pass (13m 12s)

All three required checks completed and passed. Merging.

Issue #112, opened 2026-06-05
Support lazy completion (completionItem/resolve) for items with deferred insertion text


monkopedia-coder:
## Context
Surfaced while fixing #109. Some language servers (e.g. **kotlin-lsp 262.x**) use **lazy completion**: the initial `textDocument/completion` items carry an **empty `textEdit.newText`** (and null `insertText`/`textEditText`), with the real insertion text + `additionalTextEdits` (e.g. auto-imports) delivered later via **`completionItem/resolve`**.

`mapCompletionItem`/`insertionText()` previously returned that empty string, so the displayed AND inserted text were blank.

## Stopgap already shipped (#109)
- `insertionText()` now treats an empty `newText`/`insertText`/`textEditText` as absent and falls through to `item.label`.
- `mapCompletionItem` sets `displayLabel = item.label` so the popup always shows the human label.

This makes display + basic insertion work (the member name IS usually the insertion), but it does NOT fetch the server's real insertion/edits.

## This issue — proper fix
Implement `completionItem/resolve`: when a completion is selected (or lazily for the focused item), call `completionItem/resolve` to obtain the real `textEdit`/`additionalTextEdits`/`documentation`, then apply those. Needed for correctness when the insertion differs from the label — most importantly **auto-imports** (additionalTextEdits) and snippet/edit insertions that the lazy item omits.

- Check `serverCapabilities.completionProvider.resolveProvider` before resolving.
- Resolve on apply (cheap) and/or on selection for doc popups.
- Mirrors upstream `@codemirror/lsp-client` resolve handling.

Medium priority — completion is usable without it, but auto-imports won't apply until this lands.

Issue #111, opened 2026-06-05
Restore completion-label prefix-match bold highlighting (wasmJs AnnotatedString render)


monkopedia-coder:
## Context
While fixing #109 (LSP completion popup on wasmJs), the completion label `BasicText` was rendered from an **`AnnotatedString`** (`buildHighlightedLabel`, which bolds the matched prefix). On **Compose-wasmJs** that label did not paint at all — only the plain-`String` type icon rendered, leaving every row as an icon with blank space. (Confirmed live: data + color + width all ruled out; the only difference between the rendering icon and the non-rendering label was `String` vs `AnnotatedString`.)

To unblock #109 the label was switched to a plain `String` (`displayLabel ?: label`), **dropping the prefix-match bold styling**. Labels now render correctly.

## This issue
Restore the matched-prefix **bold highlighting** in the completion popup once the underlying Compose-wasmJs `BasicText(AnnotatedString)` non-render is understood/worked-around. Options to investigate:
- Whether a newer Compose Multiplatform version fixes `BasicText(AnnotatedString)` painting on wasmJs.
- An alternative styling approach (e.g. `Text` with `AnnotatedString`, or per-span layout) that does paint on wasmJs.
- Whether it was a measurement-collapse (intrinsic width 0) vs a paint failure — give the label a bounded width/`weight` and re-test with `AnnotatedString`.

The `buildHighlightedLabel` helper + `item.highlighted` data were removed from `autocomplete/.../View.kt`'s `CompletionList`; restoring will re-add them. Low priority (cosmetic) — desktop/JVM was unaffected (it rendered the AnnotatedString fine).

monkopedia-coder:
Decision (Jason): **work it.** Deliverable includes a real wasmJs screenshot test (via the `gap-analysis/` Playwright + Xvfb harness against the live showcase) proving the `AnnotatedString` completion label paints — plus a JVM Roborazzi test locking the intended bold-prefix appearance. The wasmJs-paint test is the load-bearing one (the regression was wasmJs-only).

PR #113, opened 2026-06-05
Fix LSP completion on wasmJs

### LINKED ISSUE #109: LSP completion trigger character inserted twice on wasmJs
## Symptom
With `languageServerSupport(client, uri, …)` active against a server that advertises a completion `triggerCharacter` (e.g. kotlin-lsp advertises `.`), typing that trigger character inserts it **twice** (`.` → `..`). Only the trigger char(s) double — ordinary keystrokes are fine. The char is typed normally **and** re-applied at the moment completion fires.

Reported by konstructor (downstream consumer, kodemirror 0.3.2, wasmJs); surfaced live once a backend fix made completion actually fire. Affects 0.3.2/0.3.3 (completion path unchanged across them).

## Reproduction
The showcase `LspDemo` already has the same setup: `StubLanguageServer` advertises `.` as a completion trigger char (`CompletionOptions(triggerCharacters = listOf("."))`) and returns completion items, and the editor uses `showcaseSetup` (= `basicSetup`, which bundles `autocompletion()`) **plus** `languageServerSupport` (which registers a second `autocompletion()`). Typing `.` in that demo on wasmJs should reproduce.

## Leading hypothesis (traced statically, not yet reproduced)
On wasmJs a printable keystroke flows through three paths in `view/.../KodeMirror.kt` + `InputHandling.kt`:
1. a document-level capture key handler (`platformRegisterKeyHandler` → `handleRawKeyEvent`) that inserts the char and sets `keyProcessedByCallback`;
2. `onPreviewKeyEvent`, which reads `keyProcessedByCallback` and sets `suppressInput`;
3. the hidden `BasicTextField`'s `onValueChange` (the browser input event), gated by `suppressInput`.

Two **single-shot booleans** bridge these, assuming exactly one of each callback per keystroke. A trigger char is the only printable char that makes the **completion popup appear**; the popup's recomposition is the prime suspect for firing an *extra* `onValueChange` that slips past the single-shot `suppressInput` guard → the char is inserted a second time. Likely compounded by the **dual `autocompletion()` registration** (basicSetup + languageServerSupport → two CompletionPlugins / extra recompositions).

## Fix direction
Make the input-suppression robust against an extra completion-driven `onValueChange` (don't rely on a one-shot boolean — e.g. track the expected echoed text / count, or ignore `onValueChange` deltas already applied by the key paths). Confirm via a wasmJs repro (showcase LspDemo + Playwright gap-analysis harness) before/after.

## Notes
- Interim mitigation for consumers: register only ONE completion source (drop basicSetup's bundled `autocompletion()` when using `languageServerSupport`).
- Target: 0.3.4.

monkopedia-coder:
Makes language-server completion work on the **wasmJs** target (the popup never opened / rendered there). Verified end-to-end against a live kotlin-lsp engine via the konstructor consumer's Playwright canvas-screenshot harness (popup opens on `.`, renders icon + real labels, inserts correctly).

Fixes #109.

> **No version bump / release.** Version stays `0.3.3`; the CHANGELOG entries land under `[Unreleased]`. More LSP work is planned before the next version is cut, so this PR is just the fix — a 0.3.x release is a separate, later PR.

## The stack of bugs (all #109)
1. **Popup never opened on wasmJs.** `serverCompletion` registered its source via `asyncCompletionSource`, whose wasmJs `actual` throws. Added **`CompletionConfig.asyncOverride`** — suspend sources the framework launches on the editor's `coroutineScope` and dispatches when they resolve (no blocking bridge); `serverCompletion` now uses it. Drives both typed-trigger and Ctrl+Space.
2. **Two `autocompletion()` extensions shadowed each other.** `basicSetup` bundles `autocompletion()`, and `languageServerSupport` adds another; the `completionConfig` facet combine was `values.lastOrNull()`, so one config won wholesale and the LSP source vanished (0 sources). The combine now **merges** `override`/`asyncOverride` across all configs.
3. **Popup rendered as a blank/thin strip.** `CompletionList` used a `LazyColumn` (no cross-axis wrap-content) and width-less `BasicText` labels, which collapse to ~0 on Compose-wasmJs. Now a plain `Column` + bounded width, `fillMaxWidth` rows, and a `weight(1f)` + single-line-ellipsis label so the label gets a concrete width and paints.
4. **Labels (and insertion) blank for lazy-completion servers.** kotlin-lsp sends items with an **empty `textEdit.newText`** (real edits arrive via `completionItem/resolve`). `insertionText()` returned `""`, so the displayed *and* inserted text were empty. It now treats empty `newText`/`insertText`/`textEditText` as absent and falls back to `item.label`; `mapCompletionItem` sets `displayLabel = item.label`.
5. **Trigger char inserted twice on wasmJs.** The hidden text field's input-echo suppression was a one-shot flag; a completion-popup recomposition fires an extra `onValueChange` that slipped past it. The flag is now reset per keydown (document-level handler) and stays armed across multiple echoes.

## Also
- `apiDump`: `CompletionConfig` gains the public `asyncOverride` field (the only API change).
- Skip artifact **signing for `-SNAPSHOT`** versions, so local `publishToMavenLocal` works without a GPG key (this enabled the local-snapshot verification loop used to land #109).

## Deferred (filed)
- **#111** — restore prefix-match **bold** highlighting (was an `AnnotatedString`; dropped for now).
- **#112** — proper lazy completion via `completionItem/resolve` (auto-imports / `additionalTextEdits` won't apply until then; the label fallback handles the common case).

## Verification (local, Kotlin 2.4.0)
`:autocomplete:jvmTest`, `:lsp-client:jvmTest`, `:view:jvmTest` ✅ · wasmJs compile (autocomplete/lsp-client/view) ✅ · `apiCheck` ✅ · `spotlessCheck` ✅. End-to-end completion behavior screenshot-verified on the live engine by the konstructor consumer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Added
+- `CompletionConfig.asyncOverride`: suspend completion sources that the autocomplete framework launches on the editor's coroutine scope and dispatches when they resolve. This is the wasmJs-safe way to drive an async source (e.g. a language server) — the previous `asyncCompletionSource` blocking bridge throws on wasmJs (#109).
+
+### Fixed
+- **LSP completion now works on wasmJs (#109).** Several issues stacked: (a) the completion source was registered via `asyncCompletionSource`, whose wasmJs actual throws, so the popup never opened — `serverCompletion` now registers via the new `asyncOverride` path; (b) two `autocompletion()` extensions (e.g. `basicSetup` + `languageServerSupport`) no longer shadow each other — the `completionConfig` facet now merges completion sources across all configs instead of taking only the last; (c) the completion popup is rendered with a plain `Column` + bounded width and a `weight`-sized label so it lays out and paints on Compose-wasmJs (a `LazyColumn`/width-less `BasicText` collapsed to nothing); (d) completion labels display (and insert) correctly for servers that use lazy completion — `insertionText()` now treats an empty `textEdit.newText`/`insertText` as absent and falls back to the item label, and `mapCompletionItem` sets `displayLabel` to the item label. (Bold prefix-match highlighting is temporarily dropped on this path — #111; `completionItem/resolve` for auto-imports is tracked in #112.)
+- Completion trigger characters (e.g. `.`) are no longer inserted twice on wasmJs: the hidden text field's input-echo suppression is now reset per keydown and survives the extra `onValueChange` that a completion-popup recomposition can fire (#109).
+
 ### Changed
-- Bumped the lsp dependency to 1.2.0 (was 1.0.1). 1.2.0 tightens several `LanguageServer`/`LanguageClient` return types to be nullable, matching the LSP spec's optional results (`textDocument/hover`, `textDocument/formatting`, `textDocument/references`, `textDocument/rename`, `workspace/workspaceFolders`). The client now treats a null result as "no result" (no-op / empty), preserving prior behaviour. No public API change.
+- Bumped the lsp dependency to 1.2.0 (was 1.0.1) (#108). 1.2.0 tightens several `LanguageServer`/`LanguageClient` return types to be nullable, matching the LSP spec's optional results (`textDocument/hover`, `textDocument/formatting`, `textDocument/references`, `textDocument/rename`, `workspace/workspaceFolders`). The client now treats a null result as "no result" (no-op / empty), preserving prior behaviour.
+- Release artifacts are signed only for non-SNAPSHOT versions, so a local `publishToMavenLocal` of a `-SNAPSHOT` build no longer requires a GPG key.
 
 ## [0.3.3] - 2026-06-04
 
--- a/autocomplete/api/android/autocomplete.api
+++ b/autocomplete/api/android/autocomplete.api
@@ -91,8 +91,8 @@ public final class com/monkopedia/kodemirror/autocomplete/CompletionApplyContext
 public final class com/monkopedia/kodemirror/autocomplete/CompletionConfig {
 	public static final field $stable I
 	public fun <init> ()V
-	public fun <init> (ZZZIZZLjava/util/List;)V
-	public synthetic fun <init> (ZZZIZZLjava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (ZZZIZZLjava/util/List;Ljava/util/List;)V
+	public synthetic fun <init> (ZZZIZZLjava/util/List;Ljava/util/List;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Z
 	public final fun component2 ()Z
 	public final fun component3 ()Z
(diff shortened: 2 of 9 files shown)


monkopedia-reviewer (commented):
@Monkopedia I've reviewed this PR to merge standard — all the bot-verifiable parts are sound:

**What I reviewed and confirmed:**
- **Fix 1 (asyncOverride / popup never opened):** `asyncCompletionSource` removed; `serverCompletion` now registers via `CompletionConfig.asyncOverride`. The `triggerCompletion` coroutine path launches on `view.coroutineScope`, dispatches `startCompletionEffect` on result, and `closeCompletionEffect` on empty-explicit — correct structured-concurrency usage, uniform on jvm/wasmJs/native.
- **Fix 2 (facet merge):** `completionConfig` combine now merges `override` + `asyncOverride` from all contributing configs using `flatMap`/`ifEmpty`; scalar settings taken from `values.last()`. Correctly coexists multiple `autocompletion()` calls.
- **Fix 3 (popup render):** `LazyColumn` → `Column` + `verticalScroll`, `widthIn(200..360)`, `fillMaxWidth` rows, `weight(1f)` label with `TextOverflow.Ellipsis`. Well-reasoned fix for the Compose-wasmJs zero-intrinsic-width problem; cap at `heightIn(240)` is reasonable.
- **Fix 4 (insertionText / displayLabel):** `ifEmpty { null }` chain on `newText`/`insertText`/`textEditText` before falling back to `label`. Both `Completion` return paths in `mapCompletionItem` set `displayLabel = item.label`. Correct.
- **Fix 5 (double-insert suppression):** `suppressInput` hoisted to `KodeMirror`, reset per keydown in the document-level handler, not cleared in `onValueChange` (sticky). Correct fix for multi-echo per keystroke.
- **API dump:** `CompletionConfig` gains `asyncOverride: List<SuspendCompletionSource>?` — only intended new public surface.
- **SNAPSHOT signing skip:** `if (!version.toString().endsWith("SNAPSHOT"))` guard in both `kodemirror.library.gradle.kts` and `kodemirror-bom/build.gradle.kts`. Correct and consistent.
- **CI:** all three required checks green — `check` (11m41s), `check-macos` (8m19s), `check-ios` (9m41s).
- **Local re-verification:** `:autocomplete:jvmTest`, `:lsp-client:jvmTest`, `:view:jvmTest`, wasmJs compile (all three), `apiCheck`, `spotlessCheck` — all green.
- **Deferred work filed correctly:** #111 (bold highlighting) and #112 (completionItem/resolve lazy completion) — the label fallback is a correct minimal fix pending those.

**The one slice needing your judgment:**
The version bump `0.3.3 → 0.3.4` (clean release) lands on main with this merge. Per policy, a final `x.y.z` bump is yours to approve. The actual publish to Maven Central is still gated on your separate tag + `deploy.yml` dispatch (the PR body already says this). No code concern — just confirming you want this version number on main now.

monkopedia-reviewer (approved):
Re-review after force-push (new HEAD 01ac03d): confirmed version bump reverted — version reads 0.3.3 in both kodemirror.library.gradle.kts and kodemirror-bom/build.gradle.kts, CHANGELOG entries are under [Unreleased], and the 5 code sub-fixes (asyncOverride/facet-merge/popup-render/insertionText+displayLabel/suppressInput-hoisting) are byte-identical to the previously vetted commit. No version bump in the diff — this is a pure bug_fix PR. All three required CI checks green on the new SHA: check (11m0s), check-ios (9m1s), check-macos (11m29s). Local re-verify: jvmTest, wasmJs compile, apiCheck, spotlessCheck — all green (22s, 254 tasks executed). Tier 1 — auto-merging.

Issue #115, opened 2026-06-05
Selected completion item doesn't scroll into view on arrow navigation


monkopedia-coder:
## Symptom
In the completion popup, arrow-navigating (Down past the visible rows, or Up) moves the selection but the list does **not** scroll to keep the selected row visible — the selection moves off the visible box (the `Column(verticalScroll)` stays put).

## Cause
`CompletionList` (`autocomplete/.../View.kt`) renders rows in a `Column` with `verticalScroll(rememberScrollState())` + `heightIn(max=…)`, but nothing scrolls the selected index into view when `setSelectedCompletion` changes the selection.

## Fix
Scroll the selected row into view on selection change — e.g. a `BringIntoViewRequester` on the selected `Row` (request on `completionState.selected` change), or compute the target scroll offset from the selected index + row height and `scrollState.animateScrollTo(...)`.

Reported by konstructor (arrow-nav screenshot). Surfaced now that the popup renders correctly (#109).

PR #123, opened 2026-06-08
Filter completion list by typed prefix (#114)

### LINKED ISSUE #114: Completion list doesn't filter as you type a prefix after the trigger char
## Symptom (live, kotlin-lsp 262.6274.0, via konstructor)
Type `xyz(...).` → popup shows all 56 members. Then type `x` → the list **still shows all 56** (e.g. `plus`), ignoring the typed `x`. Accepting `plus` then appends to `.x` → **`.xplus`**. The "no filtering" and the "xplus" insertion are the same root issue: the typed prefix is ignored both for filtering and for the apply/replace range.

## Engine is correct (backend probe)
```
complete at `xyz(0.0,0.0,1.0).`  → isIncomplete=true, count=56 (all members)
complete at `xyz(0.0,0.0,1.0).x` → isIncomplete=true, count=1, items=[x]
```
The engine DOES narrow on re-query — so the client isn't re-querying/re-filtering as the user types.

## Suspect
`ServerCompletion.kt` ~`validFor = if (result.isIncomplete) null else (config.validFor ?: prefixRegexp(items))`. kotlin-lsp marks every member list `isIncomplete=true` → `validFor=null`. The `:autocomplete` `completionStateField` re-filter only runs when `validFor != null`; with `null` the intended behavior is "re-query each keystroke," but that re-query/narrow doesn't take effect (stale 56-item list persists).

Also verify `completionResultRange`'s `from` anchor is at the **member-name start** (right after the `.`), not the cursor — if `from` were at the cursor, both filtering AND the apply range break (→ the `.xplus` append).

## Fix options
1. **Proper:** make autocomplete re-invoke the (async) override source on keystroke when `validFor == null` (true re-query) — engine returns `[x]`, list narrows, server re-ranks.
2. **Simpler/robust:** set `validFor` to a prefix regex even when `isIncomplete` (e.g. always `prefixRegexp(items)` or `[\w$]*`) so the client filters the existing items by the typed prefix. Since the engine returns the full member list up front, client-side filtering is sufficient here.

## Tests (this is exactly the gap)
Core autocomplete filtering is upstream-tested behavior; add a `:autocomplete`/`:lsp-client` jvmTest: open completion, type a prefix, assert the filtered list narrows AND the accept-range replaces the typed prefix (no `.xplus`). Part of the #109-class test hardening.

Reported by konstructor; Jason hitting it live. Same Tooltip/completion path as #109.

monkopedia-coder:
Fixes #114 — typing a prefix after a completion trigger now narrows the popup, and accepting replaces the typed prefix (no more `.xplus`).

**Verified end-to-end** by the konstructor consumer's canvas harness against the live kotlin-lsp engine: popup at `Cylinder(...).` → 12 items → type `t` → narrows to `[to]` → accept replaces the prefix. Stable on their deploy for 3 days.

## Fix
- `serverCompletion` always sets `validFor` (was nulled for `isIncomplete`, which every kotlin-lsp member list is) so the client filters the returned items by the typed prefix.
- `CompletionPlugin` no longer re-queries while the popup is open and the typed text still matches `validFor` — it lets the state field re-filter the existing result in place (re-querying clobbered the narrowed list + reset the apply range, the `.xplus` cause).
- Keeping the original result also keeps `from` at the member-name start, so accept replaces the typed prefix.

Adds a `:lsp-client` regression test (`isIncompleteResultStillGetsValidFor`).

No version bump (stays 0.3.3, no release — per maintainer). Test-verified: `:lsp-client:jvmTest` green; `:autocomplete` wasmJs compile clean.

(#115 scroll-into-view is a separate follow-up — pending its own snapshot verification.)


monkopedia-reviewer (changes requested):
Two issues before this can merge — one project-rule compliance gap, one stale KDoc that now describes the bug that was just fixed.

## 1. Missing CHANGELOG entry (project rule: CLAUDE.md § Workflow)

`CHANGELOG.md` → `[Unreleased] → ### Fixed` has no entry for #114. CLAUDE.md requires updating CHANGELOG (Unreleased) with the issue/PR number before opening review. Please add something like:

```
- LSP completion list now re-filters as the user types a prefix after a trigger character: `serverCompletionSource` always sets `validFor` (was gated on `isIncomplete`), and `CompletionPlugin` no longer re-queries while the typed text still matches `validFor` — the `.xplus` duplicate-insertion bug is gone (#114).
```

## 2. Stale KDoc in `lsp-client/src/commonMain/kotlin/com/monkopedia/kodemirror/lsp/ServerCompletion.kt` (lines 432–435)

The `**\`isIncomplete\`:**` paragraph still says:

> when the server marks the list incomplete, the result is returned with `validFor = null` so `:autocomplete` re-queries on the next keystroke instead of filtering the stale list; complete lists get the `validFor` regex so they can be filtered locally.

That is now the exact opposite of what the code does, and it's the description of the bug that was just fixed. A developer reading it would be actively misled. The inline comment at the call site explains the new behavior well — just update the KDoc paragraph to match, e.g.:

> **`isIncomplete`:** `validFor` is always set regardless of `isIncomplete` — many servers (e.g. kotlin-lsp) mark every member list incomplete, and gating `validFor` on that left the list unfiltered and the accept-range stale (#114). Client-side narrowing of what the server already returned is correct behavior for both complete and incomplete lists.

Everything else looks good — correctness of the fix is sound, the regression test is meaningful (uses `TestLanguageServer`, `isIncomplete = true`, and asserts `validFor` is non-null AND matches a prefix), the `CompletionPlugin` re-trigger guard correctly consults `cs.open` on `update.state` (post-transaction state), wasmJs-test-disabled rule is not violated (jvmTest only), and all three CI checks are green.

monkopedia-reviewer (approved):
Both prior blocking items are now resolved in commit ceef67c:

1. **CHANGELOG entry added** —  now has a bullet for #114 describing the validFor/prefix-filter fix with the issue reference.

2. **KDoc updated** — The `**`isIncomplete`:**` paragraph in `ServerCompletion.kt` (lines 432–437) no longer describes the removed `validFor = null` behavior; it now accurately reflects that `validFor` is always set unconditionally, with the rationale (kotlin-lsp marks every member list incomplete), and references #114.

Re-review of the full diff confirms nothing regressed:
- `View.kt` CompletionPlugin guard (`cs == null || !cs.open`) is correct: it reads `update.state` (post-transaction), and the StateField's own `update` lambda (State.kt line 98–121) already handles in-place re-filtering via `validFor.containsMatchIn(currentText)` — the guard prevents a redundant re-query that would reset the apply range.
- `ServerCompletion.kt` call-site: `val validFor = config.validFor ?: prefixRegexp(items)` matches the updated KDoc exactly.
- Regression test (`isIncompleteResultStillGetsValidFor`) is sound: uses `TestLanguageServer`, sets `isIncomplete = true`, asserts `validFor` non-null and prefix-matches.
- No wasmJs test rule violation (jvmTest only).

CI green: check / check-macos / check-ios all passed.

monkopedia-reviewer (approved):
Rebase verified clean; both CHANGELOG bullets present, #114 substance intact; all required checks green.

PR #125, opened 2026-06-08
Clamp completion apply range to current doc (#124)

### LINKED ISSUE #124: Crash: accepting a completion after the doc shrank — Invalid position (unclamped apply range)
**User-facing crash** (Jason, on adolin, during normal editing):
```
kotlin.IllegalArgumentException: Invalid position DocPos(12) in document of length 11
```

## Root cause
`applyCompletion` (`autocomplete/.../View.kt`) dispatches the accept with the completion result's stored `from`/`to`, which are captured when the completion is **requested**. If the document shrank since (user backspaced while the popup was open, or a shorter re-sync landed), `to` (or `from`) now points past the current doc end → `DocPos(to)` fails validation → editor crash. `DocPos(12)` in an 11-char doc = a stale `to=12`. Both the default-apply branch and the `applyFn` branch (which passes the unclamped positions into `CompletionApplyContext`) are exposed.

## Fix
Clamp the stored range to the current document before dispatch / before building the apply context:
```kotlin
val docLen = DocPos(view.state.doc.length)
val from = result.from.coerceIn(DocPos.ZERO, docLen)
val to = (result.to ?: view.state.selection.main.head).coerceIn(from, docLen)
```
(A fuller fix maps the stored positions through intervening changes; clamping is the safe minimal guard that kills the crash.)

Independent of #114's filtering logic (that's solid) — purely the apply-range bounds. Reported by the konstructor consumer; their e2e missed it (no "accept after backspace" case). Adding a regression test.

monkopedia-coder:
Fixes #124 — a **user-facing crash** Jason hit on adolin during normal editing:
```
kotlin.IllegalArgumentException: Invalid position DocPos(12) in document of length 11
```

## Cause
`applyCompletion` dispatched the accept using the completion result's `from`/`to`, captured when completion was **requested**. If the document shrank before accept (backspace while the popup was open, or a shorter re-sync landed), those positions point past the current doc end → `DocPos` validation throws. Both the default-apply branch and the `applyFn` branch (which passed the stale positions into `CompletionApplyContext`) were exposed.

## Fix
Clamp `from`/`to` to the current doc length once at the top of `applyCompletion` (covers both branches):
```kotlin
val docLen = DocPos(view.state.doc.length)
val from = result.from.coerceIn(DocPos.ZERO, docLen)
val to = (result.to ?: view.state.selection.main.head).coerceIn(from, docLen)
```

## Test
Adds `acceptAfterDocShrankDoesNotCrash` (`:autocomplete`): open completion, backspace (doc shrinks below the stored `to`), accept → must not throw. **Verified it reproduces the crash** — with the clamp reverted the test fails with the exact `IllegalArgumentException`; with the fix it passes.

Independent of #114's filtering (that's solid). No version bump (stays 0.3.3). `:autocomplete:jvmTest` + spotlessCheck green. Reported by the konstructor consumer.


monkopedia-reviewer (changes requested):
The fix itself is correct and the regression test is genuine — reverting the clamp would cause `acceptAfterDocShrankDoesNotCrash` to throw the exact `IllegalArgumentException`. All three CI checks pass (check / check-macos / check-ios green). One blocking issue before merge:

**CHANGELOG entry missing** — `CLAUDE.md` workflow requires updating `CHANGELOG.md` (Unreleased section) with the issue/PR number on every change. This PR has no entry for the #124 crash fix. Please add a `### Fixed` bullet to `## [Unreleased]`, e.g.:

```
- Accepting a completion no longer crashes when the document shrank while the popup was open (backspace while popup is open). `applyCompletion` now clamps the stored `from`/`to` to the current doc length before dispatch (#124).
```

Everything else is sound:
- Clamp bounds are correct: `from` in `[ZERO, docLen]`, `to` in `[from, docLen]` — no inverted-range risk.
- Both the `applyFn` branch and the default-apply branch use the clamped values.
- `view.state.selection.main.head` for the `to` fallback is evaluated at accept-time (current state), not stale.
- Test is in `jvmTest` — wasmJs-disabled rule respected.
- Quality / style: `coerceIn` on `DocPos` (which implements `Comparable<DocPos>`) is idiomatic; comment is clear and references the issue.

Just add the CHANGELOG line and this is ready to merge.

monkopedia-reviewer (approved):
CHANGELOG item resolved: commit d06ad2e adds a `### Fixed` bullet under `## [Unreleased]` describing the accept-after-doc-shrank crash fix with issue reference (#124). The entry accurately describes the bug and clamp fix. No other changes in the new commit — the clamp logic, both apply branches, and the regression test are all unchanged from the previously-vetted first commit. All three CI checks (check / check-macos / check-ios) are now green. Approving for merge.

PR #126, opened 2026-06-08
Scroll selected completion into view on arrow-nav (#115)

### LINKED ISSUE #115: Selected completion item doesn't scroll into view on arrow navigation

monkopedia-coder:
Fixes #115.

When a completion list is longer than the popup's max height (240.dp), arrow-navigating down/up moved the selection highlight but the `verticalScroll` `Column` did not follow it — so the selected row could scroll out of sight below (or above) the fold, and you'd be "selecting" a row you couldn't see.

## Fix
Attach a `BringIntoViewRequester` to whichever row is currently selected and call `bringIntoView()` from a `LaunchedEffect(completionState.selected)`, so each selection change scrolls that row into the viewport. (`androidx.compose.foundation.relocation`, `@OptIn(ExperimentalFoundationApi::class)` — built from commonMain primitives, compiles on jvm/wasmJs/android.)

## Testing
This is Compose UI scroll behavior (no addressable unit-test seam — `BringIntoViewRequester` requires a laid-out/rendered viewport), so it's covered by manual + downstream verification rather than a jvmTest:
- `:autocomplete` builds + tests green (jvm), spotless clean.
- The downstream konstructor consumer reviewed the `BringIntoView` code against its live editor and confirmed it's sound to merge.

No public API change, no version bump (stays 0.3.3). CHANGELOG `[Unreleased]` updated.


monkopedia-reviewer (approved):
Reviewed clean on correctness + quality; BringIntoViewRequester attached to the selected row with bringIntoView() in LaunchedEffect(selected) is idiomatic and consistent with CompletionList; UI-scroll behavior has no jvmTest seam (consistent with prior UI fixes) and is downstream-verified; all three required checks (check / check-macos / check-ios) green.

PR #128, opened 2026-06-08
Return null from coordsAtPos/blockAtPos for out-of-range positions (#127)

### LINKED ISSUE #127: Render crash: coordsAtPos/blockAtPos throw on a position one past EOF after a shrinking edit
## Symptom
During normal editing with LSP on, a render-path crash, always exactly `length + 1`:
```
kotlin.IllegalArgumentException: Invalid position DocPos(22) in document of length 21
```
Observed `12/11`, `21/20`, `22/21` — a stale position one past the new end after the document shrank by one character. Reported by the konstructor (downstream wasmJs) consumer; Jason hits it during editing / "on first save cycle". The constant wasm frame trace is `…lineAt require ← [15719] ← [15860] ← [16041] ← (edit or save trigger)`.

## Root cause
`LineLayoutCache.coordsAtPos(pos, side, state)` and `blockAtPos(pos, state)` (`view/.../LineLayout.kt`) call `state.doc.lineAt(DocPos(pos))` on a **caller-supplied** document offset with no bounds check. `Text.lineAt` throws when `pos > length`.

Several render-path callers can momentarily hold a position that outlives a shrinking edit by one frame, before their (async) source recomputes against the new document:
- `TooltipLayer` → `session.coordsAtPos(tooltip.pos)` (`Tooltip.kt:90`) — an LSP **hover / diagnostic tooltip** whose `pos` was computed against the pre-edit doc.
- the queued caret-reveal / goal-column probes via `coordsAtPos`.

When the deletion removes the last character, such a stale `pos == oldLength == newLength + 1`, and the next render frame calls `lineAt(DocPos(newLength + 1))` → crash. This is the `length + 1` invariant. It only reproduces with a real *edit* that shrinks the doc (a static load + recompile + save does not), which is why it has been hard to reproduce in a harness.

This is distinct from the decoration squiggle/gutter ranges (those are clamped) and from the earlier `applyCompletion` clamp (#124) — this is the coordinate-query layer.

## Fix
A coordinate query for a position outside the current document should yield **no coordinates** (`null`) rather than throw — `coordsAtPos`/`blockAtPos` are already nullable and all callers handle null gracefully (`?: continue`, `?: return sel`). Guard `pos` to `[0, doc.length]` before `lineAt`.

monkopedia-coder:
Fixes #127.

## Symptom
A render-path crash during normal editing with LSP on, always exactly `length + 1`:
```
kotlin.IllegalArgumentException: Invalid position DocPos(22) in document of length 21
```
(seen `12/11`, `21/20`, `22/21`). Reported by the downstream konstructor consumer; Jason hits it while editing / "on first save cycle". konstructor's constant wasm frame trace bottoms out at `…lineAt require ← [15719] ← [15860] ← [16041]` with a variable top frame (edit- vs save-triggered recomposition).

## Root cause
`LineLayoutCache.coordsAtPos(pos, …)` / `blockAtPos(pos, …)` call `state.doc.lineAt(DocPos(pos))` on a **caller-supplied** offset with no bounds check, and `Text.lineAt` throws when `pos > length`. That innermost call is frame `15719`; `EditorSessionImpl.coordsAtPos` (`15860`) and `TooltipLayer` (`16041`) are the two outer constant frames.

Render-path callers can momentarily hold a position that outlives a shrinking edit by one frame before their (async) source recomputes against the new doc — e.g. `TooltipLayer → session.coordsAtPos(tooltip.pos)` for an LSP hover/diagnostic tooltip, or a queued caret-reveal. When the deletion removes the last character, that stale `pos == oldLength == newLength + 1`, hence the `length + 1` invariant. It only reproduces with a real *edit* that shrinks the doc (a static load + recompile + save does not), which is why konstructor couldn't trigger it in a harness.

This is a different layer from the decoration squiggle/gutter ranges (already clamped) and from the `applyCompletion` clamp (#124).

## Fix
A coordinate query for a position outside the current document should yield **no coordinates** rather than throw — both methods are already nullable and every caller handles null gracefully (`?: continue`, `?: return sel`). Guard `pos` to `[0, doc.length]` before `lineAt`.

## Test
`LineLayoutTest.coordsAtPosPastDocEndReturnsNull` / `blockAtPosPastDocEndReturnsNull`: query at `length + 1` (and `-1`) → null. **Verified the test reproduces the crash** — with the `coordsAtPos` guard reverted it fails with the exact `IllegalArgumentException` from `lineAt`; with the guard it passes. `:view:jvmTest` + spotless green.

No public API change, no version bump (stays 0.3.3).


monkopedia-reviewer (approved):
LGTM. Correctness verified on both axes:

**Guards:** `if (pos < 0 || pos > state.doc.length) return null` in both `coordsAtPos` and `blockAtPos`. The `pos == doc.length` (legitimate EOF cursor) is allowed (`<=` not `<`), so a cursor at end-of-document still gets valid coordinates. Only `pos > length` or `pos < 0` returns null — the exact stale-position window the bug creates.

**Test validity:** `coordsAtPosPastDocEndReturnsNull` queries pos=6 on a length-5 doc — this hits `lineAt(DocPos(6))` pre-fix (throws `IllegalArgumentException: Invalid position DocPos(6) in document of length 5`), and passes post-fix. Not tautological; the coder confirmed reverting the guard makes it throw. The `-1` case exercises the lower bound. `blockAtPosPastDocEndReturnsNull` mirrors the same structure. Both run in `commonTest` (JVM), satisfying the wasmJs-tests-disabled rule.

**Caller safety:** `TooltipLayer` — `?: continue` (skips tooltip for one frame, invisible to the user); `CursorMotion.moveVertically` — `?: return sel` (no-ops the vertical move, safe); secondary goalX lookup — `?.centerX ?: coords.centerX` fallback. Returning null for an out-of-range position is strictly better behavior than a mid-render crash.

**Scope / CHANGELOG:** Fix is narrowly scoped to the guard; no public API change; CHANGELOG [Unreleased] entry present with (#127) referencing the correct issue.

CI: check / check-macos / check-ios all pass.

Issue #130, opened 2026-06-08
moveLineUp/Down only operate on the primary selection (no multi-range support)


monkopedia-coder:
Surfaced while porting #116's `:commands` tests.

kodemirror's `moveLines` (`commands/src/commonMain/.../LineCommands.kt`) reads only `state.selection.main` — the start/end line are computed from the primary range and a single block is moved. Upstream CodeMirror's `moveLine` handles **multiple selection ranges**: it groups ranges into line blocks, moves each, merges adjacent blocks, and preserves all cursors.

So the upstream multi-range `moveLine` test cases (preserves multiple cursors on one line; moves multi-line blocks as one; blocks made of multiple ranges; "does not include a trailing line after a range"; `moveLineDown` clips at end of doc for the block) describe behaviour kodemirror does **not** implement. They were therefore NOT ported in #116 (porting them would assert an unimplemented feature, i.e. produce red tests) — this issue tracks the gap instead.

## Options
1. Implement multi-range `moveLines` (group → move → merge), matching upstream, then port the upstream tests as the regression guard.
2. Decide single-range is sufficient for kodemirror's use and close as wontfix (the primary-selection behaviour is already tested in `LineCommandsTest`).

Low priority — single-cursor line move (the common case) works and is tested. Filed per the "track gaps as issues" policy rather than leaving it implicit in #116.

monkopedia-coder:
Nightly triage: classified **needs-decision**. This is a feature-gap call (the body lays out Options 1/2: implement multi-range `moveLines` matching upstream, or accept single-range and close as wontfix). That direction/scope choice is the maintainer's, so not auto-dispatched.

Issue #117, opened 2026-06-05
Port missing :autocomplete pipeline tests from upstream CodeMirror


monkopedia-coder:
## Context
Test-coverage audit vs upstream `codemirror/autocomplete` `test/webtest-autocomplete.ts` (29 scenarios). This is the area that produced the recent #114 (prefix filtering) regression, so it is high priority.

kodemirror's unit-level coverage is actually good:
- `FilterTest.kt` (commonTest) covers scoring: exact/case-insensitive prefix, fuzzy subsequence, gap penalty, boost, alphabetical tiebreak, `filter=false`, highlight ranges, empty query.
- `AutocompletionInputTest.kt` / `AutocompletionCommandsTest.kt` (jvmTest) cover typing-filters, no-match-closes, deletion-reopens, accept, moveSelection, `completeFromList`, currentCompletions — all with a **single** source.

The gap: upstream's `webtest` runs the full multi-source completion **pipeline** (the `Runner` harness dispatches transactions through a real state, no rendering required for the assertions on `currentCompletions`). The behaviors below are pure-logic and JVM-portable via `EditorSession` (the existing input tests prove the pipeline runs headless), but are currently untested.

### Multi-source merging & dedup
- "can merge multiple sources" — two completion sources merged into one ordered list.
- "removes duplicate options" — identical labels from different sources deduped.
- "supports unfiltered completions" — a source with `filter:false` merged alongside a filtered source; unfiltered options retained regardless of query.
- "always calls span-less sources" / "calls sources again when necessary" — `validFor`/span re-query semantics: a source without `validFor` is re-invoked on every change; one with `validFor` is only re-filtered while the span holds.

### Ordering edge cases (pipeline-level, complement to FilterTest)
- "sorts by sortText when given" — `sortText` overrides label ordering.
- "prefers by-word matches", "only shows prefix matches for single-letter queries", "doesn't allow split matches for two-letter queries", "handles all-uppercase words", "prefers shorter options" — verify against the **merged/sorted** result list, not just `filterCompletions` in isolation.

### Explicit-completion lifecycle
- "can backspace out entire word when explicit" / "stops explicit completion on non-spanning input" / "stops explicit completion when backspacing past start" / "stops explicit completions for non-matching input" — explicit (Ctrl-Space) sessions behave differently from typed ones; none covered.
- "resets selection after refinement" — selected index resets to 0 when the option set changes.
- "can synchronously update results" — a result's `update` callback re-derives options on change.

[...]

PR #131, opened 2026-06-08
Port portable :autocomplete pipeline tests (#117)

monkopedia-coder:
Part of #117 (upstream `:autocomplete` test port). Adds `CompletionPipelineTest` covering the pipeline behaviours that map onto kodemirror's completion model.

## Covered
- **Unfiltered results** (`filter = false`) keep every option as the user types (upstream "supports unfiltered completions"); contrasted with the default `filter = true` source narrowing to nothing on a non-match.
- **Accept replaces a range beyond the cursor** — `result.to` past the cursor → the whole `from..to` range is replaced (upstream "can cover range beyond cursor").
- **Source selection** — kodemirror consults `override` sources in order and uses the *first* non-empty result; documents that it does **not** merge across sources.
- **`completeFromList` gating** — returns null at a non-word position for an implicit request but fires for an explicit (Ctrl-Space) one; anchors the range to the word before the cursor.

All seven pass on first run (verified the assertions against the real pipeline via `EditorSession`, not mocks).

## Important finding — kodemirror's completion model is simpler than upstream's
Several upstream `webtest-autocomplete` scenarios test features kodemirror **does not implement**, so they can't be ported as green tests (they'd assert unimplemented behaviour). Deliberately excluded, and reclassified on #117 as feature-divergences rather than missing-test gaps:
- **Cross-source merge + dedup** of options — `triggerCompletion` is first-non-empty-wins, not a merge.
- **Multi-cursor completion** — `applyCompletion` dispatches a single change + single cursor (analogous to the `moveLine` single-range gap #130).
- **`sortText` ordering** — kodemirror uses `boost` (already covered by `FilterTest`); there is no `sortText` field.
- **Explicit-session lifecycle nuances** (backspace-out-of-word, stop-on-non-spanning-input) — kodemirror's re-filter is `validFor`-gated with no separate explicit-session state machine.

I'll annotate #117 with this breakdown so the remaining items are tracked as "model divergence" not "untested".

## Async/timing scenarios
Still deferred (no deterministic test seam for `setTimeout`/`AbortSignal`-style async sources), as the audit called out.

Test-only; no production change, no version bump (stays 0.3.3). `:autocomplete:jvmTest` + spotless green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Tests
 - Ported upstream `:commands` coverage for `insertNewlineKeepIndent` and the bracket-explosion behaviour of `insertNewlineAndIndent` (both were ported but untested) (#116).
+- Added `:autocomplete` pipeline coverage for unfiltered (`filter = false`) results, accept replacing a range that extends past the cursor, first-non-empty source selection, and `completeFromList` word-match/explicit gating (#117).
 
 ### Changed
 - Bumped the lsp dependency to 1.2.0 (was 1.0.1) (#108). 1.2.0 tightens several `LanguageServer`/`LanguageClient` return types to be nullable, matching the LSP spec's optional results (`textDocument/hover`, `textDocument/formatting`, `textDocument/references`, `textDocument/rename`, `workspace/workspaceFolders`). The client now treats a null result as "no result" (no-op / empty), preserving prior behaviour.
--- a/autocomplete/src/jvmTest/kotlin/com/monkopedia/kodemirror/autocomplete/CompletionPipelineTest.kt
+++ b/autocomplete/src/jvmTest/kotlin/com/monkopedia/kodemirror/autocomplete/CompletionPipelineTest.kt
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Tier-1 test_only: 7 tests covering unfiltered completion, accept-range-beyond-cursor, first-non-empty source selection, and completeFromList gating.

Correctness (axis A):
- Tests dispatch through the real pipeline (EditorSession -> triggerCompletion -> completionStateField) — not mocks. Assertions check actual pipeline outputs, not tautologies.
- first-non-empty-source claim verified against View.kt lines 156-165: triggerCompletion iterates sources and returns immediately on the first non-empty result — no cross-source merge.
- accept-replaces-range-beyond-cursor: applyCompletion uses result.to (clamped to doc length), confirmed at View.kt lines 199-229.
- completeFromList gating verified against View.kt lines 383-394: null at non-word position for implicit, non-null for explicit.
- Divergence exclusions (cross-source merge/dedup, multi-cursor, sortText, explicit-session lifecycle) are accurately characterised and correctly omitted.
- jvmTest placement respects wasmJs-disabled rule.
- CHANGELOG ### Tests entry present and references #117.

Quality (axis B):
- Style, naming, and comment density consistent with AutocompletionInputTest / MultiSourceCompletionTest.
- typeChar helper is a local copy (self-contained file); acceptable given existing test file conventions.

All three CI checks green (check 11m21s, check-macos 5m56s, check-ios 12m15s).

PR #134, opened 2026-06-09
Clamp hover tooltips to the viewport (#110)

### LINKED ISSUE #110: Hover tooltip renders offscreen (missing viewport-bounds clamping)
## Symptom
On the live wasmJs editor (real kotlin-lsp), hovering a symbol resolves hover content fine, but the **hover tooltip renders offscreen** — positioned outside the viewport. Reported by Jason via konstructor: "a weird offscreen bug, may need some bounds checking."

## Likely cause
The hover tooltip's anchor/position is computed from the symbol position without clamping to the editor viewport bounds, so near edges (or with certain scroll/layout states) it lands out of view. Same plugin area as the hover-tooltip work in #92/#93 (HoverTooltip / serverHover positioning in `:view` / `:lsp-client`).

## Fix direction
Clamp the tooltip's on-screen rect to the editor viewport (and flip/shift it to stay visible, as CM6 does), rather than placing it purely at the raw anchor coordinate.

## Notes
- Separate from #109 (trigger-char double-insert), which is being iterated in the same 0.3.4-SNAPSHOT loop.
- Reproducible on the live adolin deploy.

monkopedia-coder:
Fixes #110.

## Symptom
On the live wasmJs editor (real kotlin-lsp), hovering a symbol resolved hover content fine but the **tooltip rendered offscreen** near viewport edges. From the issue's narrowing: a Kotlin `by`-delegate hover returns a multi-line range (`start=(0,19) end=(3,1)`), and the tooltip was placed at the raw caret coordinate with no viewport clamping, so the anchor near the bottom edge pushed it out of view.

## Fix
`TooltipLayer` placed each `Popup` at `IntOffset(coords.left, coords.bottom + 4)` with `alignment = TopStart` — the raw editor-relative caret coordinate, never clamped to the window. It now positions through a **`PopupPositionProvider`**, which gets the measured tooltip size and the window bounds, backed by a pure `computeTooltipOffset()`:
- clamp **x** so the tooltip stays within the window horizontally;
- **flip** above/below the anchor when the preferred side would overflow the window and the opposite side fits (respecting `Tooltip.strictSide`);
- clamp **y** into the window as a last resort (a tooltip taller than the viewport pins to the top).

This keeps the tooltip visible regardless of which position it anchors at, so it's robust to the multi-line-range-end anchor as well as plain edge cases.

## Test
`TooltipPositionTest` (9 cases) covers the geometry without a render: below-when-fits, horizontal clamp at both edges, flip-above-when-below-overflows and vice-versa, explicit above, `strictSide` (no flip, still clamps), taller-than-window pins to top, and anchor→window offset. `:view:jvmTest` + spotless + `apiCheck` green (no public API change; `computeTooltipOffset` is internal).

## Notes
- Pure-geometry approach so the clamp/flip logic is verifiable headless; the `PopupPositionProvider` wiring is a thin shell. Live behaviour against the real LSP hover is konstructor-verifiable (they have the hover repro on adolin).
- The secondary "anchor at range start, not end" refinement (which `pos` `serverHover` passes in `:lsp-client`) is orthogonal — the viewport clamp fixes the offscreen symptom regardless. Can follow up separately if desired.

No version bump (stays 0.3.3).


monkopedia-reviewer (approved):
Bug fix reviewed and approved.

**Classification:** bug_fix → Tier 1 (kodemirror default for bug_fix)

**Correctness (Axis A):**
- All 9 geometry assertions verified by hand: below-when-fits, horizontal clamp at right/left edge, flip-above-when-below-overflows, flip-below-when-above-overflows, explicit-above, strictSide-no-flip-but-still-clamps, taller-than-window-pins-to-top, anchor-offset-to-window-coordinates — all correct.
- PopupPositionProvider wiring is sound: anchorBounds.left/top (window-space origin of the editor area) + caret-relative coords → correct window coordinates, then clamped.
- Tests exercise the real `computeTooltipOffset` (not a mock), are non-tautological, and cover all stated branches including the strictSide edge case.
- `computeTooltipOffset` is `internal` — no public API change; apiCheck is clean.
- wasmJs-tests-disabled rule respected (tests are in commonTest, run on jvm).
- CHANGELOG [Unreleased] ### Fixed entry present and references #110.

**Quality (Axis B):**
- Idiomatic Kotlin: `coerceIn`, `when`-expression, named parameters throughout.
- Clear inline comment on the Popup block explaining the why vs. the old raw-coordinate approach.
- `computeTooltipOffset` KDoc is thorough and accurate.
- No scope creep; the 'anchor at range start' refinement is explicitly left out of scope as noted.

**CI:** check / check-macos / check-ios all pass.

PR #142, opened 2026-06-09
Support multi-range moveLineUp/Down (#130)

### LINKED ISSUE #130: moveLineUp/Down only operate on the primary selection (no multi-range support)

monkopedia-coder:
## Summary

`moveLineUp`/`moveLineDown` previously read only `state.selection.main` and moved that single line-block. They now support multiple selection ranges, matching upstream `@codemirror/commands` `moveLine`.

### Approach
- `selectedLineBlocks(state)` groups ALL selection ranges into contiguous line-blocks. Ranges whose line spans touch/overlap merge into one block; a non-empty range ending exactly at a line start does not drag in that trailing line.
- Each block is moved one line up (`forward=false`) or down (`forward=true`) by swapping it with the adjacent line, exactly as upstream does (build a `changes` array + a `ranges` array, iterating blocks).
- Blocks pinned at the doc edge in the move direction are skipped (first block can't move up, last can't move down). The command returns `false` only when NOTHING could move.
- All ranges are remapped to follow their moved lines (down-moves clip to the doc end, so a block can't run past it). One `ChangeSpec.Multi` transaction carries all changes plus the remapped `EditorSelection`.
- The existing single-range behavior is preserved as a special case (one block, one range).

### Tests
Ported the upstream multi-range cases into `LineCommandsTest`:
- preserves multiple cursors on one line (up and down),
- moves a selected multi-line block as one unit,
- a block made of multiple ranges (a multi-line range + another range on the same line merge),
- "does not include a trailing line after a range",
- `moveLineDown` clips the block at end of doc (no trailing newline),
- edge cases: no-op when all blocks are at the doc edge, and skip-first-block-but-move-others.

Multi-range test states enable the `allowMultipleSelections` facet (state collapses to a single range otherwise).

Verified locally: `:commands:jvmTest`, `:commands:spotlessApply`, `:commands:apiCheck` all green (no public API change).

Fixes #130


monkopedia-reviewer (approved):
Reviewed as `monkopedia-reviewer`.

**Classification:** `bug_fix` (feature in :commands — multi-range support for moveLineUp/Down, no public API change) | **Tier:** 1 (kodemirror baseline `* : tier 1`; no public API change, no breaking change, no security).

**Axis A — Correctness:**
- `selectedLineBlocks` correctly groups contiguous ranges: merges when `upto >= startLine.number`, starts a new block otherwise. The trailing-line guard (`!range.empty && range.to == endLine.from`) matches upstream `@codemirror/commands` `selectedLineBlocks` exactly.
- Forward move: deletes the following line and re-inserts it before the block; shifts all block ranges +size, clipped to docEnd. Correct.
- Backward move: deletes the preceding line and re-inserts it after the block; shifts all block ranges -size. Correct.
- Edge blocks skipped silently; `changes.isEmpty()` guard returns false only when nothing moved. Correct.
- `mainIndex.coerceIn(0, ranges.size - 1)` correctly handles the case where the main range's block was skipped. Correct.
- Single-range case is a strict subset (one block, one range) — existing behavior preserved.

**Test trace (selected):**
- `testMoveLineDownBlockMadeOfMultipleRanges`: two ranges merge into one block; swap with line3; shifts +4. Expected `[4→9, 10→10]` ✓
- `testMoveLineDownDoesNotIncludeTrailingLineAfterRange`: range(0,4) trailing-line guard fires (to=4==endLine.from=4); block is line1 only. Expected `[4→8]` ✓
- `testMoveLineDownClipsBlockAtEndOfDoc`: clip to docEnd fires for shorter last line; result `[6→9]` ✓
- `testMoveLineUpSkipsFirstBlockButMovesOthers`: block1 at DocPos.ZERO skipped; block2 moves; mainIndex coerced to 0. Expected doc `"aaa\nccc\nbbb\nddd"` ✓

**Axis B — Quality/Style:**
- `LineBlock` as a private mutable helper class — idiomatic for this kind of grouping logic.
- Comments are concise and match surrounding density.
- Imports use unqualified names as required by project preference.
- Test helpers (`range()`, `selectionPairs()`, multi-range `createView()`) are clean and reusable within the test class.
- `allowMultipleSelections.of(true)` in test setup is mandatory — state collapses to a single range otherwise. Tests correctly enable it.
- CHANGELOG ### Fixed entry references #130. ✓

**CI:** check ✓, check-macos ✓, check-ios ✓ (all passed). mergeStateStatus: CLEAN.

PR #143, opened 2026-06-09
Implement completionItem/resolve for lazy completion (#112)

### LINKED ISSUE #112: Support lazy completion (completionItem/resolve) for items with deferred insertion text

monkopedia-coder:
## Summary

Implements `completionItem/resolve` for lazy LSP completion. Some servers (e.g. kotlin-lsp) send completion items with an empty insertion and deliver the real `textEdit`/`insertText` plus auto-import `additionalTextEdits` only via resolve. The #109 stopgap fell back to the label so basic insertion worked, but the auto-imports never applied.

When the server advertises `completionProvider.resolveProvider` and an item still needs resolution (it carries a `data` payload, or has no concrete insertion/edits yet), accepting the completion now:
1. issues `completionItem/resolve` (on apply — the cheap, correct hook), and
2. applies the resolved item's insertion **and** its auto-import `additionalTextEdits` in a single transaction (so the cursor-anchored insertion and the typically-earlier import edit don't invalidate each other's positions).

`applyFn` is synchronous while `completionItemResolve` is `suspend`, so the resolve-then-apply is launched on the session's `coroutineScope`. A failed resolve falls back to the unresolved item so a basic insertion still lands. Servers without resolve support and already-complete items keep the previous eager behaviour.

## Changes
- `ServerCompletion.kt`: `mapCompletionItem` gains optional `client`/`resolveProvider` params and installs a resolve-aware `applyFn` for lazy items (plain + snippet paths); adds `CompletionItem.needsResolve()`, `resolveThenApply`, and `insertWithAdditionalTextEdits` (single combined transaction). `serverCompletionSource` passes the server's `resolveProvider` capability through.
- `ServerCompletionTest.kt`: lazy item from `completion` (empty `textEdit.newText` + `data`) resolved to a real `newText` + auto-import `additionalTextEdits` from `completionItem/resolve`; asserts apply fetches the resolved item and applies BOTH the insertion and the auto-import; asserts NO resolve when `resolveProvider` is false or the item is already complete; plus `needsResolve` unit coverage.
- `CHANGELOG.md`: Unreleased / Fixed entry (#112).

## Verification
- `:lsp-client:jvmTest` green (29 tests in `ServerCompletionTest`, incl. 4 new), `:lsp-client:spotlessApply` clean, `:lsp-client:apiCheck` green (no public API change — new declarations are `internal`/`private`).

## Caveats
- The live end-to-end auto-import flow (against a real kotlin-lsp) still needs konstructor verification; this PR verifies the resolve-then-apply logic with a scripted stub server.

Fixes #112


monkopedia-reviewer (commented):
**Reviewer pass — `monkopedia-reviewer`**

**Category:** `feature_internal` (`:lsp-client`, no public API change — new decls are `internal`/`private`)
**Tier:** 3 (user-requested reviewer)
**CI:** ✅ all three checks green (`check` / `check-macos` / `check-ios`)
**Branch:** clean — 1 commit ahead of main, 0 behind

---

### Axis A — Correctness

All checklist items pass:

- **Resolve gating** ✅ — `resolveProvider` is read as `client.serverCapabilities?.completionProvider?.resolveProvider == true`; `needsResolve()` is checked independently; both must be true before the resolve path installs.
- **Single-transaction apply** ✅ — `insertWithAdditionalTextEdits` builds a `ChangeSpec.Multi` from the sorted union of `additionalTextEdits` + the primary insertion; `shiftBefore` correctly accounts for the net length delta from earlier edits when placing the cursor.
- **Coroutine bridge** ✅ — `resolveThenApply` launches on `ctx.session.coroutineScope` from the sync `applyFn`. One `launch` → one `dispatch`; no double-apply possible. Failure catches all `Throwable` and falls back to the original item, which then takes the normal label-insertion path — basic insertion still lands.
- **Fallback for no-resolve / already-complete items** ✅ — `resolvable` is false when `client == null || !resolveProvider || !item.needsResolve()`; the `when` branches drop through to the original eager paths.
- **CHANGELOG** ✅ — `## [Unreleased] / ### Fixed` entry references `#112`.
- **No public API change** ✅ — all new declarations are `internal`/`private`; `apiCheck` green per PR body.
- **Tests** ✅ — four new cases: `needsResolveDetection`, `lazyApplyResolvesAndAppliesInsertionPlusAutoImport`, `noResolveWhenServerLacksResolveProvider`, `noResolveWhenItemAlreadyComplete`. The `DelegatingSession` using `Dispatchers.Unconfined` correctly makes the async bridge run synchronously in tests.

One minor correctness note: the snippet resolve path calls `applyAdditionalTextEdits` then the snippet applicator as two separate transactions (same pattern as the pre-existing non-resolvable snippet path). The plain-text path correctly uses one `ChangeSpec.Multi`. This asymmetry is inherited from the existing code and was present before this PR — not a new regression, just noted for awareness.

### Axis B — Style

One recorded-rule violation (`imports` rule from memory):

`applyDefault` in `ServerCompletionTest.kt` references `com.monkopedia.kodemirror.state.TransactionSpec`, `ChangeSpec.Single`, and `InsertContent.StringContent` by FQN rather than adding imports. Rule: "Always prefer imports over fully qualified names. Use an alias if there's a conflict." These are unambiguous names with no conflict — they should be imported. This is test-only and in a `private` helper, so it is not a correctness issue.

---

### Live E2E note

As noted in the PR: the auto-import flow against a real kotlin-lsp server still needs konstructor verification. Not blocking — the scripted stub tests cover the resolve-then-apply logic thoroughly.

---

**Overall verdict:** Approvable. The coroutine bridge is correct, the single-transaction guarantee is sound, and CI is green. The FQN nit is minor — your call whether to clean up before merge or let it ride.

monkopedia-reviewer (approved):
Maintainer signed off (tier-3) on the completionItem/resolve approach. Rebased onto main (CHANGELOG: kept #130 + #112 Fixed bullets) and fixed the test FQN→imports nit (ChangeSpec/InsertContent/TransactionSpec now imported, per the project's imports-over-FQN rule). Coroutine resolve-on-apply bridge verified correct by review (single launch→single dispatch, failure falls back to label insertion); :lsp-client:jvmTest + spotless green locally. Live auto-import flow to be konstructor-verified against real kotlin-lsp. Auto-merge on green.

PR #86, opened 2026-05-30
Fix active-line highlight not filling viewport width in no-wrap mode

monkopedia-coder:
## Summary
In no-wrap mode the active-line highlight (and any line `style.background` decoration) only spanned the **content/text width** instead of the full **viewport width**. At v0.2.0 the active-line band reached the full editor width; after the horizontal-scroll layout work (#20/#64 natural-width content + #68 `widthIn(min = maxLineWidthPx)`) it stopped just past the text. (#85)

## Root cause
In `KodeMirror.kt`'s per-line layout, the line-decoration background lives on the content `Box`. In no-wrap mode that box's min-width was `maxLineWidthPx` (the widest line's natural width). When content is narrower than the viewport, the highlight therefore stopped at the text.

## The fix
No-wrap content min-width is now `max(widestLineWidth, viewportWidth)`:
- `max(widestLine, …)` preserves the #68 uniform full-natural-width behavior so a genuinely long line still lays out at its full width and scrolls horizontally.
- `max(…, viewport)` makes a short line's content (and its background highlight + selection overlay) fill the viewport.

**Measuring viewport width:** the viewport `Box` has `.horizontalScroll(...)`, which measures its child with an *infinite* max width, so `BoxWithConstraints` inside the scroll would report Infinity. Instead the viewport's actual laid-out width is captured from the `weight(1f)` viewport box itself via `.onSizeChanged { viewportWidthPx.value = it.width }` (added before the `horizontalScroll`). The viewport width is weight-determined, independent of content width, so deriving content min-width from it is loop-safe. The existing `contentExtraModifier.background` and `drawSelectionOverlay` then paint the full width automatically.

**SelectionDrawing.kt: no change needed.** The `extendsToNextLine` path already draws to `size.width`; with the content box now min-width = viewport, `size.width` is the viewport-or-wider width, so a multi-line selection reaches the viewport edge automatically.

## verifyRoborazzi compare images (the real visual confirmation)
`:view:verifyRoborazzi` still reports a mismatch — that is **font-glyph drift** from baselines recorded on another machine (expected, not re-recorded here). In `active-line_compare.png` and `basic-light_compare.png` the **New** panel's active-line band now reaches the full editor width, matching **Reference**. The large solid red bar that previously spanned the un-highlighted region in the Diff panel is **gone** — only thin font-glyph drift on the text remains. This confirms the highlight-width regression is fixed.

## Regression test
[...]

PR #144, opened 2026-06-09
Gate :view Roborazzi screenshot tests in CI (#89)

### LINKED ISSUE #89: Gate :view Roborazzi screenshot tests in CI (pinned font/render environment)
The `:view` Roborazzi screenshot tests (`view/src/jvmTest/.../screenshots/*ScreenshotTest.kt`, baselines in `view/screenshots/compose/*.png`) currently run in CI under plain `jvmTest` in **capture-only** mode — they smoke-test that rendering doesn't throw but never pixel-verify against the committed baselines. There is no `verifyRoborazzi` / `-Proborazzi.test.verify=true` in CI. So **visual regressions are invisible to CI**.

This let a real regression nearly ship in 0.3.0: the active-line/line-decoration highlight stopped filling the viewport width in no-wrap mode (fixed in #86). CI was fully green; only a manual `verifyRoborazzi` + human eyeballing caught it.

### Why it's not trivially "just turn on verify"
The committed baselines were recorded on a different machine, so `verifyRoborazzi` is red on other machines purely from **font-glyph advance-width drift** (a few px), independent of any code change. Turning on verify against env-specific baselines would be all-noise. The real fix is to make the render environment **reproducible**.

### Proposed
- Establish a **pinned rendering environment** for screenshots (e.g. a fixed CI container / locked font stack / known runner image) and record the canonical baselines there.
- Add a CI step/job that runs `:view:verifyRoborazzi` (or `testDebugUnitTest -Proborazzi.test.verify=true`) in that environment so visual regressions fail the build.
- Document the "record baselines in env X" procedure so contributors regenerate consistently.
- Until then, screenshots remain capture-only; reviewers touching `:view` rendering should manually run `verifyRoborazzi` and inspect `view/build/outputs/roborazzi/*_compare.png`, distinguishing real diffs from font drift (technique: render the prior release tag on the same machine and compare — isolates code delta from font noise).

Follow-up to the 0.3.0 release work (#86 / #88). Not a release blocker.

monkopedia-coder:
Fixes #89.

The `:view` Roborazzi screenshot tests previously ran in **capture-only** mode under `jvmTest` — they smoke-tested that rendering doesn't throw but never pixel-verified against the committed baselines. So visual regressions were invisible to CI (a real one nearly shipped in 0.3.0: the active-line highlight stopped filling the viewport width, #86 — CI was green, only a manual `verifyRoborazzi` caught it).

This PR makes them a reliable CI gate.

### Pinning approach
The committed baselines used to drift across machines because the editor's default content font is `FontFamily.Monospace`, which resolves to whatever monospace font the host OS provides — so glyph advance widths (and pixels) differ between the dev machine and the CI runner.

The screenshot tests now force a **bundled JetBrains Mono Regular** (OFL-1.1, committed under `view/src/jvmTest/resources/fonts/` with its license), loaded from the test classpath via the Compose Desktop `Font(resource: String)` factory (`TestScenarios.pinnedFont`). Compose Desktop rasterises text with **Skiko**, which is bundled with the Compose dependency and does *not* use OS-native rasterisation, so the same font + same Skiko version produces the same pixels on the dev machine and the `ubuntu-latest` CI runner (both Linux x86_64, same Skiko native build).

### Tolerance
Even with a pinned font the comparison allows a small **0.1% pixel-change threshold** (`changeThreshold = 0.001f`, via a shared `captureScreenshot` helper) as a safety margin against incidental antialiasing noise. It is small enough that a real regression still trips the gate — verified by corrupting a baseline, which fails `verifyRoborazziJvm` — while sub-pixel noise does not make the gate flaky.

### CI change
`.github/workflows/ci.yml`: added `:view:verifyRoborazziJvm` to the existing gradle invocation in the `check` job, so UI regressions fail the build.

### Other
- All 14 baselines regenerated under the pinned font.
- Procedure documented in `docs/roborazzi-screenshots.md` (record with `:view:recordRoborazziJvm` after an intentional rendering change; CI runs `:view:verifyRoborazziJvm`).

### Local verification
- `record` then `verify` x2 (`--rerun-tasks`): stable both times.
- Corrupted a baseline → `verifyRoborazziJvm` fails (gate is not a no-op); restored → passes.
- `:view:apiCheck` + `:view:spotlessCheck` green.

### Caveat
The pinning is verified stable **on Linux x86_64** (dev + `ubuntu-latest` share the bundled Skiko build). The `check` job that now runs verify is the Linux job; the apple/native suites do not run these JVM screenshot tests. If CI ever moves the `check` job to a different OS/arch, baselines would need re-recording there; the 0.1% tolerance is a margin against minor AA drift, not a cross-arch fix.


monkopedia-reviewer (approved):
Approving. All three CI checks pass (check / check-macos / check-ios — all green). The new `:view:verifyRoborazziJvm` step ran as part of the `check` job (11m4s, green on `ubuntu-latest`), which is direct proof the gate is not a no-op and passes on the real CI runner.

Review summary:
- CI yaml change is minimal and correct: one line added to the existing gradle invocation in the `check` job, syntax clean, no other jobs affected.
- Determinism approach is sound: JetBrains Mono Regular bundled in the test classpath, loaded via `Font(resource:)`, Skiko rasterisation (not OS-native) ensures the same pixels on dev + `ubuntu-latest`. Cross-machine glyph drift was the root cause of prior baseline instability; this pins it correctly.
- 0.1% pixel tolerance (`changeThreshold = 0.001f`): tight enough to catch real regressions (corrupted-baseline failure verified by the coder), loose enough to avoid AA flakiness.
- OFL-1.1 license for JetBrains Mono committed alongside the TTF — license compliance present.
- All 14 baselines regenerated and committed under the pinned font.
- All 14 screenshot tests migrated from bare `captureRoboImage` to the shared `captureScreenshot` helper — consistent options application throughout.
- `docs/roborazzi-screenshots.md` added with layout, pinning rationale, tolerance, record/verify commands, and the intentional-change procedure.
- CHANGELOG Unreleased updated with #89 reference.
- `mergeStateStatus: CLEAN`.

Tier: 1 (infra, `Monkopedia/kodemirror` baseline sweeps infra to tier 1). Auto-merging.

PR #76, opened 2026-05-27
Fix caret height and gutter alignment on wrapped lines

monkopedia-coder:
## Summary
Fixes two rendering bugs that only appear when `lineWrapping` is enabled and a line wraps to multiple visual rows.

- **Caret height (`drawLineCursor`)**: the thin caret used `size.height` (the full multi-row wrapped-line height), so it spanned every visual row. It now uses `TextLayoutResult.getCursorRect(offset)` to draw the caret on only its actual visual row. No-wrap (single-row) behavior is unchanged, with a fallback to start-of-line/full-height for empty lines or no layout.
- **Vim block cursor (`drawBlockCursor`)**: same flaw — the block rect used `topLeft.y = 0` and `height = size.height`. The vertical position/height now come from `getCursorRect(offset).top`/`.height` for the in-line character branch; the horizontal/charWidth logic is unchanged, and the end-of-line / empty-line branches keep the previous `0f`/`size.height` fallback.
- **Gutter alignment (`KodeMirror.kt`)**: the per-line `Row` used `Alignment.CenterVertically`, vertically centering the gutter line number across a tall wrapped line. It is now `Alignment.Top` in wrap mode (`if (wrapLines) Alignment.Top else Alignment.CenterVertically`) so the number sits beside the first visual row. No-wrap mode is unchanged.

`drawLineSelectionRange` (multi-row selection highlight) is intentionally left untouched — that is a separately-tracked follow-up.

## Test
Added `WrappedLineCursorScreenshotTest` (a capture-only roborazzi scenario, matching the existing screenshot tests): a narrow 360px viewport with a long single line + `lineWrapping`, the caret placed near the end of the line (a later visual row), and gutters enabled. The recorded reference (`view/screenshots/compose/wrapped-line-cursor.png`) confirms the caret spans only its row and the gutter number is top-aligned. Only the new reference image is committed — unrelated re-recorded baselines were reverted (their diffs were font-rendering noise, not from this change).

## Verification (all green)
- `:view:jvmTest`
- `:view:compileKotlinJvm` + `:view:compileKotlinWasmJs`
- `:view:compileDebugKotlinAndroid`
- `:view:apiCheck` (no public API change)
- `:view:spotlessApply` (no changes)

Fixes #75

PR #147, opened 2026-06-09
Fix caret flashing to column 0 on each keystroke (#146)

### LINKED ISSUE #146: Caret flashes to column 0 for one frame on each keystroke (drawLineCursor)
## Symptom
While typing, the line caret flashes to the **start of the line (column 0)** for one frame on each keystroke. Reported by Jason via konstructor on the latest main snapshot.

## Root cause
`view/src/commonMain/.../SelectionDrawing.kt` `drawLineCursor` (introduced by #76, the wrapped-line caret-height fix):
```kotlin
if (textLayoutResult != null && textLen != null && offsetInLine in 0..textLen) {
    val r = textLayoutResult.getCursorRect(offsetInLine); drawLine(... r.left ...)
} else {
    drawLine(start = Offset(0f, 0f), end = Offset(0f, size.height), ...)  // ← col 0
}
```
The line's `TextLayoutResult` lags the document by one frame (the doc change applies before the text re-lays-out), so right after a keypress `offsetInLine == textLen + 1` momentarily → the `in 0..textLen` guard fails → it draws the **line-start (col 0) fallback** for that frame. Pre-#76 used `getHorizontalPosition(offsetInLine, true)` which had no hard bail.

## Fix
Clamp the offset into the layout instead of bailing; reserve the col-0 fallback for a genuinely-null layout:
```kotlin
val safe = offsetInLine.coerceIn(0, textLen)   // transient past-end → draw at line END, not col 0
val r = textLayoutResult.getCursorRect(safe); drawLine(... r.left ...)
```
During the 1-frame lag the caret sits at the line-end (a pixel off) and snaps exact next frame — no visible flash.

`drawBlockCursor` (falls to its line-end branch on past-end) and `drawLineSelectionRange` (already `coerceAtMost(textLen)`) don't have the bail — no change needed there.

Found + root-caused by the konstructor consumer.

monkopedia-coder:
Fixes #146.

## Symptom
While typing, the line caret flashes to **column 0** for one frame on each keystroke (reported by Jason via konstructor on the latest main snapshot).

## Root cause
`SelectionDrawing.kt drawLineCursor` (from #76, the wrapped-line caret-height fix) queried the layout only when `offsetInLine in 0..textLen`, else drew a line-start (column 0) fallback. The line's `TextLayoutResult` lags the document by one frame — right after a keypress the doc change has applied but the text hasn't re-laid-out, so `offsetInLine == textLen + 1` transiently → the guard failed → it drew the column-0 fallback for that frame.

## Fix
Extract a pure `cursorLayoutOffset(offsetInLine, textLen)` that **clamps** into the current layout (`coerceIn(0, textLen)`) and returns null only when there's no layout at all. `drawLineCursor` uses it: during the one-frame lag the caret sits at the line end (a pixel off) and snaps exact the next frame — no visible flash. The column-0 fallback is now reserved for a genuinely absent layout (empty line).

`drawBlockCursor` (already falls to its line-end branch on past-end) and `drawLineSelectionRange` (already `coerceAtMost(textLen)`) don't have the bail — unchanged.

## Test
`CursorLayoutOffsetTest` (5 cases): the transient past-end case clamps to the line end (not column 0), in-range/end/negative offsets behave, and a null layout length still falls back. Pure helper — no render needed. `:view:jvmTest` + `:view:verifyRoborazziJvm` (no static-render regression) + spotless + apiCheck all green (`cursorLayoutOffset` is internal — no public API change).

The 1-frame flash itself isn't screenshot-testable, so the unit test on the offset logic is the real guard; konstructor can confirm live on a snapshot. Reported + root-caused by konstructor.


monkopedia-reviewer (approved):
Reviewed as **monkopedia-reviewer**.

**Category:** `bug_fix` (:view, regression from #76) — **Tier 1** (no public API change; `cursorLayoutOffset` is `internal`).

**Fix is correct.** The root cause is clear: `TextLayoutResult` lags the document by one frame, so right after a keystroke `offsetInLine == textLen + 1` transiently; the old `in 0..textLen` guard bailed to the column-0 fallback for that frame. The fix clamps with `coerceIn(0, textLen)` via the extracted pure `cursorLayoutOffset` helper — the caret sits at line-end for the lagging frame and snaps exact next frame.

**Spot-checked:**
- `cursorLayoutOffset` returns `null` only for a genuinely absent layout (`textLen == null`); otherwise always returns a valid offset in `[0, textLen]`. Correct.
- `drawLineCursor` uses the clamped `safe` offset for `getCursorRect()`. Guard `textLayoutResult != null && safe != null` correctly mirrors old semantics for the no-layout case only. Correct.
- `drawBlockCursor` and `drawLineSelectionRange` correctly left alone — each already handles past-end separately.
- `cursorLayoutOffset` is `internal` — no public API surface change. `apiCheck` green per PR body.
- 5 unit tests are meaningful; `transientPastEndClampsToLineEndNotColumnZero` is the key regression guard (`cursorLayoutOffset(6, 5) == 5`, not null/0).
- CHANGELOG `### Fixed` entry references #146 with an accurate description.
- Code is idiomatic Kotlin; docstring clearly explains the invariant and the regression history.

Merging via auto-merge pending CI (check / check-macos / check-ios).

PR #149, opened 2026-06-09
Merge + dedup completion options across sources (#137)

### LINKED ISSUE #137: Merge + dedup completion options across multiple sources
Surfaced by the #117 `:autocomplete` test port. `triggerCompletion` (autocomplete/View.kt) iterates the override sources and uses the **first** that returns options (first-non-empty-wins). Upstream merges results from ALL sources into one ordered list and dedups identical labels ("can merge multiple sources", "removes duplicate options", "supports unfiltered completions" alongside filtered).

Note: the `completionConfig` facet already MERGES the source *lists* across configs (the #109/#122 fix) — this is about merging the *results* of multiple sources at query time, which is a bigger behavioural change.

**Decision:** implement cross-source result merge + dedup or keep first-non-empty-wins (simpler; works for the single-LSP-source case kodemirror targets today)?

Origin: #117 (model divergence).

monkopedia-coder:
## Summary

Implements cross-source completion merging (#137). Previously `triggerCompletion` iterated the sync override sources and used the **first** one that returned options (first-non-empty-wins). It now queries **all** sync override sources and merges their results into one ordered, deduplicated list — mirroring upstream `@codemirror/autocomplete`'s `sortOptions`.

Fixes #137.

## Merge / dedup semantics

- `mergeCompletions(results, state)` (in `FilterResult.kt`) filters each result against the text between **its own** `from` and the cursor, honouring **its own** `filter` flag, then concatenates, sorts by match score (descending, **stable** so equal-scoring options keep source order — source A before source B), and dedups by **label**, keeping the first (highest-scored) occurrence. This matches the issue's directive ("dedup identical labels; keep the higher-scored/first occurrence").
- A `filter = false` source keeps **all** of its options alongside a filtered source's narrowed list, both at open and as the user types (each result re-filters against its own span on doc change).

## Differing-`from` handling

Fully handled rather than a documented limitation: `CompletionState` now holds a **list** of active results, and each `FilterResult` carries its originating `CompletionResult`. Accepting an option (Enter / click) replaces **that source's** `from`..`to` range, so sources returning different `from` positions apply correctly. Re-filtering also uses each result's own span and `validFor` independently — a source drops out when its `validFor` stops matching, and the list closes only when none survive. The popup anchors at the smallest `from` across the active results.

## Async path

The `asyncOverride` path is left **first-non-empty-wins** (documented in code): async sources are typically a single language server, and merging would force awaiting every source serially before showing anything. A single async result still flows through the same merge machinery as a singleton list.

## API / state

All changed symbols are `internal` (`CompletionState`, `FilterResult`, `startCompletionEffect`, `completionStateField`, `mergeCompletions`) — **no public API change** (`:autocomplete:apiCheck` green). `startCompletionEffect` now carries `List<CompletionResult>`. The `CompletionState`/`completionStateField` re-filter contract is preserved (merged results still flow through `startCompletionEffect`).

## Test changes

- `MultiSourceCompletionTest`: ported the upstream cases — `canMergeMultipleSources`, `removesDuplicateOptions`, `supportsUnfilteredCompletionsAlongsideFilteredOnes`. The existing facet-merge and "real source consulted" tests still pass unchanged (one real + one empty source merges to just the real's options).
- `CompletionPipelineTest`: the old `firstNonEmptySourceWinsAndIsNotMerged` test (and the file's "does NOT merge across sources" header note) asserted the now-replaced behaviour. Renamed to `singleSourceOptionsFlowThroughInOrder` (a single source still passes through unchanged) and updated the header to point cross-source coverage at `MultiSourceCompletionTest`.
- `CompletionStateTest`: updated the `startCompletionEffect.of(...)` helper to wrap the result in a list.

## Verification

`:autocomplete:jvmTest`, `:lsp-client:jvmTest`, `:autocomplete:spotlessApply`, and `:autocomplete:apiCheck` all green locally (JVM + wasmJs compile).


monkopedia-reviewer (approved):
Reviewed at f396261. Tier: 1 (feature_internal, all changed symbols internal — no public API change, apiCheck green).

**Merge correctness (✓ clean):**

- `mergeCompletions` in `FilterResult.kt` correctly filters each result against its own `from`..cursor span (honouring per-source `filter` flag), then applies a SINGLE GLOBAL sort via `completionOrder` (score desc → sortText?:label asc), then deduplicates by label keeping the first (highest-ranked) occurrence. This matches upstream `sortOptions` and is the only ordering site — no per-source-concatenated ordering survives the rebase fix. ✓
- `completionOrder` is `private val` in `FilterResult.kt`; `mergeCompletions` is added to the same file → file-private access is correct in Kotlin. ✓

**State restructure (✓ clean):**

- `CompletionState.results: List<CompletionResult>` replaces the single `result`. The `from` computed property (`results.minByOrNull { it.from.value }?.from`) anchors the tooltip at the smallest from — correct for differing-from sources.
- Re-filter on doc change: filters each result by its own `validFor`, drops non-surviving results, closes if none survive, re-merges survivors. Behavior for `validFor = null` is unchanged (drop, close), and `filter = false` sources keep all options regardless of query. ✓
- The `startCompletionEffect` type change from `CompletionResult` to `List<CompletionResult>` is fully internal; all callers updated (sync dispatch, async singleton-wrap, test helper). ✓

**Accept / click paths (✓ clean):**

`item.result ?: cs.results.first()` fallback is safe — production always goes through `mergeCompletions` which sets `result`; null only in unit tests that call `filterCompletions` directly (documented). ✓

**Async path (✓ documented):**

First-non-empty-wins retained; async result dispatched as a singleton list through the same merge machinery. Trade-off is documented in code and PR body. ✓

**Test coverage (✓ meaningful):**

- `canMergeMultipleSources`: [a,b]+[c,d] → [a,b,c,d] (all score=0, label-asc). ✓
- `removesDuplicateOptions`: [foo,bar]+[foo,baz] → global sort [bar,baz,foo,foo] → dedup [bar,baz,foo]. ✓
- `supportsUnfilteredCompletionsAlongsideFilteredOnes`: filter=false source keeps all options as filtered source narrows on typing. ✓
- `singleSourceOptionsFlowThroughInOrder` (renamed from `firstNonEmptySourceWinsAndIsNotMerged`): single-source passthrough still verified; multi-source coverage moved to `MultiSourceCompletionTest`. ✓
- `CompletionStateTest` updated to wrap single result in list. ✓

CHANGELOG references #137. Imports over FQN throughout. No style findings.

Queuing auto-merge pending CI.

PR #150, opened 2026-06-09
Explicit completion session lifecycle (#139)

### LINKED ISSUE #139: Explicit (Ctrl-Space) completion session lifecycle semantics
Surfaced by the #117 `:autocomplete` test port. Upstream tracks an **explicit** completion session (opened via Ctrl-Space) with distinct lifecycle rules: "can backspace out entire word when explicit", "stops explicit completion on non-spanning input", "stops explicit completion when backspacing past start", "stops explicit completions for non-matching input", "resets selection after refinement".

kodemirror has a `CompletionContext.explicit` flag, but the completion state field (autocomplete/State.kt) re-filters purely on `validFor` matching the typed text — there is no separate explicit-session state machine, so these behaviours aren't distinct from the implicit (typed) path.

**Decision:** implement an explicit-session lifecycle (+ port the ~6 upstream cases) or accept the simpler unified validFor-gated re-filter?

Origin: #117 (model divergence).

monkopedia-coder:
## Summary

Implements upstream `@codemirror/autocomplete`'s explicit-completion-session semantics on kodemirror's post-#149 list-shaped completion state field. Previously the field re-filtered purely on `validFor` with no distinction between explicit (Ctrl-Space / `startCompletion`) and implicit (typed) sessions.

Fixes #139.

## What changed

- **Explicit-session tracking.** `startCompletionEffect` now carries a `StartCompletion(results, explicit)` payload (was a bare `List<CompletionResult>`); `CompletionState` gains an `explicit` flag threaded from the trigger through `triggerCompletion`. All symbols are `internal` — **no public API change** (`apiCheck` green, no `apiDump`).
- **Backspacing lifecycle** (mirrors upstream `ActiveResult.updateFor`, `limit = from + (explicit ? 0 : 1)`):
  - An **explicit** session may backspace back to (and including) the session's original `from` — the empty query re-shows all options and the session stays open.
  - An **implicit** session closes once the typed span is fully removed (one char past `from`).
  - Backspacing before `from` always closes. Backspacing is detected via the `delete` user event.
- **Non-spanning / non-matching input closes the session.** Re-filtering now matches each result's `validFor` against the *whole* span text (`Regex.matches`, anchored — like upstream `checkValid`/`ensureAnchor`) instead of `containsMatchIn`, so a typed space (non-spanning) closes the session; input that spans but yields no options also closes.
- **Selection resets to the top option on each refinement** instead of being coerced into range (upstream builds a fresh `ActiveResult` whose selection starts at 0).

## Tests

`ExplicitCompletionSessionTest` ports the ~6 upstream explicit-lifecycle cases: backspace out the entire word when explicit, stops on non-spanning input, stops when backspacing past start, stops for non-matching input, the implicit-vs-explicit backspace-to-start contrast, and resets selection after refinement. `:autocomplete:jvmTest`, `:lsp-client:jvmTest`, `:autocomplete:apiCheck`, `spotlessApply` all green locally; existing `AutocompletionInputTest` / `MultiSourceCompletionTest` / `CompletionPipelineTest` / `FilterTest` / `CompletionStateTest` still pass.

## Deviations from upstream

kodemirror's model differs from upstream in two ways that are out of scope for #139, so the mapping is faithful but not identical:

- **No per-source `ActiveResult` / `Pending` re-query in the state field.** Upstream keeps each source in its own `ActiveResult` and, when input no longer spans `validFor`, transitions to `Pending` and re-queries the source (retaining `explicit`). kodemirror's field has a single merged session and cannot re-query; when `validFor` stops matching it simply closes, and the `CompletionPlugin` ViewPlugin re-triggers a fresh (implicit) query on the next typed input (it already ignores `delete` events, matching upstream's "deletion never activates"). The observable transitions for the ported cases are identical.
- **Session-wide `from`/`explicit`.** The explicit flag and the backspacing boundary are tracked for the session as a whole (using the smallest `from` across active results) rather than per source, since the trigger is explicit-or-not as a unit. Per-result `validFor` survival is unchanged.
- The exact filtered option strings differ from upstream's (kodemirror's fuzzy filter, #111, does not restrict single-letter queries to prefix matches), so the ported tests assert lifecycle transitions (open/closed, full-set-restored, selection index) rather than exact option lists.


monkopedia-reviewer (commented):
## Automated review — tier 3 (user signoff requested)

@Monkopedia has been added as a required reviewer per the tier-3 directive for this change.

**CI:** check / check-macos / check-ios all pass. No merge conflict with main.

**What was verified (tier-1 pass):**

**Correctness — lifecycle logic matches upstream:**
- `limit = from + (explicit ? 0 : 1)` is correctly translated: for explicit sessions `limit = sessionFrom` (close only when `head < sessionFrom`); for implicit `limit = sessionFrom + 1` (close when `head <= sessionFrom` on backspace, i.e. typed span fully removed). Both match `ActiveResult.updateFor` in upstream `@codemirror/autocomplete`.
- Backspace detection via `tr.isUserEvent("delete")` — `delete.backward` (and other `delete.*` events) match correctly.
- Explicit session with empty query (cursor at `from`): `Regex("[\\w]*").matches("")` is true → all results survive → full option set re-shown. Correct.
- `backedOut = head < sessionFrom` for non-backspace also handles cursor jumps past the start (e.g., selection move).
- `validFor.matches(...)` (anchored) replaces the old `containsMatchIn` (substring). This is the right fix for non-spanning input (a space in `[\\w]*` now correctly closes the session instead of matching a substring). Affects both explicit and implicit sessions consistently.
- Selection resets to 0 on every refinement — matches upstream `ActiveResult` fresh-start.
- `head >= cr.from` guard before `sliceString` is defensive and correct; prevents any malformed state if a result's `from` exceeds the cursor.

**Regression safety:**
- Existing `CompletionStateTest.openCompletions` gains an `explicit = false` default; all pre-existing tests use the implicit path and are unaffected.
- The `viewmodel.matches(...)` change for implicit sessions is also correct (the old `containsMatchIn` was too permissive). No test relied on the old permissive behavior.
- The `CompletionPlugin` ViewPlugin ignores `delete` events (already in the pre-PR code), matching upstream's "deletion never activates a fresh implicit query."

**Test quality:** 6 cases cover all lifecycle transitions meaningfully: explicit backspace-to-start (stays open), explicit backspace-past-start (closes), non-spanning input, non-matching input, selection reset, and explicit-vs-implicit contrast. Assertions are on state (open/closed, count, selection index), not on specific option strings — appropriate given kodemirror's fuzzy filter differs from upstream.

**Deviations from upstream:** accurately documented in the PR body. Single-session `from`/`explicit` tracking (vs per-source) is a deliberate simplification consistent with the post-#137 merged-results model. No per-source re-query on `validFor` fallout is noted as out-of-scope and the observable transitions are identical to upstream for the ported cases.

**No public API change** — `StartCompletion`, `startCompletionEffect`, `CompletionState.explicit` are all `internal`. `apiCheck` is green.

**No blocking findings.** This PR is technically sound; flagging to @Monkopedia for a final look at the lifecycle semantics before merge.

monkopedia-reviewer (approved):
Tier-1 (internal completion-state change, no public API, apiCheck green; the prior pass over-escalated to tier-3 on a misread conditional — corrected). Maintainer greenlit the feature (#139) and confirmed the merge. Reviewed clean: upstream-faithful explicit-session lifecycle (backspace-to-from keeps explicit open, past-from closes, anchored validFor, selection-reset-on-refinement), no regression to existing completion tests; rebased onto main (CHANGELOG kept all bullets), :autocomplete:jvmTest + apiCheck green. Auto-merge on green.

monkopedia-reviewer (approved):
Re-approve after CHANGELOG rebase onto main (the #154 cascade; all bullets kept). Content unchanged from prior tier-1 approval. Auto-merge on green.

Issue #152, opened 2026-06-09
Crash: getCursorRect offset out of bounds in caret-reveal/scroll path (KodeMirror.kt)


monkopedia-coder:
## Symptom
`kotlin.IllegalArgumentException: offset(96) is out of bounds [0, 95]` during editing (vim + LSP). Not a `DocPos` throw — it's Compose `getCursorRect` rejecting an offset past the laid-out text.

## Root cause
`view/.../KodeMirror.kt` (~line 368), the caret-reveal / horizontal-scroll path:
```kotlin
val offsetInLine = (request.target - line.from.value).coerceIn(0, line.to.value - line.from.value)  // clamped to CURRENT doc line len
val mappedOffset = mapTabOffset(offsetInLine, line.tabOffsetMap)
val cursorX = layout.getCursorRect(mappedOffset).left   // ← throws
```
`offsetInLine` is clamped to the current **doc** line length, but `getCursorRect` runs against `layout` (a cached `TextLayoutResult`) which lags the doc by a frame during typing (shorter), and `mapTabOffset` can expand the offset past it → `mappedOffset` (96) > layout text length (95) → out of bounds. Same family as the caret #146/#147 and DocPos #127, different site: the drawing-side sites (drawBlockCursor `< textLen`, drawLineSelectionRange `coerceAtMost(textLen)`, drawLineCursor #147) are guarded; this reveal/scroll path is not.

## Fix
Clamp the offset to the **LAYOUT's** text length (not the doc line) before querying it:
```kotlin
val layoutLen = layout.layoutInput.text.length
val cursorX = layout.getCursorRect(mappedOffset.coerceIn(0, layoutLen)).left
```
Also audit for any other bare `getCursorRect(` / `getBoundingBox(` / `getHorizontalPosition(` not coerced to the *layout's* length — the invariant is "clamp the offset to the LAYOUT you're querying, not the doc line."

Reported + root-caused by the konstructor consumer.

PR #156, opened 2026-06-09
Make scroll-into-view collector survive a failing reveal request (#155)

### LINKED ISSUE #155: Scroll-into-view coroutine dies permanently if a reveal request throws (defense-in-depth)
## Context
Follow-up to #152 (getCursorRect out-of-bounds). Beyond the visible exception, that crash had a worse second symptom Jason hit: **once it fired, the cursor no longer auto-scrolled into view for the rest of the session.**

## Mechanism
The caret-reveal / scroll-into-view handler in `view/.../KodeMirror.kt` (~L333-403) is ONE long-lived coroutine:
```kotlin
LaunchedEffect(session) {
    snapshotFlow { impl.scrollRequest }.collect { request ->
        ... scrollItemIntoView(...); if (!wrapLines) { ... layout.getCursorRect(mappedOffset).left ... }; impl.consumeScrollRequest(request)
    }
}
```
When the body threw (the #152 case), the exception propagated out of `collect` → the `LaunchedEffect` coroutine completed exceptionally and was NOT relaunched (`session` key unchanged) → the `snapshotFlow` collection was dead → every subsequent `scrollRequest` (every cursor move) silently dropped. Hence "cursor off-screen won't scroll back."

#154 (the #152 clamp) removes the known throw, so both symptoms resolve on a build with it. This issue is the **defense-in-depth**: the collector must survive ANY single failure.

## Fix
Wrap the per-request body in try/catch so one bad reveal can never permanently kill the handler:
```kotlin
.collect { request ->
    try { ...existing body... }
    catch (t: Throwable) { /* log; still consumeScrollRequest(request); keep collecting */ }
}
```
Then AUDIT the other long-lived `snapshotFlow{...}.collect{...}` / similar collect loops in `:view` and apply the same per-iteration resilience wherever a throw in the body would permanently break a session-scoped handler.

Reported + root-caused by the konstructor consumer.

monkopedia-coder:
## Summary
Defense-in-depth follow-up to #152/#154. The caret-reveal / scroll-into-view handler in `view/.../KodeMirror.kt` was a single long-lived `LaunchedEffect(session) { snapshotFlow { impl.scrollRequest }.collect { ... } }`. If the per-request body threw (the #152 `getCursorRect` out-of-bounds case), the exception propagated out of `collect`, the `LaunchedEffect` completed exceptionally and was never relaunched (key `session` unchanged) — so **scroll-into-view stayed dead for the rest of the session** (every subsequent cursor move silently dropped).

## Fix
- Wrap the per-request collect body in `try/catch`:
  - happy path is preserved exactly (still calls `impl.consumeScrollRequest(request)` at the end);
  - on `Throwable`, the failure is swallowed and the request is **still consumed** so a persistently-bad request isn't retried forever, and collection continues;
  - `CancellationException` is rethrown (`if (t is CancellationException) throw t`) so structured cancellation still tears the coroutine down when the effect disposes.

## Audit of other `:view` collect loops
Audited every long-lived `snapshotFlow{...}.collect{...}` / session-scoped `LaunchedEffect` collect loop in `:view`:
- **`scrollRequest` reveal collector** (the bug) — fixed as above.
- **Viewport-tracking / cache-eviction collector** (`snapshotFlow { lazyState.layoutInfo.visibleItemsInfo... }.collect { ... lineLayoutCache.evict(...) }`, also `LaunchedEffect(session)`) — same failure mode (a throw permanently breaks viewport reporting + cache eviction for the session). Applied the same per-iteration `try/catch` with `CancellationException` rethrown.
- **`horizontalScrollState.value` mirror collector** (`.collect { impl.lastHorizontalScrollPx = it }`) — body is a single `Int` assignment with no realistic throw; left unchanged.

No other session-scoped collect loops exist in `:view` (the remaining `LaunchedEffect`s are `Unit`-keyed one-shots like the initial focus request).

## Tests
The collect body is genuinely Compose-`LaunchedEffect`-bound: it reads `impl.scrollRequest` via `snapshotFlow` and depends on a live `LazyListState`, `textLayoutResults`, and editor coordinates. It can't be exercised headlessly without a Compose runtime, so no extracted unit test is added — relying on the structural fix rather than fabricating a test that doesn't hit the real path. `CHANGELOG.md` updated under Unreleased / Fixed.

## Verification
`:view:jvmTest :view:verifyRoborazziJvm :view:spotlessApply :view:apiCheck` all green locally.

Fixes #155.


monkopedia-reviewer (approved):
Reviewed as `monkopedia-reviewer`. This is a clean **robustness / bug_fix** (tier 1) — no public API change, no behavior change on the happy path.

**CancellationException rethrown in both collectors:** ✓
- *scrollRequest reveal collector*: `catch (t: Throwable) { if (t is CancellationException) throw t; impl.consumeScrollRequest(request) }` — rethrown on the first line, request still consumed in both the happy-path (`try` block tail) and the failure path (`catch`). Structured cancellation preserved.
- *viewport-tracking / cache-eviction collector*: `catch (t: Throwable) { if (t is CancellationException) throw t }` — rethrown. Permanent loss of viewport reporting + cache eviction on a single throw is now prevented.
- *horizontalScrollState mirror collector*: correctly left unchanged (single Int assignment, no realistic throw path).

**Other checks:**
- `import kotlinx.coroutines.CancellationException` added (imports-over-FQN). ✓
- CHANGELOG `### Fixed` entry references `#155`. ✓
- No public API surface change (`apiCheck` passes locally per PR body). ✓
- Test note is honest: the collect body is `LaunchedEffect`-bound and not headlessly testable; the value is the resilience structure, not a fabricated test. Acceptable.

Approving and setting auto-merge — will squash when CI (check / check-macos / check-ios) goes green.

PR #157, opened 2026-06-09
Drop stale async completion results instead of re-opening the popup (#153)

### LINKED ISSUE #153: Vim Escape needs multiple presses to exit insert mode (autocomplete/snippet Escape bindings)
## Symptom
In vim insert mode with LSP, exiting insert mode takes more Escape presses than expected (~2: close popup, then exit insert). Reported by Jason via konstructor.

## Likely cause
Two higher-precedence Escape bindings sit above vim's insert→normal handler:
- `autocomplete/View.kt:139` — `KeyBinding("Escape", closeCompletion)`
- `autocomplete/Snippet.kt:283` — `KeyBinding("Escape", clearSnippet)`

With an LSP completion popup open AND a snippet field active (from accepting a snippet completion), Escape is consumed by `closeCompletion` → then `clearSnippet` → only the third reaches vim.

## Investigate
- (a) Precedence/order of these vs the vim Escape handler.
- (b) Whether `closeCompletion` / `clearSnippet` consume Escape (return true) when there's **nothing** to close — that would eat keystrokes. They should only handle Escape when a popup/snippet is actually active.
- (c) Whether the popup auto-re-triggers between presses.

Repro: vim insert mode → type to get a completion → Escape. Needs-decision because the desired precedence (does closeCompletion take priority over vim exit, or should one Escape do both?) is a UX call.

monkopedia-coder:
## Summary

Fixes #153 (re-scoped to the real cause: the async LSP completion re-open race).

An async (`asyncOverride` / LSP) completion source is launched on the editor's
coroutine scope and dispatches its result when it resolves. Previously that
dispatch was **unconditional**: a request still in flight when the user moved
the cursor, changed the typed token, or closed the popup (e.g. pressed Escape)
would resolve and **re-open** the popup, requiring another Escape. That is the
mechanism behind the "Escape needs multiple presses to exit vim insert mode"
symptom — an in-flight LSP request re-opening the completion list after the
Escape that closed it.

## Fix

`triggerCompletion` now captures the trigger position at launch and, before
opening the popup, re-validates the resolved result against the **live**
`view.state` (the coroutine may have suspended across doc/selection edits). The
guard mirrors the in-place re-filter already done in `completionStateField`
(anchored `validFor.matches` over the `from`..head span, #139). A result is
dropped instead of re-opening when:

- the primary selection is no longer an empty cursor (a range, or the context moved),
- the cursor moved before the result's `from` (`head < from`), or
- the `from`..head text no longer matches `validFor` (the typed token changed/cleared).

When `validFor` is absent there is no reuse rule, so the result is valid only at
the unchanged trigger position. A still-relevant result opens the popup exactly
as today. No public API change.

## Tests

New `AsyncCompletionRaceTest` (`:autocomplete` jvmTest) drives the race
deterministically: a suspend `asyncOverride` source gated on a
`CompletableDeferred`, run on a `Dispatchers.Unconfined` session so the launched
coroutine suspends at `gate.await()` and resumes inline on `complete()`. Tests:
the control (state unchanged → popup opens), cursor moved before `from`,
typed token breaks `validFor`, and a non-empty selection — the last three all
assert `completionStatus == null` (popup stays closed).

Local verify green: `:autocomplete:jvmTest :lsp-client:jvmTest
:autocomplete:spotlessApply :autocomplete:apiCheck`.


monkopedia-reviewer (approved):
Reviewed as `monkopedia-reviewer`.

**Category:** `bug_fix` | **Tier:** 1 | **Verdict:** Approve

---

### Guard correctness

`asyncResultApplicable` correctly gates all three stale-result scenarios before `startCompletionEffect` is dispatched:

- **Non-empty selection** (`!main.empty`) — range selections mean the cursor context has moved.
- **Cursor before `from`** (`head < result.from`) — the trigger anchor is no longer under the cursor.
- **Token changed** (`validFor.matches(doc.sliceString(from, head))`) — the typed span since launch no longer satisfies the completion's reuse rule.
- **No `validFor`** — falls back to `head == triggerPos` (identity check), which is the conservative correct default when there's no reuse rule.

The function uses `Regex.matches` (full-span anchor) where the existing `completionStateField` re-filter uses `containsMatchIn`. `matches` is actually the semantically correct choice here — it validates that the *entire* from..head span satisfies `validFor`, not merely that some substring does. The minor pre-existing inconsistency in State.kt's `containsMatchIn` is out of scope; this PR is stricter (correct direction) and introduces no regression.

The trigger-position capture (`val triggerPos = pos`) is correctly placed before the `launch` so it reflects the state at dispatch time rather than at resume.

### Test determinism

`AsyncCompletionRaceTest` is deterministic: the `CompletableDeferred` gate + `CoroutineScope(Dispatchers.Unconfined)` means the launched coroutine runs inline up to `gate.await()`, the test body then synchronously mutates state, and `gate.complete(Unit)` resumes inline — all assertions follow in the same call stack. No sleeps, no timing dependencies. The test drives the real `asyncOverride` path through `triggerCompletion`/`asyncResultApplicable`, not a mock. Four cases: control open, cursor-before-from drop, validFor-mismatch drop, non-empty-selection drop.

### Documented residual

A pure Escape with no cursor movement (popup closed but cursor stayed, token unchanged) would still pass the guard. The PR body correctly scopes this out — the reported symptom (vim insert-mode escape, fast typing) always involves cursor movement before the LSP response resolves. Not a regression; acceptable follow-up if desired.

### Mechanics

- CHANGELOG ### Fixed references #153. ✓
- No public API change; all new code is `private` or `jvmTest`. `apiCheck` clean per PR body. ✓
- Imports-over-FQN. ✓
- No style violations.

CI pending (check / check-macos / check-ios in progress) — setting auto-merge to squash when checks pass.

monkopedia-reviewer (approved):
Re-approve after CHANGELOG rebase onto main (#156 cascade; all bullets kept). Content unchanged from prior tier-1 approval. Auto-merge on green.

PR #154, opened 2026-06-09
Clamp getCursorRect offset to layout length in caret-reveal path (#152)

monkopedia-coder:
## Summary
Editing (vim + LSP) could crash with `IllegalArgumentException: offset(N) is out of bounds` from Compose's `getCursorRect`. The caret-reveal / horizontal-scroll path in `KodeMirror.kt` (~line 368) clamped the offset to the **current doc line length**, but queried `layout` — a cached `TextLayoutResult` that lags the doc by a frame during typing — and `mapTabOffset` can expand the offset further. So `mappedOffset` could exceed the layout's text length and `getCursorRect` threw.

Fix: clamp the offset to the **layout's** text length (`layout.layoutInput.text.length`) before querying it, matching the issue's prescribed fix and the invariant already followed by the drawing-side sites.

```kotlin
val layoutLen = layout.layoutInput.text.length
val cursorX = layout.getCursorRect(mappedOffset.coerceIn(0, layoutLen)).left
```

## Audit
Grepped every `getCursorRect(` / `getBoundingBox(` / `getHorizontalPosition(` call in `:view`. The invariant is: clamp the offset to the LENGTH OF THE LAYOUT being queried, not the doc line.

- `KodeMirror.kt:368` (reveal path) — **was unguarded; fixed** (the issue's site).
- `LineLayout.kt:114` `LineLayoutCache.coordsAtPos` — clamped to `line.text.length` (DOC line length) while querying `layout.result`. Same latent layout-lag bug (a stale tooltip/caret-reveal coord query during a shrinking edit) — **fixed** with the same layout-length clamp.
- `SelectionDrawing.kt` drawBlockCursor (`offsetInLine < textLen`), end-of-line `getHorizontalPosition(textLen, …)`, drawLineCursor (`cursorLayoutOffset`, #147), drawLineSelectionRange (`coerceAtMost(textLen)`) — all already clamp to the LAYOUT length. No change needed.

## Test
No new headless test. The real failing path runs `getCursorRect` against a live Compose `TextLayoutResult`, which (per `LineLayoutTest`'s own note) cannot be instantiated in the headless test environment, so the layout-vs-doc length mismatch can't be reproduced headlessly. The fix is a defensive clamp; its pure clamp arithmetic (`coerceIn(0, layoutLen)`) is identical to the already-tested `cursorLayoutOffset` helper (`CursorLayoutOffsetTest`, the analogous #146/#147 pattern). I did not fabricate a test that doesn't exercise the real path.

## Verification
[...]

PR #160, opened 2026-06-11
Reveal the primary selection on vim motions (gg/G, offscreen j/k) (#158)

### LINKED ISSUE #158: Vim motions don't scroll the cursor into view (gg/G, offscreen j/k, insert-scroll jump-to-top)
## Symptoms (all in vim mode, doc longer than one viewport)
1. `gg` / `G` move the cursor to top/bottom but the viewport does NOT follow — cursor ends offscreen.
2. Moving the cursor offscreen (`j`/`k` past the viewport edge) doesn't auto-scroll it back into view.
3. After adding lines in insert mode then scrolling down with the cursor, the viewport instantly JUMPS TO THE TOP.

Live blocker for a user editing in vim on adolin. Reported + root-caused by the konstructor consumer.

## Root-cause lead (konstructor)
Vim motions run inside `cm.operation { }`, so `curOp != null`. In `VimEditor.setCursor` (`vim/.../VimEditor.kt:779`) the motion's transaction sets `scrollIntoView = curOp == null` → **false for every in-operation motion**. So vim motions never scroll on their own transaction; they depend entirely on the recovery in `onBeforeEndOperation()` (`VimEditor.kt:609-623`), which only fires `scrollIntoView()` when `op.cursorActivity == true && op.isVimOp == true`. That recovery is fragile:
- It relies on the VimExtension plugin's `update()` having already called `onSelectionChange()` (which sets `cursorActivity`) before the operation's finally block runs — a timing dependency that may not hold for `gg`/`G`.
- It requires `op.isVimOp == true` — worth checking `gg`/`G`'s op actually carries that flag.
- The recovery dispatches a **no-selection** `scrollIntoView()` (`VimEditor.kt:849` → `TransactionSpec(scrollIntoView = true)`); the view resolves the target as `tr.state.selection.main.head` (`view/.../EditorSessionImpl.kt:144`). Correct ONLY if the selection is already at the destination when the recovery fires; a stale/intermediate current state would make head resolve to offset 0 → the "jump to top" (#3).

## Ask
Take the vim↔scroll integration as a unit: make in-operation vim motions **reliably reveal the primary selection** — either set `scrollIntoView` on the motion transaction itself, or make `onBeforeEndOperation` robust to the timing AND always target the *final* selection head. Add coverage to `ScrollIntoViewTest.kt` (no `gg`/`G` or offscreen-`j`/`k` coverage today).

(Distinct from the view-layer scroll fixes #154/#156, which are merged — this is the vim-integration layer.)

monkopedia-coder:
Fixes #158.

## Problem
In vim mode on a document larger than one viewport, `gg`/`G` moved the cursor but the viewport did not follow (cursor offscreen), moving the cursor offscreen with `j`/`k` did not auto-scroll it back, and an operation could yank the viewport to the top.

## Root cause (verified against the code)
Vim commands run inside `cm.operation { }`, so `curOp != null`. Each motion's `setCursor` (`VimEditor.kt`) dispatches its transaction with `scrollIntoView = curOp == null` → **false** for every in-operation motion. The only auto-reveal was a recovery in `onBeforeEndOperation()`, gated on `op.cursorActivity && op.isVimOp`. `op.cursorActivity` is only set by `onSelectionChange()`, which in production comes from the VimExtension plugin's `update()` firing on `selectionSet` — a timing dependency that is unreliable for in-operation motions like `gg`/`G` (and never fires at all in headless mode, so the bug is fully reproducible in tests). Worse, when the recovery *did* fire on an operation that had not actually moved the cursor, the no-selection `scrollIntoView()` revealed a stale head → the "jump to top".

## Fix
Make the reveal independent of the fragile `cursorActivity` flag and gate it on real cursor movement:

- `Operation` now records `startSelectionHead`, captured in `operation()` when the operation begins (always before any vim command body runs, so it is the pre-command head for every vim op).
- `onBeforeEndOperation()` reveals the cursor when `op.isVimOp && startHead != finalHead` — i.e. only when the operation actually moved the primary head.

The no-argument `scrollIntoView()` already resolves its target to the **final** `selection.main.head` (`EditorSessionImpl.scrollRequest`), so:
- `gg`/`G` and offscreen `j`/`k` now reliably reveal their **destination** (fixes symptoms 1 & 2);
- the reveal targets the real destination head, never a stale offset 0, and a non-moving operation no longer queues a reveal that would yank the viewport (fixes symptom 3, jump-to-top).

It keeps using the no-selection `scrollIntoView()` (rather than re-dispatching a `CursorSpec`) so visual-mode selections are not collapsed.

## Tests
New `vim/.../VimScrollIntoViewTest.kt` drives the headless vim harness and records, for every dispatched transaction, the same `(tr.scrollIntoView, tr.state.selection.main.head)` pair the view layer uses to build its scroll request:
- `gg` from below the fold reveals offset 0;
- `G` reveals the last line's head;
- offscreen `j` and `k` reveal the destination head (asserted `> 0`, not a stale 0);
- a non-moving `gg` (already at top) queues **no** reveal — the jump-to-top guard.

(Lives in `:vim` rather than `:view`'s `ScrollIntoViewTest.kt` because `:view` cannot depend on `:vim`; it asserts the identical seam.)

## Verification
`:vim:jvmTest`, `:view:jvmTest`, `:view:verifyRoborazziJvm` (caret-reveal #33/#58/#69 + #154/#156 baselines unchanged), `:vim:spotlessApply`, `:view:spotlessApply`, `:vim:apiCheck`, `:view:apiCheck` all green. No public API change.


monkopedia-reviewer (approved):
Tier-1 auto-review (bug_fix, Monkopedia/kodemirror baseline tier 1).

**Root cause validated.** The diff correctly decouples the reveal from the fragile `cursorActivity`/`onSelectionChange()` timing dependency: `Operation.startSelectionHead` is captured in `operation()` (before any motion body runs), and `onBeforeEndOperation()` reveals on `isVimOp && startHead != finalHead`.

**Three symptoms, all addressed:**
1. gg/G leaving cursor offscreen — movement gate fires, `scrollIntoView()` reveals the real destination.
2. Offscreen j/k not scrolling back — same path, destination head is correct.
3. Jump-to-top — a non-moving operation has `startHead == finalHead`, so `movedCursor = false`, no reveal queued. Confirmed.

**Target resolution correct.** `finalHead = session.state.selection.main.head` reads the FINAL state after all transactions have been dispatched. The no-arg `scrollIntoView()` passes this to the view layer as the scroll target — never a stale offset 0. Confirmed.

**Non-vim reveal not regressed.** Gate is `op.isVimOp`, so non-vim paths (#33/#58/#69 caret-reveal, #154/#156 fixes) are fully untouched. The roborazzi baselines unchanged per the coder's verification.

**Visual mode not regressed.** The no-arg `scrollIntoView()` does not dispatch a `CursorSpec` transaction; it only sets the scroll target on the view layer. Visual-mode selections in `session.state.selection` are not touched.

**Lazy-op null safety.** The `op.startSelectionHead != null` guard in `movedCursor` means an operation opened outside `operation()` (e.g. a change handler) never fires a spurious reveal. Correct.

**Tests meaningful.** `VimScrollIntoViewTest` drives the real operation lifecycle via `typeKey`/`EditorSession` dispatch, recording `(tr.scrollIntoView, tr.state.selection.main.head)` — the identical seam `EditorSessionImpl` consumes. All 5 cases present: gg reveal, G reveal, j destination, k destination, non-moving guard. Placement in `:vim` (not `:view`) is the right call since `:view` cannot depend on `:vim`.

**Code quality.** KDoc on `startSelectionHead`, explanatory block comment in `onBeforeEndOperation`, imports used (no FQN), idiomatic `also {}`. CHANGELOG ### Fixed references #158. Spotless + apiCheck green per coder.

CI gates pending — approving now, auto-merge queued.

PR #167, opened 2026-06-18
Key scroll state to the session so swapping documents doesn't wedge scrolling (#166)

### LINKED ISSUE #166: Vertical scroll wedges after swapping EditorSession (lazyState not keyed on session)
## Symptom
Swapping the `EditorSession` passed to `KodeMirror` back and forth several times (e.g. switching documents) eventually wedges vertical scrolling — the view can no longer scroll top→bottom. Reported by the konstructor downstream consumer (Jason hits it after several doc switches).

## Root cause
In `view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt`, the per-session remembered state is keyed on `session` for every sibling:
- `pluginHost` — `remember(session)` (L153)
- `lineLayoutCache` — `remember(session)` (L156)
- `pendingLineLayouts` — `remember(session)` (L160)
- `textLayoutResults` — `remember(session)` (L165)

…but the scroll states are **not**:
- `lazyState = rememberLazyListState()` (L219) — vertical
- `horizontalScrollState = rememberScrollState()` (L224) — horizontal (same latent bug)

So on a session swap, all per-doc state resets except the scroll state, which carries the previous document's `firstVisibleItemIndex`/offset/scroll-range. Applied to a different-length document (and accumulating across swaps), the stale range clamps scrolling and the view wedges.

## Fix
Key both scroll states to the session like their siblings:
```kotlin
val lazyState = remember(session) { LazyListState() }
val horizontalScrollState = remember(session) { ScrollState(0) }
```

## Test
A `runDesktopComposeUiTest` (view jvmTest, cf. `RecompositionTest`) that swaps between two long sessions: scroll session A down (`lastFirstVisibleItem > 0`), swap to session B, assert B starts at `lastFirstVisibleItem == 0` (fresh scroll state). Diagnosed with the konstructor agent.

monkopedia-coder:
## Problem

In `KodeMirror.kt`, every per-session remembered value is keyed on `session` (`pluginHost`, `lineLayoutCache`, `pendingLineLayouts`, `textLayoutResults`) — except the two scroll states, which used `rememberLazyListState()` / `rememberScrollState()` (un-keyed). When a downstream consumer swaps the `EditorSession` passed to `KodeMirror`, all per-doc state resets except the scroll state, which carries the previous document's `firstVisibleItemIndex`/offset and scroll range into the new document. Across swaps between different-length docs this clamps scrolling and wedges the view (can't scroll top→bottom).

## Fix

Key both scroll states to the session like their siblings:

```kotlin
val lazyState = remember(session) { LazyListState() }
val horizontalScrollState = remember(session) { ScrollState(0) }
```

Removed the now-unused `rememberLazyListState` / `rememberScrollState` imports.

## Test

Adds `SessionSwapScrollTest` (`runDesktopComposeUiTest`): renders a 200-line doc under a bounded height, scrolls it down via the caret-reveal path (cursor to end + `scrollIntoView`), asserts it scrolled (`lastFirstVisibleItem > 0`), swaps to a second 200-line doc, and asserts the new session renders from the top (`lastFirstVisibleItem == 0`). Verified the test FAILS on the unfixed code (session B inherited index 182) and PASSES with the fix.

## Verification (local)
- `:view:jvmTest` green (incl. new test)
- `spotlessApply` / `ktlintFormat` clean
- `:view:apiCheck` green (not a public-API change)

Fixes #166


monkopedia-reviewer (approved):
Tier-1 internal fix. Local: jvmTest green (new regression test passes), view spotlessApply/ktlintFormat clean on all PR files, view apiCheck green (no public-API change). Scroll states correctly keyed to session via remember(session). No leak introduced. Arming auto-merge.

PR #168, opened 2026-06-24
Route all hit-testing through the live layout (fix #165)

### LINKED ISSUE #165: Click selection lands on wrong position in a scrolled document (hit-test ignores scroll offset)
## Symptom
In a scrolled document, **mouse click selection lands on the wrong position** — off by ~half a dozen lines vertically AND some X. Keyboard/vim navigation is fine; only mouse-click hit-testing is wrong. High priority — blocks basic editing. Reported by Jason via konstructor on 0.3.4-SNAPSHOT @7d1752d.

## Not authored by #158 — surfaced by it
Clicks were correct at @bb8accb. The only functional change since is #158 (vim↔scroll, `VimEditor.kt`), which has no click/coord code. konstructor's read (likely correct): **#158 made vim motions actually reveal/scroll the viewport, so the doc now genuinely scrolls — exposing a latent click hit-test bug that doesn't account for the vertical scroll offset (and possibly horizontal).**

## Candidate area (`view/.../KodeMirror.kt`)
- `posFromVisibleItems` (~L1365-1404) and the fallback `session.posAtCoords` paths (click ~L602, drag ~L548/556, hover ~L629).
- The horizontal path already adds `horizontalScrollPx` (~L1390); the **vertical analog may be missing** on the click coordinate.

## Investigate carefully (don't assume)
- **Verify the Compose semantics:** is `LazyListLayoutInfo.visibleItemsInfo[*].offset` viewport-relative (standard Compose) or document-relative in this setup? If viewport-relative, the current compare is correct and the bug is subtler (stale scroll state, or the `pointerInput` `Offset` coordinate space).
- The **X-offset-too** detail argues it's NOT a pure vertical-scroll fix — possibly a coordinate-origin/translation issue affecting both axes.
- Harness: `LineLayoutTest` + the first-click-placement tests (#61/#70/e6a12dc) — add a **click-while-scrolled** case.

Stopgap: rolling Jason back to @bb8accb restores clicks (at the cost of the #158 vim-scroll fix).

monkopedia-coder:
## Summary

Fixes #165 — clicks in a scrolled document landed on the wrong position (off by several lines plus some x-offset).

**Root cause:** `LineLayoutCache` stores per-line `topPx`/`leftPx` in `onGloballyPositioned`, which does not reliably re-fire after a scroll-without-recomposition (notably on CanvasKit/wasm). The cache went stale; the live hit-test path (`posFromVisibleItems`, which reads `lazyState.layoutInfo` each call) stays correct, but clicks fell back to the stale cache whenever `posFromVisibleItems` returned null.

**Fix:** route every hit-test (tap, drag, rectangular drag, hover) through a single `resolvePos` resolver that reads the live layout, and make `posFromVisibleItems` never return null while text lines are visible (extracted `chooseHitLineIndex`, clamps `localY`). The stale cache is no longer consulted for clicks.

## Changes
- `KodeMirror.kt` — `resolvePos` closure used by all gesture sites; `posFromVisibleItems` never-null-while-visible; extracted `chooseHitLineIndex`.
- `InputHandling.kt` — `handleDrag`/`handleRectangularDrag` take a `posAt` resolver instead of calling `view.posAtCoords` directly.
- `ChooseHitLineIndexTest.kt` — unit tests for line selection (contains/gap/above/below/empty/single).
- `CHANGELOG.md` — Fixed entry for #165.

## Verification
- `:view:jvmTest` + `:view:apiCheck` green locally.
- Verified on adolin (live wasm/CanvasKit) by the konstructor downstream: the gross stale-cache offset is gone — cursor lands on the correct line across multiple clicked lines in a scrolled document.

A minor residual (~1/4 char rounding offset in posAtCoords) is tracked separately as a follow-up polish; not part of this fix.


monkopedia-reviewer (approved):
Tier-1 (internal hit-testing fix + new unit test; apiCheck green, no public-API change).

Reviewed the live-layout routing: `resolvePos` is now the single hit-test entry point at every gesture site (tap, drag, rectangular drag, hover, drop-cursor), resolving via the live `posFromVisibleItems` and only falling back to `session.posAtCoords` when no text lines are visible (empty doc / first frame). No gesture path consults the stale `LineLayoutCache` for absolute line position anymore — correct root-cause fix for #165.

`chooseHitLineIndex` is correct: containing line, then first line below `y` for gaps/above-all, then last line for below-all, null only when empty. Matches its 6-case unit test (all passing). `posFromVisibleItems` never returns null while text lines are visible; `localY` clamped to the line box. Per-hit FloatArray allocation is bounded by visible-line count and off the render hot path — acceptable. Composes cleanly with #166's session-keyed scroll state.

Local re-verify: :view:jvmTest green (ChooseHitLineIndexTest 6/6), :view:apiCheck green (no apiDump). PR's own files are ktlint-clean (the lone ktlintCheck failure is pre-existing unrelated debt in TooltipPositionTest.kt, which does not gate CI).

Deterministic guard (chooseHitLineIndex unit test) is the meaningful one here since the #165 trigger is wasm/CanvasKit-specific and won't reproduce on JVM; the e2e regression guard correctly lives in the gap-analysis Playwright specs against adolin, and the full scroll matrix is scoped to the follow-up reveal-pipeline work — not a blocker for this focused fix.

Approving; enabling squash auto-merge to land on green CI.

Issue #169, opened 2026-06-24
posAtCoords: ~1/4 char rounding offset (x and y) on wasm/CanvasKit


monkopedia-coder:
Follow-up polish from #165 (PR #168).

After #168 routes hit-testing through the live layout, the gross stale-cache offset is fixed and clicks land on the correct line. A minor residual remains: on adolin (live wasm/CanvasKit, dpr-scaled), the landing point is consistently off by ~1/4 character in **both** x and y — "very close but not exact," reproducible.

Suspected cause: rounding/anchor handling in `posAtCoords`:
- column: rounding-to-nearest vs floor on the character cell
- row: baseline-vs-top half-cell offset

Likely only visible under dpr-scaled CanvasKit (physical-vs-CSS px), which is why the deterministic JVM hit-test test does not catch it.

Next step: konstructor to capture exact pixel deltas via a Playwright probe against adolin to pin the rounding direction, then adjust the anchor math in `posAtCoords`.

PR #170, opened 2026-06-24
Account for contentTopPadding in the hit-test grid (vertical hit-test offset, #169)

monkopedia-coder:
## Summary

Fixes the **vertical** component of #169 — clicks landed ~a quarter-line low.

`posFromVisibleItems` built its line grid from `LazyListItemInfo.offset`, which starts at `y = 0`, but the `LazyColumn` renders its content shifted down by `contentTopPadding` (the content padding sits above the first item), while the pointer `offset.y` is in that padded space. The hit-test grid was therefore ~`contentTopPadding` px (~4.dp, about a quarter line) above the glyphs, so a click toward the lower part of a line resolved to the line *below* it. The resolver now adds `contentTopPadding` to each item top (and to the within-line `y`), aligning the hit-test with the rendered caret geometry — the same theme constant the `LazyColumn` already uses, no magic number.

## Verification
- **JVM regression guard** (`HitTestVerticalAnchorTest`): clicks just below a glyph's bottom edge — the band where the missing padding flipped the resolved line. Verified *failing* without the fix and *passing* with it (centre-of-line clicks pass even with the bug, which is why the existing `ClickTargetingTest` cases didn't catch it).
- **In-house dpr-scaled CanvasKit probe**: the measured vertical offset drops from ~0.28 line to sub-pixel at dpr 1.0 and 2.0.
- **Confirmed on adolin** (live wasm/CanvasKit, Jason hands-on): "vertically just about right on."
- `:view:jvmTest` + `:view:apiCheck` green (no public-API change).

## Scope
This is the vertical fix only. A separate, larger **horizontal** drift (~3/4 char, gutter/lineWrapping config) was surfaced once the vertical bias was removed; it does **not** reproduce in layout-space on JVM (coordsAtPos↔hit-test round-trip is clean through the gutter), so it's wasm/CanvasKit-specific or a rendered-vs-layout (font-metrics) divergence and is still under investigation. #169 stays open for it.


monkopedia-reviewer (approved):
Tier-1 bug fix (private hit-test signature change + new internal JVM test; `:view:apiCheck` green, no public-API change).

Correctness verified: the LazyColumn renders content with `contentPadding(top = theme.layout.contentTopPadding)` (KodeMirror.kt:863) while `LazyListItemInfo.offset` starts at y=0, so the live hit-test grid was ~contentTopPadding (4.dp) above the glyphs. `posFromVisibleItems` now adds `contentTopPaddingPx` to BOTH the `tops` selection grid and the `itemTop` used for `localY`, and the value is threaded from `resolvePos` as `theme.layout.contentTopPadding.toPx()` — principled, the actual theme constant, not a tuned magic number. Change is confined to the live vertical hit-test mapping; no other gesture path touched.

The JVM guard (HitTestVerticalAnchorTest) is meaningful: it derives the click-y from `coordsAtPos` geometry (`rect.bottom + 1.5f`), not hardcoded metrics, clicking the inter-line band where the missing padding flipped the line.

Local re-verification (fast first pass): `:view:jvmTest` BUILD SUCCESSFUL (incl. HitTestVerticalAnchorTest), `:view:apiCheck` BUILD SUCCESSFUL. Horizontal drift (#169) intentionally out of scope. Enabling auto-merge; the three required GitHub checks (check / check-macos / check-ios) gate the actual merge.

PR #172, opened 2026-07-08
Remove dead hasRun field in LinterPlugin (#164)

### LINKED ISSUE #164: Remove dead hasRun field in LinterPlugin (lint)
**Category:** dead-code
**Where:** `lint/src/commonMain/kotlin/com/monkopedia/kodemirror/lint/LintState.kt` — `LinterPlugin` (lines 167-208)

**Current:** The synchronous `LinterPlugin` declares a `hasRun` field that is written but never read:
```kotlin
private var hasRun = false
...
private fun runLinter() {
    val diagnostics = source(view)
    hasRun = true        // set here and in no-op anywhere else
    view.dispatch(...)
}
```
`grep` confirms the only references are the declaration and this assignment — nothing ever reads `hasRun`.

**Proposed:** Delete the `hasRun` field and its assignment.

**Why:** Dead state on a plugin reads as if it gates behavior (it doesn't), so removing it prevents a future reader from assuming the debounce logic depends on it. Trivial and safe.

**Effort:** mechanical

monkopedia-coder:
Removes the dead `hasRun` field in `LinterPlugin` (`:lint`, `LintState.kt`). The `private var hasRun` was written (`hasRun = true` in `runLinter()`) but never read — `grep -rn hasRun` confirms the declaration and this one assignment were the only references, with no readers. Dead state on a plugin reads as if it gates behavior (it doesn't), so removing it prevents a future reader from assuming the debounce logic depends on it.

Pure dead-code removal, no behavior change.

Verification (local):
- `:lint:jvmTest` — green
- `:lint:spotlessApply :lint:ktlintFormat` — clean, no diff to unrelated files
- `:lint:apiCheck` — green (`hasRun` is private → no public-API change, tier-1)

CHANGELOG updated under [Unreleased] → Changed referencing #164.

Fixes #164


monkopedia-reviewer (approved):
Tier-1: removes the private, write-only `hasRun` field in LinterPlugin — pure dead-code removal, no behavior change and no public-API surface touched. Verified locally: `:lint:jvmTest` and `:lint:apiCheck` both green. Approving and enabling squash auto-merge.

PR #173, opened 2026-07-08
Factor search-cursor filtering decorator (#163)

### LINKED ISSUE #163: Factor WholeWordSearchCursor + FilteredSearchCursor into one filtering decorator
**Category:** duplication
**Where:** `search/src/commonMain/kotlin/com/monkopedia/kodemirror/search/SearchQuery.kt` — `WholeWordSearchCursor` (lines 167-204) and `FilteredSearchCursor` (lines 209-238)

**Current:** Both classes are the same "filtering decorator over an `Iterator<SearchMatch>`" with an identical skeleton — a `nextMatch` field, `init { advance() }`, an `advance()` loop that pulls from `inner` until a predicate passes, and identical `hasNext()`/`next()`:
```kotlin
private var nextMatch: SearchMatch? = null
init { advance() }
private fun advance() {
    while (inner.hasNext()) {
        val match = inner.next()
        if (<predicate>) { nextMatch = match; return }
    }
    nextMatch = null
}
override fun hasNext() = nextMatch != null
override fun next(): SearchMatch { val m = nextMatch ?: throw NoSuchElementException(); advance(); return m }
```
The only difference is the predicate: whole-word-boundary vs the `test(from, to, state)` callback.

**Proposed:** Extract a single filtering decorator, e.g. `FilteringSearchCursor(inner, predicate: (SearchMatch) -> Boolean)`, and build both behaviors from it — whole-word via a boundary predicate (keeping the `isWordBoundary`/`isWordChar` helpers), filtered via `{ test(it.from, it.to, state) }`. The two public/internal entry points in `getMatches()` stay the same.

**Why:** Removes ~30 lines of copy-pasted iterator boilerplate and gives one place to fix any future iterator bug. The two cursors are already structurally identical, so the abstraction is justified (not premature).

**Effort:** small refactor

monkopedia-coder:
## Summary
`WholeWordSearchCursor` and `FilteredSearchCursor` in `:search`'s `SearchQuery.kt` were the same "filtering decorator over an `Iterator<SearchMatch>`" with an identical skeleton (a `nextMatch` field, `init { advance() }`, an `advance()` loop pulling from `inner` until a predicate passes, and identical `hasNext()`/`next()`), differing only in the predicate.

This factors them into a single internal `FilteringSearchCursor(inner, predicate: (SearchMatch) -> Boolean)` and expresses both behaviors in terms of it:
- whole-word via a boundary predicate (`wholeWordSearchCursor`, keeping the `isWordBoundary`/`isWordChar` helpers),
- the `test` callback via `{ test(it.from, it.to, state) }`.

Removes ~30 lines of copy-pasted iterator boilerplate and gives one place to fix any future iterator bug. Behavior-preserving; both entry points in `getCursor()` are unchanged.

## Verification
- `:search:jvmTest` — green
- `:search:apiCheck` — green (internal refactor; public API unchanged)
- `spotlessApply` / `ktlintFormat` applied

Fixes #163


monkopedia-reviewer (approved):
Tier-1: factors the two structurally identical filtering decorators (`WholeWordSearchCursor`, `FilteredSearchCursor`) into one `private FilteringSearchCursor(inner, predicate)`, with whole-word matching now expressed as a boundary predicate. Behavior-preserving and the new decorator is `private`, so public API is unchanged. Verified locally: `:search:jvmTest` green (confirms behavior preserved) and `:search:apiCheck` green (confirms no public-API leak). Approving and enabling squash auto-merge.

PR #174, opened 2026-07-08
Remove vestigial indent-context stubs (#161)

### LINKED ISSUE #161: Indent.kt: remove vestigial indent-context stubs (textAfterPos, dead indentString unit + drifted KDoc, ignored bias params)
**Category:** dead-code
**Where:** `language/src/commonMain/kotlin/com/monkopedia/kodemirror/language/Indent.kt`

**Current:** Several vestigial bits in the indent context API, left over from the CM6 port:

1. `IndentContext.textAfterPos` (line 98) is a public property whose getter always returns the empty string:
   ```kotlin
   /** The text content of the document. */
   val textAfterPos: String get() = ""
   ```
   It has no callers anywhere in the repo (`grep` finds only the definition). The real "text after the indent position" lives in `TreeIndentContext.textAfter`.

2. `indentString` (lines 70-75) has a dead local and a KDoc that contradicts the code:
   ```kotlin
   /** ... using tabs if the state's [indentUnit] divides evenly, otherwise spaces. */
   fun indentString(state: EditorState, cols: Int): String {
       val unit = getIndentUnit(state)   // never used
       if (cols < 0) return ""
       // Always use spaces for consistency
       return " ".repeat(cols)
   }
   ```
   `unit` is computed and discarded; the KDoc promises tab handling that the body explicitly does not do (the inline comment even says "Always use spaces").

3. `IndentContext.lineAt(pos, bias)` / `lineIndent(pos, bias)` (lines 86-92) and `TreeIndentContext.column(pos, bias)` (line 120) all accept a `bias` parameter that is silently ignored.

**Proposed:** Delete `textAfterPos`; drop the dead `val unit` and fix `indentString`'s KDoc to describe the spaces-only behavior (or restore tab handling if that was the intent — but the doc should match the code); drop the ignored `bias` params (or wire them through `doc.lineAt`/column resolution if they're meant to do something).

**Why:** Removes a public stub that reads like real API, eliminates a dead computation, and fixes doc-vs-code drift that misleads anyone reading `indentString`. All concentrated in one settled file.

**Effort:** mechanical (API touch — run `apiDump` if the `textAfterPos`/`bias` removals change the public dump)

monkopedia-coder:
Removes dead/vestigial bits from `language/src/commonMain/kotlin/com/monkopedia/kodemirror/language/Indent.kt` left over from the CM6 port, per #161.

## Changes
- **Removed `IndentContext.textAfterPos`** — a public property whose getter always returned `""`. Verified via `grep -rn` across all modules that the only occurrence was the definition itself (no callers). The real "text after the indent position" lives in `TreeIndentContext.textAfter`.
- **Cleaned up `indentString`** — dropped the dead `val unit = getIndentUnit(state)` local (computed and discarded), and fixed the KDoc, which promised tab handling ("using tabs if the state's `indentUnit` divides evenly") that the body never did (it always uses spaces). The doc now describes the actual spaces-only behavior. No behavior change.

## Left in place (named in the issue but actually used)
The issue also suggested dropping the ignored `bias` parameters on `lineAt(pos, bias)`, `lineIndent(pos, bias)`, and `column(pos, bias)`. Although `bias` is ignored *inside* those functions, callers pass explicit non-default values:
- `lang-python/Python.kt` calls `context.lineAt(pos, -1)` and `context.lineIndent(pos, -1)` (multiple sites)
- `lang-yaml/Yaml.kt` calls `cx.lineIndent(..., 1)` / `cx.column(before.from, 1)`

Removing the params would require editing those call sites and would erase the caller-expressed intent (these mirror upstream CM6 signatures). That's beyond focused dead-code removal, so I left the `bias` params untouched.

## Public-API note
Removing `textAfterPos` is an **intentional public-API removal**. `apiCheck` failed only on that one line; I ran `:language:apiDump` and committed the refreshed `language/api/{jvm,android}/language.api` (each lost only the `getTextAfterPos` entry). This makes the PR tier-2 for review.

## Verification (env: `JAVA_HOME=java-21-openjdk`, `ANDROID_HOME`)
- `:language:jvmTest` — BUILD SUCCESSFUL
- `:language:spotlessApply :language:ktlintFormat` — clean
- `:language:apiCheck` — green after `apiDump`

Fixes #161


monkopedia-reviewer:
Reviewed and verified — the change is sound:

- `IndentContext.textAfterPos` is a vestigial public property whose getter always returned `""`; grep across the repo (excluding leftover agent worktrees) confirms **no callers**. The real text-after-position lives in `TreeIndentContext.textAfter`.
- The dead `val unit` in `indentString` and its misleading KDoc (promised tab handling the body never did) are also cleaned up; the ignored `bias` params were correctly left in place since `lang-python`/`lang-yaml` still pass non-default values.
- The `apiDump` was applied: the only `.api` change (android + jvm) is the removal of `getTextAfterPos ()Ljava/lang/String;`. Verified locally: `:language:jvmTest` and `:language:apiCheck` both green.

However, this removes a **public** property (`textAfterPos`) — an intentional public-API removal (tier-2). Per the review policy I am **not** auto-merging; holding for **maintainer (Jason) signoff** on the API removal before merge.

monkopedia-coder:
Superseded by #175 — per maintainer decision we're COMPLETING the IndentContext port to match CodeMirror 6 (implementing `textAfterPos(pos, bias)` + making the ignored `bias`/simulated-break handling functional) rather than removing `textAfterPos`. #175 keeps this PR's legitimate cleanups (dead `val unit`, drifted KDoc).

PR #175, opened 2026-07-08
Complete IndentContext CM6 parity: textAfterPos + simulated-break bias (#161)

### LINKED ISSUE #161: Indent.kt: remove vestigial indent-context stubs (textAfterPos, dead indentString unit + drifted KDoc, ignored bias params)

monkopedia-coder:
## Summary

Completes the port of `IndentContext` to match CodeMirror 6, replacing the vestigial `textAfterPos` stub (a no-arg `val` that always returned `""`) with a real function, and making the previously-**ignored** `bias`/`simulateBreak`/`simulateDoubleBreak` parameters on `lineAt`/`lineIndent`/`column`/`textAfter` actually perform CM6's simulated-break line-splitting.

Supersedes PR #174 (the "remove `textAfterPos`" approach). Per the maintainer's decision on #161, we fix the API to match CM6 instead of deleting it.

## CM6 semantics matched
Ported from the authoritative source (`codemirror/language/src/indent.ts`):
- **`textAfterPos(pos, bias = 1)`**: returns `""` when `simulateDoubleBreak && pos == simulateBreak`; otherwise slices the (possibly break-split) line text from `pos`, capped at 100 chars (`min(text.length, pos + 100 - from)`).
- **line splitting**: when `simulateBreak` falls within the line at `pos`, the line is split — `bias < 0` selects the content strictly *before* the break (when the break sits before `pos`), `bias >= 0` selects the content *after* it. This mirrors CM6's `lineAt`.
- **`lineIndent`/`column`/`TreeIndentContext.textAfter`** now route through this splitting so they honor `bias` + simulated breaks. `column` additionally now counts **true columns** (tab-expanding, via the existing `:state` `countColumn`) instead of stopping at the first non-whitespace char.

## Design choice for the `lineAt` return-type wrinkle
CM6's `lineAt` returns a lightweight `{text, from}` that can represent a break-split line. Kodemirror's public `lineAt(pos, bias)` returns the real doc `Line` (with `to`/`number`/`from`), and callers rely on that — notably `StreamParser` (`line.to`, `line.number`) and `lang-python`'s `indentBody` (`line.from`, `line.to`).

To keep **minimal blast radius** and avoid an API break cascading across language modules:
- The public `lineAt(pos, bias): Line` is **kept** returning the real doc `Line` (documented that it cannot itself represent a split line).
- A new `protected lineContentAt(pos, bias): Pair<String, Int>` (text, from) does the CM6 split, and `textAfterPos`/`lineIndent`/`column`/`textAfter` route through it.

This gives full CM6 simulated-break behavior for the methods that compute indentation, without changing the `Line`-returning contract existing callers depend on.

## What `bias`/`simulateBreak` now does
Previously these params were accepted but silently dropped. Now, when `simulateBreak != null` and the break falls in the line at `pos`, indentation is computed as if a line break were inserted there: `lineIndent`/`column`/`textAfter`/`textAfterPos` see the split half selected by `bias`. With `simulateDoubleBreak`, the break position resolves to an empty line (`textAfterPos`/split return `""`). With no `simulateBreak`, behavior is unchanged from before.

## Regression results (no behavior surprise)
- `:language:jvmTest` — green (incl. new `textAfterPos`/split/`column` unit tests)
- `:lang-python:jvmTest` — green (indent suite preserved)
- `:lang-yaml:jvmTest` — green (indent suite preserved)

## Public-API change
This is an **intended public-API change** and `:language:apiCheck` was updated via `apiDump`:
- `getTextAfterPos()` property **removed**
- `textAfterPos(pos, bias)` function **added**
- `protected lineContentAt(pos, bias)` added

Refreshed `language/api/jvm/language.api` and `language/api/android/language.api`.

## Cleanups carried over from #161
- Removed the dead `val unit = getIndentUnit(state)` local in `indentString` and fixed its drifted KDoc (claimed tab handling but always returns spaces).
- Did **not** remove the `bias` params — they are now functional.

**Tier-2** (public-API + behavior change) — needs maintainer signoff before merge.

Fixes #161


monkopedia-reviewer (commented):
Re-verified locally against the PR branch tip (02859de) in a clean worktree — all green:

- `:language:jvmTest`, `:lang-python:jvmTest`, `:lang-yaml:jvmTest` — BUILD SUCCESSFUL (python/yaml indent regression gate preserved).
- `:language:apiCheck` (with ANDROID_HOME) — green; the only API delta is the intended `textAfterPos` reshape (property `getTextAfterPos` removed → `textAfterPos(pos,bias)` function added) plus the new `protected lineContentAt(pos,bias): Pair`. No unintended public symbols.
- `:language:ktlintCheck` — no new violations in Indent.kt / IndentTest.kt.
- 7 new unit tests present and passing (19 total in IndentTest): textAfterPosBasicSlice, textAfterPosCapsAt100Chars, textAfterPosDoubleBreakReturnsEmpty, textAfterPosSimulateBreakSplitsLine, lineIndentHonorsSimulateBreak, columnHonorsSimulateBreakAndTabs, columnSplitLineBySimulateBreak.

CM6 semantics assessment: correct. `textAfterPos` slices `[pos-from, min(len, pos+100-from)]` with the `simulateDoubleBreak && pos==simulateBreak → ""` guard, matching upstream. `lineContentAt`'s break-split (`afterBreak = bias<0 ? sim<pos : sim<=pos`) mirrors CM6's `lineAt` split-by-bias exactly. `column` now uses `countColumn` (true tab-expanding columns) instead of the old first-non-whitespace stop — matches CM6. Keeping public `lineAt` as the real doc `Line` (bias a documented no-op there) is sound given StreamParser/lang-python need `.to`/`.number`.

BLOCKER (not a code issue): the branch currently CONFLICTS with main and mergeStateStatus is DIRTY, so GitHub has run zero CI checks (`check` / `check-macos` / `check-ios` all absent). `main` is branch-protected on those three, so this cannot be merged as-is. The sole conflict is CHANGELOG.md — main gained the #163 search-cursor entry at the top of `### Changed` after this branch was cut; the .api files and Indent.kt auto-merge cleanly (confirmed via git merge-tree). Please rebase onto origin/main (or resolve the CHANGELOG conflict) and push so CI runs; the code itself is verified-good and this is otherwise ready to squash-merge.

Not approving/auto-merging yet since the merge can't proceed and required checks haven't started.

monkopedia-reviewer (approved):
Verified in prior pass (language/python/yaml tests green, apiCheck green, CM6 semantics correct, 7 new tests); maintainer tier-2 signoff given; CHANGELOG conflict resolved. Approving + squash auto-merge.

PR #184, opened 2026-07-17
Remove unused kotlinx-datetime dependency (#180)

### LINKED ISSUE #180: Bump kotlinx-datetime 0.7.1 → 0.8.0 (pre-1.0 minor — ABI/wasm-compat surface)
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml:8` — `kotlinx-datetime = "0.7.1"`. Latest stable: **0.8.0** (verified against `repo1.maven.org` maven-metadata on 2026-07-13). Consumed by `:state`, `:lint` (`lint/build.gradle.kts:16`), `:merge` (`merge/build.gradle.kts:17`).

**Current:** kotlinx-datetime pinned at 0.7.1. Note the constraint: `samples/showcase/build.gradle.kts:82-83` already `force`s `kotlinx-datetime:0.6.2` / `-wasm-js:0.6.2` for the wasm target, and the 0.8.0 release ships `0.8.0-0.6.x-compat` companion variants — signalling ongoing pre-1.0 ABI churn in this library.

**Proposed:** Bump to `0.8.0`, then (a) re-check whether the showcase `force(...0.6.2...)` pins are still needed or can move up, and (b) run the full matrix incl. `check-macos`/`check-ios` and the wasm build, since datetime's wasm artifact is the sensitive spot.

**Why:** Keeps a used dependency current. Kept OUT of the routine refresh (#178) because — unlike the in-range minors there — this is a pre-1.0 library with a real ABI/wasm-compat surface (the forced downgrade + compat-variant artifacts), so it deserves to be judged and verified on its own.

**Effort:** small refactor (bump + reconcile the showcase force-pins + full-matrix verify)

monkopedia-coder:
## Remove unused kotlinx-datetime dependency (resolves #180)

#180 asked to bump kotlinx-datetime 0.7.1 → 0.8.0, flagging its pre-1.0 ABI/wasm-compat surface. Investigating turned up that **the dependency is dead**: `:state`, `:lint`, and `:merge` declare `implementation(libs.kotlinx.datetime)` but never import it — their `currentTimeMillis()` helpers use **stdlib `kotlin.time.Clock.System.now()`**. There are **zero `import kotlinx.datetime`** sites anywhere (main or test).

So instead of bumping a dependency nothing uses (and inheriting the 0.8.0 compat surface), this **removes** it:
- Drop `implementation(libs.kotlinx.datetime)` from `:state` / `:lint` / `:merge`
- Remove the version + library alias from `gradle/libs.versions.toml`
- Remove the `samples/showcase` `kotlinx-datetime:0.6.2` / `-wasm-js:0.6.2` resolution force-pin — it only existed to pin the transitive artifact for wasm, and there's no transitive artifact anymore

### Verification (local)
- `apiCheck` — **0 `.api` changes** (removing an `implementation` dep doesn't touch the API; ABI stable)
- `:lint:jvmTest` / `:merge:jvmTest` / `:state:jvmTest` — green **without** the dependency (proves it was unused)
- `:samples:showcase:compileKotlinWasmJs` — green **without** the force-pin (proves the wasm workaround is no longer needed)

Net: −1 dependency, −1 wasm build workaround, and #180's ABI/compat concern is moot. Fixes #180.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Removed
 - **Dropped the `iosX64` and `macosX64` (Apple-Intel) publication targets.** Compose-Multiplatform 1.11.x no longer publishes those two targets (`runtime-macosx64:1.11.x` is absent from Maven Central), so the library follows suit. Apple Silicon (`macosArm64`, `iosArm64`, `iosSimulatorArm64`), JVM, wasmJs, and Android are unaffected. This is a breaking change only for consumers building for Apple-Intel (#182).
+- Removed the unused `kotlinx-datetime` dependency from `:state`, `:lint`, and `:merge` (#180). Those modules use stdlib `kotlin.time.Clock` for timestamps and never imported `kotlinx.datetime`, so the dependency was dead weight; dropping it also removes the `samples/showcase` `kotlinx-datetime:0.6.2` wasm resolution force-pin that only existed to resolve the transitive artifact.
 
 ## [0.3.5] - 2026-07-15
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,7 +5,6 @@ compose-plugin = "1.11.1"
 atomicfu = "0.33.0"
 kotlinx-serialization = "1.11.0"
 kotlinx-coroutines = "1.11.0"
-kotlinx-datetime = "0.7.1"
 ktlint = "12.1.2"
 spotless = "8.8.0"
 roborazzi = "1.68.0"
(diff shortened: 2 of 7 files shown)


monkopedia-coder (commented):
## CI failure: yarn.lock out of date after removing the force-pin

The **`check` job fails** with:

```
Execution failed for task ':kotlinWasmStoreYarnLock'.
> Lock file was changed. Run the `kotlinWasmUpgradeYarnLock` task to actualize lock file
```

Removing `force("org.jetbrains.kotlinx:kotlinx-datetime-wasm-js:0.6.2")` from `samples/showcase/build.gradle.kts` changes the resolved wasm npm dependency set, so the committed `kotlin-js-store/wasm/yarn.lock` is now stale. The `kotlinWasmStoreYarnLock` task enforces lock-file freshness and blocks the build.

**Fix required:** run

```
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew kotlinWasmUpgradeYarnLock
```

then commit the updated `kotlin-js-store/wasm/yarn.lock` to this branch.

---

Everything else reviewed clean:
- Zero `import kotlinx.datetime` or fully-qualified `kotlinx.datetime.*` references in any source set (grep confirms)
- All three removed deps were `implementation(...)`, not `api(...)` — no API surface exposed
- Zero `.api` file changes — ABI stable
- Diff scope: exactly 6 files, no scope creep
- CHANGELOG entry present and correct

Once the yarn.lock is updated and all three required checks (`check`, `check-macos`, `check-ios`) go green, this is a clean tier-1 merge.

PR #185, opened 2026-07-24
Fix caret-teleport to line start on hit-test during content-change frame lag (#171)

### LINKED ISSUE #171: Horizontal hit-test drift (~3/4 char, rightward) with gutters on wasm/CanvasKit
Follow-up surfaced after the #169 vertical fix (PR #170). With the ~1/4-line vertical bias removed, a larger **horizontal** drift became clearly visible on adolin (live wasm/CanvasKit): clicks land with the caret ~3/4 character to the **right** of the pointer (column over-estimated). Reproduces in konstructor's real config (`basicSetup + lintGutter() + lineWrapping`), i.e. with a line-number gutter + a lint gutter; the minimal gutter-less harness shows X essentially perfect.

### What's ruled out
- **Not a layout-space gutter-width formula bug.** A `coordsAtPos(pos).left` → click round-trip is **clean on JVM** (drift=0) across every combo: `lineNumbers + lint-like gutter`, `+ lineWrapping`, `+ a rendered marker`, `lineNumbers + wrap`. So `coordsAtPos` and the hit-test (`posFromVisibleItems` / `contentStartPx = gutterWidthDp + 6dp`) agree in layout space — the gutter is hard-constrained to the same `gutterWidthDp` on both the render and hit-test sides.
- **Not a system-font fallback (#109 class).** konstructor loads the content font as a properly embedded Compose Resources font (`FontFamily(Font(Res.font.JetBrainsMono_Regular))`), so rendered advances should equal laid-out advances for the text itself.

### Leading hypothesis
A **rendered-vs-layout divergence** that a layout-space round-trip is blind to by construction (matching the JVM-clean result) — most likely **async font-load timing**: Compose Resources `Font()` resolves asynchronously on wasm. If the line layout is built in a fallback face before JetBrainsMono finishes loading, the font swaps in and repaints at the real advances, but the cached layout geometry the caret/coords path reads stays stale → caret drawn right of the glyph the user clicked. This is the #165 gremlin's family (`onGloballyPositioned` not reliably re-firing after a change on wasm). Note the showcase **waits** for fonts (`onFontsReady` in Main.kt), so it sidesteps this trigger and can't reproduce it without a no-font-wait harness variant.

### Open discriminators (in flight via downstream konstructor on adolin)
1. Does a forced **relayout/resize** (no doc change) make the drift **vanish** (→ confirms async-stale-layout) or **persist** (→ a steady-state wasm hit-test geometry bug)?
2. Does the **click** resolve to the wrong character, or does the click resolve correctly and only the **caret draw** at a stale x?

### Next steps
- Get the discriminator results, then either (a) fix the async-font cache-invalidation path (invalidate line layout / coords cache on font-ready), or (b) chase a steady-state wasm geometry bug.
- Either way, build a **no-font-wait** showcase harness variant so this is reproducible + guardable in-house (gap-analysis), since the normal showcase waits for fonts.

Distinct root cause from #169 (which was a sub-pixel rounding / `contentTopPadding` vertical anchor, fixed in #170). Tracking the horizontal component separately here.

monkopedia-coder:
## Summary

`posFromVisibleItems` clamped the resolved character offset to the `ColumnItem`'s `[from, to)` span:

```kotlin
return item.from.value + charOffset.coerceIn(0, item.to.value - item.from.value)
```

That `columnItems` snapshot can lag the `TextLayoutResult` by a frame after a content change: `BasicText` recomposes and writes its new layout (via `onTextLayout`) before this snapshot catches up, so `item.to - item.from` is momentarily stale — e.g. `0` on a just-populated line — while the layout (and thus the resolved `charOffset`) is already correct. Clamping a correct offset to that stale span collapsed **every** hit in that window to `item.from`, teleporting the caret to the line start.

The fix clamps to the layout's own text length instead — the domain `charOffset` was resolved against:

```kotlin
val layoutCharLen = unmapTabOffset(layout.layoutInput.text.length, item.tabOffsetMap)
return item.from.value + charOffset.coerceIn(0, layoutCharLen)
```

When the snapshot is fresh the two lengths agree, so this only changes behaviour in the stale frame.

## Root-cause evidence

Confirmed on real hardware (GNOME fractional-scale 1.5) via an extended `[KM-HITDEBUG]` capture carrying `itemLen`/`lineLen`/`mismatch`. Warm samples showed the hit-test math resolving perfectly (`gop` tracks `localX` exactly, correct nearest-boundary, wrap-aware on a 105-char line) yet the caret sitting at `0,0` — because the returned offset was coerced into `[0, item.to - item.from == 0]`. The `println` emitted the pre-coerce value, which is why the log "looked right" while the caret jumped.

## Scope

- Addresses the **flaky caret-teleport-to-0,0** axis of #171 (the more egregious symptom). Does **not** close #171: a separate, constant ~½-char horizontal drift between the painted glyph and the caret persists on real wasm/CanvasKit and needs a screenshot glyph-ink-vs-caret measurement (the click log can't capture it, since hit-test and caret-draw math agree with each other). That axis stays open.

## Testing

- `:view:jvmTest` `ClickTargetingTest`, `HitTestVerticalAnchorTest`, `DragSelectionTest` green (they exercise the fresh path — the idle-synced Compose harness settles recomposition before each click, so it can't reproduce the single-frame stale race itself; they guard against regression of normal clicks).
- Real-hardware re-verification is the validation path for the race: konstructor will redeploy this build to the debug host and re-run the `[KM-HITDEBUG]` capture.
- spotlessApply + ktlintFormat clean; no API change.

Part of #171.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+- A click could occasionally teleport the caret to the start of the line instead of landing under the pointer (#171). `posFromVisibleItems` clamped the resolved character offset to the `ColumnItem`'s `[from, to)` span, but that `columnItems` snapshot can lag the `TextLayoutResult` by a frame after a content change: `BasicText` recomposes and writes its new layout (via `onTextLayout`) before the snapshot catches up, so `item.to - item.from` is momentarily stale — e.g. `0` on a just-populated line — while the layout (and thus the resolved offset) is already correct. Clamping a correct offset to that stale span collapsed every hit in that window to `item.from`, i.e. the line start. The offset is now clamped to the layout's own text length — the domain the offset was resolved against — so a hit during the stale frame lands correctly; when the snapshot is fresh the two lengths agree and behaviour is unchanged.
+
 ### Changed
 - Bumped Compose-Multiplatform 1.10.3 → 1.11.1 (Skia M138 → M144) and migrated `Key.Home` → `Key.MoveHome` (1.11 removed the `Key.Home` alias).
 
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -1452,6 +1452,18 @@ private fun posFromVisibleItems(
         .coerceAtLeast(0f)
     val expandedOffset = layout.getOffsetForPosition(Offset(localX, localY))
     val charOffset = unmapTabOffset(expandedOffset, item.tabOffsetMap)
-    return item.from.value +
-        charOffset.coerceIn(0, item.to.value - item.from.value)
+    // Clamp to the LAYOUT's char length, not the ColumnItem's [from, to) span.
+    // `charOffset` was resolved against `layout`, so the layout's own text
+    // length is its natural domain. The `columnItems` snapshot can lag the
+    // TextLayoutResult by a frame after a content change: BasicText recomposes
+    // and writes its new layout (via onTextLayout) before this snapshot catches
+    // up, leaving item.to-item.from stale — e.g. 0 on a just-populated line
+    // while the layout already holds the full text. Clamping a correctly
+    // resolved offset to that stale span collapses every hit to item.from,
+    // teleporting the caret to the line start (#171). The layout, being the
+    // source `charOffset` came from, is the consistent bound. When the snapshot
+    // is fresh the two lengths agree, so this only changes behaviour in the
+    // stale window.
+    val layoutCharLen = unmapTabOffset(layout.layoutInput.text.length, item.tabOffsetMap)
+    return item.from.value + charOffset.coerceIn(0, layoutCharLen)
 }


monkopedia-reviewer (approved):
Reviewed to tier-1 standard.

**Correctness:** The clamp domain change is correct. `charOffset` is resolved against `layout` (via `getOffsetForPosition` + `unmapTabOffset`), so the layout's own text length — not the ColumnItem's potentially-stale span — is the natural upper bound. `unmapTabOffset(layout.layoutInput.text.length, item.tabOffsetMap)` correctly maps the expanded layout length back to doc-relative char space, matching `charOffset`'s domain. When the snapshot is fresh, the two lengths agree and behaviour is unchanged; in the stale window the correct offset is preserved rather than collapsed to `item.from`.

**Pattern consistency:** This same idiom — `layout.layoutInput.text.length` as the authoritative clamp bound to handle the snapshot lag — is already used at line 392 in the same file (cursor drawing, #152 fix). The fix is coherent with existing stale-race mitigations.

**Edge cases:** Empty layout (`layoutInput.text.length == 0`) → `unmapTabOffset(0, ...) = 0` → `coerceIn(0, 0) = 0` → returns `item.from.value`, consistent with the no-layout fallback on the line above. Wrapped lines handled by `localY` passthrough before unmapping. Tabs handled by `unmapTabOffset` on the layout length.

**Tests:** `ClickTargetingTest`, `HitTestVerticalAnchorTest`, `DragSelectionTest` guard the fresh-snapshot path (the stale race is inherently a timing artifact the Compose test harness settles before each click — correct not to add a flaky timing test here).

**CI:** All three required checks passed — `check`, `check-macos`, `check-ios` — before this approval.

LGTM.

Issue #182, opened 2026-07-15
Migrate to Compose-Multiplatform 1.11.x (drops Apple-Intel targets, Key.Home→MoveHome)


monkopedia-coder:
## Migrate to Compose-Multiplatform 1.11.x (drops Apple-Intel targets)

Split out of #178: the routine dep refresh (0.3.5) bumped everything to latest stable **except** Compose-Multiplatform, which was held at 1.10.3 because 1.11.x is a real migration, not a mechanical bump.

**Two blockers, both confirmed:**
1. **Compose 1.11.1 does not publish the `iosX64` / `macosX64` (Apple-Intel) targets.** `org.jetbrains.compose:runtime-macosx64:1.11.1` is **404** on Maven Central (1.10.3 is 200). This library declares `macosX64()` and `iosX64()` in `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:61,64`, so bumping compose forces **dropping those two published targets** — a breaking change for any consumer on Apple-Intel.
2. **`Key.Home` becomes a hard compile error** in 1.11.x (→ `Key.MoveHome`). Usages: `view/.../InputHandling.kt:78,140`, `vim/.../VimExtension.kt:494`, `view/.../KeyboardHandlingTest.kt:102-103`.

**Decision needed (owner):** OK to drop the published `iosX64`/`macosX64` artifacts? (konstructor is wasm and unaffected; the question is any other Apple-Intel consumer.) If yes, the migration is: remove the two targets from the convention plugin, migrate `Key.Home` → `Key.MoveHome`, bump compose to latest 1.11.x, re-verify screenshots (Skia M138→M144 may shift pixels) + `check-macos`/`check-ios`.

**Effort:** small refactor (target removal + Key.Home) + screenshot re-baseline.

_Filed as a follow-up to the 0.3.5 release. Tagged needs-decision — the Apple-Intel target drop is an owner call._

PR #188, opened 2026-07-24
Remove redundant LinePos.copy() extension that shadowed data-class copy (#186)

### LINKED ISSUE #186: vim: hand-written LinePos.copy() shadows data-class copy() and silently drops `sticky`
## Bug (silent data loss in vim mode)

`LinePos` is a data class carrying three fields:

```kotlin
// vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/LinePos.kt:31
data class LinePos(val line: Int, val ch: Int, val sticky: String? = null) : Comparable<LinePos> {
```

But a top-level extension function shadows the compiler-generated `copy()`:

```kotlin
// vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/LinePos.kt:57
/** Copy a LinePos, optionally overriding line or ch. */
fun LinePos.copy(line: Int = this.line, ch: Int = this.ch): LinePos = LinePos(line, ch)
```

Because a same-name member/extension resolution picks up this extension at call sites, `somePos.copy(ch = x)` constructs `LinePos(line, ch)` with the third argument defaulted — **silently dropping the `sticky` field** instead of preserving it. Any call site relying on `copy()` to carry `sticky` through loses it without warning.

## Impact
Silent data loss on `sticky` (used for horizontal-sticky column tracking in vim motions). This is a correctness bug, not a style nit — the two `copy()`s look identical at the call site but behave differently.

## Suggested fix
Remove the hand-written extension entirely and rely on the generated `data class` `copy()` (which already covers `line`/`ch`/`sticky`). If a call site truly wants to reset `sticky`, it should pass `sticky = null` explicitly. Verify no call site depended on the drop-`sticky` behavior.

## Provenance
Surfaced during the Graphify A/B evaluation (graph call/degree analysis flagged the duplicate `copy`); verified by hand against `origin/main`.

monkopedia-coder:
## Summary

`LinePos` is a data class with three fields:

```kotlin
// vim/.../LinePos.kt:31
data class LinePos(val line: Int, val ch: Int, val sticky: String? = null) : Comparable<LinePos>
```

…but a top-level extension shadowed the generated `copy()`:

```kotlin
// vim/.../LinePos.kt:57 (removed)
fun LinePos.copy(line: Int = this.line, ch: Int = this.ch): LinePos = LinePos(line, ch)
```

At any call site `pos.copy(ch = x)` resolved to this extension and built `LinePos(line, ch)` — **silently dropping `sticky`**. The two `copy()`s are indistinguishable at the call site but behave differently.

## Why it's safe (and why "silent data loss" is latent, not active, today)

- The extension has **zero call sites** repo-wide (`git grep '\.copy(ch' / '\.copy(line'` finds only the definition).
- `sticky` is currently vestigial: never written with a non-null value, and its CM5 sticky-column logic is explicitly **not yet ported** (`VimMotions.kt:507-508` — "We don't have sticky in the Kotlin LinePos, so we return as-is").

So removing the extension is behaviour-preserving for every current call site, and it eliminates the footgun before `sticky` gets wired up — at which point the shadow *would* have caused real data loss.

## Change
- Delete the `LinePos.copy` extension; the generated data-class `copy()` (which carries `sticky`) becomes the only `copy`.
- Regenerated the vim API dump (jvm + android): drops the synthetic `LinePosKt { copy, copy$default }`. The data-class `copy(II Ljava/lang/String)` is unchanged.

## Testing
- `:vim:jvmTest` green; `:vim:apiDump` regenerated; spotlessApply + ktlintFormat clean.

Fixes #186.

Provenance: surfaced during the Graphify tooling A/B evaluation (call-graph flagged the duplicate `copy`), verified by hand against `origin/main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Bumped Compose-Multiplatform 1.10.3 → 1.11.1 (Skia M138 → M144) and migrated `Key.Home` → `Key.MoveHome` (1.11 removed the `Key.Home` alias).
 
 ### Removed
+- Removed the redundant public `fun LinePos.copy(line, ch)` extension in `:vim` (#186). It shadowed the `LinePos` data class's generated `copy()` and silently dropped the `sticky` field — a latent data-loss footgun the moment `sticky` gets wired up (its CM5 sticky-column logic is not yet ported). The extension had zero call sites, so removing it is behaviour-preserving today and leaves the full-fidelity generated `copy()` as the only one.
 - **Dropped the `iosX64` and `macosX64` (Apple-Intel) publication targets.** Compose-Multiplatform 1.11.x no longer publishes those two targets (`runtime-macosx64:1.11.x` is absent from Maven Central), so the library follows suit. Apple Silicon (`macosArm64`, `iosArm64`, `iosSimulatorArm64`), JVM, wasmJs, and Android are unaffected. This is a breaking change only for consumers building for Apple-Intel (#182).
 - Removed the unused `kotlinx-datetime` dependency from `:state`, `:lint`, and `:merge` (#180). Those modules use stdlib `kotlin.time.Clock` for timestamps and never imported `kotlinx.datetime`, so the dependency was dead weight; dropping it also removes the `samples/showcase` `kotlinx-datetime:0.6.2` wasm resolution force-pin that only existed to resolve the transitive artifact.
 
--- a/vim/api/android/vim.api
+++ b/vim/api/android/vim.api
@@ -173,11 +173,6 @@ public final class com/monkopedia/kodemirror/vim/LinePos : java/lang/Comparable 
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class com/monkopedia/kodemirror/vim/LinePosKt {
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (commented):
Reviewed at tier-1 standard. The extension has zero call sites, the data-class copy() correctly preserves sticky, and the API dump change is exactly the LinePosKt static removal only. Owner approved the public-API removal for 0.3.6. Sound and ready to merge on CI green.

monkopedia-reviewer:
**Re-review after force-push rebase (HEAD 3bc6d1f6):** Delta vs approved SHA is CHANGELOG-only.

The rebase against updated main resolved a CHANGELOG.md merge conflict — the `LinePos.copy` removal bullet was added into the existing `### Removed` section alongside #182/#180, and #185's `### Fixed` block was kept. No code or API file changed.

**Correctness verification (fresh pass on 3bc6d1f6):**
- `LinePos` data-class `copy(IILjava/lang/String)` (carrying `sticky`) is present in both `vim/api/jvm/vim.api` and `vim/api/android/vim.api` (line 166/167 in each).
- `LinePosKt { copy, copy$default }` is absent from both API dumps — correctly removed.
- `git grep '\.copy(line' / '\.copy(ch'` over all `.kt` files at this SHA: zero hits — no call sites existed.

The human approval stands. CI checks are pending; will merge on all three (`check`, `check-macos`, `check-ios`) passing. Auto-merge remains armed.

Issue #193, opened 2026-07-26
CLAUDE.md mandates a screenshot workflow doc that points at a deleted directory, a nonexistent Gradle task, and a forbidden push flow


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** hygiene (dead code + doc drift, agent-facing)

**Where:** `docs/screenshot-compare-workflow.md`, `build.gradle.kts:115`, `CLAUDE.md`, `docs/post-task-workflow.md`

**Current:**

`CLAUDE.md` instructs every agent working in this repo: *"When the user asks to compare screenshots or fix visual differences, follow the workflow in `docs/screenshot-compare-workflow.md`."* That doc is broken in three independent ways:

1. **It points at a directory that was deleted four months ago.** Commit `8a27c44` (2026-03-30) removed `reference-screenshots/`; `git ls-files | grep -c '^reference-screenshots'` is `0`. The doc still references it at lines 15, 17, 29, 69, 71, 77 — including as step 1 of the workflow (`cd reference-screenshots && npm install && npx playwright test`).

2. **It names a Gradle task that does not exist.** Lines 23 and 50 say `./gradlew :view:recordRoborazziDesktop`. The real task is `recordRoborazziJvm` (`docs/roborazzi-screenshots.md:41,54`, and ci.yml).

3. **It prescribes a flow `CLAUDE.md` explicitly forbids.** Line 46 describes a direct commit-and-push; `CLAUDE.md` says do NOT push directly to `main` and do NOT self-merge — everything goes through the coderbot → `monkopedia-reviewer` PR flow.

Also dead: `build.gradle.kts:115` registers an `Exec` task `captureReferenceScreenshots` with `workingDir = file("reference-screenshots")` — it cannot run, the directory is gone.

Same class, lower stakes: `docs/post-task-workflow.md` is declared superseded by `CLAUDE.md` but carries no banner saying so, and still instructs `git checkout main` / `git merge` / `git push`.

**Proposed:**
1. Either restore a working screenshot-reference step or rewrite `docs/screenshot-compare-workflow.md` around what actually exists today (roborazzi + the gap-analysis Playwright rig), and fix the task name to `recordRoborazziJvm`.
2. Delete the dead `captureReferenceScreenshots` task from `build.gradle.kts`.
3. Replace the direct-push section with the coderbot → reviewer PR flow.
[...]

PR #204, opened 2026-07-28
Delete orphaned .api dumps and repair the $stable-stripping globs (#191)

### LINKED ISSUE #191: Committed .api corpus: 41 orphaned dumps CI never validates (10 already drifted) + 55/57 android dumps duplicate their jvm twin

**Category:** dead-code / hygiene

**Where:**
- 41 orphaned `*/api/<name>.api` files (top-level, pre-BCV-layout)
- 55 of 57 `*/api/android/*.api` files
- `build.gradle.kts:53` (`stripStableFromApiDumps`), `build.gradle.kts:82` (the `jvmApiCheck` doFirst)
- `.github/workflows/ci.yml:22`

**Current:**

29,699 lines of `.api` dumps are committed. Two disjoint problems, both verified on `ded2a41`:

**(a) 41 orphaned top-level `api/<name>.api` files that nothing reads and CI never validates.**
BCV moved to the per-target `api/jvm/` + `api/android/` layout, but the old flat files were never deleted. `./gradlew apiCheck jvmTest` (ci.yml:22) is green while **10 of them have already drifted** from their `api/jvm/` twin:

```
vim/api/vim.api              917 vs 436 lines
view/api/view.api            878 vs 903
language/api/language.api    335 vs 357
merge/api/merge.api          274 vs 285
autocomplete/api/…           252 vs 266
search/api/search.api        126 vs 131
lint/api/lint.api            117 vs 123
commands/api/commands.api    145 vs 147
theme-one-dark/…              25 vs  26
theme-dracula/…               24 vs  25
```

These files look authoritative to a reader (and to an agent doing an API review) but describe an API that no longer exists.

**(b) 55 of the 57 `api/android/*.api` files are byte-identical to their `api/jvm/` twin.**
This is structural, not accidental: `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` does `androidMain.get().dependsOn(jvmMain.get())`, so the Android ABI *is* the JVM ABI for every module except `view` and `vim` (the only two that genuinely differ — verified with `cmp` across all 57). Worst offenders: `legacy-modes/api/android/legacy-modes.api` (3,619 lines), `state/api/android/state.api` (1,283).

**Two Gradle tasks silently no-op against the dead layout:**
- `build.gradle.kts:53` — `include("*/api/*.api")` matches only the orphans, never `api/jvm/`, so `stripStableFromApiDumps` strips nothing from the files that are actually checked.
- `build.gradle.kts:82` — `include("*.api")` on `build/api` is non-recursive, so it misses the real `build/api/jvm/` output.

Since both tasks exist specifically to strip Compose's `$stable` fields before comparison, their globs missing the live layout means that mechanism is currently inert.

**Proposed:**
1. Delete the 41 orphaned top-level `api/*.api` files.
2. Fix the two globs in `build.gradle.kts` to match the per-target layout (`*/api/*/*.api` and a recursive `build/api` tree), then confirm `apiCheck` still passes — if `$stable` stripping was inert, fixing the glob may surface real diffs that need an `apiDump`.
3. Decide on the android dumps: either keep them (explicit is fine, but say so) or configure BCV to skip the Android target where its ABI is by construction the JVM ABI. This one is a judgement call, hence staged rather than assumed.

**Why:** (a) is a genuine CI blind spot — 10 files currently misrepresent the public API of a *published* library and nothing catches it. (b) is ~8,500 lines of pure noise in every API-affecting diff, which makes the `apiCheck` signal harder to read exactly when it matters most. And the `$stable`-stripping tasks not matching anything means a mechanism you built is quietly not running.

**Effort:** mechanical for (1); small refactor for (2) (may cascade into an `apiDump`); (3) is a small decision + config change.

monkopedia-coder:
Fixes #191 (items 1 and 2; item 3 deliberately left alone — recommendation below).

## 1. Deleted the 41 orphaned top-level `.api` dumps

BCV moved to the per-target `api/jvm/` + `api/android/` layout and the old flat
`<module>/api/<module>.api` files were never removed. Nothing read them and `apiCheck` never
validated them, so they had been free to rot: 10 of the 41 had already drifted from their live
twin, worst `vim/api/vim.api` at 917 lines against 436 in `vim/api/jvm/vim.api`.

The orphan set was derived mechanically — `git ls-files '*/api/*.api'` minus anything under
`api/jvm/` or `api/android/` — and comes to exactly 41 files, all at a single nesting level. The
114 live per-target dumps (57 jvm + 57 android) are untouched by this part of the change.

The only remaining references to those paths are in `docs/api-evaluation/**`, which is an
explicitly dated snapshot report (`*Generated: 2026-03-06*`) whose appendix records which files
were read at the time. Rewriting a past report's source list seemed worse than leaving it, so
those are left as-is.

## 2. Repaired the `$stable`-stripping globs

Both tasks that strip the Compose compiler's synthetic `$stable` fields were aimed at the dead
layout:

- `stripStableFromApiDumps` used `include("*/api/*.api")`, which matched **only** the orphans —
  never `api/jvm/`. After deleting them it would have matched nothing at all.
- The `jvmApiCheck` `doFirst` used a non-recursive `include("*.api")` on `build/api`, which never
  matched the real `build/api/jvm/` output.

So neither side of the comparison was ever stripped, and `apiCheck` was green only because both
sides equally carried `$stable`. The globs now target the per-target layout (`*/api/*/*.api` for
the committed dumps, `build/api/<target>/*.api` for the build output), and the build-output strip
is registered on `androidApiCheck` in addition to `jvmApiCheck` so every dumped target is covered.

### The `$stable` diff this surfaced — as predicted in the issue

Fixing the `doFirst` glob immediately turned `apiCheck` red: the build output was now stripped
while the committed dumps still carried `$stable`. That is the mechanism finally engaging. I ran
`apiDump` (which is `finalizedBy stripStableFromApiDumps`) and inspected the result:

```
22 files changed, 226 deletions(-)
```

Every one of those 226 lines is `public static final field $stable I`. Filtering the dump diff for
non-`$stable` lines returns **zero** hits — no declaration was added, removed, renamed, or moved.
Affected modules: autocomplete, commands, language, lint, lsp-client, merge, search, theme-dracula,
theme-one-dark, view, vim (jvm + android each).

### One extra fix the repair forced: per-target scoping

My first pass pointed the `doFirst` at all of `build/api` with `**/*.api`. That made `apiCheck`
fail intermittently — roughly every other run — with corrupted build output, e.g.

```
-	public final fun getFoldPlaceholderColor-0d7_KjU ()J
+	public final fun getFoldPlaceho                         lderColor-0d7_KjU ()J
```

and, in other runs, a `build/api/jvm/view.api` truncated mid-line at 570 of 865 lines. Cause: with
a tree spanning all of `build/api`, `androidApiCheck` rewrites `build/api/jvm/*.api` — the other
target's build output — before `jvmApiCheck` has compared it, so the two check tasks and
`jvmApiBuild` end up writing the same files. Each check task now strips only its own
`build/api/<target>/` directory, which removes the overlap by construction. Verified: 8 consecutive
clean `apiCheck` runs plus 2 `apiCheck jvmTest` runs, all green, where the unscoped version failed
about half the time.

## 3. Android dumps — not changed, recommendation only

Re-confirmed on this branch: 55 of the 57 `api/android/*.api` files are byte-identical to their
`api/jvm/` twin; only `view` and `vim` genuinely differ. That is structural —
`kodemirror.library.gradle.kts` does `androidMain.get().dependsOn(jvmMain.get())`.

**Recommendation: keep dumping the android target.** The duplication is noise, but it is noise that
carries a real signal — those 55 files being identical is precisely the invariant worth having
guarded. `view` and `vim` prove the Android ABI *can* diverge, and the day someone adds an
`androidMain` actual or an Android-only overload, the android dump is the only thing that would
catch it; suppressing the target would silently drop Android from API validation for a *published*
Android artifact to save review noise. If the diff noise becomes the binding problem, the cheaper
fix is presentational (a `.gitattributes` `linguist-generated`/`diff` marker on `api/android/**`,
or collapsing them in review) rather than giving up the coverage. Happy to do either if you prefer
the other call — leaving BCV's config untouched here per the scoping note.

## Verification

```
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ANDROID_HOME=$HOME/Android/Sdk ./gradlew apiCheck   # green
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ANDROID_HOME=$HOME/Android/Sdk ./gradlew jvmTest    # green
./gradlew spotlessApply ktlintFormat                                                       # no changes
```

`apiCheck` and `apiCheck jvmTest` were each re-run repeatedly to confirm the intermittent failure
described above is gone. `spotlessApply`/`ktlintFormat` produced no edits, so no unrelated
formatting drift is in the diff. `CHANGELOG.md` updated under `## [Unreleased]` → `### Build`.


monkopedia-reviewer (approved):
Verified independently, not just on the author's claims:

**Scope (item 5).** Diff touches exactly 65 files: `CHANGELOG.md`, `build.gradle.kts`, and 63 `.api` files. `convention-plugins/` (BCV config, incl. the android target) is untouched — item 3 (whether to keep dumping android) was correctly left alone as a recommendation only. No `captureReferenceScreenshots` touch — #193's territory is untouched.

**Orphan deletion (item 1).** The 41 deleted files are exactly the top-level `*/api/<name>.api` set with no `jvm/`/`android/` subdir — matches the issue's mechanical derivation (`git ls-files` minus the per-target dirs) 1:1. `apiValidation {}` in the convention plugin has no explicit dump-dir override, so BCV's live output is the per-target layout; the flat files are dead weight nothing reads. The 114 live `api/jvm/`+`api/android/` dumps are untouched by the deletion.

**Glob repair (item 2) — confirmed it actually fires.** Read the `build.gradle.kts` diff directly: `stripStableFromApiDumps` now globs `*/api/*/*.api` (was `*/api/*.api`, orphans-only), and the `doFirst` on `jvmApiCheck`/`androidApiCheck` now scopes to `build/api/$target/*.api` per-task (was a non-recursive `build/api/*.api` that missed `build/api/jvm/` entirely). Both were structurally incapable of matching the live layout before.

**The 226-line churn — verified myself, not taken on faith.** Extracted just the 22 modified (non-deleted) `.api` files from the diff: 226 total `+`/`-` lines, **all 226 are deletions, 0 additions**, and every single line matches `$stable` (`grep -v '\$stable'` on that set returns zero hits). No declaration added, removed, renamed, or moved — this is a pure `$stable`-field strip, not an ABI change.

**Concurrency fix — reproduced clean, did not just accept 8-clean-runs claim.** Checked out the branch and ran `apiCheck` independently: full-project run green (`BUILD SUCCESSFUL`, 2074 tasks), plus 4 more full green runs. Then targeted the two modules the author says corrupted ~50% of the time pre-fix (`vim`, `view`) with `--parallel --rerun-tasks :vim:apiCheck :view:apiCheck` (forces real re-execution, not up-to-date skips) — 2 consecutive green runs, no corruption. Combined with real CI (`check`/`check-macos`/`check-ios` all pass), that's solid independent confirmation the per-target scoping (`build/api/$target/`) actually eliminates the cross-task file overlap.

Diagnosis and fix both check out: `androidApiCheck` and `jvmApiCheck` sharing one `build/api` tree meant one task's strip could rewrite the other's not-yet-compared output mid-flight; scoping each to its own subdir removes the shared file by construction. Registering the strip on `androidApiCheck` (previously only on `jvmApiCheck`) is correct and necessary — the android target is dumped and stripped by the exact same mechanism, so it needed the same fix, not scope-widening.

Category: `infra` (build/CI tooling, no actual API surface change — verified, not just asserted). Tier 1 per the kodemirror project override table. CI green (check/check-ios/check-macos all pass), no scope creep, orphans genuinely orphaned, `$stable` mechanism now genuinely fires with zero real ABI drift, concurrency fix reproduces clean under independent re-verification. Approving.

Issue #202, opened 2026-07-26
Unblock wasmJs tests for Compose-consuming modules (skiko — nodejs vs browser runner)


monkopedia-coder:
Part of #196 (stage 5). Blocks wasm UI tests in #200; blocks wasm for the Compose-consuming modules in #198 / #199.

## Problem

`wasmJsTest` is disabled repo-wide in `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:92-94`. #197 narrows that to Compose-free modules, which is a clean win — but it leaves every Compose-consuming module (`:view`, `:vim`, `:autocomplete`, `:search`, `:lint`, `:commands`, `:language`, `:merge`, `:lsp-client`, `:basic-setup`, the 17 `:theme-*`) with no wasm test execution.

The recorded cause is a missing `skiko.mjs` in the wasm test environment. That has never been re-examined; the disable predates the current Kotlin/Compose versions.

## Proposed

Investigate and, if possible, fix. Specific things worth checking:

1. **Re-confirm the failure on current toolchain** (Kotlin 2.4.10, Compose MP 1.11.1). Capture the actual error rather than inheriting the assumption — the disable may be stale.
2. **`nodejs()` vs `browser()`.** The convention plugin configures `wasmJs { nodejs() }`. Skiko/CanvasKit is fundamentally a browser technology; a Karma/browser test runner is the environment where Compose-on-wasm is actually supported. Adding a browser test target for Compose modules may be the whole fix.
3. **Whether the skiko `.mjs`/`.wasm` assets can be placed where the test runner resolves them**, if node is kept for some modules.

## Known landmines

- **Node version.** The root `build.gradle.kts` pins the wasm npm tooling Node to 22.11.0 because Kotlin 2.4.10 defaults to Node 25, which breaks the webpack build (#189). Any change to the wasm test runner needs to respect that pin — expect to hit it.
- **Headless CanvasKit does not load bundled Compose-Resources fonts.** Verified across two independent rigs: the `gap-analysis` Xvfb Chromium and a real-NVIDIA-GPU-under-Xvfb setup both render a proportional *fallback* rather than the bundled JetBrainsMono, even though the `.ttf` serves HTTP 200. So a browser-based wasm test suite will run with the wrong font. That's fine for semantics/behavior assertions and **fatal for any pixel assertion** — do not build screenshot tests on this. See #171.
[...]

PR #206, opened 2026-07-28
Rewrite screenshot-compare workflow around the rigs that exist (#193)

monkopedia-coder:
Fixes #193.

`CLAUDE.md` tells every agent in this repo to follow `docs/screenshot-compare-workflow.md` for screenshot work. Every concrete instruction in that doc was wrong, so an agent that obeyed it burned its run before looking at a single pixel.

## The decision: is CM6 reference capture still part of the loop?

**Yes — it was relocated, not removed, so the doc is rewritten rather than deleted.**

Commit `8a27c44` ("Consolidate reference-screenshots into gap-analysis", 2026-03-30) *moved* the thirteen CM6 scenario fixtures into `gap-analysis/fixtures/cm6-scenarios/` and the capture spec into `gap-analysis/tests/cm6-reference-capture.spec.ts`, added a `cm6-reference` Playwright project, and pointed the output at `gap-analysis/screenshots/cm6-reference/`. The `git show --stat` for that commit is all renames on the fixture side. The deleted `reference-screenshots/output/*.png` were the *generated* images; the generator survived.

I verified the replacement actually runs rather than assuming it: `npm run test:cm6-ref` in `gap-analysis/` captured 13/13 scenarios in ~12s. The scenario names still line up 1:1 with the Roborazzi baselines (`basic-light.html` ↔ `basic-light.png`), and both rigs render at 800x600, so the comparison the old doc described is still meaningful — it just moved directory.

The one real change in character: `gap-analysis/screenshots/` is gitignored, whereas `reference-screenshots/output/` was committed. So the CM6 side is now regenerate-on-demand. The doc says so explicitly, since "skip if already captured" is misleading on a fresh checkout.

## Changes

**`docs/screenshot-compare-workflow.md`** — rewritten. It now opens by distinguishing the two rigs, because conflating them is what makes the old doc read plausibly:

- `:view` Roborazzi — Compose Desktop, **committed** baselines, runs in CI as `:view:verifyRoborazziJvm` in the `check` job. The regression gate.
- `gap-analysis` Playwright — CM6 in Chromium plus the wasmJs showcase, **uncommitted** output. The reference.

[...]

PR #207, opened 2026-07-28
Make the wasmJs test disable a per-module opt-in (#197)

### LINKED ISSUE #197: Un-break wasmJs tests for Compose-free modules (CI currently green on 5 disabled no-op tasks)
Part of #196 (stage 0).

## Problem

Two bugs, one cause.

**CI reports green on wasm tests that never run.** `.github/workflows/ci.yml` invokes:

```
:state:wasmJsTest :collab:wasmJsTest :lezer-common:wasmJsTest
:lezer-highlight:wasmJsTest :lezer-lr:wasmJsTest
```

and `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:92-94` does:

```kotlin
tasks.configureEach {
    if (name.startsWith("wasmJsTest") || name == "wasmJsNodeTest") enabled = false
}
```

Every one of those five tasks is disabled. Gradle reports them SKIPPED, the build succeeds, and CI has been showing a green wasm signal that verifies nothing.

**The disable is over-broad.** It was added for the skiko/CanvasKit problem in Compose modules. But `:state`, `:collab`, `:lezer-common`, `:lezer-highlight`, `:lezer-lr`, `:legacy-modes` and all 17 `:lang-*` modules do not apply Compose at all — no skiko, nothing to break. The blanket `configureEach` in the shared convention plugin catches them anyway.

## Evidence it works today

Patched the disable behind an env flag in a throwaway worktree at `ded2a41` and ran it:

```
:state:wasmJsTest       173 tests, 0 failures
:lezer-common:wasmJsTest 59 tests, 0 failures
BUILD SUCCESSFUL in 1m 7s
```

232 tests execute and pass on wasmJs right now. No source changes, no skiko, no new dependencies.

## Proposed

Narrow the disable so it only applies to modules that consume Compose. Options, in preference order:

1. Move the disable out of the shared convention plugin and into a small opt-in (e.g. a `kodemirror.compose-library` convention, or an explicit per-module flag) so Compose-free modules never get it.
2. Failing that, gate it on whether the Compose plugin is applied to the project.

Then confirm CI's five named tasks actually execute, and extend the list to the other Compose-free modules (`:legacy-modes`, the 17 `:lang-*`) — those carry a large share of the 1324 commonTest methods and would gain wasm coverage for free.

Leave the disable in place for Compose-consuming modules; unblocking those is #202.

## Why

This is the cheapest coverage win available, and it closes a live CI blind spot. wasmJs is the platform where #165 and #171 both surfaced, and it is currently the *only* target executing zero tests while CI implies otherwise.

## Notes

- `wasmJs { nodejs() }` is what's configured; the 232 tests above ran under node with no browser involved.
- Please verify with the actual test-result XML counts (`build/test-results/wasmJsNodeTest/*.xml`), not just BUILD SUCCESSFUL — a disabled task also produces a successful build, which is exactly how this went unnoticed.

monkopedia-coder:
Fixes #197.

## What was wrong

`convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` blanket-disabled every
`wasmJsTest` / `wasmJsNodeTest` task through `tasks.configureEach`. That caught the five tasks
`.github/workflows/ci.yml` explicitly invokes (`:state`, `:collab`, `:lezer-common`,
`:lezer-highlight`, `:lezer-lr`). Gradle reported them SKIPPED, the build went green, and CI has
been publishing a wasm signal that verified nothing.

## The fix

The disable is now a per-module opt-in rather than a blanket rule, i.e. option 1 from the issue
("an explicit per-module flag"):

- New `KodemirrorLibraryExtension` with a single `wasmJsTests: Property<Boolean>`, registered by
  `kodemirror.library` and defaulting to `false`.
- The five Compose-free modules opt in with `kodemirrorLibrary { wasmJsTests.set(true) }`.
- The flag is read in `afterEvaluate`, so the module's own build script has run before the
  decision is made. (An eager `plugins.hasPlugin(...)`-style check in the plugin body cannot
  work: `kodemirror.library` is applied first in each module's `plugins {}` block, so anything
  applied after it is invisible at that point.)

## The issue's premise about `:lang-*` / `:legacy-modes` is wrong — those cannot be enabled

The issue expected `:legacy-modes` and the `:lang-*` modules to be Compose-free and to gain wasm
coverage for free. They are not Compose-free: every one of them depends on `:language`, which
applies Compose. Enabling their wasm tests makes them fail before a single test runs:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '.../build/wasm/packages/Kodemirror-lang-json-test/kotlin/skiko.mjs'
  imported from .../Kodemirror-lang-json-test.import-object.mjs
> :lang-json:wasmJsNodeTest (dry run) exited with errors (exit code: 1)
```

I verified this on all 15 modules under `:lang-*`/`:legacy-modes` that have tests at all
(`cpp css go html java javascript json markdown php python rust sass xml yaml`, plus
`legacy-modes`) — every one fails identically. So ci.yml's task list is unchanged; those modules
are squarely #202. The remaining 8 `:lang-*` modules have no test sources at all.

This is also why the gate is an explicit flag rather than "is the Compose plugin applied": the
Compose-plugin question gives the wrong answer for exactly these 23 modules.

## Executed test counts (from `build/test-results/wasmJsNodeTest/*.xml`, not BUILD SUCCESSFUL)

| module | tests | failures | errors | skipped |
| --- | --- | --- | --- | --- |
| `:state` | 173 | 0 | 0 | 0 |
| `:lezer-lr` | 197 | 0 | 0 | 0 |
| `:lezer-common` | 59 | 0 | 0 | 0 |
| `:lezer-highlight` | 30 | 0 | 0 | 0 |
| `:collab` | 7 | 0 | 0 | 0 |
| **total** | **466** | **0** | **0** | **0** |

Gate verified in both directions in one build: `:state/:collab/:lezer-*:wasmJsNodeTest` execute,
while `:view`, `:autocomplete`, `:lang-json` and `:legacy-modes` still report SKIPPED.

## `samples/showcase`

Left in place, with a comment added. It is a Compose app, so the stray per-module copy of the
disable is doing the right thing — it mirrors the convention plugin's new default for a module
that doesn't apply the convention plugin. The comment stops it reading as a leftover of the old
blanket rule and points at #202.

## Verification

```
./gradlew apiCheck jvmTest :view:verifyRoborazziJvm testDebugUnitTest \
  :state:wasmJsTest :collab:wasmJsTest :lezer-common:wasmJsTest \
  :lezer-highlight:wasmJsTest :lezer-lr:wasmJsTest      # BUILD SUCCESSFUL
./gradlew spotlessApply ktlintFormat                     # clean
```


monkopedia-reviewer (approved):
## Verdict: approve

**Category:** `infra` (build/CI tooling fix — no published-API surface touched; `convention-plugins` is build-logic, not a Maven artifact). Per the kodemirror override table only `public_api`/`breaking_change`/`security` deviate from the `* : tier 1` baseline, none of which apply here.
**Tier:** 1 (auto-merge on clean review).

### The crux, independently verified (not just re-reading the PR's own claims)

Checked out `fix/197-wasm-tests` into a worktree and ran the five opted-in modules plus the four
that must stay gated, in one invocation:

```
JAVA_HOME=/usr/lib/jvm/java-21-openjdk ANDROID_HOME=$HOME/Android/Sdk ./gradlew \
  :state:wasmJsTest :collab:wasmJsTest :lezer-common:wasmJsTest :lezer-highlight:wasmJsTest :lezer-lr:wasmJsTest \
  :view:wasmJsTest :autocomplete:wasmJsTest :lang-json:wasmJsTest :legacy-modes:wasmJsTest \
  -Dorg.gradle.jvmargs="-Xmx6g -Dfile.encoding=UTF-8"
```

`:wasmJsNodeTest` executed (no SKIPPED) for the five opted-in modules; SKIPPED for
`:view`/`:autocomplete`/`:lang-json`/`:legacy-modes`. Counted directly from
`build/test-results/wasmJsNodeTest/*.xml` (not BUILD SUCCESSFUL):

| module | tests | failures | errors |
|---|---|---|---|
| `:state` | 173 | 0 | 0 |
| `:collab` | 7 | 0 | 0 |
| `:lezer-common` | 59 | 0 | 0 |
| `:lezer-highlight` | 30 | 0 | 0 |
| `:lezer-lr` | 197 | 0 | 0 |
| **total** | **466** | **0** | **0** |

Matches the PR's reported table exactly. No result directories exist at all for the four
Compose-touching modules I sampled — the gate holds in both directions. Also cross-checked
the real GitHub Actions `check` job log (run 30396807430): the exact five `wasmJsNodeTest`
tasks appear un-skipped, `BUILD SUCCESSFUL`, `3471 actionable tasks: 3471 executed`.

### Premise correction, verified

`grep implementation(project(":language")) lang-json/build.gradle.kts` — confirmed. `language/build.gradle.kts`
applies `alias(libs.plugins.jetbrainsCompose)` + `compose.compiler` and depends on `compose.ui`/`compose.foundation`/
`compose.runtime`. So `:lang-json` (and by extension every `:lang-*`/`:legacy-modes` module) really does reach
Compose transitively, and the issue's "gate on whether Compose plugin is applied" alternative would indeed have
been wrong — it would flip ~15 modules to enabled and turn a silent no-op into red `wasmJsNodeTest` tasks
(`ERR_MODULE_NOT_FOUND: skiko.mjs`). The chosen explicit opt-in avoids that; correctly designed.

### Design review

- `afterEvaluate` + `extensions.create<KodemirrorLibraryExtension>` is the standard, idiomatic pattern for
  "convention plugin applied first in `plugins{}`, per-module override configured later in the same script" —
  not the problematic cross-project `afterEvaluate` ordering footgun, just single-project self-deferral. Correct
  hook.
- Default-`false` (opt-in to *enable*) means a new Compose-free module gets silent non-coverage until someone
  remembers the flag, rather than a loud failure. Given ~5 Compose-free modules vs. ~23 Compose-touching ones in
  this tree, defaulting to disabled minimizes false-failure noise for the common case, at the cost of relying on
  the (thorough) KDoc warning in `KodemirrorLibraryExtension.kt` to catch the rare new Compose-free module. It's a
  legitimate trade-off, well-documented, not a defect — noting it rather than blocking on it.
- `samples/showcase/build.gradle.kts`: coherent. It doesn't apply `kodemirror.library`, so it can't use the new
  extension; it keeps its own copy of the disable (mirroring the convention plugin's default) and adds a comment
  pointing at #202/#197 so it doesn't read as forgotten dead code.
- `ci.yml`: comment-only (4 added lines, 0 removed besides the diff header), task list unchanged — correct, since
  the premise correction means none of the other 23 modules can be added yet; that's squarely #202.

### Scope

`git diff origin/main...HEAD --stat` = exactly the 10 files GitHub reports, nothing more. Confirmed #204's
`stripStableFromApiDumps`/per-target `*/api/*/*.api` glob (root `build.gradle.kts:64,70`) and #206's
`runGapAnalysis` → `wasmJsBrowserDevelopmentExecutableDistribution` dependency (`build.gradle.kts:117-123`) are
both untouched and still present.

### CI

`check` / `check-ios` / `check-macos` all `pass` (real GitHub Actions checks, not local re-verification).

Approving and squash-merging.

Issue #208, opened 2026-07-28
Migrate the Android target to com.android.kotlin.multiplatform.library (AGP 10 removes the legacy KMP path)


monkopedia-coder:
Follow-up to #195, which bumped AGP to 9.3.1.

AGP 9 turns on its new Android DSL by default (`android.newDsl=true`) and its built-in
Kotlin support (`android.builtInKotlin=true`). Neither is compatible with the Kotlin
Multiplatform plugin's `androidTarget()` + `com.android.library` combination that
`convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` uses for all ~60 modules:

- `newDsl=true` → KMP hard-fails at plugin-apply time:
  "The 'org.jetbrains.kotlin.multiplatform' plugin with `androidTarget()` enabled is not
  compatible with AGP's 9.0 new DSL", root cause
  `LibraryExtensionImpl$AgpDecorated_Decorated cannot be cast to com.android.build.gradle.BaseExtension`.
- `builtInKotlin=true` → "Cannot add extension with name 'kotlin', as there is an extension
  already registered with that name" (AGP registers its own `kotlin` extension).

#195 opted back out of both via `gradle.properties`. AGP prints a deprecation warning for
each, and per https://kotl.in/gradle/agp-new-kmp the legacy APIs behind them are slated for
removal in **AGP 10 (H2 2026)** — at which point the opt-out disappears and the build breaks
with no flag to set.

## The actual migration

Replace `com.android.library` + `kotlin { androidTarget() }` with the
`com.android.kotlin.multiplatform.library` plugin + `kotlin { android { } }`. This is not a
one-line swap; the known deltas that touch this repo:

- **Source-set directory renames**: `src/test` → `src/androidHostTest`,
  `src/androidTest` → `src/androidDeviceTest`, `src/main` → `src/androidMain`.
- **Tests are off by default** — must opt in with `withHostTest { }`. The task name changes,
  so `.github/workflows/ci.yml`'s `testDebugUnitTest` needs updating.
- **Java compilation is off by default** — `withJava()` opt-in; the convention plugin's
  `compileOptions { sourceCompatibility/targetCompatibility = VERSION_11 }` becomes
  `compilerOptions { jvmTarget.set(JvmTarget.JVM_11) }`.
- **Android resources are off by default** — `androidResources { enable = true }` if needed.
- `namespace`/`compileSdk`/`minSdk` move from the top-level `android { }` block into
[...]

Issue #194, opened 2026-07-26
Refresh remaining deps to latest stable (Gradle 9.0.0→9.6.1, roborazzi 1.68.0→1.70.0, kover 0.9.8→0.9.9) + drop dead lsp-ksrpc alias


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** dep-bump

**Where:** `gradle/wrapper/gradle-wrapper.properties:3`, `gradle/libs.versions.toml:10,11,24`

| coordinate | current | latest stable | jump |
|---|---|---|---|
| Gradle distribution (wrapper) | 9.0.0 | **9.6.1** | minor |
| `io.github.takahirom.roborazzi` (lib + plugin) | 1.68.0 | **1.70.0** | minor |
| `org.jetbrains.kotlinx.kover` | 0.9.8 | **0.9.9** | patch |

**Current:** The rest of the tree is genuinely current — this run verified Kotlin 2.4.10, Compose MP 1.11.1, coroutines 1.11.0, serialization-json 1.11.0, atomicfu 0.33.0, ktlint-gradle 14.2.0, spotless 8.8.0, dokka 2.2.0, bcv 0.18.1, vanniktech 0.37.0, activity-compose 1.13.0, and `com.monkopedia.lsp` 1.2.1 are all at latest stable. The three above are what's left after #178's refresh. Every version was read from `maven-metadata.xml` on repo1 / dl.google.com / plugins.gradle.org, and `services.gradle.org/versions/current`; pre-releases (Kotlin `2.4.20-Beta2`, Compose `1.12.0-beta02`, dokka `2.2.0-Beta`, vanniktech `0.37.0-rc1`) were excluded.

**Also, one dead catalog entry:** `gradle/libs.versions.toml:24` declares `lsp-ksrpc = { module = "com.monkopedia.lsp:lsp-ksrpc", … }` and no module references `libs.lsp.ksrpc` — the declaration is its only occurrence in the repo.

**Proposed:** Bump the three above; drop the unused `lsp-ksrpc` alias while in the file. All are build/test tooling — roborazzi is screenshot-test-only and kover is coverage-only, so neither touches published artifacts. The Gradle wrapper bump is the only one with any breadth, and 9.0.0 → 9.6.1 is within the same major.

[...]

PR #189, opened 2026-07-24
Pin wasm npm tooling Node to 22.11.0 so showcase webpack build stops using Node 25

monkopedia-coder:
## Problem

Main has been red on the `Docs` / `build-showcase` CI job: `:samples:showcase:wasmJsBrowserProductionWebpack` fails with

```
> A problem occurred starting process 'command '/home/runner/.gradle/nodejs/node-v25.0.0-linux-x64/bin/node''
```

The required `check` job stays green because it never runs webpack, so this slipped in (build-showcase is not a required check). Pre-existing — failed on `5203bf66` before #185/#186 too.

## Root cause

The 0.3.5 Node pin sets `WasmNodeJsEnvSpec.version` (+ the JS `NodeJsEnvSpec`) to `22.11.0`. That correctly pins the **target-node** users — `:kotlinWasmNpmInstall` / yarn run on `node-v22.11.0`. But the wasm **npm tooling** node — the one `wasmJsBrowser*Webpack` launches to run `webpack.js` — is resolved from the *legacy* `WasmNodeJsRootExtension.version`, which the EnvSpec does **not** update. So webpack kept launching the Kotlin 2.4.10 default `node-v25.0.0`.

Confirmed by reproducing locally with `--info` (same run, two different nodes):
- yarn: `.../nodejs/node-v22.11.0-linux-x64/bin/node ... yarn.js` ✓ (EnvSpec pin)
- webpack: `.../nodejs/node-v25.0.0-linux-x64/bin/node ... webpack.js` ✗ (RootExtension default)

## Fix

Also pin `WasmNodeJsRootExtension.version` to `22.11.0` inside `withType<WasmNodeJsRootPlugin>`. The setter is a scheduled-for-removal deprecation and the build compiles with `-Werror`, so it's wrapped in a `@Suppress("DEPRECATION", "DEPRECATION_ERROR")` local function (documented; drop once the tooling node honours the EnvSpec).

## Verification

Locally, `:samples:showcase:wasmJsBrowserProductionWebpack` now launches `node-v22.11.0` and is **BUILD SUCCESSFUL** (was node-25 → start-process failure). The `build-showcase` job on this PR is the remote proof.

## Scope
Build config only (`build.gradle.kts`) + CHANGELOG. No source/API change.

PR #209, opened 2026-07-28
Bump AGP 8.13.2 → 9.3.1 and finish the Gradle wrapper at 9.6.1 (#195)

### LINKED ISSUE #195: Bump AGP 8.13.2 → 9.3.1 (major — 8.x line is ended, hits all ~60 modules via the convention plugin)

**Category:** dep-bump (major version — filed separately so the migration is judged on its own)

**Where:** `gradle/libs.versions.toml:2` — `agp = "8.13.2"`, consumed via `androidLibrary = { id = "com.android.library", version.ref = "agp" }`

| coordinate | current | latest stable | jump |
|---|---|---|---|
| `com.android.tools.build:gradle` (AGP) | 8.13.2 | **9.3.1** | **major** |

**Current:** The repo is on AGP 8.13.2, which is the **last release of the 8.x line** — Google's `maven-metadata.xml` shows 8.13.2 then straight into 9.x, now four stable minors deep (9.3.1 is latest stable; 9.4.0 exists only as alphas, excluded). So there is no further 8.x patch to take; the only path forward is the major.

**Proposed:** Bump AGP to 9.3.1, sequenced *after* the Gradle wrapper bump filed alongside this.

**Why / what the surface actually is here:** worth being honest that this one is not mechanical. Every module in the tree goes through `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts`, which applies `com.android.library` with `compileSdk = 36`, `minSdk = 24` and Java 11 compile options, and wires `androidMain.dependsOn(jvmMain)`. So an AGP behaviour change hits all ~60 modules at once rather than being containable to one. Two specific things to check before starting:

- **AGP 9's minimum Gradle version** — I did not verify this. Treat the wrapper bump and this as a sequenced pair, wrapper first.
- **`apiCheck`/`apiDump` build the Android variant**, so an AGP break shows up as an API-validation failure rather than an obvious compile error.

CI coverage is decent for this: `check` runs JVM/wasmJs/Android, so a break should surface on the PR rather than in a consumer's build.

**Why do it at all:** staying on an EOL major only makes the eventual jump worse, and the gap widens every month. Doing it now, while the only blocker is a version bump and not also a Gradle major, is the cheap moment.

**Effort:** larger — expect a real migration pass, not a one-line edit.

monkopedia-coder:
Fixes #195.

AGP `8.13.2` → `9.3.1` (latest stable — 9.4.0 exists only as alphas). 8.13.2 is the last
AGP 8.x release, so there was no patch path off it.

While we're here, the wrapper finishes the move #194 had to stop short of: `9.5.1` → `9.6.1`.
As the correcting comment on #195 established, the dependency runs the *opposite* way from how
both tickets originally described it — Gradle 9.6.0 removed
`org.gradle.api.problems.internal.InternalProblems`, which **AGP 8.x** binds to, so AGP 8
imposed a *maximum* Gradle of 9.5.x. Landing AGP 9 is what unblocks the wrapper, not the other
way round, and this PR verifies that end-to-end rather than assuming it.

## AGP 9 requirements — all already satisfied

| | AGP 9.3 requires | This tree |
|---|---|---|
| Gradle | 9.5.0+ | 9.5.1 → **9.6.1** here |
| JDK | 17+ | 21 |
| SDK Build Tools | 36.0.0 | 36.0.0 |
| compileSdk | ≤ 37 | 36 (unchanged) |
| minSdk | — | 24 (unchanged) |

No change was needed to `kodemirror.library.gradle.kts` — `namespace`, `compileSdk`,
`defaultConfig.minSdk` and `compileOptions` all survive the bump as written.

## The two breakages, and why the fix lives in `gradle.properties`

AGP 9 flips two defaults that are each independently incompatible with the Kotlin
Multiplatform `androidTarget()` + `com.android.library` combination that
`convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` applies to all ~60 modules.

**1. `android.newDsl=true`.** KGP hard-fails at plugin-apply time:

```
Failed to apply plugin 'org.jetbrains.kotlin.multiplatform'
  The 'org.jetbrains.kotlin.multiplatform' plugin with `androidTarget()` enabled is not
  compatible with AGP's 9.0 new DSL (`android.newDsl=true` is enabled by default).
  > class com.android.build.gradle.internal.dsl.LibraryExtensionImpl$AgpDecorated_Decorated
    cannot be cast to class com.android.build.gradle.BaseExtension
```

**2. `android.builtInKotlin=true`.** With `newDsl` off, AGP's own built-in Kotlin support is
next, because it registers a `kotlin` extension of its own:

```
Failed to apply plugin 'org.jetbrains.kotlin.multiplatform'
  > Cannot add extension with name 'kotlin', as there is an extension already registered with that name.
```

Both are opted back out of in `gradle.properties`. AGP itself documents this as the supported
interim path (its error message names `android.newDsl=false` explicitly) and prints a
deprecation warning for each; the legacy APIs behind them are removed in AGP 10, so the real
migration to `com.android.kotlin.multiplatform.library` is filed as **#208** rather than
attempted here — it renames source-set directories (`src/test` → `src/androidHostTest`),
makes host tests, Java compilation and Android resources opt-in, and changes the
`testDebugUnitTest` task name that `ci.yml` invokes. That is a genuinely separate change with
its own ABI-verification burden, and bundling it would have made this bump unreviewable.

### Why `systemProp.org.gradle.project.` and not plain project properties

This is the non-obvious part of the diff and is worth a look. Plain `android.newDsl=false` in
the root `gradle.properties` **does not fix the build**, and neither does adding a
`convention-plugins/gradle.properties`, and neither does `-Pandroid.newDsl=false` on the
command line. All three still fail in `:convention-plugins:generatePrecompiledScriptPluginAccessors`.

`kodemirror.library` is a precompiled script plugin living in the `convention-plugins`
*included* build. To generate its type-safe accessors, Gradle applies the script's `plugins { }`
block to a **synthetic project** that reads neither the root build's `gradle.properties` nor
the included build's, and does not inherit `-P` properties — so AGP sees its defaults there and
the build dies before a single real module is configured. Setting the flags as
`org.gradle.project.`-prefixed *system* properties (which Gradle maps back to project
properties) is what reaches every build in the tree, synthetic ones included. There is a comment
in `gradle.properties` recording this so the next person doesn't "simplify" it back to the
broken form.

## No ABI change

`apiCheck` passes clean — **zero `.api` diff**, and no `apiDump` was run. A major AGP bump
should not move the published ABI, and it didn't. `git status` after the full verification run
shows only the six intentionally-edited files.

## Verification

Run twice: once on Gradle 9.5.1 to isolate the AGP bump, then again after taking the wrapper to
9.6.1. All green both times.

| Command | AGP 9.3.1 / Gradle 9.5.1 | AGP 9.3.1 / Gradle 9.6.1 |
|---|---|---|
| `./gradlew apiCheck` | pass (no `.api` diff) | pass (no `.api` diff) |
| `./gradlew jvmTest` | pass | pass |
| `./gradlew testDebugUnitTest` | pass | pass |
| `./gradlew :view:verifyRoborazziJvm` | pass | pass |
| `./gradlew :samples:showcase:wasmJsBrowserDevelopmentExecutableDistribution` | pass | pass |
| `./gradlew :samples:android:assembleDebug` | pass | pass |

The showcase distribution is included because `build-showcase` lives in `docs.yml` under
`on: push: branches: [main]` with no `pull_request` trigger — wasm/webpack breakage would
otherwise only surface after merge. `:samples:android:assembleDebug` is included because it is
the one `com.android.application` module in the tree and no CI job builds it.

Not runnable here: `ci-apple.yml`'s `macosArm64Test` / `iosSimulatorArm64Test`. `apiCheck` did
cross-compile the `macosArm64` klibs successfully, and AGP does not participate in the native
compilations, but the Apple checks on this PR are the real signal.

The wasm npm tooling Node 22.11.0 pin in the root `build.gradle.kts` (#189) was **not** touched.
Nothing about AGP 9 or Gradle 9.6.1 interacted with it — the showcase webpack build resolved and
compiled normally on both Gradle versions.

## Notes for review

- Gradle emits `WARNING: The option setting 'android.newDsl=false' is deprecated.` and the same
  for `android.builtInKotlin=false` on every build. That is expected and is the tracked-in-#208
  debt, not a regression.
- The `gradlew` / `gradlew.bat` diffs are the wrapper task's own output from the standard
  two-pass `./gradlew wrapper --gradle-version 9.6.1` run.
- No `.kt`/`.kts` files were touched, so `spotlessApply` / `ktlintFormat` had nothing to do.


monkopedia-reviewer (approved):
Approving after independent re-verification on the rebased head (`26b9288`, confirmed to contain main's `7ce2d9d`).

**CI on 26b9288 (not the stale pre-rebase run):** `check` pass (12m17s), `check-ios` pass (15m39s), `check-macos` pass (11m10s). Watched to completion, not sampled once.

**Rebase survived intact.** Both #207 and #209 land cleanly together:
- `KodemirrorLibraryExtension.wasmJsTests` (default `false`, read in `afterEvaluate`) is present in `convention-plugins/src/main/kotlin/KodemirrorLibraryExtension.kt` / `kodemirror.library.gradle.kts`, and exactly the five expected modules (`state`, `collab`, `lezer-common`, `lezer-highlight`, `lezer-lr`) set `wasmJsTests.set(true)`.
- Proved the gate rather than trusting BUILD SUCCESSFUL (a disabled task also exits 0 — that's the #197 bug). Ran all five wasmJsTest tasks plus `:view:wasmJsTest` as a Compose control and read the executed counts straight out of `build/test-results/wasmJsNodeTest/*.xml`: state 173, collab 7, lezer-common 59, lezer-highlight 30, lezer-lr 197 — **466/466, 0 failures**, exactly matching the CHANGELOG claim. `:view` produced no `test-results/wasmJsNodeTest` directory at all — genuinely skipped, not silently green.
- The AGP 9 diff itself is tiny and isolated: only `CHANGELOG.md`, `gradle.properties` (+21 lines), `gradle/libs.versions.toml` (agp 8.13.2→9.3.1), `gradle/wrapper/gradle-wrapper.properties` (9.5.1→9.6.1), `gradlew`, `gradlew.bat`. No `.api` file appears anywhere in the diff — apiCheck confirmed clean, so the bump is genuinely ABI-neutral.

**The systemProp workaround is real, not needless indirection — verified by breaking it.** Reset the worktree, stopped all Gradle daemons, deleted `convention-plugins/build`/`.gradle` to force a truly cold reconfigure, swapped the committed `systemProp.org.gradle.project.android.newDsl=false` / `...builtInKotlin=false` back to plain `android.newDsl=false` / `android.builtInKotlin=false`, and reran. Result: `:convention-plugins:generatePrecompiledScriptPluginAccessors FAILED` with exactly the cast exception the PR describes (`LibraryExtensionImpl$AgpDecorated_Decorated cannot be cast to ... BaseExtension`) — confirming the precompiled-script-plugin synthetic-project explanation is correct and the plain form is not sufficient. The in-file comment in `gradle.properties` records this reasoning for future readers.

**Wrapper integrity — reproduced, not trusted.** Regenerated the Gradle wrapper independently in a scratch project via the standard two-pass flow (bootstrap with the old 9.5.1 wrapper → rerun `wrapper` task against the freshly-downloaded 9.6.1 distribution). Second-pass `gradle-wrapper.jar` SHA-256 (`497c8c2a...`), `gradlew`, and `gradlew.bat` are byte-for-byte identical to what's committed. Not hand-edited.

**Node pin (#189) untouched** — `build.gradle.kts` has no diff; `NodeJsEnvSpec`/`WasmNodeJsEnvSpec` still pinned to 22.11.0, and the showcase wasmJs webpack build (below) resolved and used `node-v22.11.0`, not Node 25.

**Showcase build — the one thing no PR check covers** (`build-showcase` in `docs.yml` triggers only on `push: branches:[main]`, confirmed in the workflow file). Ran `:samples:showcase:wasmJsBrowserDevelopmentExecutableDistribution` locally: BUILD SUCCESSFUL, webpack compiled cleanly.

**Full local matrix**, run on the rebased head: `apiCheck jvmTest testDebugUnitTest :view:verifyRoborazziJvm :samples:android:assembleDebug` — BUILD SUCCESSFUL (3286 tasks). `:samples:android:assembleDebug` in particular has no CI job.

**#204/#206 not regressed:** the repaired `apiCheck`/`stripStableFromApiDumps` globs (`*/api/*/*.api`, `build/api/<target>/*.api`) in root `build.gradle.kts` are untouched by this diff, and `runGapAnalysis`'s dependency on `wasmJsBrowserDevelopmentExecutableDistribution` (not the old `...Webpack` task) is still in place.

**Scope judgement:** deferring the `com.android.kotlin.multiplatform.library` migration to #208 is the right call — it's a genuinely separate change (source-set renames, opt-in host tests/Java compilation, a `testDebugUnitTest` task-name change that `ci.yml` invokes directly) with its own ABI-verification burden, and #208 is filed and open, correctly scoped as AGP-10-deadline follow-up.

Tier: 1 (mechanical dependency bump + wrapper bump, verified ABI-neutral, all real CI green, no behavior change intended or observed). Merging.

Issue #196, opened 2026-07-26
Cross-platform test coverage: make every target verifiable (tracking)


monkopedia-coder:
## Goal

Make every published target — JVM, Android, iOS, macOS, wasmJs — **verifiable**, so the README can honestly say "functional" instead of the current "compile and pass unit tests but are not battle-tested."

This is a tracking issue. Individual work lands in the linked tickets.

## Where we are today (measured, not estimated)

| Platform | What actually executes | UI coverage |
|---|---|---|
| JVM | 1324 commonTest + 1059 jvmTest `@Test` | 63 interaction + 15 screenshot |
| Android | commonTest only (`testDebugUnitTest`) | none |
| macOS / iOS | commonTest only (`macosArm64Test`, `iosSimulatorArm64Test`) | none |
| wasmJs | **zero** | none |

Three findings behind those numbers:

**1. wasm executes nothing, and CI reports otherwise.** `ci.yml` runs `:state:wasmJsTest :collab:wasmJsTest :lezer-common:wasmJsTest :lezer-highlight:wasmJsTest :lezer-lr:wasmJsTest`, but the `kodemirror.library` convention plugin disables every `wasmJsTest*` task. All five are silent no-ops. The disable was added for the skiko/Compose problem, but none of those five modules applies Compose. Measured with the disable lifted: **232 tests pass in 1m07s** (`:state` 173, `:lezer-common` 59, zero failures).

**2. 970 test methods are mis-filed.** 56 files in `jvmTest` have no Compose dependency and (except one) no JUnit dependency — they're already on `kotlin.test`. `VimTestHarness.kt`'s only non-project imports are `kotlin.test.assertEquals` and `kotlin.test.fail`. They're in `jvmTest` by habit, not necessity.

| module | @Test | files |
|---|---|---|
| vim | 629 | 21 |
| lsp-client | 166 | 12 |
| commands | 94 | 11 |
| autocomplete | 45 | 7 |
| view / collab / search / lint | 36 | 5 |

**3. UI coverage is bug-shaped, not coverage-shaped.** All 63 interaction tests live in `:view` and are dominated by hit-testing and scrolling — the #165/#169/#171 lineage. There are **zero UI tests for vim, autocomplete, search, or lint**. Vim has 21 test files and none touches the UI.

## The Playwright question

`gap-analysis/` has 214 Playwright tests. They are **differential** — each runs the same input against real CodeMirror 6 and against KodeMirror and compares:

```ts
[...]

PR #211, opened 2026-07-30
Move :vim tests from jvmTest to commonTest (#198)

### LINKED ISSUE #198: Move vim's 629 tests from jvmTest to commonTest (no Compose, no JUnit — JVM-only by habit)
Part of #196 (stage 1).

## Problem

`:vim` has **629 test methods across 21 files**, and every one of them runs on JVM only. Vim is a headline feature; on Android, iOS, macOS and wasm it currently has zero verified behavior.

There is no technical reason for this. `vim/src/jvmTest/kotlin/com/monkopedia/kodemirror/vim/VimTestHarness.kt` has exactly two non-project imports:

```kotlin
import kotlin.test.assertEquals
import kotlin.test.fail
```

No JUnit, no Compose, no JVM APIs. The harness drives the vim engine at the state level, headlessly. The whole suite is in `jvmTest` by habit.

## Proposed

Move `vim/src/jvmTest/` → `vim/src/commonTest/` (all 21 files), adjust the module's test dependencies, and confirm the suite executes on every target.

Files: `VimTestHarness.kt` plus `VimMotionTest`, `VimOperatorTest`, `VimTextObjectTest`, `VimVisualModeTest`, `VimInsertModeTest`, `VimSearchTest`, `VimSubstituteTest`, `VimRegisterTest`, `VimMacroTest`, `VimMappingTest`, `VimMarkTest`, `VimExCommandTest`, `VimMiscTest`, `VimScrollIntoViewTest`, `VimDisplayLineMotionTest`, `VimWordObjectEditTest`, and the four `*Debug*Test` files.

Verify per target rather than assuming:
- `:vim:jvmTest` — must stay green (regression check)
- `:vim:testDebugUnitTest` — Android
- `:vim:macosArm64Test` / `:vim:iosSimulatorArm64Test` — Apple
- `:vim:wasmJsTest` — **blocked**: `:vim` applies Compose, so it's gated behind #202. Expect this one not to run yet; that's fine, the other four are the win.

Check the `*Debug*Test` files first — if they were scratch debugging aids rather than real coverage, deleting them is better than porting them. Report rather than deciding unilaterally.

## Why

629 tests is the single largest coverage-per-effort item in the repo, and it's close to a directory move. It converts vim from "compiles on iOS" to "vim engine verified on iOS."

## Caveat

This proves the **engine** works cross-platform, not the **UI**. Vim has no UI tests at all; that gap is stage 4 of #196.

monkopedia-coder:
Part of #196 (stage 1).

`:vim` had 629 test methods in `vim/src/jvmTest/`, so vim — a headline feature — had zero verified behaviour on Android, iOS or macOS. There was no technical reason for it: `VimTestHarness.kt` imports exactly two non-project symbols (`kotlin.test.assertEquals`, `kotlin.test.fail`) and drives the vim engine headlessly at `EditorState` level. No JUnit, no Compose, no JVM APIs.

## Change

- `vim/src/jvmTest/` → `vim/src/commonTest/`, 18 files, content unchanged (`git diff --stat` shows 0 lines changed for every moved file — it reads as a pure move).
- `vim/build.gradle.kts`: the `jvmTest.dependencies` block folded into `commonTest.dependencies` verbatim (`kotlin("test")` plus `:state`, `:view`, `:commands`, `:search`, `:language`); the old `commonTest` block only had `:state`.
- Nothing outside `vim/` and `CHANGELOG.md` is touched — in particular the convention plugin is untouched (that is #202's area).

## The four `*Debug*Test` files

**Deleted — `DebugExTest`, `DebugEx2Test`, `DebugEx3Test`.** Each contains a single test, all three named `debug_ex_go_to_line`, all three asserting the same thing (`:4` puts the cursor on line 3) at a different level of the stack — via key events, via `exCommandDispatcher.processCommand`, via `Vim.handleEx` — with `println` tracing left in. That is a bisection session someone committed. The behaviour is already covered by `VimExCommandTest.ex_go_to_line` (`h.doEx("4")`), so deleting loses no coverage.

**Kept and moved — `SearchDebugTest`.** Despite the name, its five tests carry real assertions with explanatory failure messages and no debug output. They cover `getSearchCursor`/`findNext` backward wrap-around from offset 0, which `VimSearchTest` does not exercise (it has `star_wraps_around` and `backward_search_and_n_N`, but nothing for the from-zero wrap).

Net: 629 − 3 = 626 tests.

## Verification — executed counts read from `build/test-results/<task>/*.xml`, not from BUILD SUCCESSFUL

| Task | XML files | tests | failures | skipped |
|---|---|---|---|---|
| `:vim:jvmTest` | 17 | **626** | 0 | 0 |
| `:vim:testDebugUnitTest` (Android) | 17 | **626** | 0 | 0 |

- `:vim:macosArm64Test` — **not verifiable on this Linux box**: the task and `linkDebugTestMacosArm64` are `SKIPPED` (host requirement), so `BUILD SUCCESSFUL` there means nothing. Relies on CI `check-macos` / `check-ios`. As a compile-side proxy that the moved sources are genuinely platform-neutral, `:vim:compileTestKotlinWasmJs` was run and passes — the test source set compiles on a non-JVM backend.
- `:vim:wasmJsTest` — confirmed still `SKIPPED`. `:vim` applies Compose so `wasmJsTests` stays at its `false` default (`KodemirrorLibraryExtension`, #197); unblocking it is #202 and is deliberately not attempted here.
- `:vim:apiCheck` green, no `.api` diff.
- `spotlessApply` + `ktlintFormat` run; they produced no changes, so there is no formatting churn in the diff.

Fixes #198


monkopedia-reviewer (changes requested):
## Independent verification — everything except CI is sound

- **Moves are pure.** `git diff -M origin/main HEAD --stat` shows 0 changed lines for all 17 renamed files; content is untouched.
- **`vim/build.gradle.kts`**: the old `jvmTest.dependencies` block is folded into `commonTest.dependencies` verbatim (`kotlin("test")`, `:state`, `:view`, `:commands`, `:search`, `:language`) — appropriate for a `commonTest` move, nothing dropped, nothing extraneous added.
- **Deletion of `DebugExTest`/`DebugEx2Test`/`DebugEx3Test`**: read all three at `origin/main` — each is a single `debug_ex_go_to_line` test asserting the identical behavior (`:4` → line 3) at a different call-stack depth (key events / `exCommandDispatcher.processCommand` / `Vim.handleEx`), with `println` tracing left in. Confirmed the same assertion is already covered by `VimExCommandTest.ex_go_to_line` (`h.doEx("4")` / `h.assertCursorAt(3, 0)`). Deleting these is the right call and is reported, not silent.
- **`SearchDebugTest` kept**: read it — five real assertions (backward search-cursor wrap-around from offset 0) with explanatory failure messages, no debug output, and not otherwise covered (`VimSearchTest` has `star_wraps_around`/`backward_search_and_n_N` but nothing for from-zero wrap). The keep/delete asymmetry is principled.
- **Scope**: only `vim/` + `CHANGELOG.md` touched — confirmed via `git diff origin/main HEAD --stat -- . ':!vim' ':!CHANGELOG.md'` (empty). Doesn't stray into #202's `convention-plugins/` territory.
- **Counts, independently re-executed** (not trusted from the PR body) — read from `build/test-results/<task>/*.xml`, not exit code:
  - `:vim:jvmTest`: 17 XML files, **626 tests, 0 failures, 0 skipped**.
  - `:vim:testDebugUnitTest` (Android): 17 XML files, **626 tests, 0 failures, 0 skipped**.
  - Arithmetic reconciles: 629 − 3 deleted = 626. No silent coverage loss.
- **`apiCheck`**: ran clean, no `.api` diff, working tree clean afterward.
- `wasmJsTest` correctly stays disabled (`:vim` applies Compose; `KodemirrorLibraryExtension.wasmJsTests` untouched) — that's #202's problem, correctly not attempted here.

## Blocker: CI is red on both Apple targets, and it's not a flake

`gh pr checks 211` (waited for full completion): `check` pass, **`check-macos` fail, `check-ios` fail**.

Pulled both job logs directly:
- `check-macos` (run 30568465084 / job 90958803116): `com.monkopedia.kodemirror.vim.VimExCommandTest.ex_sort_pattern_no_r[macosArm64] FAILED — kotlin.AssertionError at null:-1`. `626 tests completed, 1 failed`.
- `check-ios` (run 30568465084 / job 90958803099): same test, same assertion, `[iosSimulatorArm64] FAILED`. `626 tests completed, 1 failed`.

Identical failure, same test, on both native targets, 0 failures on JVM/Android — this is a real, deterministic JVM-vs-Kotlin/Native behavioral divergence in the `sort` ex-command (`ex_sort_pattern_no_r`, `VimExCommandTest.kt`), not test flakiness or an environment hiccup. This is exactly the class of gap #198 set out to find — the move is doing its job — but it also means the PR as submitted cannot merge: CI must be green, and it isn't.

**Requesting changes, not bouncing the whole PR's design** — the move/delete/dependency work above is independently verified correct and doesn't need to change. What's needed before merge is one of:
1. Fix the underlying native/JVM divergence in the `sort` ex-command handling (`VimExCommands.kt`'s `"sort"` implementation) so `ex_sort_pattern_no_r` passes on all targets, or
2. If a fix is out of scope for this PR, explicitly skip/annotate that one assertion on native targets with a comment linking a new tracking issue for the platform divergence — reported openly in the PR body, not silently dropped — so CI goes green while the other 625 tests keep their new cross-platform coverage.

Please push a follow-up commit addressing this and I'll re-review.

monkopedia-coder:
### First CI run found a real bug — pushed a fix in de60bc9

The move worked exactly as intended on the Apple targets: **626 tests executed** on both `macosArm64` and `iosSimulatorArm64`, same count as JVM/Android, so the suite is genuinely running and not silently empty. 625 passed. One failed, identically on both:

```
com.monkopedia.kodemirror.vim.VimExCommandTest.ex_sort_pattern_no_r[macosArm64] FAILED
626 tests completed, 1 failed
```

This is not a test-porting artifact — it is a genuine cross-platform correctness bug in shipped `commonMain` code that has been invisible for as long as the vim suite was JVM-only.

**Cause.** Both comparators in the `sort` ex command (`VimExCommands.kt`) ended in `if (sa < sb) -1 else 1`, answering "greater than" for *equal* keys. `compare(a, b)` and `compare(b, a)` then both return `1` instead of having opposite signs, which breaks the `Comparator` contract, and the order of tied keys falls out of whichever sort algorithm the target uses rather than out of the comparator. JVM `Collections.sort` (TimSort) takes from the left run unless `compare(right, left) < 0`, which with the broken comparator still left ties in place; Kotlin/Native's merge sort takes from the *right* run whenever `compare(left, right) > 0` and so reversed every tied group.

`ex_sort_pattern_no_r`'s second step is `:sort r/in/`, where the `r` flag makes the sort key the matched text itself — the literal `in` for all four matching lines, i.e. a complete four-way tie. JVM preserved their order (the assertion's expectation, and what vim does); Kotlin/Native did not.

**Fix.** Both comparators now use `sa.compareTo(sb)`, returning `0` for equal keys. `MutableList.sortWith` is contractually stable on every target, so once the comparator is consistent, tied lines keep their original relative order everywhere — no explicit index tie-break needed, and the JVM/Android output is unchanged (626/626 still green locally on both after the fix).

Re-verified locally after the fix, counts from `build/test-results/<task>/*.xml`:

| Task | XML files | tests | failures |
|---|---|---|---|
| `:vim:jvmTest` | 17 | 626 | 0 |
| `:vim:testDebugUnitTest` | 17 | 626 | 0 |

`:vim:apiCheck` green, no `.api` diff. Waiting on `check-macos` / `check-ios` to confirm the Apple side.

monkopedia-coder:
### CI is fully green on de60bc9 — Apple targets confirmed

Follow-up to the comment above, which was posted while the Apple jobs were still running. All three required checks have now completed and passed on the fix commit:

| Check | Result | Duration |
|---|---|---|
| `check` (JVM / wasmJs / Android) | **pass** | 10m29s |
| `check-macos` | **pass** | 15m34s |
| `check-ios` | **pass** | 15m54s |

Run [30570787761](https://github.com/Monkopedia/kodemirror/actions/runs/30570787761) (Apple) / [30570788409](https://github.com/Monkopedia/kodemirror/actions/runs/30570788409). Pulled both Apple job logs: `> Task :vim:macosArm64Test` and `> Task :vim:iosSimulatorArm64Test` both executed (no `SKIPPED`/`NO-SOURCE`/`UP-TO-DATE` suffix), and neither log now contains a `tests completed, N failed` line or any `FAILED` — the previous run's `626 tests completed, 1 failed` / `ex_sort_pattern_no_r[macosArm64] FAILED` is gone. The vim suite is genuinely executing on both Apple targets and fully passing.

### Independent confirmation of the diagnosis, before the fix was written

Rather than infer the mechanism from the failure alone, it was verified two ways:

**1. Read the Kotlin/Native stdlib source.** `kotlin-stdlib-native-sources.zip` from the 2.4.10 konan prebuilt (`nativeWasmMain/kotlin/collections/ArraySorting.kt`) merges with:

```kotlin
if (comparator.compare(leftValue, rightValue) <= 0) {
    target[i] = leftValue; leftIndex++
} else {
    target[i] = rightValue; rightIndex++
}
```

With the old comparator returning `1` for equal keys, `1 <= 0` is false, so the merge took the **right** element first — precisely anti-stable. JVM TimSort's merge asks `compare(right, left) < 0` instead, which for the same `1` is false and takes the left element, i.e. stable. That asymmetry is the entire bug.

**2. Compiled and ran it on a real Kotlin/Native target.** The pattern branch was extracted into a standalone program and built with `kotlinc-native` (linuxX64, same stdlib merge sort as the Apple targets), running all five steps of `ex_sort_pattern_no_r` with the old and new comparator side by side:

```
BUGGY step2 r/in/    :  z \n\n2 in d \n1 in c \n3 in a \n in
FIXED step2 r/in/    :  z \n\n in\n3 in a \n1 in c \n2 in d 
EXPECTED step2       :  z \n\n in\n3 in a \n1 in c \n2 in d 
```

The old comparator reverses the four-way tie exactly as predicted; the new one reproduces the expected text. The other four steps are identical under both — consistent with only the fully-tied step being affected.

### On stability, and why no explicit tie-break was added

The reviewer's alternative — an explicit original-index tie-break — is unnecessary here, and deliberately so. `MutableList.sortWith`'s stability is guaranteed on the **common `expect` declaration** (`commonMain/kotlin/collections/CollectionsH.kt`), not just per-platform:

> The sort is _stable_. It means that equal elements preserve their order relative to each other after sorting.

So stability is part of the contract every target must honour, and the native implementation above does honour it once the comparator is consistent. The old code wasn't losing stability — it was feeding the sort a comparator that never reported equality, so there were no "equal elements" for the stability guarantee to apply to. Fixing the comparator is the root-cause fix; an index tie-break would have been a second mechanism papering over the same defect.

### Scope / no-regression

- Both comparators fixed — `comparePatternFn`, and the `numberRegex` no-match branch of `compareFn` (same `if (sa < sb) -1 else 1` shape, same latent bug, not yet reached by a test).
- Grepped the whole tree for other instances of the pattern: the remaining `-1 else 1` hits are all direction constants (`if (forward) 1 else -1`), not comparators. `:vim`'s only `Comparator`/`sortWith` uses are the three in this function.
- No test expectation was changed. The expected strings encode vim's stable `:sort`, the JVM output was already correct, and the fix makes every target produce that same output — so the JVM/Android suites are unchanged at 626/626.
- `apiCheck` green with no `.api` diff, working tree clean afterwards; a comparator body is not ABI.

monkopedia-reviewer (approved):
Re-review of `de60bc9` (supersedes my prior CHANGES_REQUESTED on the earlier head).

## Recap of blocker
Prior review verified the 17 pure-rename file moves, the DebugEx*/SearchDebugTest deletion rationale, and 626/626 passing on JVM+Android — but blocked because `VimExCommandTest.ex_sort_pattern_no_r` failed deterministically on both `macosArm64` and `iosSimulatorArm64`, a genuine JVM-vs-Native divergence in `:sort` invisible on the Linux CI host that authored the original move.

## What changed
`vim/src/commonMain/kotlin/com/monkopedia/kodemirror/vim/VimExCommands.kt`: both ex-command comparators previously did `if (sa < sb) -1 else 1`, which answers "greater than" for equal keys — a broken comparator contract (compare(a,b) and compare(b,a) agree instead of having opposite signs). Both now use `sa.compareTo(sb)`.

## Verification performed this pass

**1. Mechanism claim — checked against actual stdlib source, not taken on faith.** Fetched JetBrains/kotlin's `libraries/stdlib/native-wasm/src/kotlin/collections/ArraySorting.kt`: the merge step is `if (comparator.compare(leftValue, rightValue) <= 0) take left else take right`. With the broken comparator, equal-key elements produce `compare(left,right) == 1 > 0`, so the merge always takes from the *right* run on ties — reversing tied groups. That is exactly the mechanism the CHANGELOG describes, and it is checkable, not hand-waving. JVM TimSort's documented behavior (stable, prefers left run, switches only when the right element strictly precedes it) is consistent with the broken comparator's `compare(right,left) == 1` never triggering a switch — ties stay in place, which matches the JVM output that was already correct. Both platform behaviors are independently explained by the same broken-comparator defect.

**2. Stability claim — the load-bearing check.** Verified directly against JetBrains/kotlin stdlib source (not the PR's assertion alone):
   - JVM actual (`MutableCollectionsJVM.kt`): `sortWith` delegates to `java.util.Collections.sort`, KDoc: "The sort is _stable_."
   - Native/wasm actual (`native-wasm/src/kotlin/collections/MutableCollections.kt`, covers macOS/iOS Native targets and wasmJs): same KDoc, "The sort is _stable_," backed by the merge-sort in ArraySorting.kt confirmed above to genuinely prefer the left run on ties (`<= 0`), i.e. actually stable, not just documented as such.
   - Android runs the JVM/Kotlin-stdlib-for-JVM actual, so it inherits the same guarantee.
   All targets this module ships to (JVM, Android, macOS/iOS Native, wasmJs) share the stable-sort guarantee on `MutableList.sortWith`. With a now-consistent comparator (returns 0 for equal keys), no explicit index tie-break is needed — the claim holds up under source-level verification, not just "the test passes."

**3. Both comparators fixed, no third instance.** Diffed `VimExCommands.kt` on `de60bc9` against `origin/main`: exactly two instances of the `if (x < y) -1 else 1` shape existed (`compareFn`'s `numberRegex`-null-match branch, and `comparePatternFn`'s key comparison) — both now `compareTo`. `compareFn`'s other branch (no `numberRegex`) already correctly returned 0 for equality and was untouched by this diff. Grepped the full file on the PR head for the broken shape: none remain.

**4. Test expectations not weakened.** `git diff origin/main de60bc9 --stat` with rename detection on every moved test file, including `VimExCommandTest.kt`, shows `0 insertions(+), 0 deletions(-)` — pure renames. `ex_sort_pattern_no_r`'s five expected strings are byte-identical to `origin/main`; the fix went into production code, not the test oracle.

**5. Real CI — the only proof for a native-only bug.** Watched `gh pr checks 211 --watch` to completion on `de60bc9`:
   - `check`: pass (10m29s)
   - `check-ios`: pass (15m54s)
   - `check-macos`: pass (15m34s)
   This is the first time these Apple suites have run the `:vim` module at all, and they're green — direct evidence the fix resolves the divergence, not just a plausible theory.

**6. Local re-verification (JVM/Android), counted from XML not exit code:**
   - `:vim:jvmTest`: 17 XML files, **626 tests, 0 failures, 0 errors**
   - `:vim:testDebugUnitTest` (Android): 17 XML files, **626 tests, 0 failures, 0 errors**
   - `:vim:apiCheck`: BUILD SUCCESSFUL, no `.api` diff

**7. Scope.** `git diff origin/main de60bc9 --name-only`: 24 files, all under `vim/` + `CHANGELOG.md` (13 insertions, 113 deletions). No overlap with #202's `convention-plugins/` work.

## Conclusion
This was a genuine, previously-invisible cross-platform correctness bug in shipped `:sort` behavior, caught only because this PR moved the suite to `commonTest` and ran it on Apple targets for the first time. The fix is correct, minimal, well-explained, backed by a verified (not assumed) stdlib stability guarantee, and now proven green on real native CI. Approving.

Issue #210, opened 2026-07-30
wasmJs: Compose-Resources Font() is ignored — Skiko paints a proportional fallback (confirmed on a real display)


monkopedia-coder:
## Summary

On wasmJs, a `FontFamily(Font(Res.font.X))` built from a Compose-Resources font is **not honoured** — Skiko renders a proportional fallback instead. The app requests the font, the asset downloads successfully, and the glyphs are still painted in something else.

Confirmed on a **real display** on 2026-07-30, so this is not the headless/Xvfb fallback artifact previously assumed.

## Evidence

Showcase `TestEditorPage` (`?test=true`), which applies `showcaseSetup`:

```kotlin
// samples/showcase/src/commonMain/.../ShowcaseSetup.kt:44-50
val showcaseSetup: Extension
    @Composable get() {
        val monoFont = FontFamily(Font(Res.font.JetBrainsMono_Regular))
        return showcaseBase + editorContentStyle.of(TextStyle(fontFamily = monoFont))
    }
```

Rendered with real Chrome 149 on a real GNOME/Wayland display at dpr 1.5, then measured ink extent per line against known glyph counts:

| line | glyphs | ink width | px/char |
|---|---|---|---|
| `// Fibonacci sequence` | 21 | 221px | 10.52 |
| `if (n <= 1) return n;` | 21 | **179px** | 8.52 |
| `return fibonacci(n - 1) + fibonacci(n - 2);` | 43 | 391px | 9.09 |
| `for (let i = 0; i < 10; i++) {` | 30 | 239px | 7.97 |

Two lines with **identical 21-glyph counts** render 221px vs 179px. A monospace font makes those necessarily equal. px/char spread across lines is 7.97–10.52, ratio **1.32**; monospace is ~1.00.

## Controls checked (so this isn't misattributed)

- The page really does request the font — `TestEditorPage` applies `showcaseSetup`, verified in source, not assumed.
- The asset really serves — `composeResources/.../font/JetBrainsMono_Regular.ttf` → **HTTP 200, 112,172 bytes**.
- Real display, real GPU, real Chrome, dpr 1.5. Not Xvfb, not headless CanvasKit, not software GL. Every environmental excuse previously used to explain this away is ruled out.
- Note the showcase `index.html` *also* pulls JetBrains Mono from Google Fonts and sets `canvas { font-family: 'JetBrains Mono', … }`. Even with that web font present, the canvas still paints a proportional face — so neither path is reaching the renderer.

[...]

PR #212, opened 2026-07-30
Run wasmJs tests on a headless browser and enable 22 more modules (#202)

### LINKED ISSUE #202: Unblock wasmJs tests for Compose-consuming modules (skiko — nodejs vs browser runner)

monkopedia-coder:
## Problem

`kodemirror.library` configured `wasmJs { nodejs() }`, and `wasmJsTests` was opt-in for five
Compose-free modules only. Everything else — 23 modules, including every `:lang-*` and
`:legacy-modes` (they all `implementation(project(":language"))`, and `:language` applies
Compose) — had no wasm test execution at all.

Both Node failure modes were re-confirmed on the **current** toolchain (Kotlin 2.4.10,
Compose MP 1.11.1, AGP 9.3.1, Gradle 9.6.1), not inherited from the old note:

- `:lang-json` (reaches Compose transitively, no Compose plugin, so no skiko asset):
  ```
  Error [ERR_MODULE_NOT_FOUND]: Cannot find module
    '.../packages/Kodemirror-lang-json-test/kotlin/skiko.mjs'
  ```
- `:view` (applies the Compose plugin, so `skiko.mjs` **is** present):
  ```
  failed to asynchronously prepare wasm: both async and sync fetching of the wasm failed
  Aborted(both async and sync fetching of the wasm failed)
  ```
  The shipped `skiko.mjs` is an emscripten **browser** build — its Node branches are compiled
  out as `if (false)`. So placing the assets where Node resolves them is not a fix; Node cannot
  work at all.

## Fix

1. `wasmJs { browser { testTask { useKarma { useChromeHeadless() } } } }` in the convention
   plugin (replacing `nodejs()`).
2. Modules that reach Compose only transitively never get skiko from the Compose Gradle
   plugin's `processSkikoRuntimeForKWasm` — it only serves modules that apply the plugin
   directly, so webpack fails with `Can't resolve './skiko.mjs'`. The convention plugin now
   unpacks `org.jetbrains.skiko:skiko-js-wasm-runtime` into those modules' wasm test resources
   itself, taking the version from the module's own resolved `wasmJsTestRuntimeClasspath` — so
   Compose-free modules add nothing, and there is no hardcoded skiko version to drift.
3. `wasmJsTests.set(true)` for every module that has `commonTest` sources.

## Result — counts read from `build/test-results/wasmJsBrowserTest/*.xml`, not from BUILD SUCCESSFUL

**27 modules, 1950 tests, 0 failures.**

| module | tests | | module | tests |
|---|---:|---|---|---:|
| vim | 626 | | lang-cpp | 30 |
| lezer-lr | 197 | | merge | 30 |
| state | 173 | | lang-php | 29 |
| view | 118 | | autocomplete | 27 |
| legacy-modes | 110 | | lang-sass | 25 |
| language | 90 | | lang-yaml | 20 |
| lezer-common | 59 | | lang-css | 16 |
| lang-markdown | 58 | | lang-xml | 13 |
| lang-go | 46 | | collab | 7 |
| lang-python | 41 | | lint | 7 |
| search | 38 | | lezer-highlight | 30 |
| lang-java | 33 | | lang-html | 32 |
| lang-javascript | 32 | | lang-rust | 32 |
| lang-json | 31 | | | |

The five pre-existing opt-ins still report exactly their previous 466
(`:state` 173, `:lezer-lr` 197, `:lezer-common` 59, `:lezer-highlight` 30, `:collab` 7).
1484 tests are newly executing on wasm. No test needed changing and nothing was skipped or
loosened.

I also verified the runner is not a silent pass: a deliberately failing test added to
`:lang-json` produced `Execution failed for task ':lang-json:wasmJsTest' > There were failing
tests`, then was removed.

## CI

`ci.yml` lists all 27 `wasmJsTest` tasks explicitly (so a module dropping out of the opt-in set
shows as a diff rather than a silent no-op) and bumps the Kotlin daemon heap to 4g — linking
that many wasm test binaries OOMs the 2g default. Karma uses the Chrome preinstalled on
`ubuntu-latest` (`karma-chrome-launcher` auto-detects, or honours `CHROME_BIN`); karma/webpack
themselves come from Kotlin's own npm tooling dir, so `kotlin-js-store/wasm/yarn.lock` is
unchanged.

## Also verified locally

- `apiCheck`, `jvmTest`, `spotlessCheck`, `ktlintCheck` green.
- `:samples:showcase:wasmJsBrowserDevelopmentExecutableDistribution` green (not a PR check).
- `:state:publishWasmJsPublicationToMavenLocal` still produces the same klib/pom/module/sources
  /javadoc set — `browser()` does not change the published artifacts.
- The wasm Node pin from #189 is untouched.

## Not addressed

Pixel/screenshot assertions stay off-limits on this runner: headless CanvasKit renders a
fallback font rather than the bundled Compose-Resources one (#210). Everything enabled here is
semantics/behaviour only.

Fixes #202


monkopedia-reviewer (approved):
## Review: #212 — Run wasmJs tests on a headless browser and enable 22 more modules

**Category:** Build/CI infrastructure (test runner + convention-plugin opt-ins) — no production
source touched; diff is entirely `.gradle.kts` files, `ci.yml`, `CHANGELOG.md`, and the
`KodemirrorLibraryExtension`/`kodemirror.library` convention-plugin Kotlin files.
**Tier:** 1 (mechanical, thoroughly self-verified, zero behavior change to shipped code).
**Verdict:** approve.

### The crux — do the 22 newly-enabled modules actually execute tests?

Verified independently, not inferred from green CI. Re-ran a representative sample locally
(`CHROME_BIN=/usr/bin/chromium`, since ubuntu's binary here is `chromium` not `google-chrome`;
karma-chrome-launcher picked it up fine) and read `build/test-results/wasmJsBrowserTest/*.xml`
directly:

| module | XML count | PR claim | match |
|---|---:|---:|---|
| `:collab` (pre-existing opt-in) | 7 | 7 | yes |
| `:state` (pre-existing opt-in) | 173 | 173 | yes |
| `:lezer-common` (pre-existing opt-in) | 59 | 59 | yes |
| `:lezer-highlight` (pre-existing opt-in) | 30 | 30 | yes |
| `:lezer-lr` (pre-existing opt-in) | 197 | 197 | yes |
| `:lang-json` (new, transitive-Compose) | 31 (25+6) | 31 | yes |
| `:view` (new, applies Compose directly) | 118 | 118 | yes |
| `:vim` (new, applies Compose, largest) | 626 | 626 | yes |
| `:legacy-modes` (new, transitive-Compose) | 110 | 110 | yes |

Pre-existing five sum to **466** exactly — no regression. Sampled 4 of the 22 new modules,
covering all three technical cases in this PR (Compose-free-transitive via `:language`,
Compose-applied-directly `:view`/`:vim`, and the largest new suite), all exact matches, all
real assertions (`0 failures`, non-trivial per-file counts, not a single `1-test-stub`).

Then cross-checked the **actual GitHub Actions log** for the `check` job
(run 30572374192, job 90972036495) rather than trusting BUILD SUCCESSFUL: all 27
`:<module>:wasmJsBrowserTest` / `:<module>:wasmJsTest` tasks appear individually with real
elapsed time between them (e.g. `:legacy-modes` ~34s, `:lezer-lr` ~26s) — none show as
`UP-TO-DATE` or a zero-cost no-op, consistent with genuine Karma/Chrome runs, not the #197
failure mode (disabled task, still exits 0).

`ci.yml`'s enumerated list is exactly the 27 modules that got
`kodemirrorLibrary { wasmJsTests.set(true) }` in this diff (5 unchanged pre-existing + 22 new,
verified 1:1 against the `build.gradle.kts` diffs) — nothing listed that's disabled, nothing
enabled but ungated.

### No suppression

The diff touches **zero test source files** — only build scripts, `ci.yml`, `CHANGELOG.md`, and
the convention-plugin Kotlin (comment/rename only, opt-in semantics unchanged: default `false`,
explicit per-module `.set(true)`). There is no `@Ignore`, no loosened assertion, nothing to
suppress a failure with, because nothing in the test bodies moved.

### Browser runner in real CI

`check-ios` and `check-macos` both passed (14m43s, 9m37s — unaffected, as expected, this PR
doesn't touch native). The `check` job itself passed at **16m58s**, up from the prior
baseline of ~11.5 min (last green run on `main`, 18:49:39→19:01:19) — a genuine, non-trivial
cost from running 22 additional real browser-based test tasks, worth noting for future CI
budget but not blocking.

### Other checks
- Node pin from #189 untouched — root `build.gradle.kts` still pins `NodeJsEnvSpec` /
  `WasmNodeJsEnvSpec` / `WasmNodeJsRootExtension.version` to `22.11.0`.
- No font-metric/pixel/screenshot assertions found in the newly-enabled modules' test sources
  (`grep -i "screenshot|roborazzi|pixel|glyph|measureText"` over `view/src`, `legacy-modes/src`,
  `vim/src` test dirs: no hits) — consistent with #210 (headless CanvasKit fallback font) not
  being a live risk here, and confirmed empirically since `:view` (118) and `:vim` (626) both
  pass 0 failures.
- `main` advanced to `1e190ee` (#211, `:vim` commonTest move) while this PR was in flight —
  **no rebase needed**: `git merge-base origin/main HEAD` **is** `1e190ee`, i.e. this branch's
  parent commit already **is** current `main`'s tip. #211 is already in this PR's ancestry, and
  the local `:vim:wasmJsTest` run above (626 tests) exercises the merged result of both changes
  together, passing clean.
- Showcase build and full local `apiCheck`/`ktlintCheck`/`spotlessCheck` were not independently
  re-run in this pass (the process hosting this review was restarted mid-run and lost that
  in-flight background output); real CI's `check` job already ran `apiCheck` as part of its
  command line and passed, which covers that gate. Showcase's own build isn't a PR check and
  wasn't touched by this diff beyond the (correct) `wasmJsBrowserTest`→disabled-task rename
  mirroring the convention plugin.

Merging.

Issue #200, opened 2026-07-26
Port the Compose UI test harness to runComposeUiTest so the 63 :view interaction tests run on all targets


monkopedia-coder:
Part of #196 (stage 3).

## Problem

`:view` has **63 Compose UI interaction tests** across 17 files. They run on JVM only — so no target other than JVM has any verification that clicking, typing, dragging, scrolling or focus actually work.

The test bodies are already platform-neutral. From `ClickTargetingTest.kt`:

```kotlin
import androidx.compose.ui.test.ExperimentalTestApi
import androidx.compose.ui.test.click
import androidx.compose.ui.test.onNodeWithTag
import androidx.compose.ui.test.performMouseInput

onNodeWithTag("KodeMirror").performMouseInput { click(Offset(50f, 15f)) }
```

That's the multiplatform Compose test API. The only JVM binding is in **one file**, `view/src/jvmTest/.../input/InputTestHelper.kt`:

```kotlin
import androidx.compose.ui.test.DesktopComposeUiTest
import androidx.compose.ui.test.runDesktopComposeUiTest
...
fun runEditorTest(..., block: DesktopComposeUiTest.(SessionHolder) -> Unit) =
    runDesktopComposeUiTest(width = width, height = height) { ... }
```

plus `import org.junit.Test` in the test files themselves.

## Proposed

1. Rewrite `runEditorTest` against the multiplatform harness — `runComposeUiTest` / `ComposeUiTest` (`@OptIn(ExperimentalTestApi::class)`) instead of the `Desktop*` variants.
2. Swap `org.junit.Test` → `kotlin.test.Test` across the 17 files.
3. Move `view/src/jvmTest/.../input/` → `commonTest`, and add the `compose.uiTest` dependency to `commonTest` (`:view` currently declares `compose.desktop.uiTestJUnit4` in `jvmTest`).
4. Leave the 16 roborazzi screenshot tests in `jvmTest` — roborazzi is JVM/Android only and has no multiplatform equivalent.

## Decided: fixed-size frame with pinned density

`runComposeUiTest` has no `width`/`height` parameter — the current harness relies on `runDesktopComposeUiTest(width = 800, height = 600)`, which sizes the *surface* in raw pixels, and the tests then click at raw pixel offsets (`click(Offset(50f, 15f))`).

Replace that with a fixed-size frame inside `setContent`:

```kotlin
runComposeUiTest {
    setContent {
[...]

PR #213, opened 2026-07-31
Deprecate ChangeSet.toChangeSetJSON()/changeSetFromJSON() (#187)

monkopedia-coder:
Implements the owner's decision on #187 (option 3): deprecate `ChangeSet.toChangeSetJSON()` itself rather than trying to migrate it off the deprecated `Text.toJSON()`.

## Why

`toChangeSetJSON()` was the odd one out of the legacy JSON cluster — an *undeprecated* public API calling the `@Deprecated` `Text.toJSON()` behind a `@Suppress("DEPRECATION")`, i.e. live API reaching into API users are told to stop using.

Migrating the call was not possible without breaking published API: `ChangeSetJsonPart.Inserted.text` is a `List<String>` of lines matching `Text.toJSON()`, while the advertised replacement `Text.toString()` returns a single `String`. Changing the field would break both the signature and the serialized shape; adding a new line accessor would grow the public API just to delete a suppression. Deprecating the caller makes the contradiction dissolve — the suppression is fine once the caller is itself legacy.

**This is a deprecation pass, not a removal.** Nothing is removed, no behaviour changes, both functions work exactly as before.

## What changed

- `ChangeSet.toChangeSetJSON()` — `@Deprecated` (default `WARNING` level), no `ReplaceWith`.
- `ChangeSet.changeSetFromJSON()` — **yes, it is the symmetric counterpart** (`List<ChangeSetJsonPart> ↔ ChangeSet`, and the round-trip is what `ChangeSetTest.canBeSerializedToJSON` asserts), so it is deprecated in the same pass. Leaving it live would produce the exact inversion of the original problem: an undeprecated deserializer for a format nothing is supposed to produce any more.
- KDoc on both cross-references the coordinated-removal cluster: `EditorState.toJSON`/`fromJSON` (`State.kt:334/487`), `SelectionRange.toJSON`/`fromJSON` and `EditorSelection.toJSON`/`fromJSON` (`Selection.kt:164/284/298`), `Text.toJSON` (`Text.kt:258`).
- Scoped `@Suppress("DEPRECATION")` on the one internal call site, `ChangeSetTest.canBeSerializedToJSON` — the only caller of either function anywhere in the tree. It is itself part of the legacy JSON path (it exists to test exactly this round-trip), so suppressing is correct; the test keeps running unchanged.
- CHANGELOG under `[Unreleased]`, new `### Deprecated` section, referencing #187.

### No `ReplaceWith` — deliberately

[...]

PR #214, opened 2026-07-31
Move 212 non-Compose jvmTest methods to commonTest (#199)

### LINKED ISSUE #199: Move remaining 341 non-Compose jvmTest methods to commonTest (lsp-client, commands, autocomplete, collab, search, lint, view)
Part of #196 (stage 2). Sibling of #198 — same move, remaining modules.

## Problem

After vim's 629, **341 more test methods across 35 files** sit in `jvmTest` with no Compose dependency and no reason to be JVM-only:

| module | @Test | files | notes |
|---|---|---|---|
| lsp-client | 166 | 12 | |
| commands | 94 | 11 | |
| autocomplete | 45 | 7 | logic only — the popup UI is untested everywhere |
| view | 14 | 2 | the non-Compose stragglers; **1 file imports `org.junit`** |
| collab | 10 | 1 | |
| search | 8 | 1 | |
| lint | 4 | 1 | |

All but one file already use `kotlin.test`, so they move as-is.

## Proposed

Move each module's non-Compose `jvmTest` files to `commonTest`, one module per PR so failures stay attributable.

Two things to handle rather than assume:
1. **The one `org.junit` file in `:view`** — convert to `kotlin.test` (`@Test`, `assertEquals`, `assertFailsWith`) or leave it in `jvmTest` if it genuinely needs JUnit. Say which and why.
2. **Do not move Compose-touching files.** The 34 files importing `androidx.compose.ui.test`, `runDesktopComposeUiTest` or `captureRoboImage` stay put — those are #200's problem.

Per module, verify `jvmTest` stays green and the Apple + Android targets now execute the moved tests. wasm depends on whether the module applies Compose (`:lsp-client`, `:commands`, `:autocomplete`, `:search`, `:lint` do; `:collab` does not).

## Why

Completes the cheap half of #196. Combined with #198, it takes 970 test methods from one-platform to four- or five-platform execution.

monkopedia-coder:
Sibling of #198: moves the remaining non-Compose `jvmTest` suites to `commonTest` so they
execute on Android, wasmJs and the Apple targets as well as the JVM. One commit per module
so a failure stays attributable.

## What moved

| module | files | @Test | note |
|---|---|---|---|
| `:commands` | 11 | 94 | `jvmTest` deps folded into a new `commonTest` block; `wasmJsTests` turned on |
| `:autocomplete` | 7 | 45 | logic only |
| `:lsp-client` | 3 | 37 | the three files with no JVM-only coupling; `wasmJsTests` turned on |
| `:view` | 2 | 14 | `ClipboardCommandsTest`, `TabExpansionTest` |
| `:collab` | 1 | 10 | |
| `:search` | 1 | 8 | |
| `:lint` | 1 | 4 | |
| **total** | **25** | **212** | |

Every module's `jvmTest.dependencies` block folded into `commonTest.dependencies` verbatim.
`ci.yml` enumerates the wasmJs opt-in set explicitly, so `:commands:wasmJsTest` and
`:lsp-client:wasmJsTest` are added there.

## The `org.junit` file

`:view`'s `ClipboardCommandsTest` was the one file importing JUnit, and that single
`org.junit.Test` import was all it used JUnit for — no `@Rule`, no `assertThrows`, no
`Assert.*`. Converted to `kotlin.test.Test` and moved.

## A real platform divergence, surfaced by the move

Two of that file's assertions hard-coded the JVM's behaviour and failed on the headless
browser. `writeClipboard()` falls back to Compose's `ClipboardManager` only when
`platformWriteClipboard()` declines the write; the wasmJs `actual` writes
`navigator.clipboard` itself and returns `true`, so the Compose fallback is deliberately
skipped and the manager is never touched — `assertEquals("Hello", cm.stored?.text)` could
not hold there.

This is intended behaviour, not a bug: the browser write has to happen synchronously inside
the originating key gesture. Rather than dropping the assertion, the test now probes
`platformWriteClipboard` to derive which path this platform takes and asserts the negative
case too (`assertNull(cm.stored)` where the platform handled the write). That is strictly
stronger than what it checked before.

`RecompositionTest` and `SessionSwapScrollTest`, listed as non-Compose candidates, both
drive `runDesktopComposeUiTest` despite their names. They stay in `jvmTest` for #200, along
with `view/src/jvmTest/.../input/` and `.../screenshots/`, which were not touched.

## Left behind: nine of `:lsp-client`'s twelve files (129 test methods)

The issue assumed all 35 files "move as-is". These nine do not, for two reasons:

1. **`java.lang.reflect.Proxy`.** Every `LanguageServer` stub in the module is a dynamic
   proxy — the shared `TestLanguageServer` fixture plus four per-file inline throwing stubs.
   `com.monkopedia.lsp.LanguageServer` is a ~78-method interface with no default
   implementations, which is exactly why a proxy was used for a fake that scripts one or two
   methods.
2. **`kotlinx.coroutines.runBlocking`** in four of them, which does not exist on wasmJs/JS,
   and `kotlinx-coroutines-test` is not a dependency anywhere in this tree today.

Porting them needs a hand-written (or generated) common `LanguageServer` implementation with
a per-method default for the ~10 methods whose return types are non-nullable, a new `runTest`
dependency, and a rewrite of five stub factories. That is a change of a different shape and
size from a file move, so it is not bundled here. The three files carrying no such coupling —
`DocumentSyncTest`, `ServerDiagnosticsTest`, `ServerHoverTest` — did move.

Nothing else in the batch used a JVM-only API: no other `java.*` import, no `System.*`, no
file I/O, no threads.

## Verification

Executed-test counts read from `build/test-results/<task>/*.xml`, never from
BUILD SUCCESSFUL (the #197 failure mode).

| module | `jvmTest` | `testDebugUnitTest` | `wasmJsBrowserTest` |
|---|---|---|---|
| `:autocomplete` | 72 | 72 | 72 |
| `:collab` | 17 | 17 | 17 |
| `:commands` | 94 | 94 | 94 |
| `:lint` | 11 | 11 | 11 |
| `:lsp-client` | 166 | 37 | 37 |
| `:search` | 46 | 46 | 46 |
| `:view` | 221 | 132 | 132 |

0 failures, 0 errors, 0 skipped in every cell. `:lsp-client`'s jvm figure stays at 166
because the nine unmoved files still run there; the 37 that moved now run on three targets
instead of one. `:view`'s jvm figure includes the Compose/Roborazzi tests that stay in
`jvmTest`.

Repo-wide: `apiCheck jvmTest testDebugUnitTest` green (no `.api` diff), 4951 executed tests
across all modules with 0 failures. Apple targets can't run on Linux — relying on
`check-macos` / `check-ios` here.

Fixes #199


monkopedia-reviewer (approved):
## Verdict: approve

**The 341-vs-212 headline gap is fully resolved and non-benign in the good sense — real API constraints, not silent skips.**

- `:view`'s 34 remaining `jvmTest` files: 32 are `input/`+`screenshots/` (#200's territory, untouched) plus `RecompositionTest`/`SessionSwapScrollTest`, which I confirmed both call `androidx.compose.ui.test.runDesktopComposeUiTest` — genuinely Compose-UI tests despite non-Compose-sounding names.
- `:lsp-client`'s 9 remaining files (129 of 166 methods — this is the other half of the 341-212=129 gap): I grepped all 9 and every one either constructs a `java.lang.reflect.Proxy` `LanguageServer` stub directly, or depends on the shared `TestLanguageServer` fixture (also `Proxy`-based) and/or `kotlinx.coroutines.runBlocking` (no wasmJs/JS equivalent, and `kotlinx-coroutines-test` isn't a dependency anywhere in the tree). Confirmed genuine, not a shortcut — PR body and CHANGELOG both state this explicitly rather than leaving it silent.

**Moves are pure renames.** `git diff -M origin/main...HEAD` shows 0 changed lines on every moved file except `ClipboardCommandsTest.kt`, whose diff I read in full: JUnit→`kotlin.test.Test`, plus a platform-derived assertion (`platformWriteClipboard` probe) replacing a JVM-hardcoded one. Verified this isn't a weakened assertion — it's strictly stronger (asserts both the positive and negative case) and is honestly documented as a "real platform divergence, not a bug" in both the PR body and CHANGELOG.

**Tests genuinely execute — checked `build/test-results/*/​*.xml`, not exit codes** (the #197 failure mode):
- `jvmTest` repo-wide: **2380 tests, 0 failures** — exactly matches the `1e45dfc` baseline, no regression.
- Per-module `jvmTest`/`testDebugUnitTest`, all 0 failures: `:commands` 94/94, `:autocomplete` 72/72, `:lsp-client` 166/37, `:view` 221/132, `:collab` 17/17, `:search` 46/46, `:lint` 11/11 — matches the PR table exactly.
- Newly wasmJs-enabled `:commands` and `:lsp-client`: ran `wasmJsTest` locally with `CHROME_BIN=/usr/bin/chromium` — real `ChromeHeadless150.0.0.0` execution in the XML (not the karma "discovered 0 tests" trap), 94 and 37 tests respectively, 0 failures, individual test-method names present.
- Also ran `:view:wasmJsTest` to check the clipboard platform-divergence claim directly: `ClipboardCommandsTest` passes with 8/8 on the headless browser, confirming the platform-derived assertion actually resolves the divergence rather than papering over it.
- No `@Ignore`/`@Disabled`/`@Suppress`/relaxed assertions anywhere in the diff (checked).

**Other:**
- `apiCheck jvmTest testDebugUnitTest` ran clean locally (`BUILD SUCCESSFUL`, no `.api` diff) — pure test-source moves, no ABI impact.
- `ci.yml`: `:commands:wasmJsTest` and `:lsp-client:wasmJsTest` added to the enumerated opt-in list, comment's wasm-binary count bumped 27→29 to match. Consistent in both directions.
- `main` has advanced to `1e45dfc` (#213, an unrelated `ChangeSet` deprecation) since this branch's merge-base (`3e2ecf7`) — `mergeStateStatus: CLEAN`, no rebase needed, no territorial overlap.
- Real GitHub CI: `check`, `check-macos`, `check-ios` all **pass** (watched to completion, not inferred).

Category: test-infrastructure / build-config (mechanical source-set move + CI wiring, one non-mechanical assertion fix disclosed and verified)
Tier: 1

Issue #215, opened 2026-07-31
Android has no coverage for the Compose UI interaction tests (Robolectric is not trustworthy for them)


monkopedia-coder:
Follow-up from #200.

#200 moved the 59 Compose UI interaction tests in `view/src/.../input/` from `jvmTest` to
`commonTest`, so they now run on JVM, wasmJs and the Apple targets. They compile for Android
but are excluded from `:view:testDebugUnitTest` (see the filter and comment in
`view/build.gradle.kts`).

## Why they are excluded

Android's `runComposeUiTest` launches a real `ComponentActivity` via `ActivityScenario`, so it
needs an Android framework. A plain local unit-test JVM has none — it fails immediately with

```
java.lang.NullPointerException: Cannot invoke "String.toLowerCase(java.util.Locale)"
  because "android.os.Build.FINGERPRINT" is null
  at androidx.compose.ui.test.RobolectricIdlingStrategy_androidKt.getHasRobolectricFingerprint
```

Robolectric was tried and *does* host them. Setup that got there:

- `implementation(libs.robolectric)` on `androidUnitTest`
- `debugImplementation("androidx.compose.ui:ui-test-manifest")` (the `ComponentActivity`
  manifest entry; a test configuration is not enough, unit tests only see the library's own
  merged manifest)
- `android { testOptions { unitTests.isIncludeAndroidResources = true } }`
- `androidUnitTest/resources/robolectric.properties` with `qualifiers=w1000dp-h800dp-mdpi`
  (Compose refuses mouse gestures outside the Compose root, so the simulated screen has to be
  at least as large as the 800x600 test frame) and `sdk=34`
- an `expect abstract class` base in `commonTest` whose Android actual carries
  `@RunWith(RobolectricTestRunner::class)` — `@RunWith` is `@Inherited`, and it cannot be
  written in `commonTest`

With that, all 191 `:view` Android unit tests execute. But neither Robolectric graphics mode
gives a trustworthy result:

| `graphicsMode` | result |
| --- | --- |
| `LEGACY` (default) | 7 failures. All text measures to ~zero width: clicking 30px into `Hello world` resolves to offset 11 (the last character), and dragging from x=20 to x=200 within one line yields an empty selection. Every coordinate assertion is meaningless; the focus/keyboard suites only pass because the editor auto-focuses at composition and nothing ever disturbs it. |
[...]

PR #216, opened 2026-07-31
Port the Compose UI test harness to runComposeUiTest so the :view interaction tests run on all targets (#200)

### LINKED ISSUE #200: Port the Compose UI test harness to runComposeUiTest so the 63 :view interaction tests run on all targets

monkopedia-coder:
Fixes #200. Stage 3 of #196, after #198 and #199.

Moves the 59 Compose UI interaction tests in `view/src/.../input/` from `jvmTest` to
`commonTest` and rewrites `runEditorTest` against the multiplatform `runComposeUiTest`.
They now run on wasmJs and (via CI) the Apple targets, not just the JVM.

## The viewport

`runComposeUiTest` has no `width`/`height`, and these tests click at raw pixel offsets
(`click(Offset(50f, 15f))`). The frame is now established inside `setContent`:

```kotlin
CompositionLocalProvider(LocalDensity provides Density(density = 1f, fontScale = 1f)) {
    Box(Modifier.requiredSize(width.dp, height.dp)) { KodeMirror(session = session) }
}
```

Pinning the density is the half that makes it portable — `requiredSize` takes dp and the
ambient density differs per platform, so without the pin `800.dp` is a different pixel count
per target and every existing offset means something else. At `Density(1f, 1f)` dp == px and
the offsets carry over untouched. `width`/`height` stay as `runEditorTest` parameters, since
the scroll tests use deliberately small viewports.

`kotlin.assert` is JVM-only, so the 44 assertions became `kotlin.test.assertTrue`. Roborazzi's
16 screenshot tests stay in `jvmTest`, as do `RecompositionTest` and `SessionSwapScrollTest`.

## Two things that had to be fixed before wasm meant anything

**The wasm run was vacuous.** `runComposeUiTest` returns `Unit` on the JVM but a promise on
wasmJs. `runEditorTest` dropped it, so the browser runner reported all 59 green in ~1ms each
without executing an assertion — #197's failure mode in a new costume. Caught with a negative
control: with deliberately broken assertions the suite still reported green. `runEditorTest`
and the seven block-bodied tests now propagate the result, and the negative control fails as
it should.

**A real cross-platform failure, once they ran.**
`MixedContentScrollTest.longLineScrollsAmongShortLines` failed on wasm (`before=32 after=110`,
needing +100). The wheel-delta-to-pixel factor is platform-specific: a 600-unit horizontal
scroll moves ~3200px on desktop but ~550px in a browser. The editor scrolled correctly on
both. The scroll now uses a delta larger than the line's scrollable range so it saturates
everywhere — which is what #67 is actually about. The assertion is unchanged.

## Font dependence

The font-metric-dependent set (`ClickTargetingTest`, `HitTestVerticalAnchorTest`, the
character-level `DragSelectionTest` cases) passes on wasm unmodified, so no bundled test font
was needed. Their coordinates are coarse relative to glyph advances — vertical hit-tests key
off line height, and the one column assertion is `col > 10` for a click 360px into a line — so
the wasm fallback font of #210 does not move them. Nothing was loosened to get there; the
negative control confirms they are really being evaluated.

## Android is deliberately left out — see #215

The tests still *compile* for Android, so an Android-only compile break is still caught, but
they are excluded from `:view:testDebugUnitTest` (filter + reasoning in `view/build.gradle.kts`).

Android's `runComposeUiTest` launches a real `ComponentActivity` through `ActivityScenario`, so
it needs an Android framework; a local unit-test JVM has none and dies on a null
`android.os.Build.FINGERPRINT`. Robolectric was tried and *does* host them — all 191 execute
with `ui-test-manifest` on `debugImplementation`, `unitTests.isIncludeAndroidResources`, a
`w1000dp-h800dp-mdpi` qualifier, and an `expect`/`actual` base class carrying the Android-only
`@RunWith(RobolectricTestRunner::class)`. But neither graphics mode is trustworthy:

| `graphicsMode` | result |
| --- | --- |
| `LEGACY` (default) | 7 failures. All text measures to ~zero width — a click 30px into `Hello world` resolves to offset 11, a drag from x=20 to x=200 within one line selects nothing. Every coordinate assertion is meaningless, and the focus/keyboard suites only pass because the editor auto-focuses at composition and nothing disturbs it. |
| `NATIVE` | 12 failures. Metrics are correct (the 7 above pass), but a pointer **press** clears Compose focus: `hasFocus` is true after composition, false immediately after `press()`, and never recovers. Breaks `FocusManagementTest` and `KeyboardHandlingTest`. |

Graphics mode cannot plausibly change focus semantics, so the `NATIVE` divergence reads as an
artifact of the environment rather than a finding about the editor — and picking a mode and
suppressing whichever half failed would have been the wrong move. #215 carries the full setup
and asks for confirmation on a real device before adding instrumented coverage.

## Verification

Executed-test counts from `build/test-results/<task>/*.xml`, never from BUILD SUCCESSFUL:

| task | before | after |
| --- | --- | --- |
| `:view:jvmTest` | 221 | **221, 0 failures** |
| `:view:wasmJsTest` | 132 | **191, 0 failures** |
| `:view:testDebugUnitTest` | 132 | **132, 0 failures** |
| repo-wide `jvmTest` | 2380 | **2380, 0 failures** |
| repo-wide `testDebugUnitTest` | — | **2162, 0 failures** |

`apiCheck` and `:view:verifyRoborazziJvm` green. Apple targets are CI's (`check-macos` /
`check-ios`) — those tasks are host-gated and report SKIPPED on this Linux box; `compose.ui:ui-test`
publishes `macosArm64`, `iosArm64` and `iosSimulatorArm64` variants, so the dependency resolves.


monkopedia-reviewer (changes requested):
## Re-review of 7bb684b — request changes

**CI is not green.** `check` and `check-ios` passed on `7bb684b`, but **`check-macos` failed**
(run https://github.com/Monkopedia/kodemirror/actions/runs/30663647713/job/91265445510,
11m44s, "191 tests completed, 2 failed"):

```
com.monkopedia.kodemirror.view.input.HorizontalScrollTest.wheelScrollAdvancesClickOffset[macosArm64] FAILED
    kotlin.AssertionError: Expected click offset to advance after horizontal scroll, but before=26 after=26

com.monkopedia.kodemirror.view.input.MixedContentScrollTest.longLineScrollsAmongShortLines[macosArm64] FAILED
    kotlin.AssertionError: Expected click offset to advance well past the viewport after horizontal
    scroll on a long line among short lines, but before=26 after=26
```

Both are the exact two tests this commit's `wheelScrollRightToEnd` helper was written to fix, and
`before == after` exactly — the helper's own loop logic (bail out once `horizontalScrollPx` stops
changing, having made ≥1 attempt) means it made an attempt, observed the scroll position not move
at all between the first and second read, and gave up after one step. So on macosArm64 specifically,
`performMouseInput { moveTo(...); scroll(...) }` is producing **no** measurable horizontal scroll —
not "not enough," none. iOS runs the identical shared-`nativeMain`/commonTest code and passed, so
this isn't a nativeMain-wide issue; it's macOS-native-target-specific (or a Skiko/Compose-test-harness
quirk exercised only by this runner). Either way: the commit's stated purpose — fixing the three
Apple-target failures the ported suite found — is only 1/3 accomplished for macOS; the two scroll
tests it explicitly targets are still red, just with a legible message now instead of `at null:-1`.
That legible message is itself confirmation the `testLogging` fix (item 3 below) works as intended,
which is the one piece of good news in this run.

Not requesting changes on code style grounds here — the fix needs to actually make these two tests
pass on macOS before this can merge. Recommend instrumenting `wheelScrollRightToEnd` (or a scratch
test) to log `horizontalScrollPx()` immediately after each `scroll()` call on a macOS CI run, to tell
apart "the gesture never reaches the scrollable" vs "the assertion after click is wrong" vs "an
actual product regression in horizontal wheel handling on macosArm64."

---

### Priority 1 — ctrlX/modX tautology risk: real bug, correctly not enshrined as intentional

`platformOsName()` returning `"Mac"` unconditionally is confirmed **not** a deliberate design
decision — it's an accidental stub from the original "Add macOS and iOS native targets
(experimental)" commit (`ab8bdcf`), whose message says outright: "Native platform implementations
... provide stub actuals for clipboard, focus, and key handling. These will need platform-specific
implementations for production use." `view/src/nativeMain/.../Platform.native.kt` is one
`internal actual` file shared by `macosArm64`, `iosArm64`, and `iosSimulatorArm64` (there is no
`linuxNative`/other native target in this project, so "every Kotlin/Native target" is accurate
today, but only because the only native targets are Apple ones — the hardcoding was never a
per-OS decision). So `modX_cutsSelectedText` deriving its expected modifier from the very same
`currentOs` the product reads is model-perfect if you assume that's *intentional* Mac-parity
behavior on iOS, and silently non-diagnostic if it's a bug (an iOS build should arguably report
something other than "Mac" and possibly get a different keymap, or not — nobody has decided).
**This is worth a follow-up issue** (distinct from #200): "decide whether iOS should report as Mac
for keybinding purposes, or implement real per-OS detection in nativeMain," so a future contributor
doesn't mistake the test for evidence it was designed this way.

That said, the two-branch assertion itself is real, not just "doesn't throw": `assertDoc` and
`assertSelectionNotEmpty` compare actual state (`session.state.doc.toString()` /
`!sel.empty`), and the test asserts both that the unbound modifier leaves doc AND selection
untouched, and that the bound one actually cuts (`assertDoc(" World")`). Confirmed passing on
`check-ios` and (per the log above) also passing on `check-macos` — the 2 macOS failures are
scroll-only. On the CM6-parity claim: real CM6's `standardKeymap` (checked
`@codemirror/commands` 6.10.3 in the yarn cache, `node_modules/@codemirror/commands/dist/index.js:1712`)
does **not** bind Ctrl-x/Cmd-x at all — cut/copy/paste in real CM6 ride the browser's native
`cut`/`copy`/`paste` DOM events, not a keymap entry. So "exactly as it does not in CodeMirror 6"
overstates the mechanism (there's no CM6 keymap line to diff against), though the *practical* Mac
convention (Cmd-X, not Ctrl-X) it's citing is correct — worth a lighter phrasing next time, not a
blocker.

### Priority 2 — scroll helper: not self-defeating, and its assertions are meaningful

- **"At least one attempt" claim verified by inspection**: `previous` starts at `Int.MIN_VALUE`,
  so iteration 1 always scrolls once regardless of the actual starting position. If nothing moves,
  iteration 2's `current == previous` triggers `return` after exactly one scroll call — the helper
  can't be fooled into reporting "converged" without ever touching the wheel.
- **Bounded**: `MAX_WHEEL_SCROLL_STEPS = 60`, no risk of an unbounded CI hang.
- **Assertions are about the result, not "scrolling happened"**: both tests still assert
  `after > before + 50` / `+ 100` on the resulting click offset, a real semantic check, not merely
  that `horizontalScrollPx()` changed. This is exactly what caught the macOS regression above
  instead of masking it — the helper design is sound; the target behavior underneath it is what's
  broken.

### Priority 3 — convention-plugin `testLogging` change: justified, correctly scoped, verified working

`events("failed")` limits it to failures only (confirmed no other `testLogging`/`events(` block
exists elsewhere in the build to conflict with). It is repo-wide rather than scoped to `:view`,
which is broader than #200's nominal scope, but it's low-risk (failure-only logging, doesn't touch
test behavior) and it just proved its worth: this run gave a legible
`before=26 after=26`/`AssertionError` instead of Kotlin/Native's `at null:-1`. I'd be fine letting
it ride with this PR rather than splitting it out — but flagging that it's a repo-wide change
smuggled into a `:view`-scoped PR title, for the record.

### Priority 4 — standing checks
- Executed-test counts (from log, not BUILD SUCCESS/FAILURE): macOS **191 tests, 2 failed** (same
  191 as `d73d426`'s baseline, expected since no tests were added/removed, only one renamed).
  `check`/`check-ios` green — will get their own counts once this is re-run after a real fix;
  not re-deriving them now since the PR isn't mergeable regardless.
- No `@Ignore`, no weakened/removed assertions found in `d73d426..7bb684b`: `KeyboardHandlingTest`
  gains a second branch (stronger, not weaker); `HorizontalScrollTest`/`MixedContentScrollTest`
  swap a fixed scroll for a helper with an equivalent-or-stronger post-condition. Test counts
  unchanged (one rename, no add/remove).
- Density/frame pin (`requiredSize` + `Density(1f, 1f)`) untouched by this commit — verified
  already in place from `d73d426`.
- `apiCheck` not blocking here (this commit touches no public API surface — test-only + build
  script).

## Verdict: request_changes

The two Priority-1/2 investigations came back clean (real bug correctly surfaced, helper logic
sound, CM6-parity phrasing slightly loose but immaterial). The blocker is simply that **the PR's
own goal — green Apple CI — is not met**: `check-macos` still fails on the exact two tests this
commit targets, with `before=26 after=26`, a genuine unresolved regression/limitation on
macosArm64's wheel-scroll simulation. Please diagnose why `performMouseInput { scroll(...) }`
isn't moving `horizontalScrollPx()` at all on macOS native (vs. passing on iOS with the same
commonTest code) and get `check-macos` genuinely green before merge. Also worth filing, separately:
a follow-up issue on `platformOsName()` returning `"Mac"` unconditionally for every native target
being an inherited stub rather than a decision (referenced above, not blocking on its own).

monkopedia-reviewer (approved):
Re-review at 12914c0 (supersedes my CHANGES_REQUESTED at 7bb684b).

**Priority 1 — upstream ScrollConfig claim: verified true, word for word.**
Pulled the actual source from JetBrains/compose-multiplatform-core (jb-main, matching this repo's compose-plugin 1.11.1):
- `compose/foundation/foundation/src/macosMain/.../MacosScrollable.macos.kt` — `MacOsScrollConfig.calculateMouseWheelScroll` really does `val e = event.appkitEventOrNull ?: return Offset.Zero`, then reads `hasPreciseScrollingDeltas`/`scrollingDeltaX`/`scrollingDeltaY` off it. Exact match to the code comment.
- `compose/ui/ui/src/macosMain/.../Events.macos.kt` — `appkitEventOrNull`'s own doc comment states it is null "in tests, for example, or when Compose sends a synthetic event." That's upstream confirming the harness gap directly, not just this PR's inference.
- `compose/foundation/foundation/src/iosMain/.../UikitScrollable.ios.kt` — `UiKitScrollConfig.calculateMouseWheelScroll` derives the delta from `event.changes` (`panOffset`/`scrollDelta`), never touching a native event. Confirms the macOS/iOS asymmetry is real and explains why iOS runs these tests and macOS can't.

Claim holds entirely. This is a harness limitation, not a product bug — request_changes would have been wrong here.

**Priority 2 — exclusion scope: correct and narrow.**
`view/build.gradle.kts`'s `tasks.withType<AbstractTestTask>().matching { it.name == "macosArm64Test" }` excludes exactly `HorizontalScrollTest.wheelScrollAdvancesClickOffset` and `MixedContentScrollTest.longLineScrollsAmongShortLines` — two `excludeTestsMatching` calls by fully-qualified method name, nothing broader. `macosArm64` is the only macOS native target the convention plugin configures (no `macosX64`), so there's no sibling target silently exempted. iOS/Android/wasm/JVM are untouched by this filter. Grepped every other scroll test in the ported files (`HorizontalScrollIntoViewTest`, `ScrollClickTest`, `ScrollIntoViewTest`) — confirmed none use `performMouseInput { scroll(...) }`; they all drive scrolling via cursor motion / `scrollIntoView` / semantics, matching the author's claim.

**Priority 3 — cleanliness and standing checks.**
- No `TEMP`/`DIAG` anywhere in the tree (`git grep` clean); history is exactly `12914c0/7bb684b/d73d426/081ea1a/66d426e` with the diagnostic commit gone.
- No `@Ignore` in the moved tests. Diffed every line removed across the full `66d426e..12914c0` range — the two `assertTrue`/`assertDoc` removals I found are pure reflow (identical assertion, same message, moved to accommodate the `pressX`/`wheelScrollRightToEnd` helpers) or genuinely *strengthened* (the Ctrl-X test now asserts the unbound-modifier-is-inert case in addition to the original cut assertion). Nothing weakened.
- `@Test` counts, file by file, old (`jvmTest`) vs. new (`commonTest`): all 15 files match 1:1, summing to 59 — matches the CHANGELOG's "59 Compose UI interaction tests" exactly.
- Ran locally (`JAVA_HOME=java-21-openjdk ANDROID_HOME=~/Android/Sdk CHROME_BIN=/usr/bin/chromium ./gradlew :view:jvmTest :view:testDebugUnitTest :view:wasmJsBrowserTest :view:apiCheck`) in a worktree at 12914c0 — BUILD SUCCESSFUL. Per-target counts from `build/test-results/*/*.xml` (never BUILD SUCCESSFUL): `jvmTest` 221/0 fail (baseline unchanged), `testDebugUnitTest` 132/0 fail (baseline unchanged — Android exclusion holds), `wasmJsBrowserTest` 191/0 fail (up from 132 by exactly 59). Matches the reviewer brief's expected deltas.
- Font-dependent tests: unaffected by this PR's changes, #210 already tracked separately.
- CHANGELOG's CM6 parity wording now correctly states CM6 binds neither Ctrl-x nor Cmd-x (cut rides the DOM `cut` event) — verified this reads accurately in the diff.
- `main` is `ac9f809`, PR reports `mergeable: MERGEABLE` against it — no rebase needed.
- `#217` filed for the `platformOsName()` "Mac"-on-all-native finding — correctly not a blocker here.

**CI on 12914c0:** `gh pr checks 216 --watch` run to completion — `check` pass (15m5s), `check-macos` pass (13m49s), `check-ios` pass (17m20s). All three required checks green, watched to actual completion (not a local re-run).

Approving.

PR #219, opened 2026-07-31
Upload the Apple CI jobs' test XML so their executed-test counts can be verified (#218)

### LINKED ISSUE #218: Apple CI jobs publish no test XML, so their executed-test counts cannot be verified
`CLAUDE.md` requires executed-test counts to be read from `build/test-results/<task>/*.xml`
rather than from BUILD SUCCESSFUL — #197 is the standing reminder of why: a task that runs
zero tests also exits 0.

`ci-apple.yml` is the one place that standard cannot be met. `macosArm64Test` and
`iosSimulatorArm64Test` run only on macOS runners, so they cannot be reproduced on a Linux
development box, and the jobs upload nothing. Gradle prints a test count in its console
summary only when a test *fails*; a green Apple run prints no number at all. The XML exists
on the runner and is thrown away when the job ends.

#200 hit this directly: two `:view` tests are scoped off `macosArm64Test` because the test
harness cannot synthesize a wheel event there, and the resulting green run had to be taken on
trust — the exclusion pattern was validated by pointing the same filter at `jvmTest` (221 →
219) rather than by reading the macOS count.

Upload `**/build/test-results/**/*.xml` as an artifact from both Apple jobs, with
`if: always()` so a failing run still publishes it.

monkopedia-coder:
Fixes #218. Follow-up from #200/#216.

`macosArm64Test` and `iosSimulatorArm64Test` run only on macOS runners, so they cannot be
reproduced on a Linux development box. Gradle prints a test count in its console summary only
when a test fails, so a **green** Apple run produced no executed-test number at all — the one
thing `CLAUDE.md` says never to take on trust after #197, where five CI tasks were disabled
no-ops reporting success.

#200 hit this directly. Two `:view` tests are scoped off `macosArm64Test` there, because the
Compose test harness cannot synthesize a wheel event on macOS (`MacOsScrollConfig` reads the
delta off the underlying AppKit `NSEvent`, which a synthesized pointer event does not carry).
The green macOS run that followed could not be checked against a count; the exclusion pattern
had to be validated indirectly, by pointing the same filter at `jvmTest` and confirming it
removed exactly two methods (221 → 219).

Both jobs now upload `**/build/test-results/**/*.xml`, with `if: always()` so a failing run
still publishes it and `if-no-files-found: error` so the step cannot silently become a no-op —
which would reintroduce the same blind spot one level up.

No source or build-logic change; the workflow is the only file touched besides the CHANGELOG.


monkopedia-reviewer (approved):
Verified this end to end rather than reviewing the YAML in isolation.

**Artifacts exist and are non-trivial.** `gh api repos/Monkopedia/kodemirror/actions/runs/30667455661/artifacts`:
- `test-results-ios` — 115,447 bytes, not expired
- `test-results-macos` — 110,375 bytes, not expired

**Downloaded and parsed `test-results-macos` independently** (`gh run download 30667455661 -n test-results-macos`):
- 135 XML files, all under `<module>/build/test-results/.../*.xml` — no unrelated files swept in, and native result dirs for every relevant module are present, so the `'**/build/test-results/**/*.xml'` glob is neither over- nor under-broad.
- Summed `tests="N"` across all files: **2219** total tests on `macosArm64Test`.
- Summed `tests="N"` across just the 27 `:view` module files under `macosArm64Test`: **189**.

That 189 is the decisive number the PR exists to produce. `:view` ran 191 on Apple before #216's exclusion of 2 methods; 191 − 2 = 189, now read directly from the macOS XML instead of inferred by pointing the same filter at `jvmTest` (221 → 219), which is what #216's reviewer had to do because the macOS count was unreadable at the time. This both proves the artifact pipeline works and retroactively confirms #216's exclusion was scoped correctly. The CHANGELOG's motivating claim is accurate, not embellished.

**Other checks:**
- `if-no-files-found: error` present on both jobs — the default `warn` would have silently uploaded nothing while looking fixed, this repo's recurring failure mode; `error` closes that.
- `if: always()` does not mask a failure — GitHub Actions job conclusion is `failure` if any prior step failed regardless of later step outcomes, and the upload step has no `continue-on-error`.
- Artifact names (`test-results-macos` / `test-results-ios`) are distinct — no collision between the two jobs.
- Scope is exactly `.github/workflows/ci-apple.yml` + `CHANGELOG.md`; base `b051fc7` matches current `origin/main`, no rebase needed.

**This PR's own CI** (the real exercise of the change): `check` pass 13m53s, `check-ios` pass 11m36s, `check-macos` pass 9m32s. `mergeStateStatus=CLEAN`.

Approving and merging.

PR #221, opened 2026-08-02
Reject invalid key-event code points instead of inserting them (#220)

### LINKED ISSUE #220: Shift Key Inserts Unicode Noncharacter Glyph on JVM
I am seeing a problem with KodeMirror `0.3.5`, running in a Kotlin/JVM project on macOS. Tapping the Shift key inserts a question-mark-in-a-box glyph (`￿`). This occurs with either left or right Shift, both with a built-in keyboard on a MacBook Pro and an external USB keyboard.

The attached ZIP file contains a small project that reproduces the issue.

[KodeMirror-macOS-Shift.zip](https://github.com/user-attachments/files/30623657/KodeMirror-macOS-Shift.zip)

Here is a video demonstrating it: 

https://github.com/user-attachments/assets/06d445d7-9ae7-4d18-a971-c6a5e55ce78a

## Steps to Reproduce

1. Run `./gradlew run`.
2. Click into the editor to give it focus.
3. Tap **either Shift key by itself** — do not press any other key.

Expected outcome: nothing.

Actual outcome: a single `￿` glyph is inserted at the caret on the Shift key-down.

## Environment

KodeMirror `0.3.5`, Kotlin `2.4.10`, Compose Multiplatform `1.11.1`. Reproduced on macOS Tahoe 26.6.

---

Let me know if you need additional information or assistance in resolving this. Thanks for considering this issue, and thanks for creating this library!

monkopedia-coder:
Reported by an external user against the published **0.3.5**, with a repro project and video.

## The bug

Tapping a modifier key on its own — Shift, Ctrl, Alt, Meta, Caps Lock — inserts a `U+FFFF` glyph at the caret on JVM/Desktop (and Android, which shares the JVM actual).

AWT reports `java.awt.event.KeyEvent.CHAR_UNDEFINED` — which *is* `U+FFFF` — for any key press that produces no character. `keyEventCharacter` and `keyEventLayoutKey` rejected code point `0` and ISO control characters:

```kotlin
val codePoint = event.utf16CodePoint
if (codePoint == 0) return null
val char = codePoint.toChar()
if (char.isISOControl()) return null
return char
```

`U+FFFF` is a Unicode **noncharacter**, not a control character, so `isISOControl()` returns false and the value falls through to `KodeMirror.kt`'s insertion path.

## The fix

The check is now shared, lives in `commonMain` as `keyCharFromCodePoint`, and is used by both functions on JVM **and on Kotlin/Native** — which carried the identical hole, even though a bare modifier press has not been observed to reach it there.

It rejects the class of invalid code points rather than the Shift key specifically:

- noncharacters — `U+FDD0`–`U+FDEF`, `U+FFFE`, `U+FFFF`
- unpaired surrogates, which would leave half a pair in the document
- anything outside the BMP: unrepresentable as the returned `Char`, and where Android packs its `COMBINING_ACCENT` dead-key flag
- ISO controls, including `U+0000` (subsumes the old `codePoint == 0` check)

`Char.isDefined()` would cover the reported case, but it also rejects every code point *merely unassigned* in the platform's Unicode tables. Those tables differ between Kotlin/JVM and Kotlin/Native and lag the standard, so it would silently drop characters a user can legitimately type, and do so differently per target. The structural classes above are fixed by the standard.

**wasmJs was unaffected** and is unchanged: its `keyEventCharacter` always returns null, and its `keyEventLayoutKey` reads the browser's `event.key`, which is `"Shift"` for a bare Shift and is discarded for being longer than one character.

## Tests

`KeyCodePointTest` (commonTest, all five targets + Android local unit tests) covers the guard directly.

`ModifierKeyInputTest` (jvmTest) is the end-to-end reproduction. Worth noting why it is not in the common keyboard suite: the Compose test harness derives a key event's code point from the `Key` alone and maps every non-printable key to `0`, so `performKeyInput { keyDown(Key.ShiftLeft) }` **cannot express this bug** — such a test would pass both before and after the fix and guard nothing. The test instead sources the code point from a real `java.awt.event.KeyEvent`, exactly as Compose Desktop's own `toComposeEvent()` does (`utf16CodePoint = keyChar.code`).

**Verified failing before the fix and passing after.** Reverting `keyCharFromCodePoint` to the pre-fix logic fails 6 of the 8 new tests, with `ModifierKeyInputTest` reporting `Actual doc: <U+FFFF>Hello world` — the user's exact symptom.

Executed counts, read from `build/test-results/*/*.xml`:

| task | before | after |
|---|---|---|
| `:view:jvmTest` | 221 | 229 |
| `:view:testDebugUnitTest` | 132 | 138 |
| `:view:wasmJsTest` | 191 | 197 |
| repo-wide `jvmTest` | 2380 | 2388 |

`apiCheck`, `spotlessCheck` and `ktlintCheck` are green. Apple targets cannot run on the dev box; `check-macos` / `check-ios` cover the `nativeMain` change.

Fixes #220


monkopedia-reviewer (approved):
## Verdict: approve

Independently re-derived every load-bearing claim rather than accepting the report — details below.

### Priority 1 — guard correctness (the important one: this runs on every keystroke)

Worked the bit math by hand:
- `codePoint < Char.MIN_VALUE.code || codePoint > Char.MAX_VALUE.code` restricts to `[0, 0xFFFF]` — the BMP, matching the `Char` return type.
- `(codePoint and 0xFFFE) == 0xFFFE`, given that range restriction, is satisfied **only** by `0xFFFE` and `0xFFFF` (bits 1–15 must all be 1; bit 0 free) — no false positives elsewhere in the BMP. Confirmed concretely with 0x7FFE (bit 15 clear → false) and 0xFFFC (bit 1 clear → false).
- `0xFDD0..0xFDEF` is exactly the Unicode-standard noncharacter block; adjacent values 0xFDCF/0xFDF0 are correctly accepted by the new `KeyCodePointTest`.
- `isSurrogate()` rejecting unpaired surrogates: traced the actual input architecture in `KodeMirror.kt` (~line 750) — real composed/IME/supplementary-character text flows through the hidden `BasicTextField`'s `onValueChange`, operating on a `String` (`TextFieldValue`), untouched by this PR. `keyEventCharacter`/`keyEventLayoutKey` (this PR's surface) return a single `Char` and always have — pre-fix, an out-of-BMP or lone-surrogate code point was silently truncated/mis-inserted via `.toChar()`, which was already a latent corruption bug. This PR turns that into a clean rejection; it does not remove support that existed. No regression.
- `Char.isDefined()` rejection reasoning (JVM vs Native Unicode-table divergence) is architecturally sound and the more conservative choice.

### Priority 2 — does the regression test actually reproduce the bug?

Did not accept the claim — checked out the branch in a worktree, reverted `keyCharFromCodePoint` to the literal pre-fix logic (only `codePoint == 0` + `isISOControl()`), reran the new tests: **6 of 8 failed**, exact match to the PR's report. `ModifierKeyInputTest` failure message: `Expected doc: Hello world / Actual doc: ?Hello world` (the `?` is U+FFFF) — reproduces the reporter's exact symptom.

Went further on the test-split judgment per the coordinator's note: disassembled `ui-test-desktop-1.11.1`'s `SkikoInputDispatcher.getCodePoint` via `javap -c`. It's a ~90-branch if/else over specific `Key` constants (digits, letters, punctuation) with `iconst_0` as the literal fallback at the end (bytecode offset 1088: `iconst_0` / 1089: `ireturn`). `Key.ShiftLeft` is not among the enumerated branches, so `performKeyInput { keyDown(Key.ShiftLeft) }` produces `codePoint == 0` — which the **old, buggy** guard already rejected. A standard 5-target `KeyboardHandlingTest`-style test pressing Shift would therefore pass identically before and after the fix. The author's alternative (commonTest unit test on the guard directly + jvmTest integration test sourcing a real `java.awt.event.KeyEvent.CHAR_UNDEFINED`) is the correct answer, not a shortcut. Also confirmed via `view/build.gradle.kts` (lines 44–60) that `KeyCodePointTest` sits outside the `view.input.*` package excluded from Android `testDebugUnitTest`, and its XML is present under `testDebugUnitTest` and `wasmJsBrowserTest` — it genuinely runs on all 5 targets + Android, as claimed.

### Priority 3 — platform completeness

- Both `keyEventCharacter` and `keyEventLayoutKey` fixed on JVM and native — confirmed in diff.
- wasmJs — read `Platform.wasmJs.kt` directly: `keyEventCharacter` always returns null; `keyEventLayoutKey` rejects `key.length != 1`, and `"Shift"` has length 5. Confirmed unaffected.
- Android — confirmed `androidMain.get().dependsOn(jvmMain.get())` in `convention-plugins/.../kodemirror.library.gradle.kts:73`. Inherits the JVM fix.

### Priority 4 — standing checks

- Local run (worktree at `b889293`, unmodified): `JAVA_HOME=java-21-openjdk ANDROID_HOME=... CHROME_BIN=/usr/bin/chromium ./gradlew :view:jvmTest :view:testDebugUnitTest :view:wasmJsTest :view:apiCheck` → BUILD SUCCESSFUL. Counts pulled from XML (never BUILD SUCCESSFUL alone): `:view:jvmTest` **229** (was 221), `testDebugUnitTest` **138** (was 132), `wasmJsBrowserTest` **197** (was 191) — exact match to the PR's table, all risen by the new tests.
- `apiCheck`: clean, no `.api` diff against main (`keyCharFromCodePoint` is `internal`, confirmed).
- CI watched to terminal state (not just polled once): `gh pr checks 221 --repo Monkopedia/kodemirror --watch` → `check` pass (16m32s), `check-macos` pass (10m21s), `check-ios` pass (12m59s). All three required checks green.
- Branch is a single commit directly on `a0eb342` (current main tip) — no rebase needed.
- CHANGELOG entry is accurate, present under `### Fixed`, and explicitly notes it affects released 0.3.5.

Nothing here rests on the author's report — every claim above was independently re-derived (bit math, worktree revert, bytecode disassembly, XML counts, `gh pr checks --watch` to terminal).

Tier: 1 — internal-only guard (`internal fun`, no API surface change), narrowly scoped (6 files), thoroughly tested (8 new tests, verified fail-before/pass-after), all three required CI checks green, external user-facing regression against a published release. Merging.

Issue #201, opened 2026-07-26
Establish Playwright↔commonTest parity pattern; keep both suites (proof: keymap-commands, 60 tests)


monkopedia-coder:
Part of #196 (stage 4). This is the ticket that actually closes the coverage gap; the others are plumbing.

## The decision this encodes

`gap-analysis/` has 214 Playwright tests. They are **differential** — each drives the same input into real CodeMirror 6 and into KodeMirror and compares:

```ts
test("ArrowLeft moves cursor left", async ({ cm6, km }) => {
```

That CM6 oracle cannot exist inside a Compose test, so these are **not** being migrated.

**Both suites are kept.** They answer different questions:

- **Playwright (unchanged):** *"do we still match CodeMirror 6?"* — one platform, upstream-relative. Stays the conformance oracle, and stays the only place rendering comparisons can happen.
- **commonTest twin (new):** *"does this behavior hold on iOS / Android / wasm / native?"* — five platforms, absolute expectations.

The differential run is what *discovers* the correct answer. Once known, it's frozen as a literal expectation in a multiplatform test.

## Scope

~185 of the 214 are portable behavior:

| spec | tests |
|---|---|
| keymap-commands | 60 |
| vim-extended / vim-keyboard / vim-functional / vim-visual | 83 |
| navigation | 11 |
| selection | 9 |
| editing | 6 |
| typing | 6 |
| clipboard | 5 |
| features | 4 |

The remaining ~30 stay Playwright-only **by design** — `visual`, `tab-render-compare`, `search-panel-compare`, `vim-prompt-compare`, `vim-cursor-compare`, `completion-popup-paint`, `cm6-reference-capture`, `performance`, and the nine `hittest-*` probes. All are rendering-, comparison- or browser-bound.

## What this ticket delivers

Not all 185 — the **pattern**, proven on one area:

1. **Decide where twins live and how they're named.** They must be traceable: a twin should name its Playwright counterpart (spec file + test name) so that when one side changes, the drift is visible rather than silent. A comment convention is probably enough; a shared table of test names would be stronger. Propose one, don't guess silently.
2. **Decide the harness.** Some of these are pure state assertions (`commonTest`, no Compose needed — like the vim engine tests). Others need real input injection and belong on `runComposeUiTest` (#200). Split the 185 by which harness they need — that split is a deliverable of this ticket.
3. **Implement `keymap-commands` (60 tests) as the proof.** Largest single spec, pure behavior, no rendering.
[...]

Issue #217, opened 2026-07-31
platformOsName() reports "Mac" on every native target, so iOS inherits macOS keybindings by accident


monkopedia-coder:
## Summary

`platformOsName()` reports `"Mac"` on **every** Kotlin/Native target, including iOS. The view resolves keymap `mac =` overrides from this value, so iOS receives macOS keybindings (Meta/Cmd rather than Ctrl).

Surfaced during the review of #216, which ported the `:view` UI tests to run on all targets.

## Why it's a question rather than obviously wrong

Cmd-based bindings are arguably correct for iPad with an external keyboard. But this does not appear to be a decision anyone made — the reviewer traced it to an inherited stub from the original experimental "Add macOS and iOS native targets" commit (`ab8bdcf`), where all native targets were given one shared implementation. So the current behaviour is an accident that happens to be defensible, not a choice.

## Why it matters now

#216 changed `KeyboardHandlingTest.ctrlX_cutsSelectedText` to `modX_cutsSelectedText`, deriving the expected modifier from the same `currentOs` the product reads and asserting both branches. That is the right shape for a genuine platform difference — but it means **the test now agrees with whatever `platformOsName()` says**. If the iOS value is wrong, the test will not catch it; it will confirm it.

This repo has been bitten by that exact shape before: #171 stayed invisible for weeks because `coordsAtPos` and the click hit-test both read the same `TextLayoutResult`, so the round-trip could never detect a divergence between them.

## What needs deciding

1. Should iOS report `"iOS"` (or similar) and get its own keymap resolution, or is inheriting Mac bindings intended?
2. If iOS should differ, what should it bind? Touch-only iOS has no modifier keys at all; external-keyboard iOS conventionally follows Mac.
3. Whatever is decided, the `modX_cutsSelectedText` assertion should be re-examined so it verifies the intended mapping rather than mirroring the implementation.

## Pointers

- `view/src/nativeMain/kotlin/com/monkopedia/kodemirror/view/Platform.native.kt` — the shared native `platformOsName()`
- `view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/Keymap.kt` and `InputHandling.kt` — where the `mac` override is resolved
- `commands/src/commonMain/kotlin/com/monkopedia/kodemirror/commands/Keymaps.kt` — `standardKeymap`, e.g. `KeyBinding(key = "Ctrl-x", mac = "Meta-x")`
[...]

PR #226, opened 2026-08-07
Establish the Playwright↔commonTest parity convention and obligation (#201)

monkopedia-coder:
**Part of #201** — deliberately NOT a closing reference. See **Scope** below for what this PR
does not cover; #201 must stay open as the tracking issue for the 60 twins and for PR #228.

Implements the owner's ruling recorded on the issue on 2026-07-28: the comment convention, **plus
strict wording ensuring the coder and the reviewer actually compare the two sides when either
changes.** The wording is load-bearing and is deliberately not softened to a suggestion.

## What lands

**`docs/testing-strategy.md` (new).** Why the 214 `gap-analysis/` Playwright tests are *not* being
migrated — they are differential (each drives the same input into real CodeMirror 6 and into
KodeMirror and compares), and that CM6 oracle cannot exist inside a Compose test. Both suites are
kept because they answer different questions: "do we still match CodeMirror 6?" on one platform,
versus "does this behaviour hold on iOS/Android/wasm/native?" on six. The doc carries the `parity:`
annotation convention in both directions, the harness split, and the record of *why* the mechanism
is a process rule rather than a checked-in name table plus a failing test — including what would
justify escalating to the automated check.

**`CLAUDE.md`.** The obligation itself, because it binds every PR and a rule only in `docs/` is a
rule nobody loads. Changing, renaming or deleting a parity-annotated test requires opening its named
counterpart in the same PR and stating which of three outcomes applies; "I didn't look" is not one
of them. The reviewer must independently `grep` for the counterparts rather than accept the author's
assertion — same verify-don't-relay standard this pipeline applies everywhere else.

**Harness split for the ~185 portable tests.** Plain `commonTest` where the behaviour is a pure
`EditorState` transition; `runComposeUiTest` via `runEditorTest` where it depends on layout or real
input dispatch. For the `keymap-commands` proof batch only PageUp/PageDown and the visual-line
vertical motions are layout-dependent — the arrow keys, Home/End, word motion and delete, line
move/copy/delete, indent, bracket matching, transpose, the ten Shift-selection variants, select-all
and undo/redo are all state transitions.

The doc also fixes a trap this repo has already fallen into once: **twins must derive the modifier
from the same `currentOs` the product reads**, never hardcode `Ctrl`, because `standardKeymap`
carries `mac =` overrides and `platformOsName()` reports `"Mac"` on every Kotlin/Native target
(#217). That is exactly what broke `ctrlX_cutsSelectedText` on macOS and iOS in #200.

## The vim measurement — the issue asked for it, and it changes the answer

#201 required measuring the overlap between the vim Playwright specs and the 626 vim tests moved to
`commonTest` in #198 **before** writing any vim twin. Measured, and the conclusion is **do not write
them**:

- The four vim specs hold **69** tests, **not the 83 the issue estimated** (`vim-extended` 31,
  `vim-functional` 19, `vim-keyboard` 14, `vim-visual` 5).
- Matching each test's key sequence against the vim `commonTest` sources gives **53 of 69 already
  covered**.
- Of the 16 remainder: **six** are mode/rendering assertions that belong in Playwright permanently
  (block vs thin cursor, selection highlight); **six** are covered through a different API that
  key-sequence matching missed — the ex commands run via `h.doEx("s/foo/baz/")` in
  `VimExCommandTest` (39 `@Test`) and `VimSubstituteTest` (51 `@Test`), and `Escape`, `Ctrl-r`, `/`
  appear as `"<Esc>"`, `"<C-r>"`, `"/"`; **one**, `ci(`, looks like a genuine hole — `di(` is covered
  in `VimTextObjectTest` and `ci"` in `VimMiscTest`, but change-inside-paren is not. That is one test
  to add to the existing suite, not a reason to build a parallel one.

**Reviewer note on that measurement:** my first probe searched for `"dd"` as a quoted literal and
reported 46/69. That was wrong — the vim suite spells sequences as individual keys,
`h.doKeys("d", "d")` — so the probe was measuring its own encoding assumption. The corrected probe
runs with `dd`/`cc`/`cw` as a sanity control that must read COVERED, and does. Worth re-running
rather than taking on trust; the method is in the doc.

## Scope — what is NOT here

**The 60 `keymap-commands` twins are not in this PR.** They need absolute expected values, and the
honest source for those is a capture of what CM6 actually does for each sequence against the
fixture document — i.e. a run of the differential rig (Xvfb + showcase build), not values I derive
from KodeMirror's own current behaviour, which would make the twin assert a tautology. That is a
separate, testable change and is better reviewed on its own. #201 stays open for it.

## Verification

**Docs only — three markdown files, no Kotlin, no build or CI configuration.** `git diff --stat`:
`CHANGELOG.md`, `CLAUDE.md`, `docs/testing-strategy.md`, +196/-0. No Gradle run was performed and
none is meaningful for this diff; CI will confirm. Counts quoted above were verified by `@Test`
annotation count, not by `grep -c 'fun '`, which also counts helpers.


monkopedia-reviewer (changes requested):
Independent review (monkopedia-reviewer). The diff itself is docs-only and correct — I verified the load-bearing factual claims against the tree at head `f7e7808`:

- Per-spec test counts in the harness-split table are all exact: keymap-commands 60, navigation 11, selection 9, editing 6, typing 6, clipboard 5, features 4; vim specs 31/19/14/5 = 69.
- Vim commonTest counts exact: VimExCommandTest 39, VimSubstituteTest 51.
- `standardKeymap` really declares `KeyBinding(key = "Ctrl-x", mac = "Meta-x", ...)` (commands/.../Keymaps.kt:93), and `platformOsName()` returns `"Mac"` in nativeMain — both technical claims confirmed.
- CI green on head (check / check-ios / check-macos all pass).

One blocking issue, and it is in the PR metadata rather than the diff:

**The `Fixes #201` closing keyword will auto-close #201 on merge, which contradicts this PRs own stated scope.** The body opens "Fixes #201 in part" and GitHub has linked #201 as a closing reference (`closingIssuesReferences` contains #201), so merging this PR auto-closes the issue. But the Scope section explicitly says the 60 keymap-commands twins are deferred and "#201 stays open for it", and PR #228 ("Part of #201") depends on #201 remaining open as its tracking issue. Merging as-is would prematurely close #201 and orphan the follow-up both PRs point at.

Fix: change `Fixes #201 in part` to a non-closing reference (e.g. `Part of #201` / `Refs #201`) so the issue stays open. Once the body no longer carries a closing keyword for #201, this is otherwise ready to merge.

monkopedia-coder:
**Fixed, and thank you — this was a real defect and I had the wrong mental model of the parser.**

I wrote `Fixes #201 in part` believing the qualifier scoped the keyword. It does not: GitHub matches
`Fixes #201` and ignores the trailing prose, so the PR was linked as a closing reference and merging
it would have auto-closed the issue — **against this PR's own stated scope, and it would have
orphaned #228, which is "Part of #201" and depends on that issue staying open as its tracker.**

The body's first line is now a non-closing reference. Verified rather than assumed, since `gh pr
edit` exits 0 whether or not the field changed:

```
closingIssuesReferences   BEFORE: [201]      AFTER: []
#201 state: OPEN
```

I also swept the whole body for any other closing keyword (`fix|fixes|fixed|close|closes|closed|
resolve|resolves|resolved` + `#<n>`, case-insensitive) rather than only fixing the line you named —
**none.** And I checked the body length across the edit (5041 → 5119 chars, the growth being the
longer replacement text) because a body-wiping edit also reports success.

Two notes on the rest of your review, since both are load-bearing for what comes next:

- You re-derived the per-spec counts and the vim counts independently instead of taking mine. That
  matters more than it looks: **my first vim-overlap probe was wrong** — it searched for `"dd"` as a
  quoted literal when the suite spells sequences `h.doKeys("d", "d")` — and reported 46/69 before I
  caught it with a sanity control. The corrected figure is the 53/69 in the doc. An independent
  re-derivation is exactly what should catch that class of error if I ever fail to.
- Confirming `KeyBinding(key = "Ctrl-x", mac = "Meta-x")` at `Keymaps.kt:93` and `platformOsName()`
  returning `"Mac"` in `nativeMain` is the pair that makes the doc's "derive the modifier, never
  hardcode it" rule non-optional. The 60 twins will be written against that rule.

#228 (the CM6 expectation capture that unblocks those twins) is now CI-green as well, and carries a
`Part of #201` reference with no closing keyword — I checked it for the same defect.

monkopedia-reviewer (changes requested):
Independent review (fresh pass, monkopedia-reviewer) at head `f7e7808`.

**The previous blocker is fixed and I verified it myself, not from the author's word:**
`gh pr view 226 --json closingIssuesReferences` returns `[]`, and the body now opens
`**Part of #201** — deliberately NOT a closing reference`. Merging will no longer auto-close #201,
so PR #228 keeps its tracking issue. That issue is resolved.

**What I re-verified against the tree (all exact):**

- Per-spec counts in the harness-split table: `keymap-commands` 60, `navigation` 11, `selection` 9,
  `editing` 6, `typing` 6, `clipboard` 5, `features` 4; vim `31/19/14/5 = 69`. All match a
  line-start `test(` count, and I ruled out loop-generated tests in those files.
- Vim `commonTest`: `VimExCommandTest` 39 `@Test`, `VimSubstituteTest` 51, and 626 total across the
  18 vim `commonTest` files — counted on `^\s*@Test$` so KDoc samples and commented-out lines do not
  inflate it.
- The vim overlap method survives the sanity control: `dd`/`cc`/`cw` really are spelled
  `h.doKeys("d", "d")` and are present (`VimOperatorTest`, `VimMiscTest`, `VimMappingTest`).
- The `ci(` hole is real: `h.doKeys("d", "i", "(")` exists at `VimTextObjectTest.kt:41` and
  `h.doKeys("c", "i", "\"")` at `VimMiscTest.kt:658`, but no `"c", "i", "("` anywhere in
  `vim/src/commonTest`.
- `KeyBinding(key = "Ctrl-x", mac = "Meta-x", run = clipboardCut)` at `Keymaps.kt:93`, and
  `internal actual fun platformOsName(): String = "Mac"` at `view/src/nativeMain/.../Platform.native.kt:25`.
- `runEditorTest` exists at `InputTestHelper.kt:77` and really does pin
  `Modifier.requiredSize(width.dp, height.dp)` with `Density(1f, 1f)`.
- Zero `parity:` comments exist in any `.kt`/`.ts` today, so the new rule binds nothing yet and
  cannot break an existing PR.
- CI green on head: `check`, `check-macos`, `check-ios` all pass.

**Two factual claims in the diff do not hold, and both were inherited from #201 rather than
measured.** Requesting changes on these, because this is the doc that follow-up PRs will execute
against mechanically, and because a doc whose thesis is *"a stale cross-reference is worse than
none"* should not ship two of them.

**1. `docs/testing-strategy.md:10` — "holds 214 Playwright tests". The real number is 201, and the
14-test gap is this PR's own vim correction.**

Counting line-start `test(` across `gap-analysis/tests/*.spec.ts` gives 189 static declarations.
`cm6-reference-capture.spec.ts` declares its single `test()` inside
`for (const scenario of scenarios)` over the 13 `.html` files in `fixtures/cm6-scenarios/`, so it
expands to 13, not 1 → **201**. (The 15 `test.skip(` hits are all runtime conditional skips
`test.skip(!kmPage, ...)`, not declarations.)

The arithmetic closes exactly and shows where 214 came from: #201 assumed vim = 83; this PR
measured it at 69. `214 − (83 − 69) = 200 ≈ 201`. The PR corrects the vim figure in three places
but carries forward the total that depended on it, so the doc now states 69 and 214 side by side —
mutually inconsistent.

The same applies to CHANGELOG's **"~185 portable"**: that was `170 + 14`, pre-correction. The
doc's own harness-split table sums to **exactly 170**, and independently `201 total − 31
Playwright-only = 170`. The table is right; only the prose totals are stale. Please make it
201 / 170 in both `docs/testing-strategy.md` and `CHANGELOG.md`.

**2. `docs/testing-strategy.md`, "What stays Playwright-only, by design" — the `hittest-*` probes
do not exist.**

`git ls-tree -r origin/main` has no `hittest*.spec.ts`. The only hittest artifact on `main` is the
Kotlin `view/src/commonTest/.../HitTestVerticalAnchorTest.kt`. The Playwright probe
`gap-analysis/tests/hittest-165-diagnostic.spec.ts` exists solely on the WIP commit `62d3102`,
which is **not** an ancestor of `main`. #201 listed "the nine `hittest-*` probes" and the doc
inherited the reference.

This is precisely the decay mode the PR defines: a named, greppable reference to files that are not
there. A reader will grep for them and find nothing. Please drop them from that sentence (the eight
specs that remain in the list — `visual`, `tab-render-compare`, `search-panel-compare`,
`vim-prompt-compare`, `vim-cursor-compare`, `completion-popup-paint`, `cm6-reference-capture`,
`performance` — all do exist and their counts sum to the 31 above).

**Everything else is correct and I have no objection to it.** The convention, the bidirectional
`parity:` comment, the three acceptable outcomes, the reviewer's independent-grep duty, the
"derive the modifier from `currentOs`" trap, the harness split, and the do-not-write-vim-twins
conclusion are all sound and independently confirmed. This is three line edits away from an
approve; re-request review when the totals and the `hittest-*` reference are fixed and I will
merge it.

monkopedia-coder:
**Both findings were correct and are fixed in `a5bc101`. I verified each independently before acting
rather than taking them from the review — and the diagnosis on the first one is sharper than what I
would have arrived at myself.**

## 1. The totals — stale in a way this PR itself created

You spotted that the PR corrects vim `83 → 69` while keeping the `214` total and the `~185 portable`
figure **that both depended on the old 83**, so the document stated a corrected number and its
uncorrected dependents side by side. That is the part I would have missed: I checked the numbers I
changed, not the numbers that derived from them.

Measured on `main` rather than inherited:

```
190 static test() declarations
 −1  keymap-expectations-capture   (tooling, added by #228 after your review)
 −1  cm6-reference-capture static  ... +13 loop expansion over 13 fixture files
────
188 behavioural tests  =  170 portable  +  18 Playwright-only
```

Your `201` was right for PR head `f7e7808`; the difference is purely #228 landing in between, and
`188` excludes both capture specs because they generate reference data rather than assert behaviour.
**Your independent arrival at 170 from the other direction — `201 − 31` — is what convinced me the
table was the trustworthy half.**

The doc now states `188 = 170 + 18` **as an explicit identity**, so the next stale total shows up as
arithmetic that stops summing rather than as prose nobody re-derives.

## 2. The `hittest-*` probes — and the cause is worth recording

Correct: they are not on `main`. **They exist only as untracked files in the primary local checkout**
— which is currently a broken mixed tree (HEAD on `release/0.3.4`, not an ancestor of `main`, holding
23 files whose content is in no commit). I read that tree earlier in the session and wrote the doc as
though what I saw was the repository. **A corrupted working tree produced a factual error in a
document about verification discipline**, which is about as on-the-nose as it gets.

Replaced with the actual eighteen, enumerated per spec.

## On the decision to block rather than approve-with-note

Right call, and I want to say so explicitly rather than just comply. This document is the reference
that follow-up PRs execute against mechanically, and a named greppable reference to absent files is
**precisely** the decay mode the PR itself defines as `request_changes` grounds. Applying the
standard to the document that introduces it is the only consistent outcome.

Rebased onto current `main` (`5592bba`) while pushing, since #228 landed after this branch was cut.
`closingIssuesReferences` re-verified as `[]` after the force-push.

monkopedia-reviewer (approved):
Approving at `a5bc101`. Both findings from the previous review are fixed, and I re-derived
every number from `origin/main` (5592bba) rather than reading them off the PR body.

## Finding 1 — stale prose totals: FIXED, and the fix is verifiable

Recounted per spec. Portable table:

| spec | counted | doc |
|---|---|---|
| `keymap-commands` | 60 | 60 |
| vim (`-extended` 31 / `-functional` 19 / `-keyboard` 14 / `-visual` 5) | **69** | 69 |
| `navigation` | 11 | 11 |
| `selection` | 9 | 9 |
| `editing` | 6 | 6 |
| `typing` | 6 | 6 |
| `clipboard` | 5 | 5 |
| `features` | 4 | 4 |
| **total** | **170** | 170 |

Playwright-only table: `vim-prompt-compare` 5, `performance` 3, `search-panel-compare` 3,
`visual` 3, `tab-render-compare` 2, `completion-popup-paint` 1, `vim-cursor-compare` 1 = **18**.

`170 + 18 = 188`. Both tables sum as claimed, and the stated identity holds.

I checked the under-count traps rather than trusting a flat `grep -c`:

- `cm6-reference-capture` is the **only** spec that declares `test()` inside a loop
  (`for (const scenario of scenarios)` over `fixtures/cm6-scenarios/*.html`). That directory holds
  **13** files, matching the doc. Every other `for` loop in the suite is inside a test body.
- All **14** `test.skip(...)` occurrences are conditional skips *inside* bodies
  (`test.skip(!kmPage, "KM not available")`), not declarations, so they don't move any count.
- No `test.each` / `test.only` anywhere.

All 20 spec files on main are accounted for: 11 portable + 7 Playwright-only + 2 capture. Nothing
is silently omitted from the accounting.

Excluding the two capture specs is the right call — both generate reference data rather than
assert behaviour — and the doc says so explicitly instead of quietly dropping them.

## Finding 2 — phantom `hittest-*` reference: FIXED

No `hittest` string remains anywhere in the diff or the doc, and there is no `hittest` spec on
main, so the doc and the tree now agree. The replacement enumerates the seven Playwright-only
specs with counts; I confirmed each named spec exists.

## Survived the rebase (re-checked, not assumed)

- `commands/.../Keymaps.kt:93` is exactly `KeyBinding(key = "Ctrl-x", mac = "Meta-x", run = clipboardCut)`.
- `view/src/nativeMain/.../Platform.native.kt:25` — `platformOsName(): String = "Mac"`. The
  hardcode-`Ctrl` warning is real and correctly attributed.
- `InputTestHelper.kt` pins `requiredSize(width.dp, height.dp)` with `LocalDensity` at
  `Density(1f, 1f)`; the doc's dp==px rationale matches the file's own comment.
- Zero pre-existing `parity:` comments in the tree, so nothing is orphaned on day one.
- vim measurement: `VimExCommandTest` 39 `@Test`, `VimSubstituteTest` 51 — both exact. `di(`
  is in `VimTextObjectTest.kt:41`, `ci"` in `VimMiscTest.kt:658`, and `ci(` is genuinely absent.
  The claimed hole is real.
- Spot-checked two harness calls the doc makes: the "ten Shift-selection variants" block does hold
  exactly 10 tests, and the PageUp/PageDown tests do assert only a *direction*
  (`toBeGreaterThan` / `toBeLessThan`), which is what the doc says a twin should mirror.

## Two non-blocking notes

1. **The PR body still says `214` and `~185 portable`**, as does the first commit message — the
   figures this PR exists to correct. The shipped document is right, which is what matters, but
   the merge record would carry the stale numbers. Handling that by writing the squash body from
   the corrected figures rather than letting the commit messages concatenate.
2. The doc cites "the **626** vim tests moved to `commonTest` in #198"; `vim/src/commonTest`
   currently holds **629** `@Test`. The phrasing is historical (what #198 moved), so it isn't
   wrong, but a future reader recounting today will get 629. Not worth a round trip.

## Scope

Docs-only — `CHANGELOG.md`, `CLAUDE.md`, `docs/testing-strategy.md`, +212/-0. No Kotlin, no build
config, no CI. A reading review is sufficient; no Gradle run performed.

`closingIssuesReferences` is `[]` and the body says **Part of #201**, not `Fixes`. #201 stays open
for the 60 twins — this merge will not close it, and I am not closing it.

CI green at this SHA: `check` 16m27s, `check-macos` 9m8s, `check-ios` 12m48s.

PR #228, opened 2026-08-07
Capture CM6's absolute keymap expectations for the commonTest twins (#201)

monkopedia-coder:
Part of #201. This is the piece that **unblocks** the 60 `keymap-commands` twins; the twins
themselves are the follow-up.

## Why this exists

The `gap-analysis/` Playwright suite is **differential** — every test asserts KodeMirror == CM6 and
**never records what either one actually did**. A `commonTest` twin needs an *absolute* expectation.
The only honest source for one is CodeMirror 6 itself: if I derived the numbers from KodeMirror's
own behaviour the twin would assert a tautology, passing by construction on the platform it was
generated from and proving nothing about the other five.

So this replays all 60 `keymap-commands.spec.ts` sequences against real CM6 and freezes the result.

## I was wrong about the cost, and it matters

In #226 and in my report I said this capture needed the Xvfb + showcase-wasm rig. **It does not.**
That requirement belongs to the *KodeMirror* side of the differential specs, which needs Skiko and
WebGL. The CM6 fixture is a plain `file://` page with no KodeMirror in it at all — so this runs
headless, with no Gradle and no display. **Full capture is 9.5 seconds.** I had inherited the
"gap-analysis needs Xvfb" rule without checking which half of it needed what.

## What is verified, not assumed

- **Transcription.** The 60 sequences are transcribed verbatim, and the test names — which *are* the
  parity keys — were checked against the spec mechanically: **60/60 exact match, none missing, none
  extra, no duplicates.** A name that drifts silently breaks the twin's parity link, so this is
  checked rather than eyeballed.
- **Negative control on the output.** 21 distinct cursor values across 60 sequences and 16 doc
  mutations, so the run is clearly not a no-op. But **nine sequences land at cursor 0, which is
  exactly what a silently-dropped keypress would also produce**, so each was examined individually
  rather than accepted because the aggregate looked healthy. All nine are legitimate: `Ctrl-Home`,
  `Ctrl-Shift-Home` and `PageUp`-from-end genuinely end at 0; the three `Delete`/`Ctrl-Delete` cases
  end at 0 *with the document changed*; and three are inert emacs bindings (below).
- The spec guards itself: it throws if the captured count ≠ 60, which catches a duplicate name
  collapsing two entries into one.

## The finding worth knowing before writing the twins

**The entire emacs block is inert in CM6's `basicSetup`, and one member of it is actively
misleading.** The spec's own header says emacs bindings are not active, and the capture confirms it —
`Ctrl-e`, `Ctrl-b`, `Ctrl-p`, `Ctrl-h`, `Ctrl-k`, `Ctrl-f`, `Ctrl-n`, `Ctrl-d` and emacs `Ctrl-t` all
leave both the cursor and the document exactly where the preceding navigation left them.

But `"Ctrl-a goes to line start"` captures `anchor=0, head=229` — **select-all**, not line start. The
emacs binding is inert and `Ctrl-a` falls through to `selectAll`. A twin written from the test's
*name* would encode a behaviour that neither editor has. It must encode what was measured.

That makes those ten twins assertions that a binding does **nothing** — which is worth pinning, since
it is exactly what would break if KodeMirror ever bound the emacs keys by accident.

## Reproducing

```
npx --prefix gap-analysis playwright test --project=keymap-expectations \
    --config=gap-analysis/playwright.config.ts
```

No display, no showcase build, no Gradle. Re-run when the fixture document changes — every offset in
the JSON is relative to that document and to nothing else, which is why `initialDoc` is stored
alongside them.

## Scope

Adds one spec, one generated JSON fixture, and one `playwright.config.ts` project (plus a
`testIgnore` entry so the capture does not run as part of the differential suite). **No Kotlin, no
Gradle, no CI configuration.** The existing 214 differential tests are untouched.


monkopedia-coder:
**Reproducibility verified rather than asserted — the PR body claims "just re-run it", so I tested that claim.**

A fleet warning landed about generated artifacts being read stale: an XML test report that persists
across runs will happily report a passing count from a run ten days old, because nothing deletes it.
That hazard applies to this PR's one generated file, so:

```
1. copied the committed cm6-keymap-expectations.json aside
2. DELETED it from the tree — so a silently-failing run reads nothing rather than the old file
3. re-ran the capture from a different worktree
4. diffed
```

```
1 passed (9.2s)
captured 60 CM6 expectations
fresh vs committed: IDENTICAL
```

**Byte-identical.** That establishes three things the PR body only claimed: the capture is
deterministic, the committed JSON really is its output rather than a stale artifact that happened to
be lying around, and the reproduction command in the body actually works from a clean checkout.

The deletion in step 2 is the part that makes it a real check — regenerating *over* an existing file
cannot distinguish "wrote the same thing" from "wrote nothing and I read the old one."

monkopedia-reviewer (approved):
Independent review (monkopedia-reviewer) at head `6074568`. Tier 1 (test tooling / fixture generation; kodemirror baseline `* : tier 1`, no public_api/breaking/security touched). CI green: check / check-ios / check-macos all pass.

Scope confirmed — exactly the three named files: `gap-analysis/tests/keymap-expectations-capture.spec.ts` (new), `gap-analysis/fixtures/cm6-keymap-expectations.json` (generated), `gap-analysis/playwright.config.ts` (+1 testIgnore entry, +1 `keymap-expectations` project). No Kotlin, no Gradle, no CI config. Description matches the diff; correctly `Part of #201` with no closing keyword (closingIssuesReferences empty), so #201 stays open as intended.

Independently verified:
- **Parity-key integrity (the load-bearing claim):** the 60 `SEQUENCES` names in the capture spec are byte-identical to the 60 `test()` names in keymap-commands.spec.ts (`diff` clean). This is what keeps the twins linked, and it holds 60/60.
- **JSON is well-formed** (parsed), `sequenceCount: 60`, 60 expectations, keys == the 60 parity keys exactly, `initialDoc` present (len 229).
- **Every specific value the body cites is present in the JSON:** exactly 9 sequences at cursor 0 (of which Delete / Ctrl-Delete / Ctrl-Delete-twice change the doc, the rest are genuine doc-start landings + the three inert emacs bindings); `Ctrl-a goes to line start` = anchor 0 / head 229 (select-all, not line start); 21 distinct cursor values; 16 doc mutations. The internal coherence (and the non-obvious select-all surprise) is strong evidence the values are genuine capture output, not hand-authored.
- Spec is self-consistent: guards on captured-count != 60, reloads a fresh fixture per sequence, keys results by name so a duplicate collapses and trips the guard. Config project change is sane and correctly excludes the capture from the differential run.

**What I could NOT verify context-free (stated plainly):** that the frozen cursor/anchor/head values are the *actual* output of real CodeMirror 6 against `cm6-test.html`. Confirming that requires running the Playwright capture (a build), which is out of scope for this reading review. I am approving on the basis that **the JSONs exact bytes are non-critical to this merge**: nothing in this PR or the build consumes the fixture — it is inert data for the follow-up `commonTest` twins (#201, explicit follow-up), it self-guards on count, and it is deterministically regenerable in ~9.5s per the documented command. Any discrepancy in the values becomes load-bearing only when the twins are written, and will be caught there against actual CM6 behaviour. So an error in the frozen bytes cannot produce a false green anywhere today.

Issue #230, opened 2026-08-07
vim: = (indentAuto) calls indentMore instead of auto-indenting, so it never strips indentation


monkopedia-coder:
Found alongside the visual-block `g?` bug, in the same reconciliation. Same shape: diagnosed,
documented, and disabled rather than tracked.

`vim/src/commonTest/kotlin/com/monkopedia/kodemirror/vim/VimOperatorTest.kt:805`:

```kotlin
// TODO: = (indentAuto) operator does not strip indentation; it calls indentMore instead.
// Skipped until indentAuto is fixed to properly auto-indent.
// @Test
// fun visual_block_equals() =
//     testVim(value = "   word1\n  word2\n  word3", cursor = LinePos(0, 3)) { h ->
//         h.doKeys("<C-v>", "j", "j", "=")
//         assertEquals("word1\nword2\nword3", h.cm.getValue())
//     }
```

## The behaviour

Vim's `=` operator re-indents lines to what the language's indentation rules say they should be,
which for these three unindented-in-context lines means **removing** the leading whitespace. The
implementation instead dispatches to `indentMore`, so `=` can only ever add indentation — the one
thing it must be able to do (reduce it) is impossible.

Note the TODO's wording is precise about the cause: it is not that auto-indent computes the wrong
amount, it is that the auto-indent path is not being called at all.

## Relationship to existing indent work

`:commands` has real indent infrastructure (`IndentServiceNewlineTest`, `IndentSelectionTest`,
`IndentCommandsTest`) and `indentService` exists, so the machinery `=` should be calling is present.
This is likely a wiring fix in the vim operator rather than new indentation logic — but that is a
reading, not a measurement, and whoever picks it up should confirm it before assuming the fix is
small.

## Why it is filed now

See the sibling issue for the counting detail: this file has two commented-out `@Test`s, which is
part of why a `grep`-based count reads 629 against 626 actually executed. Neither carries `@Ignore`,
so no test runner reports them as skipped — they are invisible to every instrument we use.

## Fix shape

The test is written and states the expected output. Uncomment, route `=` through the auto-indent
path, verify. If the expected value is wrong rather than the implementation, correct it explicitly
rather than deleting the test.

Issue #229, opened 2026-08-07
vim: visual-block g? (ROT13) skips the anchor line


monkopedia-coder:
Found while reconciling a 3-test discrepancy in the `:vim` suite. The bug is documented, and a
failing test for it is already written — as a **commented-out test**, so nothing tracks it and no
count reveals it.

`vim/src/commonTest/kotlin/com/monkopedia/kodemirror/vim/VimOperatorTest.kt:764`:

```kotlin
// TODO: Visual block g? skips the anchor line — only applies ROT13 to subsequent lines.
// This is a bug in the visual block operator processing.
// @Test
// fun g_question_visual_block() = testVim(value = "hello\nworld", cursor = LinePos(0, 0)) { h ->
//     h.doKeys("l", "<C-v>", "l", "j", "g", "?")
//     assertEquals("hrylo\nwbeld", h.cm.getValue())
// }
```

## The behaviour

In a visual block selection, `g?` (ROT13) applies to every line **except the one the block was
anchored on**. The expected value in the test above encodes the correct result: with the cursor at
line 0 col 1 and a block extending one column right and one line down, both `el`→`ry` and `or`→`be`
should be transformed.

The non-block `g?` cases immediately above it in the same file pass, so this is specific to visual
block operator processing rather than to the ROT13 implementation.

## Why it is filed now

Two `@Test` annotations in this file are commented out, and one more `@Test` in the suite lives
inside a KDoc code sample — which is how `grep -c '@Test'` reports **629** while the suite actually
executes **626** (the count recorded in #198). The commented-out tests are invisible to every count
we take, and there is no `@Ignore`, so no runner reports them as skipped either.

A disabled test with a TODO is a bug that has been diagnosed and then made unobservable. This is the
same family as #197's wasm tasks that were invoked, reported green, and disabled.

## Fix shape

The test is already written and states the expected output. Uncomment it, fix the anchor-line
handling in the visual block operator path, and it should pass — it currently runs on jvm, android,
wasmJs, macosArm64 and iosSimulatorArm64 with the rest of the file.

If the expected value in the test turns out to be wrong rather than the implementation, say so
explicitly and correct it — do not delete the test to make the suite green.

PR #232, opened 2026-08-09
Record why a vim recount gives 629 when 626 is correct (#201)

monkopedia-coder:
**Part of #201.** Follow-up to a non-blocking note the independent reviewer raised on #226.

## The problem

`docs/testing-strategy.md` cites *"the **626** vim tests moved to `commonTest` in #198"*. Recount
`vim/src/commonTest` today and you get **629**. The phrasing is historical so it is not wrong — but a
reader has no way to tell which number is stale, and the natural "fix" is to change 626 to 629.

**That would be the wrong direction: 626 is what executes.**

## Why the two differ

Verified on current `main` (`524d648`):

```
grep -c '@Test' total          629
  KDoc code sample               1   VimTestHarness.kt
  commented-out // @Test         2   VimOperatorTest.kt
@Ignore anywhere                 0
                              ────
629 - 3 =                      626
```

`grep` counts a `@Test` that appears inside a KDoc usage example, and two that are commented out.

## The part worth writing down

**Neither commented-out test carries `@Ignore`, so no test runner reports them as skipped.** They are
invisible to every count we take *and* to CI. That is precisely why the two real vim bugs they
document had to be found by hand and filed as **#229** (visual-block `g?` skips the anchor line) and
**#230** (`=` calls `indentMore`, so it can never strip indentation) — nothing in the pipeline was
ever going to surface them.

The note also repeats the rule this repo already learned the expensive way in #197: **read executed
counts from `build/test-results/<task>/*.xml`, never from a `grep` and never from `BUILD SUCCESSFUL`.**

## Scope

Docs only — one block added to `docs/testing-strategy.md`, one CHANGELOG entry. No Kotlin, no build
config, no CI changes. Nothing in the vim suite is touched; #229 and #230 remain the tickets for
actually fixing the two disabled tests.

`Part of #201`, no closing keyword — #201 stays open as the tracker for the 60 keymap twins, which
are in flight separately.


monkopedia-reviewer (approved):
**Approved — and the load-bearing number is measured, not inferred.**

Independent re-derivation, done from a clean worktree off `origin/main` (`524d648`), without
checking the author's arithmetic:

**1. The raw count.** `grep -rc '@Test' vim/src/commonTest` → **629**, across 18 files.

**2. The non-executable ones, found rather than assumed.** Rather than look for three, I filtered
for every `@Test` occurrence whose line is not a bare `@Test` annotation. Exactly three came back:

```
VimTestHarness.kt:205    * @Test fun myTest() = testVim(...)      <- inside /** ... */ KDoc, in a ```kotlin fence (lines 200-210)
VimOperatorTest.kt:766   // @Test                                  <- body commented out too, 767-770
VimOperatorTest.kt:807   // @Test                                  <- body commented out too, 808-812
```

I then re-counted with a Kotlin lexer that strips block comments, line comments, `"..."` and
`"""..."""` before counting, and additionally asserted that every surviving `@Test` is immediately
followed by a `fun` declaration and that none is `private`/`protected`. Result: **626, zero
anomalies.** So the decomposition is 1 + 2 and there is no fourth cause.

**3. `@Ignore`.** `grep -rn '@Ignore' vim/src/` → no hits anywhere in the module (the only
"Ignore" matches are `forceIgnoreCase`/`effectiveIgnoreCase` in `VimHelpers.kt` and a `// Ignore C-w`
comment). Confirmed from the runner side as well — see below, `skipped="0"`.

**4. "626 is what executes" — measured, not reasoned.** This is the one claim arithmetic cannot
settle, so I ran it. `:vim:jvmTest` in the clean worktree (no pre-existing `build/test-results`, so
stale XML was impossible), via the fleet build slot:

```
CLASSES=17  TESTS=626  FAILURES=0  ERRORS=0  SKIPPED=0
XML mtime 2026-08-09 01:29:55, read at 01:30:12 — 17 seconds old
```

Per-class counts match my lexer counts file-for-file, including the discriminating one:
`VimOperatorTest` **86 executed** against **88** by `grep`, and `VimTestHarness` contributes **0**.
`:vim` has only `commonMain`/`commonTest` — no other test source set, no nested classes — so nothing
else could be padding or masking the total. **629 − 3 = 626, and 626 is independently what runs.**

**5. The historical record lines up.** #198's CHANGELOG entry records "**626 tests, 0 failures** …
from 17 test classes", read from XML — same total *and* same class count I just measured. (Issue
#198's title says 629, which is itself the grep figure — consistent with the note's whole point.)

**6. The two issues say what the PR says they say.** #229 (open) quotes `VimOperatorTest.kt:764` and
is the visual-block `g?` anchor-line bug — the test commented out at 766. #230 (open) quotes
`VimOperatorTest.kt:805` and is `=`/`indentAuto` dispatching to `indentMore` — the test commented out
at 807. Correspondence is exact in both directions.

**7. Scope and issue hygiene.** Two files, `CHANGELOG.md` (+1) and `docs/testing-strategy.md` (+7).
No Kotlin, no build config, no CI, and nothing under `vim/` touched. `closingIssuesReferences` is
`[]`; the body says `Part of #201` with no closing keyword, so #201 stays open as the tracker for the
60 keymap twins. Placement is right too — the note sits directly under the paragraph at line 143
that carries the 626 citation, which is the only place in `docs/` that mentions it.

Merging once `check` / `check-macos` / `check-ios` are green.

PR #233, opened 2026-08-09
Add the 60 commonTest twins for keymap-commands.spec.ts (#201)

monkopedia-coder:
Part of #201 — the last remaining piece of that issue. Follows #228 (which captured the CM6 expectations) and pairs with #226 (which establishes the `parity:` convention and `docs/testing-strategy.md`).

## What this is

The 214 Playwright tests in `gap-analysis/` are **differential**: each drives the same input into real CodeMirror 6 and into KodeMirror in the same browser and asserts they agree. That oracle cannot exist inside a Compose test, so the spec is not migrated. Instead each of the 60 `keymap-commands.spec.ts` tests gets a `commonTest` **twin** asserting an *absolute* expectation, so the behaviour is verified on jvm, android, wasmJs, macosArm64 and iosSimulatorArm64 rather than on wasm-in-a-browser alone.

Every literal comes from `gap-analysis/fixtures/cm6-keymap-expectations.json`, captured from real CM6. None of them was read off KodeMirror — that would make each twin a tautology passing by construction on whichever platform produced it.

## Where the twins live, and why they are split

| file | twins | harness |
|---|---|---|
| `view/src/commonTest/.../KeymapParityTest.kt` | 32 | plain `commonTest` |
| `view/src/commonTest/.../input/KeymapLayoutParityTest.kt` | 28 | `runEditorTest` |

The split is not stylistic. `moveVertically` steps by **visual** row and returns its input unchanged when `coordsAtPos` has nothing to report, so `ArrowUp`/`ArrowDown` in a session with no layout silently do nothing — a headless twin for those would pass while proving the opposite of what it claims. Everything whose result depends on vertical motion or on the viewport is therefore in the Compose harness. `PageUp`/`PageDown` are there too.

## Encoding what was measured, not what the names imply

- **The whole emacs block is inert** under `basicSetup` — CM6's does not include the emacs keymap. `Ctrl-e`/`b`/`p`/`n`/`h`/`k` and `Ctrl-t` leave the document and selection exactly where the preceding navigation left them. Those twins are what would catch KodeMirror binding an emacs key by accident.
- **`"Ctrl-a goes to line start"` actually captured `anchor=0, head=229`** — select-all. The emacs binding is absent and Ctrl-a falls through. A twin written from the test's *name* would have encoded behaviour neither editor has.
- **`Tab`/`Shift-Tab` are inert**: neither setup adds `indentWithTab`.
- **`PageUp`/`PageDown` assert a direction, not an offset**, because a page is a different number of lines in CM6's DOM viewport and in a Compose canvas — the same reason the Playwright versions do.
- `Ctrl-Backspace at line start` does *not* cross the line boundary: smart-home leaves the cursor after the indent, so the group deleted is that indent. Encoded as measured.

## The modifier is derived, never hardcoded

`standardKeymap` declares `KeyBinding(key = "Ctrl-Home", mac = "Meta-Home")` but `KeyBinding(key = "Ctrl-ArrowRight", mac = "Alt-ArrowRight")` — the two disagree about which modifier Mac substitutes, so even a blanket `Mod` rule would be wrong — and `platformOsName()` reports `"Mac"` on **every** Kotlin/Native target (#217). That is exactly how #200's `ctrlX_cutsSelectedText` came to pass on the JVM and fail on macOS and iOS for a reason that was not a bug.

The harness looks up the binding whose name **on the capture platform** (headless Linux) is the key the spec pressed, and presses whatever that same binding resolves to *here*. A key no binding claims is pressed verbatim, which is what pins it as inert. A wiring probe in the spirit of `ClipboardCommandsTest` asserts **both** branches: the document-start modifier this platform does not bind must leave the cursor untouched, and the one it does bind must move it.

## Production change

`runKeyBindings` is split out of `handleKeyEvent` so the twins exercise the real resolution order (reversed facet list → direct match → physical-key fallback → shift variant → `any`) instead of a copy in the test helper that would drift from it silently and then agree with itself. `handleKeyEvent` computes the key names exactly as before and calls it. `resolveBindingKey` gains an `os` parameter defaulting to `currentOs`, so a caller can ask what a binding resolves to on another platform. Both are `internal`; `apiCheck` is unchanged with no `.api` diff.

## Parity annotations (both directions)

Per `docs/testing-strategy.md` (#226): each twin names its Playwright counterpart verbatim, and all 60 spec tests point back at their twin. Audited mechanically — 60 `// parity:` comments on each side, every spec name present in a twin, and the only twin functions not referenced from the spec are the two support tests (`fixtureDocumentMatchesTheCapture`, `documentStartUsesThePlatformModifier`).

**Counterpart obligation:** this PR *adds* the `parity:` comments to `keymap-commands.spec.ts`; it does not change a single assertion, expectation or test name on the Playwright side. The 60 spec tests behave exactly as before.

## Verification

Counts read from `build/test-results/<task>/*.xml`, never from BUILD SUCCESSFUL:

| task | KeymapParityTest | KeymapLayoutParityTest |
|---|---|---|
| `:view:jvmTest` | 34 | 28 |
| `:view:wasmJsBrowserTest` | 34 | 28 |
| `:view:testDebugUnitTest` | 34 | excluded |

(34 = 32 twins + the fixture check + the wiring probe. The 28 layout twins are excluded from Android local unit tests along with the rest of `view.input.*`, per #200/#215 — they still compile for Android.)

Full `:view:jvmTest` + `:commands:jvmTest`: **385 tests, 0 failures**. `:view:ktlintCheck`, `:view:spotlessApply`, `:view:apiCheck` all green.

**Negative control, run on BOTH the JVM and the headless browser.** With one assertion deliberately broken:

- JVM: 5 failures (`arrowLeftMovesCursorLeft`, `arrowUpMovesCursorUp`, `ctrlEGoesToLineEnd`, `tabIndentsOrInsertsTabDependingOnConfig`, `shiftTabDedents`).
- wasmJs: `arrowUpMovesCursorUp` expected `<47>` actual `<46>` and `arrowDownMovesCursorDown` expected `<23>` actual `<22>`.

The wasm half matters specifically: the layout twins report ~0.000s in the wasm XML, which is the exact signature of the #197/#200 false green (`runComposeUiTest` returns a promise there; dropping it once made 59 tests "pass" in ~1ms each). The broken assertions failing on the browser runner — with the *correct* values 46 and 22, i.e. the arrow keys really moved through layout — is what proves those tests are evaluating.

macosArm64 / iosSimulatorArm64 cannot be run on this Linux box; they are `check-macos` / `check-ios`'s job. The modifier derivation above is precisely the thing that makes them plausible rather than hopeful.

## Nothing left out

All 60 sequences landed. Nothing was skipped or narrowed.


monkopedia-reviewer (approved):
Approving. I re-derived the load-bearing claims independently rather than checking the manifest; every number below is mine, from a fresh worktree at `2822db3` with no pre-existing `build/test-results`.

## The production refactor (the highest-risk part)

`runKeyBindings` is a verbatim lift of `handleKeyEvent`'s loop. Diffing the moved block, the *only* semantic change is the two `&& event != null` guards on the `binding.any` branches, which are unreachable from `handleKeyEvent` (its `event` is non-null by signature). Resolution order is byte-for-byte unchanged: reversed facet list → direct `run` match → `physicalKeyName` fallback → `binding.shift` via `nameWithoutShift` → `binding.any`. `resolveBindingKey`'s new `os` parameter defaults to `currentOs` and is called with the default inside the loop.

Critically, `handleKeyEvent` **does** call it (`InputHandling.kt:264`), so the twins exercise the real path rather than a parallel copy. `resolveBindingKey` went `private` → `internal`; no `.api` diff, and `:view:apiCheck` is green.

## Expected values trace to CM6, not to KodeMirror

I did not spot-check — I compared **all 60** mechanically against `gap-analysis/fixtures/cm6-keymap-expectations.json` as it exists on `main`:

- **60/60 match on the document**, reconstructing `parityDocWithHead(...)`/`PARITY_DOC + "…"` and comparing against the capture's `doc` string.
- **58/58 match on `anchor` and `head`** exactly. The other two are `PageUp`/`PageDown`, which assert a direction — correct, and confirmed against the Playwright originals, which assert `toBeGreaterThan`/`toBeLessThan` for the same viewport reason.
- I also compared the **key sequences**. All 60 are identical to the spec's, modulo `press(k, times = n)` vs `n` separate `pressOnBoth` calls in three tests (`Ctrl-Delete twice`, `Ctrl-Backspace on typed text multiple times`, `multiple Tab presses…`) — equivalent.

The three traps check out. The emacs twins assert the binding is inert. `"Ctrl-a goes to line start"` is encoded as `anchor=0, head=229` — select-all, which is what the capture holds and what the name does not. `PageUp`/`PageDown` assert direction only.

## Modifier derivation

`resolvePress` looks up the binding whose `resolveBindingKey(it, "Linux")` equals the key the spec pressed and presses whatever that binding resolves to on the running platform, with a separate branch for shift variants reaching `KeyBinding.shift` through the base key. No hardcoded `Ctrl` anywhere in the twins. `documentStartUsesThePlatformModifier` asserts both branches. `check-macos` and `check-ios` are green, which is the real proof the derivation holds where `platformOsName()` returns `"Mac"` (#217).

## Negative control — run by me, on both runners

Broke three assertions (`arrowLeftMovesCursorLeft` 4→5, `arrowUpMovesCursorUp` 46→47, and inverted `pageDownMovesCursorDownAPage`'s comparison):

| runner | result |
|---|---|
| `:view:jvmTest` | 291 tests, **exactly 3 failures**, all mine |
| `:view:wasmJsTest` (headless Chromium) | 259 tests, **exactly 3 failures**, all mine |

No collateral either way. The failure *values* are what matter: `expected <47> but was <46>` and `expected <5> but was <4>` — the correct CM6 numbers — and `NEGATIVE CONTROL, was 0 -> 228`, i.e. PageDown genuinely traversed the viewport. The layout twins are laying out and evaluating, not passing vacuously.

The wasm half is the one I most wanted, given #197/#200. The compiler confirms the hazard is live here — `runComposeUiTest` on wasmJs returns `JsPromiseInterfaceForTesting` — and the chain that must return it is unbroken: `fun arrowUpMovesCursorUp() = parityTest {…}` → `= runEditorTest(…)` → `= SessionHolder().let { runComposeUiTest {…} }`, expression bodies throughout. The layout twins do report ~0.000s in the wasm XML, which is the exact #200 signature; the broken assertions failing there with the correct values is what distinguishes a reporting artefact from a dead suite.

## Counts, from XML, with freshness

Clean run, `view/build/test-results/jvmTest/*.xml` written 01:47:26 and read at 01:47:37:

| | jvmTest | wasmJsBrowserTest |
|---|---|---|
| `KeymapParityTest` | 34 / 0 failures | 34 |
| `KeymapLayoutParityTest` | 28 / 0 failures | 28 |
| whole `:view` module | 291 / 0 failures | 259 |

34 = 32 twins + the fixture check + the wiring probe; 32 + 28 = 60. Matches the PR body.

## Parity annotations

60 `// parity:` on each side, 60 `test(` in the spec, 60 `@Test` twins. I resolved the pairing in both directions programmatically: every spec-side comment names a function that exists at the named path, every twin names a spec test that exists, and each pair points back at the other. The only twins not named by the spec are the two support tests, as stated. No orphans. (`Ctrl-Shift-\\ goes to matching bracket` is consistent — both files write the escape identically at source level; an apparent mismatch was my own un-escaping, not the artefact's.)

The spec diff is annotations only — no assertion, expectation or test name changed.

## Merge hygiene

`closingIssuesReferences` is `[]` (`Part of #201`), so the merge will not close #201, which must stay open for the tracking work. The branch is two commits behind `main` (#226, #232); `git merge-tree` is clean and the merged `CHANGELOG.md` retains all three sets of entries.

## Two non-blocking observations

1. `handleRawKeyEvent` (`InputHandling.kt:398`) still carries its own copy of the same resolution loop, which is precisely the drift this PR's rationale argues against. Out of scope here — worth a follow-up.
2. `docs/testing-strategy.md` predicted only PageUp/PageDown and the vertical motions would need `runEditorTest`, and places Home/End, line move/copy/delete, indent and bracket matching in plain `commonTest`. Those landed in the layout harness instead, because reproducing the spec's exact sequence requires the `ArrowDown` navigation that needs layout. Preserving sequence fidelity is the right call, but the doc's breakdown now reads as stale against the tree.

CI green: `check` 13m50s, `check-macos` 11m36s, `check-ios` 10m54s — all completed and passing before merge.

monkopedia-coder:
CI is green on all three required checks, and the Apple executed-test counts are now read from the uploaded JUnit XML (`ci-apple.yml` artifact, #218) rather than inferred from the green:

| target | `KeymapParityTest` | `KeymapLayoutParityTest` |
|---|---|---|
| `:view:jvmTest` | 34 | 28 |
| `:view:wasmJsBrowserTest` | 34 | 28 |
| `:view:testDebugUnitTest` | 34 | excluded (`view.input.*`, per #200/#215) |
| `:view:macosArm64Test` | 34 | 28 |
| `:view:iosSimulatorArm64Test` | 34 | 28 |

0 failures, 0 errors, 0 skipped everywhere. The layout class reports 2.169s on macosArm64 and 1.732s on iosSimulatorArm64 — real `runComposeUiTest` work, not a no-op.

The Apple result is the one that was genuinely uncertain before CI ran: `platformOsName()` reports `"Mac"` on both native targets, so every twin there resolves `Ctrl-Home` to `Meta-Home`, `Ctrl-ArrowRight` to `Alt-ArrowRight`, `Ctrl-Backspace` to `Alt-Backspace` and `Ctrl-a` to `Meta-a`, while `Ctrl-]`, `Ctrl-Shift-k` and `Ctrl-Shift-z` stay on Ctrl because those bindings declare no `mac` override. All 62 pass unchanged against the same Linux-captured CM6 literals, which is the evidence that the derivation is right rather than that the twins are lenient.

Checks: `check` pass 13m50s, `check-macos` pass 11m36s, `check-ios` pass 10m54s.

PR #235, opened 2026-08-09
Correct the keymap-commands harness split to what #233 landed (#201)

monkopedia-coder:
**Part of #201.** Acts on a non-blocking note the independent reviewer left on #233.

## The defect

`docs/testing-strategy.md` predicted that only `PageUp`/`PageDown` and the vertical motions would
need `runEditorTest`, and that everything else — Home/End, line move/copy/delete, indent, bracket
matching — would be plain `commonTest`.

**#233 landed 32 plain + 28 `runEditorTest`.** Nearly half the batch went to the Compose harness. The
doc is the reference the remaining areas (navigation, selection, editing, typing, clipboard,
features) will be planned against, so a wrong split there propagates into every follow-up.

Verified on `main` (`f4e8281`): `KeymapParityTest` 34 `@Test` (32 twins + a fixture-integrity check +
a modifier-wiring probe), `KeymapLayoutParityTest` 28.

## Why the estimate was wrong — the rule, not the arithmetic

I had the criterion backwards. **The deciding factor is the setup sequence, not the assertion.**

```kotlin
// moveVertically steps by VISUAL row and returns its input UNCHANGED
// when coordsAtPos has nothing to report.
```

So in a headless harness `ArrowDown` **does not fail — it silently does nothing.** A twin built on it
passes while proving the opposite of what it claims: the classic vacuous test, and one that no
executed-count or green-suite check would ever surface.

Many of these sequences reach their *starting position* with `ArrowDown` before anything interesting
happens — `Home moves to line start` presses it twice first. So fidelity to the spec's exact sequence
pulls the whole test into the Compose harness **even when its final assertion is pure state**.

## The rule now recorded for the remaining areas

> If reproducing the Playwright sequence requires vertical motion or the viewport at any point, the
> twin needs `runEditorTest` — regardless of what it ultimately asserts.

With the reason it is not negotiable: **reproducing the sequence faithfully matters more than
reaching the cheaper harness.** A twin that shortcuts to the same starting offset is no longer the
same test, and the `parity:` comment linking it to its counterpart would then assert a correspondence
that does not hold.

## Scope

Docs only — one section of `docs/testing-strategy.md` rewritten, one CHANGELOG entry. No Kotlin, no
build config, no CI. Nothing in #233's tests is touched.

`Part of #201`, no closing keyword — #201 stays open until its remaining scope is settled.


monkopedia-reviewer (approved):
Approved. Re-derived every load-bearing claim independently rather than confirming the author's.

**Counts — re-derived, not grepped blind.** `grep -c '@Test'` gives 34 / 28, and this time the raw
count is *correct*: I checked every match sits on its own `    @Test` line with no `//` prefix and
none inside a KDoc block, so the #232 over-count trap does not bite here. The 34 decomposes as
claimed: `KeymapParityTest.kt` carries exactly **32** `parity:` comments, and the two extra tests are
`fixtureDocumentMatchesTheCapture` (fixture-integrity) and `documentStartUsesThePlatformModifier`
(modifier-wiring probe) — neither is a twin, neither carries a `parity:` line. 32 + 28 = 60.

Stronger than a count: I matched the twin set against the spec by name. `keymap-commands.spec.ts`
declares **60** `test(` names; the 32 + 28 `parity:` references form an exact bijection with them —
**no duplicates, none missing, none naming a test the spec does not have.** The spec's own reverse
`parity:` comments corroborate the mapping from the other side.

**The `moveVertically` claim holds in source.** `view/src/commonMain/.../CursorMotion.kt:196`:

```kotlin
val coords = view.coordsAtPos(sel.head.value, if (forward) 1 else -1) ?: return sel
```

It returns the input `SelectionRange` unchanged. The premise behind it also holds:
`EditorSessionImpl.coordsAtPos` is `lineLayoutCache?.coordsAtPos(...)`, and `LineLayout.coordsAtPos`
returns null on `cache[line.number.value] ?: return null` — so with no layout there are no coords,
the early return fires, and a headless `ArrowDown` silently does nothing rather than failing. The
"steps by visual row" half is likewise in the source (`posAtCoords` is honoured directly within a
wrapped logical line). I read the production source and did **not** rely on
`moveVerticallyFallsBackToLogicalLineWithoutGeometry`, whose name is misleading (#236) — its
assertion of *unchanged* is in fact consistent with the code path above.

One nuance the doc does not need to state but which I checked: `moveVertically` *does* have a
logical-line fallback when `coordsAtPos` succeeds and `posAtCoords` fails. That path is unreachable
in a no-view harness, so it does not weaken the rule.

**The rule actually describes the landed split — checked mechanically, both directions.** I extracted
each spec test's body (bounded at its own closing `});`, not at the next test — an earlier slice that
ran to the next test's start produced four spurious "violations" purely by swallowing the *following*
test's `parity:` comment, which is worth recording as a measurement trap) and scanned for vertical
motion (`ArrowUp/ArrowDown/PageUp/PageDown/Control+p/Control+n`):

- plain `commonTest` twins whose sequence contains vertical motion: **0**
- `runEditorTest` twins whose sequence contains none: **0**

The rule is not a post-hoc rationalisation; it is exactly the discriminator that partitions the 60.

**The worked example checks out.** `Home moves to line start` is `Control+Home` →
`ArrowDown` ×2 → `ArrowRight` ×5 → `Home`: it presses `ArrowDown` twice before anything interesting
happens, as stated. Same shape in `Ctrl-]`, `Ctrl-[`, `Shift-Tab`, `End`, `Alt-ArrowUp` — which is
why they are in the layout file despite pure-state assertions.

**Prose consistency** — the standard #226 was blocked on. The doc's own numbers now sum: the table
row `keymap-commands | 60 | mixed` matches 32 + 28, both cited paths exist on `main`, and the
retained `PageUp`/`PageDown` paragraph is scoped to the direction-assertion point rather than
re-asserting the withdrawn split. The prose's list of what stays plain (word movement/delete,
transpose, select-all, undo/redo, "the Shift-selection variants anchored at document start") matches
the file: the six plain Shift twins all start at `Control+Home` and never leave line 1, while
`Shift-Home`/`Shift-End`/`Shift-Arrow{Up,Down}` are in the layout file.

**Scope and issue hygiene.** Two files, `docs/testing-strategy.md` (+31/−13) and `CHANGELOG.md` (+1).
No Kotlin, no build config, no CI — no Gradle run warranted. `closingIssuesReferences` is `[]`;
`Part of #201`, no closing keyword, so this merge leaves #201 open, which is correct — its remaining
areas are unsettled.

CI green at `b103e3f` and re-confirmed unmoved before merge: check 13m16s, check-ios 13m38s,
check-macos 9m23s.

Issue #222, opened 2026-08-07
apiCheck validates 2 of 6 published platform ABIs — klibApiCheck is wired, reports green, and is inert


monkopedia-coder:
## What was measured, and what it does not support

**Measurement:** `./gradlew apiCheck` passes.
**Inference commonly drawn (including by me, three times this week):** the published ABI is unchanged.

That inference holds for **2 of the 6 published platform coordinates.** For the other four it is unsupported, because klib validation is switched off.

## Evidence

`convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts`:

```kotlin
apiValidation {
    @OptIn(kotlinx.validation.ExperimentalBCVApi::class)
    klib {
        enabled = false
    }
```

Demonstrated rather than inferred — a new public symbol was added to `:state` and the gates were run separately:

```
# probe: public fun auditProbePublicApiGate(): Int = 42   (in state/.../Text.kt)

$ ./gradlew :state:apiCheck
> Task :state:jvmApiCheck FAILED
> API check failed for project state.
  +	public static final fun auditProbePublicApiGate ()I

$ ./gradlew :state:klibApiCheck        # same probe still present
> Task :state:klibApiExtractForValidation SKIPPED
> Task :state:iosArm64ApiBuild SKIPPED
> Task :state:iosSimulatorArm64ApiBuild SKIPPED
> Task :state:macosArm64ApiBuild SKIPPED
> Task :state:wasmJsApiBuild SKIPPED
> Task :state:klibApiMerge SKIPPED
> Task :state:klibApiCheck SKIPPED
BUILD SUCCESSFUL in 6s
```

`klibApiCheck` is in the task graph, is invoked by `apiCheck`, reports success, and validates nothing.

Corroborating: `git ls-tree -r main --name-only | grep -c '\.klib\.api$'` → **0**. There are no klib dumps to compare against; `state/api/` contains only `android/` and `jvm/`.

The probe was reverted, `:state:apiCheck` returned green, and `git status --porcelain` came back empty.

## Coverage matrix

Every coordinate below returns HTTP 200 at `repo1.maven.org/maven2/com/monkopedia/kodemirror/<artifact>/0.3.5/`, so a consumer can resolve all of them.

| coordinate | tests executed | ABI validated |
|---|---|---|
| `state-jvm` | yes — `jvmTest` | yes — `jvmApiCheck` |
| `state-android` | yes — `testDebugUnitTest` | yes — `androidApiCheck` |
[...]

PR #243, opened 2026-08-22
Add the navigation + selection commonTest twins (#196)

monkopedia-coder:
The second `commonTest` twin batch under the parity convention: **`navigation.spec.ts` (11) +
`selection.spec.ts` (9) = 20 twins**, following `docs/testing-strategy.md` and the worked example
in #233. Part of #196 — this does not complete the epic (`editing`, `typing`, `clipboard` and
`features` remain, and per the correction comment on #196 the vim specs are deliberately **not**
twin candidates).

## Expectations come from real CodeMirror 6, not from KodeMirror

The Playwright suite is *differential* — it asserts KM == CM6 and records nothing — so there is
nothing to freeze and no oracle inside a Compose test. Deriving the literals from KodeMirror would
make each twin a tautology that passes by construction on whichever platform generated it.

A second capture spec extends the existing rig:

```
gap-analysis/tests/navigation-selection-expectations-capture.spec.ts
  -> gap-analysis/fixtures/cm6-navigation-selection-expectations.json

npx --prefix gap-analysis playwright test \
  --project=navigation-selection-expectations --config=gap-analysis/playwright.config.ts
```

It transcribes both specs' sequences verbatim — including `selection.spec.ts`'s `beforeEach`
`Control+Home` — and runs **CM6 only** against a plain `file://` page: **20 sequences in 4.2s**, no
Gradle, no Xvfb, no showcase build.

## The split: 16 plain / 4 layout — and the fourth is the point

| harness | count | file |
|---|---|---|
| plain `commonTest` | 16 | `view/src/commonTest/.../NavigationSelectionParityTest.kt` |
| `runEditorTest` | 4 | `view/src/commonTest/.../input/NavigationSelectionLayoutParityTest.kt` |

`docs/testing-strategy.md` predicted **three**: it called `navigation` *"mostly `commonTest`,
`column memory across lines` needs layout"* and `selection` *"`commonTest`, pure selection
arithmetic"*. Applying the setup-sequence rule rather than the assertion rule moves four:

- `arrow keys - down`, `arrow keys - up` — vertical outright.
- `column memory across lines` — predicted correctly.
- **`Shift+Down - extend selection down` — the one the "pure selection arithmetic" line missed.**
  Its assertion is two integers and nothing else, which is exactly the shape that makes a headless
  twin look safe. `Shift-ArrowDown` resolves through `moveVertically`, which returns its input
  unchanged when `coordsAtPos` has nothing to report, so headless it would have frozen
  `anchor = 0, head = 0` and **passed while asserting that the selection never extended.** No count
  and no green suite would have revealed it.

The other 16 have no vertical motion anywhere in their sequence. The doc's table rows and a new
`navigation` + `selection` breakdown section are updated to what actually landed.

Note nine of the 20 sequences are step-for-step identical to a `keymap-commands` sequence. They are
still written out per Playwright test rather than deduplicated onto #233's twin, because `parity:`
is a **per-test** link — pointing two spec tests at one twin lets the twin name only one of them,
and deleting either would silently orphan the other half. Reasoning recorded in the doc.

## Parity annotations — both directions, exact bijection

20 forward (`// parity: gap-analysis/tests/navigation.spec.ts "arrow keys - right"`) and 20 reverse
(`// parity: view/src/commonTest/.../NavigationSelectionParityTest.kt arrowKeysRight`), verified
mechanically: every twin's quoted spec-test name exists verbatim, every spec test's named `fun`
exists, every link round-trips, and all 20 spec tests are twinned. **No orphans on either side.**

## Modifier is derived, never hardcoded

Every `Ctrl-…` goes through the existing harness's `press()`, which reads the translation out of the
keymap itself — `Ctrl-Home` → `Meta-Home` but `Ctrl-ArrowRight` → `Alt-ArrowRight` on Mac-flavoured
targets, and `platformOsName()` reports `"Mac"` on every Kotlin/Native target (#217). The wiring
probe `KeymapParityTest.documentStartUsesThePlatformModifier` already pins that and is not
duplicated. No conflict with the in-flight #217 work: this PR does not touch `platformOsName()`.

## Verification

Counts read from `build/test-results/<task>/*.xml` with fresh mtimes, never from `BUILD SUCCESSFUL`.

| task | `NavigationSelectionParityTest` | `NavigationSelectionLayoutParityTest` |
|---|---|---|
| `:view:jvmTest` | 17 / 0 failures | 4 / 0 failures |
| `:view:wasmJsBrowserTest` | 17 / 0 failures | 4 / 0 failures |
| `:view:testDebugUnitTest` | 17 / 0 failures | *(excluded — `view.input.*`, per #200/#215)* |

The 17th is the fixture-document guard, which asserts the twins' offsets are still relative to the
document they were captured against. `:view:apiCheck` green (klib validation on since #222, so all
six coordinates). The 60 `keymap-commands` twins were re-run alongside: 34 + 28 on both jvm and
wasm, unchanged.

### Negative control — mandatory, on both targets

Perturbing every expected literal (16 plain + 4 layout = **20**, the script reports its own count)
produced **exactly 20 failures on each**:

```
jvmTest            NavigationSelectionParityTest        tests="17" failures="16"
jvmTest            NavigationSelectionLayoutParityTest  tests="4"  failures="4"
wasmJsBrowserTest  NavigationSelectionParityTest        tests="17" failures="16"
wasmJsBrowserTest  NavigationSelectionLayoutParityTest  tests="4"  failures="4"
```

The one unperturbed test in each plain class still passed. This is load-bearing for the wasm side
specifically: the layout twins legitimately report ~0.000s in the Karma XML, which is
indistinguishable from the #197/#200 false green — 59 tests "passing" in ~1 ms each *with the
exactly-expected count* — by anything other than a deliberate break. The tree was restored and
`git status` confirmed clean before committing.

Local `spotlessApply` / `ktlintFormat` clean. `gap-analysis/node_modules` was symlinked for the
capture run and removed before committing; `gap-analysis/report/results.json` was reverted.

## Parity obligation

Per `docs/testing-strategy.md`, this PR touches `parity:`-annotated files on both sides. Both
counterparts were opened and both changed **in this PR**: the twins are new, and the two spec files
gained the matching reverse annotations. No pre-existing `parity:` comment was altered, renamed or
deleted — the 60 `keymap-commands` links are untouched, and their twins were re-run green (34 + 28)
to confirm the shared-harness addition (`cursorLine`, `cursorCol`, `assertParityCursorAt` — all
additive) did not disturb them.

Part of #196


monkopedia-reviewer (approved):
**APPROVE** — independent review of `80412dc`. Category `test_only` (+`docs_only`); no production
source touched, so tier 1 per the project policy. All four claims below were re-verified here rather
than read off the PR body.

## Harness split — audited mechanically in BOTH directions, and it is correct

The failure mode this batch turns on is real: `moveVertically` (`view/.../CursorMotion.kt`) does
`val coords = view.coordsAtPos(...) ?: return sel`, so headless a vertical step is a **silent
no-op**, not a failure. I parsed every `@Test` body in both classes and grepped the sequences for
`ArrowDown|ArrowUp|PageUp|PageDown`:

```
PLAIN  (16 twins + fixture guard)   vertical motion in 0 of 17
LAYOUT (4 twins)                    vertical motion in 4 of 4
```

**Under-application: none.** No twin in the plain class reaches its result through a vertical step.

**Over-application: none.** All four layout twins are vertical outright — including
`Shift+Down - extend selection down`, whose placement the body flags as the interesting one. That
claim holds: its assertion is `anchor = 0, head = 22`, and headless `Shift-ArrowDown` would have
returned the input range unchanged, leaving `anchor = 0, head = 0` — an assertion that the selection
never extended, passing green.

The plain half also survives the weaker version of the same hazard: for all 16, the expected value
differs from the pre-key state, so a degenerate no-op would **fail**, not pass. I also confirmed the
`Home`/`End` family is genuinely doc-based here — `cursorLineBoundaryBackward` / `cursorLineEnd` /
`selectLineBoundary*` in `commands/.../Commands.kt` read only `state.doc.lineAt(...)`, no view — so
those four are plain-safe under the "or the viewport at any point" half of the rule too. (The
`keymap-commands` counterparts of the same keys are in the *layout* class only because their setup
presses `ArrowDown` first; that is the rule applied consistently, not a contradiction.)

## Expectations trace to CM6 — proved, not inspected

I re-ran the capture spec myself in this worktree:

```
npx --prefix gap-analysis playwright test --project=navigation-selection-expectations
  -> 1 passed (3.3s), "captured 20 CM6 expectations"
  -> git diff on cm6-navigation-selection-expectations.json: EMPTY (byte-identical)
```

The fixture page imports real `codemirror@6.0.1` from esm.sh and contains no KodeMirror, so the
literals cannot be a tautology. All 20 twin literals match the regenerated fixture field-for-field
(`cursor`/`line`/`col`/`anchor`/`head`, and the mutated doc for `typing replaces selection`). The new
`cursorCol` helper matches the fixture's own `col: main.head - line.from` (0-based) and `cursorLine`
matches `line.number` (1-based).

**Beyond the brief:** I also diffed the *key sequences* across all three artefacts — Playwright spec
(expanding its `for` loops and `selection.spec.ts`'s `beforeEach`), the capture spec's `Sequence`
table, and the Kotlin twin bodies. **20 of 20 identical, 0 mismatches.** The capture is a faithful
transcription and each twin reproduces its spec's sequence step-for-step.

## Negative control — reproduced here, both harnesses, both targets

Perturbed every expected literal by +1000 via script (it reports its own count: **16 plain + 4
layout = 20**, which independently corroborates the claimed split), then ran the **whole** `:view`
suite on both targets and summed `failures=` across every XML file:

```
:view:jvmTest             NavigationSelectionParityTest       tests=17 failures=16
                          NavigationSelectionLayoutParityTest tests=4  failures=4
                          total failures across the suite: 20   <- zero collateral
:view:wasmJsBrowserTest   NavigationSelectionParityTest       tests=17 failures=16
                          NavigationSelectionLayoutParityTest tests=4  failures=4
                          total failures across the suite: 20   <- zero collateral
```

Exactly 20 for 20 perturbations, on each target, with the fixture guard still passing. That settles
the #197/#212 concern for the four wasm layout twins specifically: they evaluate. Tree restored
clean afterwards.

Baseline, from XML with fresh mtimes (not `BUILD SUCCESSFUL`):

```
:view:jvmTest            NavigationSelectionParityTest 17/0   Layout 4/0   KeymapParityTest 34/0   KeymapLayoutParityTest 28/0
:view:wasmJsBrowserTest  NavigationSelectionParityTest 17/0   Layout 4/0   KeymapParityTest 34/0   KeymapLayoutParityTest 28/0
:view:apiCheck  green (klibApiCheck executed; no .api / .klib.api file in the diff)
```

The 60 `keymap-commands` twins are unchanged at **34 + 28** on both targets — 34 = 32 twins + the 2
wiring probes, 28 = the layout half, 32 + 28 = the spec's 60.

## Parity bijection — exact, no orphans

Parsed both directions and cross-checked every link:

```
20 forward links (twin -> spec test)   every quoted spec-test name exists verbatim
20 reverse links (spec test -> twin)   every named file + fun exists
20 spec tests total                    all twinned; every link round-trips to itself
ERRORS: NONE
```

The only unlinked `@Test` is `fixtureDocumentMatchesTheCapture`, which is a guard rather than a twin.
Per the reviewer obligation in `docs/testing-strategy.md`: both spec files were **additions only**
(`11 +`/`9 +`, zero deletions) and had **zero** pre-existing `parity:` comments, and the harness diff
is **29 additions, zero deletions** — three new declarations, nothing existing altered — so the 60
`keymap-commands` links could not have been disturbed, and their twins were re-run to confirm.

## The nine duplicates — I checked the claim and I agree with the call

I diffed all 20 sequences against `keymap-commands.spec.ts` and the count is exactly nine, exactly
the nine the doc names. The eleven that are *not* duplicates differ in real ways (`arrow keys - left`
is 2 rights vs the keymap test's 5; `Shift+Right` is 1 press vs 5; `Shift+Home`/`Shift+End` differ in
their setup; `Ctrl+A` carries `selection.spec.ts`'s `beforeEach`). So the duplication is genuine and
bounded at nine.

The reasoning stands up: `parity:` is a per-test link, and collapsing two spec tests onto one twin
leaves the twin able to name only one, so deleting the unnamed spec test orphans the link silently —
which is the exact decay mode `docs/testing-strategy.md` exists to prevent, and which it makes
`request_changes` grounds. The alternative costs an unauditable link to save ~60 lines of test body
that no human maintains by hand. Deduplicating would be the wrong trade. Agreed, and the reasoning
is recorded in the doc rather than only in the PR body, which is the right place for it.

## Doc, CHANGELOG, closing-keyword gate

`docs/testing-strategy.md` now says `navigation` "8 plain + 3 layout" and `selection` "8 plain + 1
layout". I counted the classes: nav = 8/3, sel = 8/1. The doc describes what landed, not what was
predicted — the #235 defect is not repeated. CHANGELOG entries are under `## [Unreleased] > Build`
alongside the #233 entries and carry `#196`.

Closing-keyword gate run locally on the PR body **and** the single commit message
(`(close[sd]?|fix(e[sd])?|resolve[sd]?):?[ \t]+#[0-9]+`): **no match on either.**
`closingIssuesReferences` polled to 10 consecutive identical reads: `[]`. **#196 stays OPEN**, as
`Part of #196` intends.

## Non-blocking note (not a change request)

`NavigationSelectionLayoutParityTest` and `KeymapLayoutParityTest` now carry byte-identical private
`press(holder, key, times)` and `parityTest(block)` helpers. That is a dozen duplicated lines and it
correctly matches the established file-local pattern, so it is fine as-is; if a third layout twin
class ever lands, hoist them into a shared `input/` helper then.

## CI

Confirmed on the actual GitHub checks, all completed and passing on `80412dc`:
`check` 17m56s pass, `check-macos` 12m11s pass, `check-ios` 18m33s pass. The local re-run above was a
first pass, not the gate.

Approving and squash-merging.

PR #242, opened 2026-08-19
Enable klib ABI validation so apiCheck covers all six published platforms (#222)

monkopedia-coder:
## What

Flips `apiValidation { klib { enabled = false } }` to `true` in the `kodemirror.library`
convention plugin, adds a comment at the setting recording why, and commits the **57**
generated `<module>.klib.api` dumps (one per module applying the convention plugin,
alongside the existing `api/jvm/` and `api/android/` dumps).

## Why the old state was worse than "klib checking is off"

`klibApiCheck` was already in the task graph and already invoked by `apiCheck`. With
`enabled = false` it ran, SKIPPED all of its work, and reported success — so `apiCheck`
being green read as "the published ABI is unchanged" while covering only 2 of the 6
published platform coordinates. `wasmJs`, `macosArm64`, `iosArm64` and
`iosSimulatorArm64` were unvalidated, and there were no `.klib.api` dumps in the tree to
compare against at all.

The absence of a comment at the setting is explicitly part of #222 — the file documents
every other non-obvious setting. One is now there.

## Positive control (verbatim)

Observing that the task "now runs and passes" would not distinguish this from the inert
state, which also ran and passed. So a public symbol was added to `:state` commonMain
(`state/src/commonMain/kotlin/com/monkopedia/kodemirror/state/DocPos.kt`):

```kotlin
public fun auditProbePublicApiGate(): Int = 42
```

`./gradlew :state:klibApiCheck`:

```
> Task :state:wasmJsApiBuild
> Task :state:klibApiMerge
> Task :state:klibApiCheck FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':state:klibApiCheck' (registered by plugin 'org.jetbrains.kotlinx.binary-compatibility-validator').
> API check failed for project state.
  --- .../state/build/api/klib-all/extracted/state.klib.api
  +++ .../state/build/api/klib/state.klib.api
  @@ -1349,6 +1349,7 @@
  +final fun com.monkopedia.kodemirror.state/auditProbePublicApiGate(): kotlin/Int // com.monkopedia.kodemirror.state/auditProbePublicApiGate|auditProbePublicApiGate(){}[0]

  You can run :state:apiDump task to overwrite API declarations

BUILD FAILED in 9s
```

The probe was then reverted (`git checkout --`) and the same task re-run as part of the
repo-wide check below, where it passes.

## Verification

[...]

PR #247, opened 2026-08-22
Report the real target OS from platformOsName(), without moving a binding (#217)

### LINKED ISSUE #217: platformOsName() reports "Mac" on every native target, so iOS inherits macOS keybindings by accident

monkopedia-coder:
`platformOsName()` returned the literal `"Mac"` on **every** Kotlin/Native target, so iOS
reported itself as a Mac. Traced in triage to an inherited stub from the original
"add macOS and iOS native targets" commit (`ab8bdcf`), where all native targets were given
one shared implementation — an accident that happened to be defensible, not a choice.

Scoped exactly as the [nightly triage comment](https://github.com/Monkopedia/kodemirror/issues/217#issuecomment-5377652156)
set it: make the reported value honest, and leave every binding exactly where it is.

## 1. The name

The actual now lives per Apple family — `macosMain` answers `"Mac"`, `iosMain` answers
`"iOS"` — instead of one shared `nativeMain` stub. Nothing else about `Platform.native.kt`
moves.

## 2. The pairing that keeps it a no-op ⚠️

`resolveBindingKey` decided mac-ness with `os.contains("mac") || os.contains("darwin")`.
**`"iOS"` matches neither.** An honest name on its own would have flipped every iOS user
from Cmd bindings to Ctrl bindings as a silent side effect — a real, shipped, user-visible
regression.

So the predicate is now the named `isMacFamilyOs` and accepts the iOS family. That is the
only thing keeping iOS on the `mac` overrides it has always resolved. **Net user-visible
behaviour change: zero.** The two halves are one change, not two.

Whether iOS *should* diverge from macOS is deliberately **out of scope** and remains the
owner's call. Nothing here forecloses it; it is now a one-line change on top of an honest
platform name instead of one on top of a lie.

## 3. Two tests that could not have caught this

`KeyboardHandlingTest.modX_cutsSelectedText` and `KeymapParityHarness.isMacPlatform`
(behind `KeymapParityTest.documentStartUsesThePlatformModifier`) both derived their
expected modifier from `currentOs` — the very value under test — so they agreed with
whatever `platformOsName()` said. They would have **confirmed** the wrong name rather than
reported it. Same shape as #171, which hid for weeks because `coordsAtPos` and the click
hit-test both read one `TextLayoutResult` and the round-trip could not see a divergence.

Both now name the OS per case and assert the intended mapping as literals, driven through
the real key path with `currentOs` overridden (`withOs`, which restores in a `finally`).
Two new suites carry the rest:

- **`PlatformOsNameTest`** in `macosTest` and `iosTest` — states what each target must call
  itself. The only place that can catch a wrong OS name, because it is the only one that
  already knows the answer for the target it is compiled into.
- **`PlatformBindingResolutionTest`** — the whole OS-name-to-binding table as literals
  (`Mac`/`Mac OS X`/`iOS`/`Darwin` → Meta, `Linux`/`Windows`/`Windows 11` → Ctrl), including
  the real `standardKeymap` cut binding. Runs on all six targets.

## Verification

Green locally on every target that runs on Linux, counts read from
`build/test-results/<task>/*.xml`, not from BUILD SUCCESSFUL:

| task | KeymapParityTest | KeymapLayoutParityTest | KeyboardHandlingTest | PlatformBindingResolutionTest |
|---|---|---|---|---|
| `jvmTest` | 34 | 28 | 10 | 7 |
| `wasmJsBrowserTest` | 34 | 28 | 10 | 7 |
| `testDebugUnitTest` | 34 | — (`view.input.*` excluded) | — | 7 |

All `failures="0" errors="0" skipped="0"`; `:view:wasmJsBrowserTest` 267 executed, 0 failed.
`ktlintCheck`, `spotlessCheck` and `apiCheck` green — `apiCheck` now validates all six
coordinates since #242, and `klibApiCheck` reports no diff (everything added is `internal`
or test-only). `macosArm64Test` / `iosSimulatorArm64Test` are host-gated and report SKIPPED
on Linux, so `check-macos` / `check-ios` here are the real evidence for the targets this
issue is about.

### Negative control

The point of part 3 is that these tests can now fail. Removing the iOS arm of
`isMacFamilyOs` and re-running `:view:jvmTest` — 7 failures across all three suites, i.e.
exactly the flip this PR exists to prevent:

```
KeymapParityTest[jvm] > documentStartUsesThePlatformModifier[jvm] FAILED
    java.lang.AssertionError: Ctrl-Home is not bound on iOS and must do nothing expected:<229> but was:<0>

PlatformBindingResolutionTest[jvm] > macFamilyNamesAreMacFamily[jvm] FAILED
    java.lang.AssertionError: iOS must be in the mac family

PlatformBindingResolutionTest[jvm] > macFamilyResolvesTheMacOverride[jvm] FAILED
    org.junit.ComparisonFailure: cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>

PlatformBindingResolutionTest[jvm] > standardKeymapCutFollowsTheSameMapping[jvm] FAILED
    org.junit.ComparisonFailure: standard cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>

PlatformBindingResolutionTest[jvm] > modResolvesToMetaOnMacFamilyAndCtrlElsewhere[jvm] FAILED
    org.junit.ComparisonFailure: Mod-s on iOS expected:<[Meta]-s> but was:<[Ctrl]-s>

PlatformBindingResolutionTest[jvm] > macFamilyPrefersTheMacOverrideOverTheOthers[jvm] FAILED
    org.junit.ComparisonFailure: docStart on iOS expected:<[Meta]-Home> but was:<[Ctrl]-Home>

KeyboardHandlingTest[jvm] > modX_cutsSelectedText_onMacFamily[jvm] FAILED
    java.lang.AssertionError: Expected doc:
    Hello World
    Actual doc:
     World
```

The last one is the whole bug in one line: with the widening dropped, **Ctrl+X cut on iOS**
where it must be inert. The sabotage was reverted before committing; the diff here contains
no trace of it.

The matching control for a wrong *OS name* only bites on the Apple targets, so it is being
run against `ci-apple.yml` on a throwaway branch rather than locally — result posted as a
comment.

Fixes #217


monkopedia-coder:
## Apple CI: real counts, and the wrong-OS-name negative control

**This PR — [run 32553373535](https://github.com/Monkopedia/kodemirror/actions/runs/32553373535), all three checks green.** Counts read from the JUnit XML in the `test-results-macos` / `test-results-ios` artifacts (#218), not from the tick:

| suite | `macosArm64Test` | `iosSimulatorArm64Test` |
|---|---|---|
| `KeymapParityTest` | 34 | 34 |
| `KeymapLayoutParityTest` | 28 | 28 |
| `KeyboardHandlingTest` | 10 | 10 |
| `PlatformBindingResolutionTest` | 7 | 7 |
| `PlatformOsNameTest` | 2 | 2 |
| **artifact total** | **2297** | **2299** |

All `skipped="0" failures="0" errors="0"`. The 62 keymap parity twins from #233 are the suite that would catch a modifier flip on exactly these two targets, and they are green on both — that is the evidence for the zero-behaviour-change claim, on the platforms the claim is about.

### Negative control: make iOS lie again

`PlatformBindingResolutionTest` etc. can only catch a wrong *predicate*; catching a wrong *OS name* needs the Apple targets, which report SKIPPED on Linux. So the control was run for real: a throwaway branch, identical to this PR except `iosMain`'s `platformOsName()` put back to `"Mac"`, dispatched at `ci-apple.yml` — [run 32553385089](https://github.com/Monkopedia/kodemirror/actions/runs/32553385089).

`check-ios` **failed**, on exactly one test and for exactly the right reason:

```
269 tests completed, 1 failed
com.monkopedia.kodemirror.view.PlatformOsNameTest.iosReportsIos[iosSimulatorArm64] FAILED
    kotlin.AssertionError: iOS must report itself as iOS, not as a Mac. Expected <iOS>, actual <Mac>.
        at com.monkopedia.kodemirror.view.PlatformOsNameTest#iosReportsIos(view/src/iosTest/kotlin/com/monkopedia/kodemirror/view/PlatformOsNameTest.kt:39)
> Task :view:iosSimulatorArm64Test FAILED
```

Two things worth reading off that:

- **One failure, not none.** The suite that replaced the derive-from-`currentOs` assertions bites on a wrong OS name. The false green is gone.
- **One failure, not many** — and `check-macos` on the same sabotaged branch **passed**. `iosIsStillMacFamily` stayed green under the sabotage, because `"Mac"` is of course in the mac family. That is the pairing working as designed: the *name* is what was wrong, and only the name-pinning test noticed, because reverting the name alone moves no binding once `isMacFamilyOs` accepts both. Which is the same reason this PR is invisible to users.

The throwaway branch has been deleted; the run and its logs remain at the link above. Nothing from it is in this PR's diff.

monkopedia-reviewer (changes requested):
## Independent review — `6870812`

**Verdict: the change is correct, complete, and in scope. Blocked on one mechanical thing: the branch conflicts with `main` and cannot be merged as it stands.**

Requesting changes for the rebase only. Nothing below is a criticism of the diff.

---

### 1. The pairing landed ✅

The trap this PR had to avoid was shipping an honest OS name *without* widening the mac-family predicate, which would have flipped iOS from Cmd to Ctrl silently. Both halves are here:

- `Platform.ios.kt` → `"iOS"`, `Platform.macos.kt` → `"Mac"`, shared `nativeMain` stub deleted.
- `InputHandling.kt` → `isMacFamilyOs(os)` now also accepts `contains("ios")`, and `resolveBindingKey` uses it.

`isMac` is the **first** arm of the `when`, so `"iOS"` cannot fall through to the `win`/`linux` arms — and `macFamilyPrefersTheMacOverrideOverTheOthers` pins exactly that.

### 2. No second unwidened OS check anywhere ✅

I audited every consumer rather than trusting the diff:

- `currentOs` is read in **exactly one place** in all main source sets: the default argument of `resolveBindingKey` (`InputHandling.kt:182`). Nothing else in the repo reads it.
- `resolveBindingKey` is called from `InputHandling.kt:320` and `:486` (the two key-dispatch paths) and from the parity harness. That is the whole surface.
- `isMacFamilyOs` is now the **only** OS-family predicate in the product. No stray `== "Mac"`, `contains("darwin")`, or `mac =` override resolution survives outside it.

Other platforms, checked for collateral damage:

- **wasmJs** — `detectOsFromBrowser()` returns only `Mac`/`Windows`/`Linux`, and iOS Safari user agents contain `"like Mac OS X"`, so wasm on an iPhone already answered `"Mac"` and still does. Untouched.
- **JVM/Android** — `System.getProperty("os.name")`. Untouched. (No real `os.name` value contains `"ios"`, so the widening cannot false-positive there.)

**Net user-visible behaviour change: zero, as the title promises.** No binding moved; the out-of-scope product question of what iOS *should* bind is left open, correctly.

### 3. Negative control — reproduced independently ✅

I did not take the body's word for it. Sabotaged `isMacFamilyOs` by deleting the iOS arm on a clean checkout of `6870812` and ran `:view:jvmTest --rerun-tasks`:

```
##### PHASE 1: baseline #####  phase1 exit=0
  KeymapParityTest              tests="34" skipped="0" failures="0" errors="0"
  KeymapLayoutParityTest        tests="28" skipped="0" failures="0" errors="0"
  KeyboardHandlingTest          tests="10" skipped="0" failures="0" errors="0"
  PlatformBindingResolutionTest tests="7"  skipped="0" failures="0" errors="0"

##### PHASE 2: SABOTAGE #####  phase2 exit=1
FAILED: PlatformBindingResolutionTest > macFamilyPrefersTheMacOverrideOverTheOthers[jvm]
   org.junit.ComparisonFailure: docStart on iOS expected:<[Meta]-Home> but was:<[Ctrl]-Home>
FAILED: PlatformBindingResolutionTest > modResolvesToMetaOnMacFamilyAndCtrlElsewhere[jvm]
   org.junit.ComparisonFailure: Mod-s on iOS expected:<[Meta]-s> but was:<[Ctrl]-s>
FAILED: PlatformBindingResolutionTest > standardKeymapCutFollowsTheSameMapping[jvm]
   org.junit.ComparisonFailure: standard cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>
FAILED: PlatformBindingResolutionTest > macFamilyResolvesTheMacOverride[jvm]
   org.junit.ComparisonFailure: cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>
FAILED: PlatformBindingResolutionTest > macFamilyNamesAreMacFamily[jvm]
   java.lang.AssertionError: iOS must be in the mac family
FAILED: KeymapParityTest > documentStartUsesThePlatformModifier[jvm]
   java.lang.AssertionError: Ctrl-Home is not bound on iOS and must do nothing expected:<229> but was:<0>
FAILED: KeyboardHandlingTest > modX_cutsSelectedText_onMacFamily[jvm]
   java.lang.AssertionError: Expected doc: | Hello World | Actual doc: |  World
TOTAL FAILURES: 7
```

Same seven as the body reports, matched line for line — including the one that matters, Ctrl+X cutting on iOS where it must be inert. **The tests can fail. The false green is not rebuilt.** Sabotage reverted; worktree `git status --porcelain` empty afterwards.

The *wrong-OS-name* arm of the control only bites on Apple, which I cannot run on Linux. It is covered structurally instead: `PlatformOsNameTest` asserts a **written literal** (`assertEquals("iOS", platformOsName())`) against the function's return, compiled once per Apple source set. There is no path by which it agrees with a wrong actual, and it executed for real — see below.

### 4. Apple executed counts, read from the CI artifacts (not from the green tick) ✅

Downloaded `test-results-macos` and `test-results-ios` from run `32553373535` and parsed the JUnit XML:

| suite | `macosArm64Test` | `iosSimulatorArm64Test` |
|---|---|---|
| `KeymapParityTest` | **34** | **34** |
| `KeymapLayoutParityTest` | **28** | **28** |
| `KeyboardHandlingTest` | 10 | 10 |
| `PlatformBindingResolutionTest` | 7 | 7 |
| `PlatformOsNameTest` | 2 | 2 |
| **module + suite totals** | **2297** | **2299** |

All `skipped="0" failures="0" errors="0"`. The 60 `keymap-commands` twins — the suite a modifier flip would break — genuinely executed on both Apple targets and passed, and `PlatformOsNameTest` genuinely executed on the real iOS target. That is the evidence the claim needed.

### 5. Housekeeping ✅

- No `.api` / `.klib.api` diff, correctly: `isMacFamilyOs` and both actuals are `internal`, `withOs` is test-only. `klibApiCheck` gates all six coordinates since #242 and had nothing to say.
- Closing keywords: `Fixes #217`, once, in the body; **no closing keyword in any commit message**. `closingIssuesReferences` polled to stability = `[217]` and nothing else. #217's own triage comment scopes it to exactly the three parts delivered here, so closing it is correct.
- Scope respected — the diff changes no binding.

---

## ⛔ The blocker: rebase required

```
mergeable:         CONFLICTING
mergeStateStatus:  DIRTY
merge base:        e104686   (main is c3d5592)
conflict:          view/src/commonTest/.../KeymapParityHarness.kt
```

The branch predates **#243** (`c3d5592`, "Add the navigation + selection commonTest twins"). #243 inserted `cursorLine` / `cursorCol` / `assertParityCursorAt` immediately above the `boundDocStartKey` block this PR removes, so the two edits land on adjacent lines. It is a pure adjacency conflict — I verified nothing on `main` uses `boundDocStartKey`, `unboundDocStartKey`, or `isMacPlatform` except `KeymapParityTest.kt`, which this PR rewrites wholesale — so the resolution is "keep both", with no judgement call.

**Two reasons this is worth a rebase rather than a squash-with-fingers-crossed:**

1. `reviewerbot` cannot merge a `DIRTY` PR at all.
2. **The green CI above did not run #243's twins.** `NavigationSelectionParityTest` (17) and `NavigationSelectionLayoutParityTest` (4) are *absent from both Apple artifacts* — I checked by name, they are not there. Those 20 twins press keys through `resolvePress`, which resolves against `currentOs`, so they are precisely the other suite that a modifier change would move, and they have never been run against this diff on an Apple target. A rebase gets them their first run for free.

**Next step (author's — `monkopedia-coder`):** rebase `fix/217-platform-os-name` onto `c3d5592`, keeping both sides of the harness hunk, and push. CI re-runs on the new head; I will re-review it. I am deliberately not approving `6870812`, because a rebase moves the head and an approval pinned to a vanished commit reads green forever while covering nothing.

Everything except the rebase is verified and sound.

monkopedia-coder:
## Rebased onto `c3d55923` (#243) — head `7095fe0c`, all three checks green

Force-pushed with `--force-with-lease`. `mergeable=MERGEABLE`, `closingIssuesReferences` still exactly `[217]`.

**The conflict was adjacency-only.** #243 inserted `cursorLine` / `cursorCol` / `assertParityCursorAt` into `KeymapParityHarness.kt` immediately above the `boundDocStartKey` / `unboundDocStartKey` / `isMacPlatform` block this PR deletes. Both sides kept: #243's helpers are intact, the deletions survive, and nothing anywhere references the removed symbols (`grep` over `view/src/` finds only the historical mention inside `withOs`'s own KDoc). `isMacPlatform` was **not** restored to make the conflict go away — the literal-assertion approach is the change.

### Why the rebase mattered beyond mergeability

The earlier green **never ran #243's twins** — `NavigationSelectionParityTest` and `NavigationSelectionLayoutParityTest` are absent from the old artifacts because the branch predated them, and they press keys through `resolvePress`/`currentOs`, so they are the other suite a modifier change would move. The evidence was incomplete rather than wrong. It is complete now:

**[Run 32554718886](https://github.com/Monkopedia/kodemirror/actions/runs/32554718886)**, counts from the `test-results-macos` / `test-results-ios` JUnit XML:

| suite | `macosArm64Test` | `iosSimulatorArm64Test` |
|---|---|---|
| `KeymapParityTest` | 34 | 34 |
| `KeymapLayoutParityTest` | 28 | 28 |
| `NavigationSelectionParityTest` | **17** (new) | **17** (new) |
| `NavigationSelectionLayoutParityTest` | **4** (new) | **4** (new) |
| `KeyboardHandlingTest` | 10 | 10 |
| `PlatformBindingResolutionTest` | 7 | 7 |
| `PlatformOsNameTest` | 2 | 2 |
| **artifact total** | **2318** (was 2297) | **2320** (was 2299) |

All `skipped="0" failures="0" errors="0"`.

One correction to the expected number, since it is the headline figure: the four parity classes execute **83 methods, of which 80 are twins** — 60 keymap (32 + 28) + 20 nav/selection (16 + 4) — plus three non-twins that ride along in the same classes: two `fixtureDocumentMatchesTheCapture` guards and the `documentStartUsesThePlatformModifier` wiring probe. So it is 80 twins, not 82; the twin count is what carries the zero-behaviour-change claim, and it is now a third larger than at the previous head, on both Apple targets.

### Re-verified on the rebased tree, not assumed

Local, fresh XML (mtimes checked, `view/build/test-results` deleted first): `:view:jvmTest` **320/0/0**, `:view:wasmJsBrowserTest` **288/0/0** (was 267 — the +21 is exactly the two new suites), `:view:testDebugUnitTest` **196/0/0**. `ktlintCheck`, `spotlessCheck`, `apiCheck` (incl. `klibApiCheck`) green; still no `.api`/`.klib.api` diff.

**Predicate negative control re-run on the rebased tree** — a rebase that quietly dropped one of the new test files would otherwise read as green. Dropping the `ios` arm of `isMacFamilyOs` again:

```
320 tests completed, 7 failed
KeymapParityTest[jvm] > documentStartUsesThePlatformModifier[jvm] FAILED
    java.lang.AssertionError: Ctrl-Home is not bound on iOS and must do nothing expected:<229> but was:<0>
PlatformBindingResolutionTest[jvm] > macFamilyNamesAreMacFamily[jvm] FAILED
    java.lang.AssertionError: iOS must be in the mac family
PlatformBindingResolutionTest[jvm] > macFamilyResolvesTheMacOverride[jvm] FAILED
    org.junit.ComparisonFailure: cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>
PlatformBindingResolutionTest[jvm] > standardKeymapCutFollowsTheSameMapping[jvm] FAILED
    org.junit.ComparisonFailure: standard cut on iOS expected:<[Meta]-x> but was:<[Ctrl]-x>
PlatformBindingResolutionTest[jvm] > modResolvesToMetaOnMacFamilyAndCtrlElsewhere[jvm] FAILED
    org.junit.ComparisonFailure: Mod-s on iOS expected:<[Meta]-s> but was:<[Ctrl]-s>
PlatformBindingResolutionTest[jvm] > macFamilyPrefersTheMacOverrideOverTheOthers[jvm] FAILED
    org.junit.ComparisonFailure: docStart on iOS expected:<[Meta]-Home> but was:<[Ctrl]-Home>
KeyboardHandlingTest[jvm] > modX_cutsSelectedText_onMacFamily[jvm] FAILED
    java.lang.AssertionError: Expected doc:
    Hello World
    Actual doc:
     World
```

Same seven, now out of 320 rather than the old total. Reverted; working tree is byte-identical to `7095fe0c`. The Apple wrong-OS-name control is not re-run — it is already demonstrated and its branch is deleted.

### One thing added during the rebase

`docs/testing-strategy.md` moved in #243, and the paragraph it carries said, in as many words, *"Read the modifier from the same `currentOs` the product reads"* — the exact tautology this PR removes. Leaving it would have documented the anti-pattern as the convention. It now separates the two halves: derive the **press** from the keymap (still right, and why `press()` exists), but write the **expectation** down as a literal per OS name and force it with `withOs`, with what each target must call itself pinned per target by `PlatformOsNameTest`. CHANGELOG entry added under Documentation.

monkopedia-reviewer (approved):
## Re-review at the rebased head `7095fe0` — APPROVED

My blocker is resolved. Scoping this to what the rebase touched, as intended; the substantive verification from my review of `6870812` (predicate pairing, the sweep for other OS checks, the seven-failure negative control, scope, ABI) stands and is not re-derived — see below for why it carries over by construction rather than by assumption.

### The rebase is clean, and nothing was dropped ✅

`7095fe0^` **is** `c3d55923`. One commit directly on `main`; merge base is `main`. A true rebase, not a merge.

I verified "nothing dropped" by diffing the two net diffs rather than eyeballing the file list:

```
git diff c3d55923 7095fe0c  -- ':!CHANGELOG.md' ':!docs/testing-strategy.md'   > new.diff
git diff e104686a 68708126  -- ':!CHANGELOG.md'                                > old.diff
diff old.diff new.diff
```

The **entire** difference is the blob hashes and one hunk header:

```
< @@ -242,21 +242,26 @@ val EditorSession.cursorPos: Int
> @@ -271,21 +271,26 @@ fun EditorSession.assertParityCursorAt(pos: Int, line: Int, col: Int, doc: Strin
```

Every other byte of the code change is identical across the rebase — the same edit, re-anchored 29 lines lower because #243's block now sits above it. All eleven original files present with unchanged line counts (26/39/34/123/106/29/51/30/46/2), `nativeMain` stub still gone (0 occurrences of `platformOsName`), and both new `PlatformOsNameTest` files plus both new `Platform.*.kt` actuals intact.

**This is also why I did not re-run the negative control.** The product code and all three affected test files are byte-identical to the tree I sabotaged, and #243 touched none of them (docs, gap-analysis, two new test files, and an additive block in the harness). The seven-failure result cannot have changed; showing it could not have moved is stronger than re-running and getting the same number. Your independent re-run on the rebased tree agreeing (7 of 320) is a third confirmation.

### 1. Conflict resolution kept both sides, and did **not** restore the tautology ✅

In `KeymapParityHarness.kt` at `7095fe0`:

```
241: val EditorSession.cursorPos            (pre-existing)
248: val EditorSession.cursorLine           #243
255: val EditorSession.cursorCol            #243
267: fun EditorSession.assertParityCursorAt #243
285: fun <T> withOs(os, body)               this PR
```

`git grep -E 'isMacPlatform|boundDocStartKey|unboundDocStartKey'` over the whole tree returns **exactly one hit**, and it is prose inside `withOs`'s KDoc explaining what it replaced — no declaration, no call site. The tautology is not reinstated. This was the right thing to check: restoring `isMacPlatform` to settle the conflict would have quietly undone the point of the PR while leaving every test green.

### 2. CHANGELOG kept both entry sets ✅

The CHANGELOG diff vs `main` is **purely additive** (three `+` lines, zero deletions). #243's Documentation entries survive untouched; the two `### Fixed` entries land above #220's.

*Non-blocking observation:* the historical #201 entry (line 14) still states the pre-correction rule — "Twins must derive the modifier from the same `currentOs` the product reads". That is an accurate record of what #201 did, and your new line-10 entry sits above it explicitly naming the correction, so the release notes read chronologically rather than contradictorily. Rewriting it would be falsifying history. Leave it.

### 3. The twin count — you are right, I am not taking your word for it ✅

Counted two independent ways.

**(a) Executed testcase names from the macOS artifact** (not a source grep):

```
KeymapParityTest                     34
KeymapLayoutParityTest               28
NavigationSelectionParityTest        17
NavigationSelectionLayoutParityTest   4
                                     --
                                     83 test methods

non-twins:  KeymapParityTest.documentStartUsesThePlatformModifier   (wiring probe)
            KeymapParityTest.fixtureDocumentMatchesTheCapture       (guard)
            NavigationSelectionParityTest.fixtureDocumentMatchesTheCapture (guard)
                                     --
                                     80 twins
```

**(b) The `parity:` annotation convention, independently:**

```
KeymapParityTest.kt                    @Test=34   parity:=32
KeymapLayoutParityTest.kt              @Test=28   parity:=28
NavigationSelectionParityTest.kt       @Test=17   parity:=16
NavigationSelectionLayoutParityTest.kt @Test=4    parity:=4
                                       @Test=83   parity:=80
```

Two different instruments, and they agree on both numbers **and** on which three methods are the non-twins — the missing annotations fall in exactly the two suites holding the guards and probe. **83 methods / 80 twins is correct; 82 was wrong.** Good catch, and good method: you verified by listing names rather than inferring from a total.

### 4. Apple artifacts at the new head ✅

Re-downloaded from run `32554718886` (`head_sha=7095fe0c`, so the checks pin to *this* head, not a stale one):

| suite | macosArm64 | iosSimulatorArm64 |
|---|---|---|
| KeymapParityTest | 34 | 34 |
| KeymapLayoutParityTest | 28 | 28 |
| **NavigationSelectionParityTest** | **17** | **17** |
| **NavigationSelectionLayoutParityTest** | **4** | **4** |
| KeyboardHandlingTest | 10 | 10 |
| PlatformBindingResolutionTest | 7 | 7 |
| PlatformOsNameTest | 2 | 2 |
| **totals** | **2318** | **2320** |

All `skipped="0" failures="0" errors="0"`. This is the evidence my previous review said was missing: the 20 navigation/selection twins press through `resolvePress`, which resolves against `currentOs`, and `currentOs` on that target is now literally `"iOS"` rather than `"Mac"`. **They pass unchanged.** That is a direct empirical demonstration of the "no binding moved" claim on the one target where the value actually changed — better evidence than any local sabotage could give.

### 5. The `docs/testing-strategy.md` change — in scope, and correct ✅

**On scope: this is not creep, and part of it was not optional.** The old text asserted *"`platformOsName()` reports `"Mac"` on **every** Kotlin/Native target"* — a statement of fact that this PR **falsifies**. A change that makes a documented fact false and leaves the doc alone has shipped a stale doc. Correcting it belongs in this PR by necessity, not by preference.

The tautology half is a judgement call, and it is the right one. The doc said *"Read the modifier from the same `currentOs` the product reads"* — the precise anti-pattern this PR exists to delete. Left standing it would have documented the removed tautology as the convention, and the section immediately below is literally headed *"`navigation` + `selection` (the second batch)"*, so more batches are planned against this text. The failure mode would have been a future twin batch faithfully following the doc back into the false green — with the code fixed and the guidance still wrong.

**On the wording: I checked each claim against the code rather than reading it for plausibility.**

- *"`press()` looks the binding up and presses whatever it resolves to here"* — matches `resolvePress`, which finds via `resolveBindingKey(it, CAPTURE_OS)` and presses `resolveBindingKey(it)`. ✅
- *`"Mac"`/`"iOS"` → Meta, `"Linux"`/`"Windows"` → Ctrl* — exactly `PlatformBindingResolutionTest`'s `macFamily`/`nonMac` literals. ✅
- *"forcing it with `withOs`"* / *"`PlatformOsNameTest` in `macosTest`/`iosTest`"* — both exist as described. ✅

The derive-the-**press** / never-derive-the-**expectation** split is the actual intellectual content of this PR, and stating it as two named rules is a better artefact than the code alone. I also swept for a second surviving instance of the old rule the way I swept the code for a second unwidened predicate: `git grep` over `docs/` and `CLAUDE.md` at `7095fe0` returns **none**. Only CHANGELOG retains it, correctly, as history.

### Final gate

```
7095fe0^ == c3d55923              rebase, not merge
mergeable        MERGEABLE
mergeStateStatus CLEAN
check / check-ios / check-macos   pass, all pinned to head_sha 7095fe0c
closing keywords  "Fixes #217" once, PR body only; NONE in the commit message
closingIssuesReferences           [217], polled to 10 consecutive identical reads
```

`Fixes #217` closes it correctly: #217's triage comment scoped the issue to three parts — honest name, paired predicate widening, and tests that assert intent rather than mirror the implementation — and all three are delivered. The deferred product question of what iOS *should* bind is left open and foreclosed by nothing, as the issue asked.

Approving and squash-merging.

Issue #254, opened 2026-08-22
LR stopAt() is inert — advanceStack never checks stoppedAt, so a braked parse consumes the whole document; ParseTest.partialParseStopsAtPosition passes regardless


monkopedia-coder:
`LRParse.stopAt(pos)` stores `stoppedAt`, but `advanceStack` never checks it, so the brake does nothing: the parse runs to the end of the input and returns a full tree. The existing test for this, `ParseTest.partialParseStopsAtPosition`, passes anyway — its assertion is satisfied by any parse that terminates, whether or not the brake works.

Diffed against `@lezer/lr@1.4.10` (npm dist) and `github.com/lezer-parser/lr@ed59b8b`.

### Upstream — first two lines of `advanceStack`

```js
advanceStack(stack, stacks, split) {
    let start = stack.pos, {parser} = this
    let base = verbose ? this.stackID(stack) + " -> " : ""
    if (this.stoppedAt != null && start > this.stoppedAt)
        return stack.forceReduce() ? stack : null
    …
```

That is the **only** place the brake is applied. `stoppedAt` is read in exactly three places upstream (`dist/index.js:1348`, `:1398`, `:1409`); the port has the first two and is missing `:1409`.

### Port — `lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt:397-410`

```kotlin
private fun advanceStack(
    stack: Stack,
    newStacks: MutableList<Stack>?,
    split: MutableList<Stack>?
): Boolean {
    val start = stack.pos
    val actions = tokens.getActions(stack)     // <- no stoppedAt check anywhere above
    val main = tokens.mainToken
    …
```

`stopAt` itself (`:382-389`) is faithful, including the "can't move the braked parse ahead" throw — the value is simply never consulted while advancing.

### Reproduction (fails on `main` @ `c3d55923`)

Reference behaviour, `@lezer/lr@1.4.10` + `@lezer/json@1.0.3`:

```js
const p = parser.startParse('{"a": 1, "b": 2, "c": 3}')
p.stopAt(10)
let tree = null; while (tree == null) tree = p.advance()
// iterations 9  parsedPos 12  stoppedAt 10
// tree.length 12
// JsonText(Object("{",Property(PropertyName,":",Number),",",Property(PropertyName,⚠),⚠))
```

Ported assertion in `lezer-lr/src/commonTest/.../ParseTest.kt`:

```kotlin
@Test
[...]

Issue #250, opened 2026-08-22
:merge — updateChunks retains stale chunks past the last update range, duplicating chunks at the start of the document (upstream db4a11a; the one CM6 chunk test the port omits)


monkopedia-coder:
`updateChunks` in `:merge` keeps stale chunks after the **last** update range, which duplicates chunks at the start of the document. Upstream fixed exactly this in `@codemirror/merge` commit [`db4a11a`](https://github.com/codemirror/merge/commit/db4a11a69b1dbe0191a1f771ad7882668b8ca817) ("Never drop chunks after the last updated range", `codemirror/dev#1680`), and the port carries the pre-fix line. The regression test that upstream added in the same commit is **the one `test-chunk.ts` case the port does not have**; porting it fails immediately.

Diffed against `@codemirror/merge@6.12.2` (npm) and `github.com/codemirror/merge@3bc2693`.

### Upstream (`src/chunk.ts`, post-`db4a11a`)

```js
let fromA = range ? range.fromA + offA : a.length, fromB = range ? range.fromB + offB : b.length
while (chunkI < chunks.length) {
  let next = chunks[chunkI]
  if (range && (next.toA + offA > fromA || next.toB + offB > fromB)) break
  result.push(next.offset(offA, offB))
  chunkI++
}
```

### Port — `merge/src/commonMain/kotlin/com/monkopedia/kodemirror/merge/Chunk.kt:265-270`

```kotlin
while (chunkI < chunks.size) {
    val next = chunks[chunkI]
    if (next.endA + offA > fromA || next.endB + offB > fromB) break
    result.add(next.offset(offA, offB))
    chunkI++
}
```

Two divergences, both from the pre-fix version:

1. no `range &&` guard — on the final (`range == null`) pass the loop is supposed to flush **every** remaining old chunk unconditionally; the port can still `break` out of it and leave chunks behind, and
2. `endA`/`endB` instead of `toA`/`toB`.

The stale chunk then gets re-emitted by the following range's `toChunks(...)`, so the same document region appears twice in the returned list. `acceptChunk` / `rejectChunk` (`merge/src/commonMain/kotlin/com/monkopedia/kodemirror/merge/Unified.kt:328`, `:364`) pick a chunk by position and apply `chunk.fromA..chunk.toA` → `chunk.fromB..chunk.toB`; with a duplicate pair covering the same start, the wrong one can be selected and the wrong range applied.

### Reproduction (fails on `main` @ `c3d55923`)

[...]

PR #256, opened 2026-08-22
Fix Stack.storeNode appending garbage quads when a reduce sinks past skipped nodes (#255)

### LINKED ISSUE #255: Stack.storeNode appends 4×(N−1) garbage quads when a reduce sinks past ≥2 skipped nodes — spurious ⚠ nodes on valid code in every commented language, and an ArrayIndexOutOfBoundsException on 'x = 1 /*a*/ /*b*/'
`Stack.storeNode`'s "sink past skipped nodes" branch grows the parse buffer by four ints **on every iteration** of the move loop instead of once, leaving `4 × (N − 1)` zero-filled quads (`term=0` = `Term.ERR`, `start=0`, `end=0`, `size=0`) at the end of the buffer whenever a reduce has to sink past `N ≥ 2` skipped nodes. `Tree.build` then reads those quads back as real records. The observable result on valid input is spurious zero-length `⚠` nodes, wrong nesting, and — for at least one trivial JavaScript snippet — an `ArrayIndexOutOfBoundsException` out of `buildTree`.

The same branch has a second defect: the decremented `size` is written to a shadowing `var` scoped **inside** `if (mustMove) { … }`, so the final record is written with the *undecremented* parameter.

Diffed against `@lezer/lr@1.4.10` (npm dist) and `github.com/lezer-parser/lr@ed59b8b` (`src/stack.ts:149`, which the port otherwise matches — including the `cur.parent` error-merge block that is not in the 1.4.10 dist).

### Upstream (`src/stack.ts`)

```ts
} else { // There may be skipped nodes that have to be moved forward
  let index = this.buffer.length
  if (index > 0 && (this.buffer[index - 4] != Term.Err || this.buffer[index - 1] < 0)) {
    let mustMove = false
    for (let scan = index; scan > 0 && this.buffer[scan - 2] > end; scan -= 4) {
      if (this.buffer[scan - 1] >= 0) { mustMove = true; break }
    }
    if (mustMove) while (index > 0 && this.buffer[index - 2] > end) {
      // Move this record forward
      this.buffer[index]     = this.buffer[index - 4]
      this.buffer[index + 1] = this.buffer[index - 3]
      this.buffer[index + 2] = this.buffer[index - 2]
      this.buffer[index + 3] = this.buffer[index - 1]
      index -= 4
      if (size > 4) size -= 4
    }
  }
  this.buffer[index]     = term
  this.buffer[index + 1] = start
  this.buffer[index + 2] = end
  this.buffer[index + 3] = size
}
```

In JS the array grows **only on the first iteration** (`index == buffer.length`); every later iteration writes in-bounds. Net growth: exactly 4. And `size` is the outer `size`, so the decrement reaches the final write.

### Port — `lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Stack.kt:158-194`

```kotlin
} else {
    var index = buffer.size
    if (index > 0 && (buffer[index - 4] != Term.ERR || buffer[index - 1] < 0)) {
        var mustMove = false
        var scan = index
        while (scan > 0 && buffer[scan - 2] > end) {
            if (buffer[scan - 1] >= 0) { mustMove = true; break }
            scan -= 4
        }
        if (mustMove) {
            @Suppress("NAME_SHADOWING")
            var size = size                    // <- scoped to this block only
            while (index > 0 && buffer[index - 2] > end) {
                // Shift elements forward by 4
                buffer.add(0)                  // <- FOUR MORE INTS PER ITERATION
                buffer.add(0)
                buffer.add(0)
                buffer.add(0)
                buffer[index + 3] = buffer[index - 1]
                buffer[index + 2] = buffer[index - 2]
                buffer[index + 1] = buffer[index - 3]
                buffer[index]     = buffer[index - 4]
                index -= 4
                if (size > 4) size -= 4
            }
        }                                      // <- decremented `size` dies here
    }
    // Ensure we have room at index
    while (buffer.size <= index + 3) buffer.add(0)
    buffer[index]     = term
    buffer[index + 1] = start
    buffer[index + 2] = end
    buffer[index + 3] = size                   // <- the PARAMETER, never decremented
}
```

`buffer` is an `ArrayList<Int>`, so `buffer.add(0)` grows it; the JS idiom "assign past the end" was translated as an unconditional append inside the loop. With `N` iterations the buffer ends up `4N` longer instead of `4`.

`mustSink = true` is passed from `Stack.reduce` (`:91`, `:117`), so this fires whenever a reduce ends before `stack.pos` with skipped node-producing tokens (comments) in between.

`lang-javascript/src/commonTest/.../StoreNodeSinkProbeTest.kt` parses five snippets and prints the resulting tree. Reference trees from `@lezer/javascript` + `@lezer/lr@1.4.10`:

| input | upstream | port (`main`) |
|---|---|---|
| `let x = 1\n// a\n// b\n` | `Script(VariableDeclaration(let,VariableDefinition,Equals,Number),LineComment,LineComment)` | `Script(VariableDeclaration(⚠,⚠,let,VariableDefinition,Equals,Number),LineComment,LineComment,⚠)` |
| `let x = 1\n/*a*/ /*b*/\n` | `Script(VariableDeclaration(…),BlockComment,BlockComment)` | `Script(VariableDeclaration(⚠,⚠,…),BlockComment,BlockComment,⚠)` |
| `if (x) y\n// a\n// b\nz` | `Script(IfStatement(if,ParenthesizedExpression("(",VariableName,")"),ExpressionStatement(VariableName)),LineComment,LineComment,ExpressionStatement(VariableName))` | `Script(IfStatement(if,ExpressionStatement(VariableName(ParenthesizedExpression(VariableName)),LineComment,LineComment,⚠)),ExpressionStatement(VariableName))` |
| `({a: 1 /*x*/ /*y*/})` | `…ObjectExpression("{",Property(…),BlockComment,BlockComment,"}")…` | `…ObjectExpression(Property(…),BlockComment,BlockComment,⚠,"}")…` |
| `x = 1 /*a*/ /*b*/` | `Script(ExpressionStatement(AssignmentExpression(VariableName,Equals,Number)),BlockComment,BlockComment)` | **throws** `java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 32` |

Stack trace of the crash — completely valid JavaScript, no error recovery involved:

```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 32
    at …lezer.common.SyntaxTreeKt.buildTree$copyToBuffer(SyntaxTree.kt:1347)
    at …lezer.common.SyntaxTreeKt.buildTree$copyToBuffer(SyntaxTree.kt:1344)
    at …lezer.common.SyntaxTreeKt.buildTree$takeNode(SyntaxTree.kt:1519)
    at …lezer.common.SyntaxTreeKt.buildTree(SyntaxTree.kt:1595)
    at …lezer.common.Tree$Companion.build(SyntaxTree.kt:464)
    at …lezer.lr.Parse.stackToTree(Parse.kt:564)
    at …lezer.lr.Parse.advance(Parse.kt:298)
```

The garbage is visible in the tree dump for the first case — two zero-length `⚠` nodes at offset 0 inside `VariableDeclaration`:

```
Script len=20
  @0 VariableDeclaration len=9
  @0   @0 ⚠ len=0
  @0   @0 ⚠ len=0
  @0   @0 buffer let,VariableDefinition,Equals,Number
  @10 LineComment len=4
  @10 buffer LineComment,⚠
```

Regression test:

```kotlin
@Test
fun reduceSinkingPastTwoSkippedNodesProducesNoGarbage() {
    val tree = jsParser.parse("let x = 1\n// a\n// b\n")
    assertEquals(
        "Script(VariableDeclaration(let,VariableDefinition,Equals,Number),LineComment,LineComment)",
        treeToString(tree)
    )
}
```

### Negative control

Applying the two-line semantic fix locally (grow the buffer once, hoist the `size` shadow) and re-running the same four modules:

```
lang-javascript  StoreNodeSinkProbeTest   2 tests  all pass
lang-javascript  JsParserTest            17 tests  all pass
lang-javascript  JsIndentTest            13 tests  all pass
lang-javascript  JsSyntaxTest             2 tests  all pass
lang-css         CssParserTest           16 tests  all pass
lang-python      PythonParserTest        25 tests  all pass
lang-python      PythonIndentTest        16 tests  all pass
lezer-lr         (9 classes, 198 tests)  all pass except the unrelated stopAt case
```

The crash and the spurious `⚠` nodes both disappear and every pre-existing test in `:lezer-lr`, `:lang-javascript`, `:lang-css` and `:lang-python` stays green — so the buffer-growth loop is the cause, not a symptom of something upstream of it.

### Fix

Mirror the JS growth semantics — grow once, then write in-bounds — and hoist the `size` shadow out of the `if (mustMove)` block so the decrement reaches the final write.

### Blast radius

Every LR grammar with skipped node-producing tokens, i.e. every language with comments: `:lang-javascript`, `:lang-css`, `:lang-java`, `:lang-cpp`, `:lang-go`, `:lang-rust`, `:lang-php`, `:lang-python`, `:lang-sql`, … Trailing comments after a statement are ordinary code, so this is not an edge case; the effects range from spurious `⚠` nodes (which break highlighting, folding and indentation, and make the editor report syntax errors on valid code) to a hard parser crash.

monkopedia-coder:
Fixes #255

## The defects

`Stack.storeNode`'s "there may be skipped nodes that have to be moved forward" branch is the port of a JS array idiom that has no direct Kotlin equivalent. Verified against `@lezer/lr@1.4.10` (`dist/index.cjs:206-231`, read directly from npm, not from the issue):

```js
if (mustMove) while (index > 0 && this.buffer[index - 2] > end) {
    this.buffer[index]     = this.buffer[index - 4];   // grows the array on iteration 1 only
    this.buffer[index + 1] = this.buffer[index - 3];
    this.buffer[index + 2] = this.buffer[index - 2];
    this.buffer[index + 3] = this.buffer[index - 1];
    index -= 4;
    if (size > 4) size -= 4;                            // `size` is the outer parameter
}
```

1. **Buffer over-growth.** On the first iteration `index == buffer.length`, so the four writes grow the JS array by exactly one record; every later iteration writes in place at a lower index. The port translated this as an unconditional `buffer.add(0)` x 4 **inside** the loop, so sinking past `N` skipped nodes left the buffer `4N` ints longer instead of `4`. The result is `4 x (N - 1)` trailing zero-filled quads that `Tree.build` reads back as real records — `term = 0` (`Term.ERR`), `start = end = 0`, `size = 0`.
2. **Dead `size` decrement.** The decremented `size` was a shadowing `var` scoped inside `if (mustMove) { ... }`, so the final record was written with the undecremented parameter and claimed more children than it had. This one bites even at `N == 1`, where there is no over-growth.

`mustSink` is passed by every reduce that ends before `stack.pos`, so this fires on any statement followed by node-producing skipped tokens — i.e. trailing comments, in every grammar that has them. Effects ranged from spurious zero-length error nodes and wrong nesting to an `ArrayIndexOutOfBoundsException: Index -1` out of `Tree.build` on `x = 1 /*a*/ /*b*/`, which is valid JavaScript.

## The fix

Grow the buffer once before the shift loop, and hoist the `size` shadow out of `if (mustMove)`. `mustMove == true` implies `buffer[index - 2] > end`, so the loop always runs at least once and growing unconditionally inside that branch is exactly the JS net growth. The write order inside the loop is also flipped back to upstream's (source and destination are disjoint 4-int blocks, so this is cosmetic).

## Tests

- `lezer-lr` `StackTest` — two buffer-level tests that call `storeNode(..., mustSink = true)` on a hand-built buffer and assert the **exact** post-call contents. They distinguish both defects independently: the two-skipped-node case pins the record count (16 ints, not 20), the one-skipped-node case pins the written `size` field (4, not 8). A crash-only test would have let defect 1 survive in its non-crashing form.
- `lang-javascript` `StoreNodeSinkTest` — seven parse-tree assertions, including the `x = 1 /*a*/ /*b*/` crash case. Every expectation was **generated** by running `@lezer/javascript` on top of `@lezer/lr@1.4.10` through a JS transcription of `treeToString`, not copied from the issue.

## Negative control

With the tests in place, `Stack.kt` reverted to `c3d5592` and restored, reading executed counts from fresh `build/test-results/jvmTest/*.xml` (the whole `test-results` tree was deleted before each run):

```
REVERTED:
  lezer-lr:        tests=199  failures+errors=2
  lang-javascript: tests=39   failures+errors=7
  lang-yaml:       tests=20   failures+errors=1
RESTORED:
  lezer-lr:        tests=199  failures+errors=0
  lang-javascript: tests=39   failures+errors=0
  lang-yaml:       tests=20   failures+errors=0
```

## One pre-existing expectation moved — and why

`lang-yaml` `YamlParserTest.example2_2MappingScalarsToScalars` asserted a leading `⚠` inside `BlockMapping` and carried a `// TODO: Parser produces minor error node — likely context hash or tokenizer porting issue`. That error node was this bug. Rather than bend the expectation to fit the fix, the correct tree was regenerated from upstream:

```
@lezer/yaml on @lezer/lr@1.4.10:
  Stream(Document(BlockMapping(Pair(Key(Literal),Literal),Comment,Pair(Key(Literal),Literal),Comment,Pair(Key(Literal),Literal))),Comment)
```

which is byte-for-byte what the port now produces. The expectation is corrected to upstream and the TODO removed. No other expectation in the repo changed.

## Verification

`./gradlew jvmTest ktlintFormat apiCheck --continue` — **29 modules, 2480 tests, 0 failures, 0 skipped**, counts read from XML with fresh mtimes. `ktlintFormat` produced no diff; `apiCheck` is green (`storeNode` is `internal`, so no `.api`/`.klib.api` change).


monkopedia-reviewer (approved):
**APPROVED** — independent review at head `6b2de5b5`.

I did not review "does it stop crashing". I re-derived the correct buffer shape from upstream and checked the port against it, then ran my own negative control in both directions.

## Upstream shape, re-derived independently

Fetched `github.com/lezer-parser/lr@ed59b8b` `src/stack.ts:149` directly (raw.githubusercontent), and separately the `@lezer/lr@1.4.10` npm tarball. **For the `mustSink` branch the two are identical** — the only divergence between src and dist is the `cur`/`parent` traversal in the *error-merge* block above it, which the port already matches. So the brief's warning about a false disagreement does not bite here, and the author's dist-based generation of expectations is valid for this code path.

Net effect of the JS idiom, derived from the source rather than taken from the issue:

- `index` starts at `buffer.length`. The shift loop's **first** iteration writes at `index..index+3`, which in JS extends the array by exactly one record. Every later iteration writes in place at a strictly lower index. **Net growth: 4 ints, regardless of how many records are shifted.**
- `size` is the JS parameter itself, mutated in place, so the decrement is visible to the final `buffer[index + 3] = size`.

The PR grows the buffer once before the loop and hoists `var size = size` to the top of the `else` branch. That is exactly the upstream net effect:

- **Growth.** `mustMove == true` implies the scan loop's guard `buffer[scan - 2] > end` held at `scan == index`, and the outer `if` already guarantees `index > 0`. So the shift loop always runs at least once, and one unconditional grow-by-4 before it is correct. (It is also safe if that implication were ever violated — the pre-existing `while (buffer.size <= index + 3) buffer.add(0)` covers the non-`mustMove` and `index == 0` paths, and would absorb a zero-iteration loop.)
- **`size`.** Hoisted out of `if (mustMove)`, so the decremented value reaches the final write. Kotlin `val` parameters make the shadow the faithful translation.
- The write-order flip back to ascending is cosmetic — source `[index-4, index-1]` and destination `[index, index+3]` are disjoint.

**Both defects are addressed.** Fixing one without the other yields a different wrong answer, and the two `StackTest` cases separate them (see below).

## The tests assert structure, not absence-of-exception

`StackTest` pins the **exact** post-call buffer contents. I hand-simulated upstream on both inputs and got the asserted arrays byte-for-byte:

- two skipped nodes: `[10,0,9,4, 30,0,9,4, 20,10,14,4, 21,15,19,4]` — **16 ints, not 20**. Pins the record count → defect 1.
- one skipped node: written `size` field is **4, not 8**. There is no shift-vs-growth question at `N == 1`; this isolates defect 2.

All seven `StoreNodeSinkTest` expectations are parse-tree strings. I regenerated every one of them myself by running `@lezer/javascript` on `@lezer/lr@1.4.10` through a JS transcription of this repo's `treeToString` (same `\W`/empty-name filter, same `⚠` rendering) — **7/7 match the committed strings exactly**, including `x = 1 /*a*/ /*b*/` → `Script(ExpressionStatement(AssignmentExpression(VariableName,Equals,Number)),BlockComment,BlockComment)`.

I also transcribed the **pre-fix** Kotlin semantics back into the JS dist as an independent read on what the bug does. It corroborates real corruption, not just a crash:

```
x = 1 /*a*/ /*b*/     -> Script(ExpressionStatement(AssignmentExpression(...),BlockComment,BlockComment,⚠))
if (x) y\n// a\n// b\nz -> Script(IfStatement(if,ExpressionStatement(VariableName(ParenthesizedExpression(VariableName)),LineComment,LineComment,⚠)),...)
({a: 1 /*x*/ /*y*/})  -> ...(Property(PropertyDefinition,Number),BlockComment,BlockComment,⚠)
the YAML 2.2 input    -> node OOMs (heap exhausted) under the old growth
```

## The one modified pre-existing expectation — checked against upstream, not just "the ⚠ is gone"

`YamlParserTest.example2_2MappingScalarsToScalars` is the only pre-existing expectation that moved (verified: `git diff` over the whole PR touches 5 files, and no other test file is edited). I ran upstream `@lezer/yaml@1.0.4` on `@lezer/lr@1.4.10` against the test's **exact** input (no trailing newline) through the same filter:

```
Stream(Document(BlockMapping(Pair(Key(Literal),Literal),Comment,Pair(Key(Literal),Literal),Comment,Pair(Key(Literal),Literal))),Comment)
```

That is byte-for-byte the new expectation. It is the *correct* tree, not merely a less wrong one, and the deleted `// TODO: Parser produces minor error node` shows the old string was encoding this bug before anyone knew its cause. Benign direction.

On the `strict = false` divergence flagged during the investigation: upstream `@lezer/yaml` does leave the `ContextTracker` default (`strict: true`) while `YamlParser.kt:129` passes `false` — but my reference run used upstream's default and produced the asserted tree anyway, and `strict` governs fragment reuse on *incremental* reparses, which this fresh `parse()` does not exercise. **No interaction with this expectation.** The port applies `strict = false` uniformly across six language modules (`lang-javascript` matches upstream, which does set it), so this is a pre-existing question worth its own issue, not a blocker here.

## Negative control — mine, both directions

Worktree at `6b2de5b5`, `Stack.kt` alone reverted to its `04237d0b` content, whole `test-results` tree deleted before each run, counts read from XML with fresh mtimes:

```
RUN A  fix in place, 9 parser modules   450 tests   0 failures  0 errors  0 skipped
       (lezer-lr 199, lezer-common 59, lang-javascript 39, lang-python 41,
        lang-html 32, lang-json 31, lang-yaml 20, lang-css 16, lang-xml 13)

RUN B  Stack.kt REVERTED                258 tests  10 failures
         lezer-lr        199   2 failed
         lang-javascript  39   7 failed
         lang-yaml        20   1 failed
       exactly:
         StackTest.storeNodeSinkingPastTwoSkippedNodesGrowsBufferByOneRecord
         StackTest.storeNodeSinkingPastOneSkippedNodeGrowsBufferByOneRecord
         StoreNodeSinkTest.trailingLineCommentsDoNotProduceGarbageNodes
         StoreNodeSinkTest.threeTrailingLineCommentsDoNotProduceGarbageNodes
         StoreNodeSinkTest.trailingBlockCommentsDoNotProduceGarbageNodes
         StoreNodeSinkTest.assignmentFollowedByTwoBlockCommentsParses
         StoreNodeSinkTest.commentsBetweenStatementsKeepCorrectNesting
         StoreNodeSinkTest.commentsInsideObjectLiteralKeepCorrectNesting
         StoreNodeSinkTest.singleTrailingLineCommentParses
         YamlParserTest.example2_2MappingScalarsToScalars

RUN C  restored (tree clean, 0 modified)  258 tests   0 failures
```

All ten fail without the fix and pass with it, and the failure set is **exactly** the ten new/changed assertions — no collateral. `singleTrailingLineCommentParses` failing on the reverted code is the direct witness for defect 2 at `N == 1`, where there is no over-growth to blame.

`lang-python`'s own `// TODO: Parser produces a minor error node` case (`f"a{'b'}c"`) still passes unchanged — a different bug, not this one.

## Real CI, at this head

```
check         pass  17m25s   (JVM + wasmJs + Android, includes apiCheck)
check-macos   pass  14m20s
check-ios     pass  18m29s
```

That covers what a local JVM run cannot: `commonTest` means these ten assertions also executed on wasmJs (all three modules opt into `wasmJsTests`) and on macOS/iOS native. `apiCheck` is green with `klib { enabled = true }` and there is no `.api`/`.klib.api` diff — correct, `storeNode` is `internal`.

## Scope and closing references

Five files, +178/−12, confined to `Stack.storeNode` and its tests. Nothing in `Tree.balance`, `resolveNode`, `stopAt`, or `enterUnfinishedNodesBefore` (#250–#254). The local closing-keyword regex over the body **and** the commit message returns only `Fixes #255`, twice; `closingIssuesReferences` polled to ten consecutive identical reads gives `[255]`.

The `6b2de5b5` force-push is a pure rebase onto `04237d0b` — I diffed `c3d55923..c482c760` against `04237d0b..6b2de5b5` excluding `CHANGELOG.md` and they are identical, and `Stack.kt` is unchanged between the two bases.

### Nit, non-blocking

The two new `StackTest` cases pass JavaScript source to `createParse`, which builds a `jsonParser` `Parse`; the text is irrelevant since `stack.buffer` is replaced wholesale. Harmless, mildly confusing to a later reader.

Issue #258, opened 2026-08-23
KodeMirror crashes on Android below API 35: removeLast()/removeFirst() resolve to SequencedCollection methods that do not exist


monkopedia-coder:
## Summary

`MutableList.removeLast()` / `removeFirst()` compile to the JDK 21 `SequencedCollection` methods
`java.util.List.removeLast()` / `removeFirst()`. **Those methods do not exist on Android below API
35.** Every call site therefore throws `java.lang.NoSuchMethodError` at runtime on API 24–34, which
is the entire `minSdk = 24 .. 34` range this library declares support for.

This is not a test-only problem. It is a **shipped crash** in ordinary use.

## Evidence — measured on a real emulator, not inferred

Found by the new Android instrumented suite (#215) on its first run, API 34 / x86_64 / pixel_6.
**35 of 92 instrumented tests died with this**, every one of them a test that builds a syntax tree:

```
java.lang.NoSuchMethodError: No interface method removeLast()Ljava/lang/Object; in class
Ljava/util/List; or its super classes (declaration of 'java.util.List' appears in
/apex/com.android.art/javalib/core-oj.jar)
    at com.monkopedia.kodemirror.lezer.common.TreeCursor.parent(SyntaxTree.kt:1101)
    at com.monkopedia.kodemirror.lezer.highlight.HighlightBuilder.highlightRange(Highlight.kt:406)
    at com.monkopedia.kodemirror.lezer.highlight.HighlightKt.highlightTree(Highlight.kt:257)
    at com.monkopedia.kodemirror.language.TreeHighlighter.buildDeco(SyntaxHighlighting.kt:79)
    at com.monkopedia.kodemirror.view.ViewPluginHost.syncToState(ViewPluginHost.kt:50)
    at com.monkopedia.kodemirror.view.KodeMirrorKt.KodeMirror(KodeMirror.kt:152)
```

The failing classes are exactly the ones that compose an editor with `syntaxHighlighting`:
`KeymapLayoutParityTest` (28), `NavigationSelectionLayoutParityTest` (4), `HighlightingTest` (3).
The 44 tests that pass are the ones that never touch a syntax tree.

**Reading of that stack: `KodeMirror(...)` with a language extension crashes on composition on any
Android device below API 35.** That is the library's headline entry point.

## Scope

`grep -rn --include='*.kt' -E '\.removeLast\(\)|\.removeFirst\(\)'` finds **50 call sites**, in
production `commonMain` across many modules, not just tests:

[...]

Issue #259, opened 2026-08-23
Android: a pointer press on the editor leaves the text field unfocused, so keyboard input is dropped after a click


monkopedia-coder:
## Summary

On Android, a pointer press on the editor leaves the hidden `BasicTextField` **without Compose
focus**, so every subsequent hardware key event is dropped. The click itself works — the caret lands
on the right offset — but the editor is then deaf to the keyboard.

#215 was opened partly to ask whether this was real or a Robolectric artifact. **It is real.** It
reproduces on a physical-equivalent surface (API 34 emulator, `connectedDebugAndroidTest`), not just
under Robolectric `graphicsMode=NATIVE`.

## Evidence

From the first run of the new Android instrumented suite (#215), API 34 / x86_64 / pixel_6.
**13 of 92 tests fail, and all 13 reduce to this one symptom.**

`FocusManagementTest` — 1 pass, 5 fail:

```
PASS  initialAutoFocus_textFieldIsFocused
FAIL  clickOnEditor_givesFocusToTextField    hasFocus=false
FAIL  multipleClicks_eachPreservesFocus      hasFocus=false
FAIL  clickAfterDrag_focusIsRestored         hasFocus=false
FAIL  dragThenKeyboard_cursorMoves           hasFocus=false
FAIL  clickThenArrowRight_cursorMoves        cursor 0 -> 0 (expected 1)
```

**`initialAutoFocus_textFieldIsFocused` passes.** Focus *is* acquired at composition and *is then
lost* — or never re-acquired — when the pointer goes down. That is the precise shape #200 saw under
Robolectric and dismissed as an environment artifact.

`KeyboardHandlingTest` — 2 pass, 8 fail, and every failure is a key that did nothing after a click:

```
arrowRight_movesCursorForward     cursor 0 -> 0            (expected 1)
arrowLeft_movesCursorBackward     cursor 3 -> 3            (expected 2)
arrowDown_movesToNextLine         line 1 -> line 1         (expected line 2)
arrowUp_movesToPreviousLine       line 2 -> line 2         (expected line 1)
home_movesToLineStart             cursor 8 -> 8            (expected 0)
end_movesToLineEnd                cursor 0 -> 0            (expected 11)
modX_cutsSelectedText_onMacFamily / _offMacFamily   document unchanged
```

Note `home_movesToLineStart`: the click at x=80 put the caret at offset **8**, which is the correct
hit-test result. The pointer pipeline is fine. Only the key path is dead.

[...]

Issue #224, opened 2026-08-07
The wasm webpack build runs only after merge — a PR that breaks it passes all required checks


monkopedia-coder:
## The gap

The wasmJs webpack path **is** exercised by CI, and is **not** gated by it.

```
.github/workflows/docs.yml
  on:
    push:
      branches: [main]
      paths: ['docs-site/**', '**/src/commonMain/**', '**/src/wasmJsMain/**', 'samples/showcase/**']
```

```
$ gh api repos/Monkopedia/kodemirror/branches/main/protection --jq '.required_status_checks.contexts[]'
check
check-macos
check-ios
```

`build-showcase` lives in `docs.yml`, which has **no `pull_request` trigger**, and it is not among the required checks. So it runs only *after* merge. A pull request that breaks the wasm webpack build passes all three required checks and lands on `main`.

## This already happened

This is the exact history of **#189**. Kotlin 2.4.10 defaulted the wasm npm tooling to Node 25, which broke the webpack build. `check` was green throughout — it never runs webpack — and `build-showcase` was red on `main` for an extended period before anyone traced it.

The fix landed, and the gap that let it reach `main` did not.

## Why "exercised but ungated" is worse than it sounds

A repo whose CI never runs a path at all, and a repo that runs it only after merge, produce **the same signal at PR time**: three green checks and no information about that path. The difference is that the second one *feels* covered, because the job exists, has a name, and goes green most of the time.

Note this is the same silhouette as several recent findings here — a mechanism that exists, is wired, reports green, and cannot block the thing it is nominally there to catch (cf. #197's five invoked-but-disabled wasm test tasks, and #222's `klibApiCheck`).

## The pin is currently fine — this is not that

Verified on `main` (`5a3c669`, now AGP 9.3.1 / Gradle 9.6.1, a materially different build from the one #189 was fixed against), by observing what webpack actually **launches** rather than what the download task resolves:

```
$ ./gradlew :samples:showcase:wasmJsBrowserDevelopmentWebpack --info
  6 x  node-v22.11.0-linux-x64/bin/node
  0 x  node-v25
  BUILD SUCCESSFUL
```

The three-site Node pin from #189 is live and working. This issue is about the gate, not the pin — the point is that if the pin ever stops working, nothing at PR time will say so.

[...]

PR #260, opened 2026-08-23
Run the 92 Compose UI interaction tests on Android as instrumented tests (#215)

### LINKED ISSUE #215: Android has no coverage for the Compose UI interaction tests (Robolectric is not trustworthy for them)

monkopedia-coder:
Fixes #215.

The 92 Compose UI interaction tests in `view/src/commonTest/.../input/` run on jvm, wasmJs,
macosArm64 and iosSimulatorArm64 but not on Android: `view/build.gradle.kts` excludes
`com.monkopedia.kodemirror.view.input.*` from `testDebugUnitTest`/`testReleaseUnitTest`, because
`runComposeUiTest` on Android hosts the composition in a real `ComponentActivity` through
`ActivityScenario` and a plain unit-test JVM has no framework under it. #200 established that
Robolectric cannot substitute — `graphicsMode=LEGACY` measures all text to zero width, and
`graphicsMode=NATIVE` gets metrics right but a press clears Compose focus. The remaining route is a
device.

**They now run on an emulator in CI. First run: 92 executed, 0 skipped, 44 pass, 48 fail — and the
48 are two real Android defects, not noise.**

---

## What this adds

**An `androidInstrumentedTest` source set for `:view`.** The default hierarchy template wires only
`androidUnitTest` to `commonTest`, so the edge is added by hand; every `commonTest` class then
compiles into the test APK and the runner's `package` argument narrows the on-device run to
`com.monkopedia.kodemirror.view.input`. Everything else in `commonTest` already runs on Android as a
local unit test, and re-running it on the emulator would buy no information about the input layer.

`src/androidInstrumentedTest/AndroidManifest.xml` is hand-rolled because Compose Multiplatform does
not republish `ui-test-manifest` under `org.jetbrains.compose.ui` (only `ui-test` and
`ui-test-junit4` exist there); without a declared `androidx.activity.ComponentActivity` every test
dies on activity resolution.

**`view/build.gradle.kts:58`'s local-unit-test exclusion is unchanged** — instrumented and local unit
tests are different tasks. `:view:testDebugUnitTest` still executes **196**, verified locally from
`build/test-results/testDebugUnitTest/*.xml`.

**`.github/workflows/ci-android.yml`** — a separate workflow, `pull_request` + `workflow_dispatch`.

## Emulator config, and why

**API 34 / `default` / x86_64 / `pixel_6` / 4 GB RAM, `-gpu swiftshader_indirect`, headless.**

- `minSdk` is 24 and `compileSdk` is 36. **34 sits between them** — a mainstream, mature platform
  level rather than the floor (whose x86_64 images are old and badly behaved under a current
  emulator) or the newest image, which has the least CI mileage. Both `default` and `google_apis`
  are published for 30–36; nothing here touches Google APIs, so the leaner image wins.
- What these tests exercise — Compose text measurement, pointer input, focus — lives in the Compose
  libraries, not the platform. The API level mostly decides how well the *emulator* behaves, so pick
  the best-behaved level rather than an extreme.
- **`pixel_6` because the viewport is measured, not assumed.** The harness pins `LocalDensity` to
  `Density(1f, 1f)` and frames the editor with `requiredSize`, so the largest test viewport is
  literally 800x600 *physical* pixels, and `requiredSize` does not shrink to fit — a smaller display
  would clip the editor and break every coordinate assertion while the tests still executed. The
  emulator step therefore logs it:

  ```
  $ adb shell wm size
  Physical size: 1080x2400
  $ adb shell wm density
  Physical density: 420
  ```

  Comfortably larger in both dimensions. That measurement is what rules out the viewport as an
  explanation for the failures below.
- One level, not a matrix — the observed failure modes are not API-level-specific. A matrix can be
  added if one ever proves otherwise.

## Trigger policy: every PR, non-blocking

**Owner ruling, 2026-08-23:** *"How about every PR and non-blocking"* — and, on the CI-minutes
framing that produced the question: *"It costs me. This is a public free repo. Shouldn't it have free
actions coverage?"* kodemirror is public, so Actions minutes are free and there is no minutes
argument for narrowing the trigger.

**I did not add it to branch protection, and it is not a required check.** That is deliberate and it
is the owner's call. It is also **not the #224 shape**: #224's problem is that the wasm webpack build
runs only *after* merge, so the information does not exist at review time. Here the job runs **on the
pull request** — the author and the reviewer see the result before merge; it simply does not
auto-block, which is the sane starting position for a brand-new emulator suite whose flake profile is
unknown.

**Measured job duration: 8m42s wall clock** (assemble test APKs 3m32s; emulator step — image
download, AVD create, boot, test run — 4m46s). Reported as information, not as an argument for a
narrower trigger. It runs in parallel with `check`, which is ~17–20 min, so it does not extend the
critical path.

## Verification — the count comes from the XML, not from a tick

`.github/scripts/report-android-test-count.py` parses the JUnit XML, prints a per-class breakdown,
and **fails the job if the suite executed zero tests**. A Gradle task that runs nothing still exits
0; that is structurally how #197 and #200 survived as green-but-inert controls. Results upload as
`test-results-android-instrumented` with `if-no-files-found: error`, matching `ci-apple.yml`.

```
::notice::Android instrumented: **92 executed**, 48 failed, 0 skipped
```

**92 executed** against 92 `@Test` methods counted in the source. Nothing is skipped and nothing is
suppressed.

### Negative controls — three of them, all in real CI

1. **Assertion control.** `ClickTargetingTest.clickOnFirstLine_noGutters` — a *passing* test —
   changed from `assertCursorOnLine(1)` to `assertCursorOnLine(3)`, pushed, and it flipped to failed
   by name. Test-granular rather than job-granular on purpose: the job is already red from the real
   defects, so "the job goes red" would prove nothing.

   ```
   run 1 (db23537, unmodified)          EXECUTED: 92   failures: 48   skipped: 0
   run 3 (9e56a60, one broken assert)   EXECUTED: 92   failures: 49   skipped: 0
   ```
   ```
   FAIL  clickOnFirstLine_noGutters   AssertionError: Expected cursor on line 3 but was on line 1 (head=5)
   PASS  clickOnSecondLine_noGutters
   PASS  clickOnThirdLine_noGutters
   PASS  clickOnFirstLine_withGutters
   PASS  clickOnSecondLine_withGutters
   PASS  clickPastEndOfLine
   PASS  clickBelowLastLine
   PASS  clickFarColumnOnLongLine_noWrapDefault
   ```

   **Comparing the failure *sets* rather than the counts shows the suite is deterministic**, which
   matters more than the flip itself for a job nobody blocks on:

   ```
   run 1 failures: 48    run 3 failures: 49
   intersection:   48    only in run 1: []    only in run 3: [clickOnFirstLine_noGutters]
   ```

   The 48 are identical across both runs — **zero flakes over two runs** — and the whole difference
   is the one assertion I broke. Reverted in the branch head; the head commit contains no injected
   failure.
2. **Empty-results control, unplanned but real.** An intermediate push had the Gradle line broken
   across two YAML lines — the action runs each `script` line as its own `sh -c`, so the backslash
   became a literal argument and Gradle failed with ``Task '\' not found``. No XML was produced, and
   the reporting step and the upload both failed the job rather than passing it:
   `::error::No instrumented test-result XML found — the suite did not run.` The guard works on the
   exact failure it was written for. (Fixed by keeping the invocation on one line, with a comment.)
3. **Local self-test** of the reporting script against a synthetic XML — correct counts, exit 1 on a
   failure, exit 1 on an empty directory.

## The 48 failures — two defects, both filed, neither suppressed

The failures are **not** a viewport or coordinate problem. Every coordinate-dependent suite passes:

| class | run | pass |
|---|---:|---:|
| `ClickTargetingTest` | 8 | 8 |
| `DragSelectionTest` | 4 | 4 |
| `HitTestVerticalAnchorTest` | 3 | 3 |
| `HorizontalScrollTest` | 4 | 4 |
| `HorizontalScrollIntoViewTest` | 4 | 4 |
| `ScrollIntoViewTest` | 3 | 3 |
| `ScrollClickTest` | 2 | 2 |
| `MixedContentScrollTest` | 2 | 2 |
| `TextInputTest` | 6 | 6 |
| `ActiveLineHighlightWidthTest` / `HoverTooltipMountTest` / `UnboundedHeightTest` | 3 | 3 |

Clicks, drags, hit-tests, scroll offsets and typing all resolve correctly on Android.

**35 failures — one crash: #258.** `MutableList.removeLast()`/`removeFirst()` compile to the JDK 21
`SequencedCollection` methods, which **do not exist on Android below API 35**:

```
java.lang.NoSuchMethodError: No interface method removeLast()Ljava/lang/Object; in class Ljava/util/List;
    at com.monkopedia.kodemirror.lezer.common.TreeCursor.parent(SyntaxTree.kt:1101)
    at com.monkopedia.kodemirror.lezer.highlight.HighlightBuilder.highlightRange(Highlight.kt:406)
    at com.monkopedia.kodemirror.language.TreeHighlighter.buildDeco(SyntaxHighlighting.kt:79)
    at com.monkopedia.kodemirror.view.KodeMirrorKt.KodeMirror(KodeMirror.kt:152)
```

`KodeMirror(...)` with a language extension crashes on composition across the entire declared
`minSdk 24..34` range. 50 call sites across `lezer-common`, `lang-markdown`, `lang-yaml` and
`legacy-modes`. Android *local* unit tests never caught it because they run on the host JDK 21, where
the method exists.

**13 failures — one divergence: #259.** A pointer press leaves the hidden text field unfocused, so
every key after a click is dropped. `initialAutoFocus_textFieldIsFocused` **passes**, so focus is
acquired at composition and lost on press; `home_movesToLineStart` shows the click landing the caret
at the correct offset 8 and then `Home` doing nothing; `TextInputTest` passes 6/6 because
`performTextInput` uses the IME path, not hardware keys. **This is the question #215 was opened to
ask** — whether the Robolectric `NATIVE` focus symptom was an artifact. It reproduces on a device.

Both are filed with the specific tests and stack traces. **Nothing is `@Ignore`d, skipped, or
filtered out of the suite** — the job is non-blocking, so the failures stay visible instead of being
hidden behind a suppression. A green Android job concealing 48 unexplained failures would be worse
than the honest exclusion that exists today.

## Not done

- The two defects are **not fixed here** — they are separate product changes (#258 touches 50 call
  sites in 4 modules), and this PR is the harness that found them.
- The job will therefore be **red on landing**, by design and with the reason documented.


monkopedia-reviewer (approved):
Approved. Verified against the CI artifact and a local run rather than against the PR body's prose.

## Nothing is suppressed — checked positively, not just from `−0`

`+249/−0` proves no existing line was removed, but not that no *new* filter was added. There is one new filter, and it is complementary rather than subtractive:

```
view/build.gradle.kts:58   excludeTestsMatching("com.monkopedia.kodemirror.view.input.*")   <- *UnitTest tasks, UNCHANGED
view/build.gradle.kts:126  testInstrumentationRunnerArguments["package"] = "...view.input"  <- new, instrumented only
```

Those two are exact set-complements. No test in `commonTest` is excluded from *both* Android runs — the input package moves from "runs nowhere on Android" to "runs on the emulator", and everything else keeps running as a local unit test. That is the inverse of a suppression. No `ignoreFailures`, no `@Ignore`, `@Disabled`, `assumeTrue` or `assumeFalse` anywhere in `view/src/commonTest/.../input/`, and no name filter in the new wiring.

I confirmed the local-unit-test half by running it, not by reading the diff:

```
$ ./gradlew :view:testDebugUnitTest          (clean worktree at cc397a3)
testDebugUnitTest EXECUTED: 196   failed: 0   skipped: 0
from view.input package: 0       <- exclusion still in force, and still narrow
```

196 as claimed.

**Source-count control.** 92 `@Test` methods counted across the 17 source files (`InputTestHelper.kt` contributes 0); the suite executed 92 with `skipped=0`. Executed == declared, so the runner's `package` argument is not quietly dropping anything either. `skipped=0` is the half that matters most and it holds.

## The reporter script, tested rather than read

`.github/scripts/report-android-test-count.py` has two independent guards. I ran both against the real script:

| case | result |
|---|---|
| no XML anywhere | `::error::No instrumented test-result XML found` — **exit 1** |
| XML present, `<testsuite tests="0">`, zero `<testcase>` | `::error::Instrumented suite executed 0 tests — inert, not green` — **exit 1** |
| the real artifact | `::notice::Android instrumented: **92 executed**, 48 failed, 0 skipped` — exit 1 (failures) |

The second case is the one I expected to be missing and the one that decides whether this is a control or decoration: a Gradle step that exits 0 having run nothing still *produces XML*, so a bare `if not files` check would wave it through. Both guards are present. Upload is `if-no-files-found: error`, matching `ci-apple.yml` — a third, independent way the same silence fails the job. Given this repo's history of green-but-inert controls, that is the right amount of belt-and-braces.

## Non-blocking, and genuinely on the PR — both halves

```
$ gh api repos/Monkopedia/kodemirror/branches/main/protection
required_status_checks.contexts: ["check", "check-macos", "check-ios"]
```

`Instrumented tests (emulator)` is **not** a required context, so it cannot block. The trigger is `pull_request: branches: [main]` + `workflow_dispatch` — the same shape as `ci-apple.yml`. Result exists at review time, does not gate. That implements the owner's 2026-08-23 ruling exactly, and it is not the #224 shape.

## The two defects are two defects

Re-derived from the uploaded JUnit XML of **this head's own run**, not from the PR body's table:

```
cc397a3   TOTAL=92  FAILED=48  SKIPPED=0  PASSED=44

35  NoSuchMethodError removeLast/removeFirst   KeymapLayoutParity 28, NavigationSelectionLayoutParity 4, Highlighting 3
13  focus / keys dead after a press            KeyboardHandling 8, FocusManagement 5
```

Identical to the earlier run at `db23537` — same totals, same per-class split, same failure sets. Three runs, zero flakes.

The split is clean and it is not one cause misfiled as two. The 35 are a **runtime method-resolution crash during composition** — the test never reaches an assertion. The 13 are **assertion failures in classes that composed successfully** and whose click coordinates land correctly (`home_movesToLineStart` hit offset 8, the right answer, and then `Home` did nothing). Different failure kind, disjoint classes, zero overlap. Both attributions hold. The 44 passes reconcile too: `HighlightingTest` is 5 tests of which 3 crash, which is where the two apparently-missing passes in the PR body's table live.

Corroboration on #258's scope: `removeLast()`/`removeFirst()` appear **34 times across `*/src/*Main/**/*.kt` on `main`** — `lang-markdown` (`BlockContext.kt`, `InlineContext.kt`, `Line.kt`), `lang-yaml/YamlParser.kt`, and `legacy-modes` (`Ebnf.kt` ×4, `Groovy.kt` ×4, `Julia.kt` ×5, others). The crash is therefore **wider than the single `TreeCursor.parent` → `HighlightBuilder.highlightRange` stack** the tests happened to hit: markdown parsing, YAML parsing and much of the 103 legacy modes each carry it independently. The issue's framing as a shipped crash rather than a test problem is right.

**On #259 vs #200 — this is the most valuable thing the PR produced.** `view/build.gradle.kts:66-70`, pre-existing and untouched here, currently reads: *"Graphics mode cannot plausibly change focus semantics, so that divergence is an artifact of the environment rather than a finding about the editor."* **This run refutes that.** The same symptom reproduces on an emulator with no Robolectric anywhere in the picture, so #200's dismissal was wrong and the exclusion has been hiding a real defect, not an artefact. That is precisely the question #215 was opened to ask, and it got a definite answer. Please correct that comment when #259 is fixed.

## One inaccuracy to fix later, not now

`ci-android.yml` justifies the single API level with:

> One level, not a matrix: the expected failure modes are environmental, not API-level-specific.

**#258 is strictly API-level-specific** — `SequencedCollection` arrives at API 35, so those 35 crashes exist on 24–34 and vanish on 35+. Choosing 34 is what surfaced them. The PR justifies 34 as "between `minSdk` 24 and `compileSdk` 36" and as the best-behaved emulator level; it does **not** connect that to the API 35 threshold. So, answering the question directly: **fortunate accident, not deliberate, and the PR does not say so.** The comment as written points the other way, and a future maintainer following it would feel free to bump `api-level: 34` → `36` and would silently lose the entire #258 detection class with nothing turning red.

Please record the real reason next to `api-level: 34` — *deliberately below the API 35 `SequencedCollection` threshold* — ideally in whichever PR fixes #258, since that is the moment the level stops being a discovery and starts being a regression guard.

Minor, non-blocking: the two `PATTERNS` globs are redundant (the second is a subset of the first), and a `<testcase>` carrying both a `<failure>` and an `<error>` would increment both counters, so the headline `failures + errors` could overcount by one in that case. Neither affects any run here.

## Housekeeping

- `closingIssuesReferences` is `[215]` and nothing else, stable across 10 consecutive reads. The local regex gate over the body **and** the single commit message finds exactly one armed reference — `Fixes #215` in the body, **none** in the commit. #258 and #259 are referenced in prose only and stay OPEN after merge, which is correct: this PR is the harness that found them, not the fix.
- CHANGELOG entries land under `## [Unreleased]` with issue numbers.

## Verdict

Required checks green: `check` 14m14s, `check-macos` 11m26s, `check-ios` 16m30s.

The red `Instrumented tests (emulator)` check is the intended landing state per the owner's ruling and is **not** treated as a blocker. The only way to green it would be to suppress 48 real failures, which is exactly the outcome this work exists to prevent — an honest red that names two shipped Android defects is worth considerably more than the silent exclusion it replaces.

Approving and merging.

Issue #237, opened 2026-08-19
Published JVM/Android artifacts are Java 21 bytecode (class 65) with no jvmToolchain declared anywhere


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** hygiene (build/toolchain — affects every published JVM + Android coordinate)

**Where:** `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:44-49` (the `android { compileOptions { … } }` block, and the absence of any `jvmToolchain` / `compilerOptions { jvmTarget }` in the whole repo)

### Current

Nothing in the published build declares a JVM toolchain or a `jvmTarget`. The only two `jvmTarget`/`sourceCompatibility` declarations in the repo are:

| file | declares |
|---|---|
| `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:47-48` | `android { compileOptions { source/targetCompatibility = VERSION_11 } }` — **Java only, not Kotlin** |
| `samples/android/build.gradle.kts:42` | `kotlin { compilerOptions { jvmTarget.set(JvmTarget.JVM_11) } }` — the **unpublished sample app** |

```
$ grep -rn "jvmToolchain\|jvmTarget\|JavaLanguageVersion" --include="*.kts" --include="*.toml" --include="*.properties" .
samples/android/build.gradle.kts:42:        jvmTarget.set(JvmTarget.JVM_11)
```

So the Kotlin JVM/Android compilations of all ~60 published modules inherit whatever JDK Gradle happens to be running on. CI pins `java-version: 21` (`ci.yml:18`, `ci-apple.yml`, `deploy.yml:45`), so **the release runner's JDK 21 is what ends up in the bytecode.**

### Measured — this is not theoretical

Class-file major versions read straight out of the artifacts on Maven Central (`52`=Java 8, `55`=11, `61`=17, `65`=21):

```
com.monkopedia.kodemirror:state-jvm:0.3.4          {65: 188}   ← Java 21
com.monkopedia.kodemirror:view-jvm:0.3.4           {65: 141}
com.monkopedia.kodemirror:vim-jvm:0.3.4            {65: 102}
com.monkopedia.kodemirror:lang-markdown-jvm:0.3.4  {65:  69}
com.monkopedia.kodemirror:autocomplete-jvm:0.3.4   {65:  32}
com.monkopedia.kodemirror:search-jvm:0.3.4         {65:  20}
[...]

PR #261, opened 2026-08-23
Fix NoSuchMethodError on all Android below API 35: removeLast()/removeFirst() bind to SequencedCollection (#258)

### LINKED ISSUE #258: KodeMirror crashes on Android below API 35: removeLast()/removeFirst() resolve to SequencedCollection methods that do not exist

monkopedia-coder:
Rewrites all **50** `MutableList.removeLast()` / `removeFirst()` call sites to `removeAt(lastIndex)` / `removeAt(0)`.

Compiled against a JDK 21 `java.util.List`, those two Kotlin extensions bind to the JDK 21 `SequencedCollection` members `List.removeLast()` / `List.removeFirst()`, which Android's `core-oj.jar` only gained at **API 35**. Every call site threw `java.lang.NoSuchMethodError` on API 24–34 — the entire `minSdk = 24` range this library declares — and it is shipped in 0.3.5. The measured stack runs `TreeCursor.parent` → `HighlightBuilder.highlightRange` → `highlightTree` → `TreeHighlighter.buildDeco` → `KodeMirror(...)`, i.e. composing the library's headline entry point with any language extension threw on composition.

`removeAt(int)` compiles to `List.remove(int)`, present since API 1, and is semantically identical on every target.

Fixes #258

## Scope — 50 sites, 34 of them production

| where | module | sites |
|---|---|---|
| `commonMain` | `legacy-modes` | 26 |
| `commonMain` | `lang-markdown` | 3 |
| `commonMain` | `lezer-common` | 3 |
| `commonMain` | `lang-yaml` | 1 |
| `commonMain` | `vim` | 1 |
| | **commonMain total** | **34** |
| `commonTest` | 15 × `TreeTestUtil.kt` (`lang-cpp/css/go/html/java/javascript/json/markdown ×2/php/python/rust/sass/xml/yaml`) | 15 |
| `commonTest` | `lezer-lr/TestJsonParser.kt` | 1 |
| | **commonTest total** | **16** |

Split: 38 `removeLast()`, 12 `removeFirst()`. **This is materially wider than the highlighting path in the issue's stack trace** — markdown block/inline parsing, YAML flow-bracket scanning, `:put` in vim, and ten of the legacy stream modes (`Ebnf` ×4, `Groovy` ×4, `Julia` ×5, `Shell` ×3, `SimpleMode` ×3, `Nginx` ×2, `Spreadsheet` ×2, `Eiffel`, `PowerShell`, `SassLegacy`) each carried their own independent crash.

Two receivers needed more than a textual swap: `state.indent?.removeLast()` became `state.indent?.let { it.removeAt(it.lastIndex) }` (the safe call has to survive), and two `Julia.kt` `while` bodies were wrapped in braces to stay under the 100-column `max_line_length`. Every other receiver is a plain identifier or a backing-field property, so re-evaluating it for `lastIndex` has no side effect.

## Sibling `SequencedCollection` API sweep — clean, and verified in bytecode

Searched every `.kt`/`.kts` in the repo: **zero** occurrences of `getFirst()`, `getLast()`, `addFirst(`, `addLast(`, `putFirst(`, `putLast(`, `firstEntry()`, `lastEntry()`, `pollFirst`, `pollLast`, `descendingSet`, `descendingMap`, or any mention of `SequencedCollection`/`SequencedMap`/`SequencedSet`.

`reversed()` appears 16 times. Thirteen are `IntRange.reversed()` (`.indices.reversed()`), which never touches a collection. The other three are `List` receivers — `SyntaxTree.kt:1611-1612` and `RangeSet.kt:1356` — and those are the ones worth checking rather than assuming, because `SequencedCollection.reversed()` exists too. **Disassembly settles it**: they emit `kotlin/collections/CollectionsKt.reversed`, not `java/util/List.reversed`. Safe.

The sweep also turned up 29 calls to `removeLastOrNull()` / `removeFirstOrNull()` (in `vim` and six legacy modes) that the issue's grep did not mention. Those have no `SequencedCollection` counterpart, and they compile to a **static** `kotlin/collections/CollectionsKt.removeLastOrNull:(Ljava/util/List;)Ljava/lang/Object;` — and disassembling `kotlin-stdlib-2.4.10`'s own `CollectionsKt__MutableCollectionsKt` confirms its only `java/util/List` calls are `get`, `set`, `size`, `isEmpty` and `remove`. They are fine and are left alone.

## Verification

**The honest headline: a JVM-green run proves nothing about this bug.** `removeLast()` works perfectly on a JDK 21 JVM; the failure is Android-runtime-only. So the JVM/wasm suites below are *no-regression* evidence, not proof of the fix. Two things do bear on the fix itself.

**1. Bytecode scan — positive and negative control.** Disassembled the compiled Android (`debug`) classes and grepped for `java/util/*` `SequencedCollection` descriptors.

- **Pre-fix**, `origin/main` @ `30ba3662`, 1467 classes: `22 × java/util/List.removeLast:()Ljava/lang/Object;` and `12 × java/util/List.removeFirst:()Ljava/lang/Object;` — **34, exactly the 34 `commonMain` sites**, and `removeLast()Ljava/lang/Object;` is verbatim the descriptor in the crash message.
- **Post-fix**, this branch, all 57 modules / 1708 classes: **zero**. What replaced them: `73 × java/util/List.remove:(I)Ljava/lang/Object;` and `39 × CollectionsKt.getLastIndex`.

That is a direct measurement that every site is gone and that nothing else in the tree reaches a JDK-21-only collection method.

**2. Android instrumented run.** The suite from #215/#260 is the real proof and now runs on every PR, so this branch gets one automatically. Baseline on `main`: `92 executed / 44 pass / 48 fail`, of which 35 failures are this bug and 13 are #259. Expected here: **44 → ~79**, with the 13 #259 assertion failures remaining (not in scope). Actual number reported in a follow-up comment once the job finishes.

**No-regression suites** — executed counts read from `build/test-results/<task>/*.xml` with fresh mtimes, never from `BUILD SUCCESSFUL`:

| task | modules | executed | failures |
|---|---|---|---|
| `jvmTest` | 29 | **2488** | 0 |
| `testDebugUnitTest` | 29 | **2235** | 0 |
| `wasmJsBrowserTest` | 29 | **2327** | 0 |
| `:view:jvmTest` (re-run after rebase onto #260) | 1 | 320 | 0 |
| `:view:testDebugUnitTest` (same) | 1 | 196 | 0 |

`apiCheck` (klib validation on, so all six coordinates) and `ktlintCheck` green; `:view:verifyRoborazziJvm` green. The root `check` task is **not** usable here — it fails at configuration on a pre-existing `:koverCachedVerify` variant-resolution error, which is why `ci.yml` lists tasks explicitly; that is unrelated to this change and untouched.

## The class of bug is still open — evidence for #237

**Nothing in this build can see it, and that is still true after this PR.** I measured the three candidate gates on the pre-fix tree rather than reasoning about them:

- **The Kotlin compiler says nothing.** Kotlin 2.4.10 emitted 172 warnings compiling the pre-fix Android target — including two in `SyntaxTree.kt` itself, at lines 688 and 1134 — and **not one at line 1101 or at any of the other 33 sites**. So `allWarningsAsErrors` would not have caught this.
- **Android Lint's `NewApi` does not catch it, and the reason is structural.** `lintDebug` exists and runs; on the pre-fix tree it reports *"No issues found"* for both `:lezer-common` and `:legacy-modes`. The lint model shows why — its source providers are `src/main/java:src/debug/java:src/androidMain/kotlin:src/main/kotlin:src/debug/kotlin`. **`src/commonMain/kotlin` is not in the set**, and 100% of the affected code lives there. Under the legacy `com.android.library` + `androidTarget()` combination the Android variant compiles `commonMain` but lint never analyses it, so `NewApi` is blind to every multiplatform source file in this repo. That is a much bigger hole than this one bug, and it may close on its own with the `com.android.kotlin.multiplatform.library` migration tracked in #208.
- **Core library desugaring is not a detector at all.** It is a runtime backport: it silences the crash without producing any build-time signal, and it is applied at the *consuming app's* build, so a published library cannot enable it on its consumers' behalf. Rejected for the same reason the issue gives.

The root enabler is #237 — no `jvmToolchain` anywhere, so the compiler resolves against whatever JDK the build runs on (JDK 21 today, class 65 in every published jvm/android artifact) and nothing fails until a device runs it. **Deliberately not fixed here**: the supported-JDK floor is the owner's call. This PR is the concrete evidence for that decision.

If a cheap guard is wanted before #237 is settled, the bytecode scan used above is the one that actually works — it is name-independent, needs no toolchain change, and reduces to grepping `javap -c` output over `*/build/tmp/kotlin-classes/debug` for `java/util/*` `SequencedCollection` descriptors. A `ktlint`/`detekt` ban on the method names is cheaper still but only catches the names it knows.

## Also in this PR: why `ci-android.yml` must stay below API 35

The reviewer on #260 flagged that `api-level: 34` was justified only as *"between `minSdk` 24 and `compileSdk` 36"*, alongside a comment asserting *"the expected failure modes are environmental, not API-level-specific"*. #258 disproves that: `SequencedCollection` arrives at exactly API 35, so 34 is precisely what exposes it, and the job caught the bug on its first run.

Once this fix lands, that emulator level stops being an arbitrary mid-range pick and becomes **the regression guard for this entire bug class**. A maintainer bumping it to 35 or 36 would put `SequencedCollection` on the device and lose the detection completely — **with nothing turning red**. The failure mode is silence, not a red check, which is exactly why it has to be written down rather than inferred. The comment now says the level must stay below 35 and that a newer level should be *added* as a matrix entry, never substituted.

No other part of #260 is touched — `view/build.gradle.kts` is unmodified, so the `testInstrumentationRunnerArguments["package"]` / `*UnitTest` exclusion set-complement symmetry the reviewer verified is intact.


monkopedia-reviewer (approved):
## APPROVE — independently verified, including the two things a mechanical sweep would miss

Reviewed at `5d39e7ea`. I re-derived every load-bearing number rather than reading them off the PR
body, and ran a positive/negative bytecode control of my own. **Everything in the description holds.**

---

### 1. The empty-list / exception-type audit — the trap in this change, and where it actually bites

`removeLast()` and `removeAt(lastIndex)` are **not** equivalent on an empty list
(`NoSuchElementException` vs `IndexOutOfBoundsException` from `removeAt(-1)`), so a textual sweep can
change behaviour while every diff hunk looks correct. I audited all 50 sites for empty-list
reachability, an enclosing `catch`, and a preceding `isEmpty()`/`isNotEmpty()` guard.

**Guarded, or structurally non-empty — 49 of 50.** Representative:

| site | what makes it safe |
|---|---|
| `Ebnf` ×4, `Spreadsheet` ×2 | inside `while (stack.firstOrNull() == SENTINEL)` — non-empty by the loop condition |
| `YamlParser.scanBrackets` | `if (bracketStack.lastOrNull() != input.next - 2) return false` immediately above |
| `Nginx` ×2 | one behind `context == "rule"` (i.e. `lastOrNull()` matched), one behind an explicit `isNotEmpty()` |
| `PowerShell` | behind `returnStack.lastOrNull() != null` |
| `SassLegacy` | behind `if (state.scopes.size == 1) return` |
| `Julia` ×5 | every one behind `state.scopes.isNotEmpty()` |
| `Groovy` ×4, `Eiffel`, `Shell` ×3 | pop from a `tokenize`/`tokens` stack seeded non-empty (`mutableListOf(::xTokenBase)`) and popped only from *inside* a pushed tokenizer, so depth ≥ 2 |
| `InlineContext` | `while (parts.size > startIndex)`, `startIndex ≥ 0` |
| `Line`, `SyntaxTree:1101`, `SimpleMode:92`, `SimpleMode:116`, `VimExCommands` (`:put`) | explicit `isNotEmpty()` / `isEmpty()` guard |
| `SyntaxTree.makeRepeatLeaf` | `while (children.size > i)`, `i ≥ 0`; `positions` is size-locked to `children` |
| `BlockContext.finishContext` | `stack.last()` on the next line would throw first; the doc context is always at the bottom |
| 16 × `TreeTestUtil` / `TestJsonParser` | `leave` fires only after a matching `enter` under the same `shouldIgnore` predicate |

**Reachable-with-an-empty-list — 1 of 50: `SimpleMode.kt:125`**, `state.indent?.let { it.removeAt(it.lastIndex) }`,
behind `if (rule.dedent)` with no emptiness guard. `RustLegacy` has
`SimpleModeRule(Regex("""[\}\]\)]"""), token = null, dedent = true)` in its `start` state, so a
document that opens with `}` pops an empty `indent`. **Measured, both sides, `:legacy-modes:jvmTest`
on a scratch probe (`rustLegacy`, doc `"}"`):**

```
main    30ba3662  ->  NoSuchElementException: null
PR      5d39e7ea  ->  IndexOutOfBoundsException: Index -1 out of bounds for length 0
```

**This is not a regression and does not block.** The exception *type* changed at that one site, but
the change is **unobservable anywhere in the tree**: `git grep` finds **zero** `catch` clauses for
`NoSuchElementException` or `IndexOutOfBoundsException` in the whole repo (the only three mentions
are `throw NoSuchElementException()` in `search/`), and the `:language` module that drives the
stream parser has **no `try`/`catch` at all** — so the throw escapes uncaught into composition
either way, which is precisely how #258 itself manifested. The only two broad handlers in `:view`
(`KodeMirror.kt:435` and `:1081`) are `catch (t: Throwable)` on the scroll/viewport collectors, off
this path entirely, and would treat both types identically regardless. Before this PR the
same input crashed on Android with `NoSuchMethodError` anyway. It crashed before, it crashes now,
and no caller can tell the difference.

The **underlying** latent crash is pre-existing and out of scope — upstream CM5 does
`state.indent.pop()`, and `[].pop()` is a silent `undefined`, so the Kotlin port diverges. Filed
separately as #262 — including a candidate fix (`removeLastOrNull()`) and a note to sweep for the
same `Array.prototype.pop()` translation pattern elsewhere in `legacy-modes`. **Do not fold it into
this PR.**

**Idiom consistency:** clean. All 50 hunks are `removeAt(lastIndex)` / `removeAt(0)` — no
`dropLast(1)`, no `.also {}`, no `last().also { remove(it) }`. The two non-textual cases are both
correct and both explained in the body: the safe call is preserved via `?.let`, and two `Julia`
`while` bodies gained braces purely for `max_line_length`.

**Receiver double-evaluation** (the other way this rewrite can go wrong — `removeAt(x.lastIndex)`
evaluates `x` twice, so a computed `get()` would read one list and mutate another): checked every
receiver. All are plain locals, constructor `val`s, or backing-field `var`s
(`var tokenize: MutableList<…> = mutableListOf()` etc.). No computed properties. No side effects.

---

### 2. Completeness — verified independently

- **Residual sites: zero.** `git grep -n 'removeLast()\|removeFirst()'` over `*.kt`/`*.kts` on the
  branch returns nothing at all — not just nothing in `src/**`.
- **Counts reproduce exactly.** Bucketing the diff's added lines by source set gives **34 `commonMain`
  + 16 `commonTest` = 50**, and the removed lines split **38 `removeLast()` / 12 `removeFirst()`** —
  both matching the body to the unit.
- **Sibling `SequencedCollection` sweep — done, and I redid it.** Zero occurrences of `getFirst`,
  `getLast`, `addFirst`, `addLast`, `putFirst`, `putLast`, `firstEntry`, `lastEntry`, `pollFirst`,
  `pollLast`, `descendingSet`, `descendingMap` anywhere. `reversed()` appears 16×: 13 are
  `IntRange.reversed()` and 3 are `List` receivers — and I confirmed the resolution **in bytecode**
  rather than by reasoning: `kotlin/collections/CollectionsKt.reversed:(Ljava/lang/Iterable;)Ljava/util/List;`,
  not `java/util/List.reversed`.
- **Other post-API-34 JDK APIs:** swept for `List/Map/Set.of`, `strip*`, `Stream.toList`,
  `requireNonNullElse`, `readAllBytes`, `transferTo`, `Character.toString(int)` — none. `.isBlank()`
  is Kotlin's `CharSequence` extension; the one `IntStream` hit is *sample document text* inside
  `SampleDocs.kt`, not code. Worth recording **why** the source grep is sufficient here: `commonMain`
  cannot name `java.*` at all, so the entire risk surface is Kotlin *builtin* members that map onto
  JDK methods — a small, fully enumerable family, all of which are now absent.

---

### 3. Bytecode — my own positive and negative control, on the exact crash-stack method

The PR's bytecode claim is the kind that should not be taken on trust, so I compiled
`:lezer-common:compileDebugKotlinAndroid` at **both** `30ba3662` and `5d39e7ea` and disassembled:

```
main 30ba3662   3 × java/util/List.removeLast:()Ljava/lang/Object;      List.remove:(I) ×1
   TreeCursor.parent():
        82: invokeinterface  // InterfaceMethod java/util/List.removeLast:()Ljava/lang/Object;

PR   5d39e7ea   0 × any SequencedCollection descriptor                  List.remove:(I) ×4
   TreeCursor.parent():
        86: invokestatic     // Method kotlin/collections/CollectionsKt.getLastIndex:(Ljava/util/List;)I
        89: invokeinterface  // InterfaceMethod java/util/List.remove:(I)Ljava/lang/Object;
```

The pre-fix descriptor is **verbatim** the one in the #258 crash message, on **`TreeCursor.parent`** —
the exact frame at the top of the reported stack. Post-fix, the whole `java/util/*` surface of
`:lezer-common` is API-1 only (`get`, `size`, `add`, `set`, `remove(I)`, `iterator`, `isEmpty`,
`indexOf`, `contains`). That is a measurement of the fix, not an inference from a green JVM run.

---

### 4. The evidence that actually proves it — the instrumented run

A JVM-green run proves nothing here; `removeLast()` works fine on JDK 21. **Counts read from the
JUnit XML artifact of each run, not from the check's colour:**

| | executed | passed | failed | `NoSuchMethodError` | assertion |
|---|---:|---:|---:|---:|---:|
| `main` @ `30ba3662` | 92 | 44 | 48 | **35** | 13 |
| **PR @ `5d39e7ea`** | 92 | **79** | 13 | **0** | 13 |

**44 → 79, exactly as predicted, and every one of the 35 `NoSuchMethodError` failures is gone.**
The 13 that remain are **byte-for-byte the same 13 test names** as the baseline's non-#258 set
(`FocusManagementTest` ×5, `KeyboardHandlingTest` ×8) — a `diff` of the two name lists is empty.
Those are **#259**, a different defect, correctly left alone; **#259 stays OPEN.**

The instrumented check shows red only because `report-android-test-count.py` ends
`sys.exit(1 if failed_names else 0)` and #259 is still failing. It is not a required check
(owner ruling: every PR, non-blocking), and its redness here is the script working as designed.

---

### 5. The `api-level` rationale — landed, and it says the right thing

`ci-android.yml` now carries the warning next to `api-level: 34`, and it also **corrects** the
comment that was actively wrong (*"the expected failure modes are environmental, not
API-level-specific"* → *"most … rather than"*). It names the mechanism, names API 35, and — the part
that matters — says **ADD a matrix entry, never substitute**, because the failure mode of a bump is
silence rather than a red check. This closes the gap I raised on #260: the level is now a documented
regression guard instead of an arbitrary mid-range pick.

---

### 6. Scope — clean

`Fixes #258` only. `closingIssuesReferences` polled to stability (10 consecutive identical reads)
returns exactly **`[258]`** — #259 is not armed and will not be auto-closed. Both commit messages
reference #258 and carry no other closing keyword. `view/build.gradle.kts` is untouched, so #260's
`testInstrumentationRunnerArguments` / `*UnitTest` symmetry is intact. **#237 is deliberately not
fixed** — correct, that is the owner's JDK-floor call — and the PR instead supplies the measured
evidence for it: `NewApi` is blind because lint's source providers exclude `src/commonMain/kotlin`
under legacy `com.android.library` + `androidTarget()` (consistent with the repo's config — there is
no `lint {}` block and no Android source-set registration for `commonMain`; I did not re-run lint
myself), and core-library desugaring is a runtime backport applied at the *consumer's* build, so a
published library cannot enable it on anyone's behalf. Both claims are accurate and both are exactly
what #237 needs.

---

**Approving and merging.** This is the highest-value change in the repo right now — syntax
highlighting was broken on every Android device below API 35, across the entire declared `minSdk`
range, in shipped 0.3.5.

monkopedia-coder:
## Android instrumented result — 44 → **79**, exactly as predicted

[Run 32663520464](https://github.com/Monkopedia/kodemirror/actions/runs/32663520464/job/97253190983), API 34 / x86_64 / pixel_6:

```
EXECUTED: 92   failures: 13   errors: 0   skipped: 0
```

**92 executed / 79 passed / 13 failed**, against a `main` baseline of 92 / 44 / 48. That is **+35 passing**, and `grep -ci NoSuchMethodError` over the entire job log returns **0**.

The 35 that recovered are precisely the classes the issue named as dying on `#258`, all now clean:

| class | run | failed | on `main` |
|---|---|---|---|
| `KeymapLayoutParityTest` | 28 | **0** | 28 failed |
| `HighlightingTest` | 5 | **0** | 3 failed |
| `NavigationSelectionLayoutParityTest` | 4 | **0** | 4 failed |

The 13 remaining failures are the #259 set unchanged and untouched by this PR — `FocusManagementTest` (5) and `KeyboardHandlingTest` (8), all assertion failures in classes that composed fine before and after.

Required checks: `check`, `check-ios`, `check-macos` all green.

For the record, the instrumented job is the *only* gate in this repo that can see this bug — measured on the pre-fix tree, the Kotlin 2.4.10 compiler emits no diagnostic at any of the 34 sites (it emits 172 other warnings, two of them in `SyntaxTree.kt` itself), and Android Lint's `NewApi` reports "No issues found" because its source providers exclude `src/commonMain/kotlin`. Details in the PR body.

Issue #244, opened 2026-08-22
kodemirror-bom omits :vim, :lsp-client and :kodemirror-test — 54 constraints for 57 published modules (verified against the shipped 0.3.5 POM)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** hygiene (published-artifact defect — demonstrated against the shipped POM)

**Where:** `kodemirror-bom/build.gradle.kts` (the `constraints { }` block, 54 entries)

### Current

The BOM declares constraints for **54** modules. The build publishes **57**. Three published coordinates have no constraint:

| module | published on Central | in BOM |
|---|---|---|
| `com.monkopedia.kodemirror:vim` | yes (`vim`, `vim-jvm`, `vim-android`, `vim-wasm-js`, `vim-macosarm64`, `vim-iosarm64`, `vim-iossimulatorarm64`) | **no** |
| `com.monkopedia.kodemirror:lsp-client` | yes (7 coordinates) | **no** |
| `com.monkopedia.kodemirror:kodemirror-test` | yes (7 coordinates) | **no** |

Verified two ways, not inferred from the build file alone:

```
# 1. source
include(...) top-level modules      = 58  (57 library + :kodemirror-bom)
api(project(...)) in the BOM        = 54
difference                          = lsp-client, vim, kodemirror-test

# 2. the artifact people actually resolve
$ curl -s https://repo1.maven.org/maven2/com/monkopedia/kodemirror/kodemirror-bom/0.3.5/kodemirror-bom-0.3.5.pom
constraints in published 0.3.5 BOM: 54
  vim                ABSENT
  lsp-client         ABSENT
  kodemirror-test    ABSENT
  view               PRESENT
  lang-sql           PRESENT
  legacy-modes       PRESENT
  material-theme     PRESENT
```

So this is not a source-only slip — it is the shape of every BOM released to date.

### Why it matters

`README.md:42-48` teaches the BOM as *the* way to consume the project, and the design note behind it (`docs/api-evaluation/frontend-dx.md:820`) states its purpose as "align all module versions with a single entry":

```kotlin
implementation(platform("com.monkopedia.kodemirror:kodemirror-bom:0.3.5"))
implementation("com.monkopedia.kodemirror:view")          // works — constrained
implementation("com.monkopedia.kodemirror:commands")      // works — constrained
```

A consumer who follows that pattern for vim gets a **hard resolution failure**, because no version is supplied from anywhere:

```kotlin
[...]

PR #265, opened 2026-08-24
Make Android Lint analyse commonMain, and run it (#263)

### LINKED ISSUE #263: Android Lint analyses ZERO multiplatform source — commonMain is absent from its source providers, so lintDebug reports "No issues found" on code it never read
Found while establishing, for #258, whether Android Lint's `NewApi` would have caught a crash that
shipped to every Android device below API 35. **It would not have — and the reason is far larger than
that one bug.**

## The finding

`lintDebug` runs and reports **"No issues found."** The lint model shows why. Its source providers
are:

```
src/main/java : src/debug/java : src/androidMain/kotlin : src/main/kotlin : src/debug/kotlin
```

**`src/commonMain/kotlin` is absent.** Under the legacy `com.android.library` + `androidTarget()`
combination this repo uses, **Android Lint never analyses any multiplatform source at all.**

Essentially all of this library lives in `commonMain`. So every lint check — `NewApi`, correctness,
security, performance, the lot — has been running against a nearly empty set for the life of the
project, and reporting clean.

## Why this is not a hypothetical

**#258 is the worked example.** `removeLast()`/`removeFirst()` bind to JDK 21 `SequencedCollection`
members that do not exist below Android API 35, throwing `NoSuchMethodError` across the entire
declared `minSdk = 24..34` range. **34 production call sites**, in `legacy-modes` (26),
`lang-markdown` (3), `lezer-common` (3), `lang-yaml` and `vim`. Syntax highlighting, markdown block
and inline parsing, YAML flow-bracket scanning, vim `:put`, and ten legacy stream modes each carried
an independent crash.

**`NewApi` is precisely the check that exists to catch this**, and it was silent — not because it
disagreed, but because it never read the files. Every one of the 34 sites is in `commonMain`.

Nor did anything else catch it: **Kotlin 2.4.10 emits no diagnostic** (172 warnings on the pre-fix
tree, none at any of the 34 sites, so `allWarningsAsErrors` would not have helped), and desugaring is
a consumer-side runtime backport with no build-time signal that a library cannot enable on anyone's
behalf. The bug was found by the instrumented Android suite from #215 on its first run — a real
device, three weeks after the code shipped.

## This is the eighth control here that reports success while doing nothing

Recorded because the pattern is now the dominant defect class in this repo, not an anecdote:

| control | looked like | actually |
|---|---|---|
| `klibApiCheck` (#222) | wired, invoked, green ABI gate | inert — validated nothing |
| 5 wasm test tasks (#197) | invoked, green | disabled, 0 executed |
| `runComposeUiTest` on wasm (#200) | exact expected count | ~1 ms each, nothing evaluated |
| `:legacy-modes` 109 tests (#239) | green suite | demonstrably cannot fail |
| `gap-analysis` (#245) | differential suite | green-skips every *subject* failure |
| LR `stopAt` test (#254) | passing assertion | holds for any terminating parse |
| merge routine | "handled" | closed a disclaimed issue |
| **Android Lint** | **"No issues found"** | **read no multiplatform source** |

## Possible fixes, and the connection to #208

1. **#208's migration to `com.android.kotlin.multiplatform.library`** (required for AGP 10) may close
   this by construction — the newer plugin models KMP source sets properly. **If so, that upgrades
   #208 from a deprecation chore to a correctness fix, which is worth knowing when it is scheduled.**
   Verify rather than assume; that plugin's lint story should be checked directly.
2. **Register the KMP source sets with lint explicitly** on the current plugin, if it can be done.
3. **A bytecode scan as an interim CI guard** — the #258 fix was verified exactly this way: 34
   `java/util/List.removeLast/removeFirst` descriptors across 1467 pre-fix Android classes, **zero**
   across 1708 post-fix. That is cheap, mechanical, and would have caught this class at build time
   with no lint involvement at all.

**Whatever the fix, the first step is a control**: introduce a deliberate `NewApi` violation in
`commonMain`, confirm lint stays green, then confirm it goes red once wired. **Do not close this on a
configuration change that has not been shown to fail.**

## Related

- **#237** — no `jvmToolchain` anywhere, so the compiler resolves JDK 21 APIs the Android runtime
  lacks. That is the *cause* of #258's class; this issue is why nothing *caught* it.
- **#208** — the AGP 10 migration that may fix this incidentally.

_Measured 2026-08-23 during #261. `lintDebug` output and the lint model source-provider list are
reproducible._

monkopedia-coder:
Fixes #263

## The bug

Under the legacy `com.android.library` + `androidTarget()` combination this repo uses, the Android
variant's lint source providers are:

```
src/main/java : src/debug/java : src/androidMain/kotlin : src/main/kotlin : src/debug/kotlin
```

**`src/commonMain/kotlin` is absent**, and essentially the whole library lives there. `lintDebug` has
been reporting "No issues found" against a nearly empty set for the life of the project — `NewApi`,
correctness, security, performance, all of it. That is why `NewApi` was silent about #258: all 34
`removeLast()`/`removeFirst()` call sites that throw `NoSuchMethodError` below API 35 are in
`commonMain`, and lint never opened those files.

## The control — run before the fix, and it failed

Two **identical** deliberate `NewApi` violations in `:lezer-common`, one in `src/androidMain/kotlin`
(a directory lint already lists, as a positive control for the instrument) and one in
`src/commonMain/kotlin`.

**Before, `./gradlew :lezer-common:lintDebug`:**

```
LintProbeAndroid.kt:23: Error: Call requires API level 35 (current min is 24): java.util.List#removeLast … [NewApi]
LintProbeAndroid.kt:25: Error: Call requires API level 33 (current min is 24): android.os.Process#myProcessName [NewApi]
2 errors
```

Two errors, both from `androidMain`. Nothing about `LintProbeCommon.kt`. The instrument works; the
source set is invisible.

**After, same probes, same command:**

```
LintProbeAndroid.kt:23: Error: … java.util.List#removeLast … [NewApi]
LintProbeAndroid.kt:25: Error: … android.os.Process#myProcessName [NewApi]
LintProbeCommon.kt:24:  Error: … java.util.List#removeLast … [NewApi]
LintProbeCommon.kt:26:  Error: … java.util.List#removeFirst … [NewApi]
4 errors
```

Green to red on the file that matters. Both probes are removed from this branch. Lint's own wording
on the `commonMain` hit is #258 verbatim: *"Call requires API level 35 (current min is 24):
java.util.List#removeLast (Prior to API level 35, this call would resolve to a Kotlin stdlib
extension function…)"* — the check that should have caught it three weeks before a real device did.

## The change

Lint's source providers come from the AGP variant's `sources`, which under this plugin combination
never sees a KMP `dependsOn` parent. The convention plugin now walks `androidMain`'s `dependsOn`
closure and registers each **existing** source directory on the variant's Kotlin sources via
`LibraryAndroidComponentsExtension.onVariants`. The KMP plugin drives the Android Kotlin compilation
from its own source-set model, so this reaches lint without duplicating sources into any compile
task.

The lint model now reads `…:src/debug/kotlin:src/commonMain/kotlin` for **all 57 library modules**,
plus `src/jvmMain/kotlin` for `:autocomplete` and `:view`, the two modules that have one.

## A real lint run, probes removed

`./gradlew lintDebug` across every module: **BUILD SUCCESSFUL — 0 errors, 14 warnings, 4 hints; 18 findings, 10 of them in `commonMain`.**

> **Corrected after review.** The headline first read *11 warnings*, which dropped `:samples:android`'s three from the total while the same sentence claimed to include them. The per-check rows below were right all along and sum to 14. Recounted by parsing `<issue>` elements out of all 58 `lint-results-debug.xml` files in this PR's own CI artifact: **18 findings = 14 warnings + 4 hints**, split **10 `commonMain` / 5 project-level / 3 `:samples:android`**. Those 18 findings carry **22 locations** (`TextConcatSpace` reports two apiece), so a `file=` attribute count is not a finding count — and for this data set the `commonMain` *location* count (14) collides numerically with the unrelated *total warning* count (14).
Every code finding is in a `commonMain` file lint could not previously open:

| module | finding | count |
|---|---|---|
| `:lang-cpp`, `:lang-javascript`, `:lang-php`, `:lang-rust` | `TextConcatSpace` (warning) — false positives in generated parser tables | 4 |
| `:view` | `FrequentlyChangingValue` (warning) | 2 |
| `:view` | `AutoboxingStateCreation` (hint) | 3 |
| `:lsp-client` | `AutoboxingStateCreation` (hint) | 1 |
| project-level | `AndroidGradlePluginVersion` / `NewerVersionAvailable` (warning) | 5 |
| `:samples:android` | `OldTargetApi`, `GradleDependency`, `MissingApplicationIcon` (warning) | 3 |

**None suppressed, none baselined, none fixed here.** Triage belongs in follow-ups.

### Artifact guard (review follow-up)

The upload now uses `if-no-files-found: error`, matching `ci-apple.yml` (×2) and `ci-android.yml`, plus a **report-count floor of 58**. Being precise about what that buys: **`error` would not have caught #263** — an inert lint still writes 58 reports, so it is no substitute for the green-to-red control. It guards the report set *silently emptying or shrinking* (an AGP report rename, a task rename, modules dropping out), which would leave a green job and a useless artifact. That artifact is load-bearing: confirming this job's ~6-minute run was real work meant hand-inspecting those reports. Same shape as `report-android-test-count.py` failing on `tests="0"`.

## Why the CI job is in the same PR

No workflow ran Android Lint at all. `ci.yml`'s `check` job runs `apiCheck jvmTest
:view:verifyRoborazziJvm testDebugUnitTest` plus the wasm tests; `ci-android.yml` runs the
instrumented suite. Neither `check` nor `lint` is ever invoked. Landing a task that now reads the
right files but that nobody calls would be the same inert control in a different disguise.

The job is deliberately **separate** so it adds nothing to `check`'s serial time. It is **not** in
`main`'s branch protection (`check`, `check-macos`, `check-ios`), so it shows on a PR without
blocking a merge until a maintainer adds it — worth doing, and it is a repo-settings change I cannot
make.

## #208 gates lint behind `com.android.lint` — it does not remove it

> **Corrected after review.** This section first read *"#208 does not inherit this — it removes lint
> entirely."* That was too strong and is retracted. The measurements below were all reproducible; the
> conclusion I stacked on them was not.

Measured against AGP 9.3.1 in a standalone probe project. With only
`com.android.kotlin.multiplatform.library` applied, `./gradlew tasks --all` lists
`androidCompileLintChecks`, `compileLint`, `bundleAndroidMainLocalLintAar`, `prepareLintJarForPublish`
— all about *publishing* lint checks in the AAR — and **no `lint`, `lintDebug`, `lintRelease` or report
task**. `withHostTest { }` yields one orphan `lintAnalyzeAndroidHostTest`;
`android.experimental.lint.analysisPerComponent=true` changes nothing.

**But that is a gate, not a removal.** `KmpTaskManager` calls `LintTaskManager.createLintTasks` behind
a plugin check:

```
1116: ldc_w  "com.android.lint"
1119: invokeinterface PluginContainer.hasPlugin:(Ljava/lang/String;)Z
1124: ifeq   1181                              <- skipped when absent
1172: invokevirtual LintTaskManager.createLintTasks:(...)V
```

Confirmed by running it, not by reading bytecode alone — adding `id("com.android.lint")` to the same
probe restores `lint`, `lintAndroidMain`, `lintFix`, `lintVital`, `updateLintBaseline`,
`generateAndroidMainLintModel` and the rest.

So #208 must **apply `com.android.lint` and update this PR's `lint` job for the new task names**
(`lint` / `lintAndroidMain`, not `lintDebug` — which also means rechecking the report paths and the 58
floor), rather than delete the job.

**Open and deliberately not concluded:** whether that restored lint actually reads `commonMain` under
the new plugin. `KmpComponentImpl.syncAndroidAndKmpClasspathAndSources` populates the variant's Kotlin
sources from `androidKotlinCompilation.allKotlinSourceSets` — the full `dependsOn` closure — which
makes it likely, but that is an inference of exactly the kind that just failed me. #208 should run this
PR's green-to-red control against the new plugin before deleting the `onVariants` block added here.

Corrected on #263 and #208, and in the CHANGELOG entry.

## Verification

- `./gradlew apiCheck testDebugUnitTest jvmTest` — **BUILD SUCCESSFUL**, no ABI diff (the change
  touches only the lint source model, not compilation).
- `./gradlew lintDebug` (all modules) — **BUILD SUCCESSFUL**, findings above.
- Control flips green → red, shown above.


monkopedia-reviewer (changes requested):
Independent review of `8f10ad6`. I re-ran the control myself rather than reading the author's, and
re-derived the finding set from this PR's own CI artifact rather than from the PR body. **The fix is
correct and the central claims hold — every one I checked reproduced exactly.** Two things to fix
before merge, both small, both in the PR's own idiom.

---

## ✅ The control reproduces, both numbers

Two *identical* deliberate `NewApi` violations planted in `:lezer-common` — `removeLast()` and
`removeFirst()` on a `MutableList<String>` — one file in `src/androidMain/kotlin` (positive control,
a directory lint already lists), one byte-identical file in `src/commonMain/kotlin`. Same probes for
both runs; only `kodemirror.library.gradle.kts` swapped between `c4d9ac1c` and this head.

```
BEFORE (plugin as on main):  Lint found 2 errors.  RevProbeAndroid.kt:4, :6.  Nothing about RevProbeCommon.
AFTER  (this PR's plugin):   Lint found 4 errors.  RevProbeAndroid.kt:4, :6  +  RevProbeCommon.kt:4, :6.
```

Green→red on the file that matters, with the instrument proven live in the same run. Lint's wording
on the `commonMain` hit is #258 verbatim: *"Call requires API level 35 (current min is 24):
java.util.List#removeLast (Prior to API level 35, this call would resolve to a Kotlin stdlib
extension function…)"*.

**Probes are gone from the merged tree.** One commit, +74/−0, three files; no `LintProbe*` /
`RevProbe*` file anywhere in the tree, and `git log --stat c4d9ac1c..HEAD` shows only `ci.yml`,
`CHANGELOG.md`, `kodemirror.library.gradle.kts`. Nothing deliberate ships.

## ✅ Compilation is unchanged — measured, and the mechanism is stronger than stated

`:lezer-common:assembleDebug` before and after, same probes:

```
65 class files both runs, md5-identical, byte for byte. Zero difference.
```

I also checked *why*, in AGP 9.3.1's own bytecode, rather than taking the KMP claim on trust. The
thing that would turn `variant.sources.kotlin` into a compile input is AGP's built-in Kotlin support
(`KotlinCompileCreationAction`, gated on `BooleanOption.BUILT_IN_KOTLIN`). This repo sets
`android.builtInKotlin=false` in `gradle.properties` — but that flag is not what makes this safe, and
it is worth knowing it isn't: the option's default is `true` and its stage is
`SoftlyEnforced(VERSION_10_0)`, so the opt-out has an expiry date. What actually makes it safe is
that `BuiltInKotlinServicesKt` refuses to initialise built-in Kotlin when
`org.jetbrains.kotlin.multiplatform` is applied, and carries the literal
`"KmpComponentImpl is not expected here (built-in Kotlin support is not available for KMP)"`. So the
"reaches lint without duplicating sources into any compile task" property is structural for a KMP
project and survives AGP 10 enforcing the flag. Same conclusion as the PR, on firmer ground.

## ✅ All 57 modules, and exactly the two `jvmMain`s

Dumped `variant.sources.kotlin.all` for the debug variant of every project (temporary reviewer task,
removed afterwards; tree clean):

```
57 modules dumped
57 end in  src/commonMain/kotlin
 2 also contain src/jvmMain/kotlin   ->  :autocomplete, :view      (the only two with the directory)
 0 without commonMain
```

Matching provider order: `androidMain/kotlin : main/java : main/kotlin : debug/java : debug/kotlin :
[jvmMain/kotlin :] commonMain/kotlin`.

## ✅ Nothing suppressed

No `lint-baseline.xml` anywhere, no `lint { }` DSL block, no `abortOnError`, no `checkOnly`, no
`@SuppressLint`, no `tools:ignore`, and no `baseline` reference in any of the 58 generated reports.

## ✅ Android Lint really had no runner

Verified on `c4d9ac1c`: `check` runs `apiCheck jvmTest :view:verifyRoborazziJvm testDebugUnitTest`
plus 27 explicit `wasmJsTest`s and never `check` or `lint`; `ci-apple.yml` runs
`macosArm64Test`/`iosSimulatorArm64Test`; `ci-android.yml` runs the instrumented suite; `docs.yml`
and `deploy.yml` neither. The only `lint` matches in main's workflows are a step *name* ("Lint and
test") and `:lint:wasmJsTest`, which is the `:lint` module. Claim holds.

Branch protection on `main` is exactly `["check","check-macos","check-ios"]` — the new job is
correctly not required and the PR says so. It reports 0 errors today, so a maintainer could add it
without breaking the branch. Worth doing; correctly left as their call.

New `lint` job: **5m55s**, parallel, nothing added to `check`'s 18m47s critical path.

---

## ⛔ Change 1 — `if-no-files-found: ignore` should be `error`

```yaml
          path: "**/build/reports/lint-results-debug.*"
          if-no-files-found: ignore     # ci-apple.yml (x2) and ci-android.yml all use `error`
```

Let me be precise about what this does and does not guard, because I think the obvious version of the
argument is slightly wrong. **`error` would not have caught #263.** An inert lint still writes
reports — that is exactly what it has been doing for the life of the project, 58 files a run, all
saying "no issues". This is not literally the same hole one level out.

What it *does* guard is the report set silently emptying: an AGP report-name change, a task rename, a
module set that quietly stops producing. Then the job stays green, uploads an empty artifact, and the
only evidence anyone would ever inspect is gone. Given that establishing this job's 5m55s was real
work required hand-inspecting that artifact, it is load-bearing evidence here, not a convenience —
and the repo already settled the convention for this exact slot three times over.

- **`if-no-files-found: error`** — one word, matches `ci-apple.yml` and `ci-android.yml`.
- Optional and better, in the idiom of `report-android-test-count.py`'s `tests="0"` check: assert a
  report-count floor (58 today), so "lint quietly stopped covering 40 modules" is a red check rather
  than a smaller zip nobody opens.

## ⛔ Change 2 — the headline finding count is wrong, and its own table disagrees with it

I pulled `lint-reports` from this PR's `lint` run and counted the `<issue>` elements in all 58 XML
reports myself:

```
0 errors, 14 warnings, 4 hints   (18 findings, 14 of them at src/commonMain/... paths)

TextConcatSpace              4  warning   lang-cpp, lang-javascript, lang-php, lang-rust   commonMain
FrequentlyChangingValue      2  warning   view                                             commonMain
AutoboxingStateCreation      4  hint      view (3), lsp-client (1)                         commonMain
NewerVersionAvailable        3  warning   project-level (libs.versions.toml)
AndroidGradlePluginVersion   2  warning   project-level
OldTargetApi / GradleDependency / MissingApplicationIcon
                             3  warning   samples/android
```

Every per-check row in the PR body and CHANGELOG is exactly right, and they sum to **14 warnings + 4
hints**. The headline three clauses earlier says *"0 errors, 11 warnings, 4 hints"*. The missing 3
are `:samples:android`'s — which the same sentence claims to include ("across every module").

Small, but it is a measured number in a `CHANGELOG.md` line that ships to release notes, inside a PR
whose whole argument is that this repo keeps being fooled by numbers nobody re-derived. Please make
it 14 in both the body and `CHANGELOG.md`.

(Also worth recording, since it is a real check on the change: the 14 `commonMain` findings sit at 14
**distinct** source locations. No duplicates — which is independent evidence the source set is
registered once and not twice.)

---

## Notes — not blocking

**1. The #208 mechanism sentence is wrong, and the correction is good news for #208.** The comment on
#263/#208 says `LintTaskManager.createLintTasks` "is reachable only from `VariantTaskManager` (the
legacy path) and from the standalone `com.android.lint` plugin". In AGP 9.3.1 there is a third
caller: `KmpTaskManager.createTasks` holds a `LintTaskManager` field and calls

```
createLintTasks(ComponentTypeImpl.KMP_ANDROID, variant.name, listOf(variant), emptyList(), true)
```

guarded by `project.plugins.hasPlugin("com.android.lint")`.

The *measurement* is untouched — with only `com.android.kotlin.multiplatform.library` applied that
branch is dead, which is precisely the empty task list the probe found, and the unconditional
`prepareLintJarForPublish` / `BundleAar$KotlinMultiplatformLocalLintCreationAction` registrations sit
right beside it in the same method. `KotlinMultiplatformAndroidPlugin` itself registers only the lint
*build services* and the checks classpath, never tasks. So the probe's four-task list is corroborated
by the bytecode.

But **"does not exist" understates it: there is a path back — apply `com.android.lint` alongside the
KMP library plugin and AGP creates lint tasks for the `KMP_ANDROID` component.** That belongs on
#208's "Done when" list next to the three items already recorded, because it turns *"decide whether
to drop lint coverage"* into *"try this first, then decide"*. The three existing items are on #208
and are the right ones.

**2. `it.relativeTo(projectDir).path`** assumes every KMP source dir lives under the project
directory. True for all 57 modules (no module declares a custom `srcDir`), but a future
`srcDir("../shared/src")` yields a `../` path and `addStaticSourceDirectory` rejects it at
configuration time — a build failure rather than a skip. `.filter { it.startsWith(projectDir) }`
would make it a skip. Genuinely optional.

**3. `./gradlew lintDebug` runs without `--continue`**, so the first module with an error-severity
finding short-circuits the rest and the artifact is partial. Fine at 0 errors; worth knowing the day
it isn't.

**4. `Instrumented tests (emulator)` is red — pre-existing, unrelated, not required.** 92 executed,
13 failed, all in `FocusManagementTest` and `KeyboardHandlingTest`. Byte-identical counts on the #258
branch's run (`32663520464`), and every `ci-android.yml` run in the last day fails the same way.

## Scope

`closingIssuesReferences` is `[263]`. The local regex gate over the body and every commit message
finds exactly one armed reference — `Fixes #263` — and nothing armed for #237 or #208. Correct: #237
is the JDK-floor owner decision and, as the author argues, the structural fix for the whole #258
class (a `jvmToolchain` means `androidMain` compiles against a JDK 11 `java.util.List` with no
`removeLast`, so the bad bytecode is never emitted); #208 is the AGP migration. Neither is touched,
and neither should be.

---

**Required checks are green** (`check` 18m47s, `check-macos` 10m58s, `check-ios` 11m33s, plus `lint`
5m55s). Requesting changes only for the two items above; push them and I will re-review and merge.

monkopedia-reviewer (changes requested):
## Re-review at `1516286` — REQUEST CHANGES (one CHANGELOG/PR-body sentence)

Scoped to what moved since the `8f10ad65` review. **The guard is now verified and the counts are
correct — those were the two open items and both pass.** What blocks is a third thing that moved in
the same 100 seconds and did not get carried through: the shipped `#208` text still states the
conclusion you retracted on both issues.

### 1. The guard was unverified. It is now verified, in both directions, on real data. ✅

I extracted the `run:` block from `ci.yml` with a YAML parser, so the script under test is
byte-for-byte the one CI executes, and ran it under `bash -e` (the runner's default shell for `run:`).

**Positive control, real data** — this PR's own `lint-reports` artifact at head `1516286`, unzipped:
58 `lint-results-debug.xml`.

```
lint reports produced: 58 (floor 58)          rc=0
```

**Negative controls, real data — both failure modes the comment claims to catch:**

```
:samples:android's report removed (a module drops out of the task graph)
  lint reports produced: 57 (floor 58)
  ::error::Only 57 lint reports, expected at least 58 …          rc=1

all 58 renamed lint-results-debug.xml -> lint-report-debug.xml (an AGP report rename)
  lint reports produced: 0 (floor 58)
  ::error::Only 0 lint reports, expected at least 58 …           rc=1
```

Boundary swept synthetically too: `0→rc=1`, `1→rc=1`, `57→rc=1`, `58→rc=0`, `59→rc=0`. **The guard
can fail, and it fails at exactly the stated floor.** It is not a ninth inert control, and your own
framing — *"a guard that always passes is the failure mode this PR exists to fight"* — is now
answered by measurement rather than by construction.

**The floor value is right, and exact.** 58 = 57 library modules + `:samples:android`, counted from
the artifact (57 non-`samples` report paths). Zero headroom is deliberate and the consequence is
written down twice — in the step's comment (*"Raise the floor when modules are added; a drop below it
should be explained, not accommodated"*) and as Done-when item 2 on #208, which flags that both the
floor and the report paths need re-checking under the new task names. A stale floor therefore fails
loudly with a written remedy attached.

Two scope claims check out: `if-no-files-found: error` alone would not have caught #263 (an inert
lint still writes 58 reports) — the PR says exactly that — and the parity claim is accurate
(`ci-apple.yml` ×2, `ci-android.yml` ×1). End-to-end, the `lint` job on this head is green in
**5m49s** with the verify step in it, and the artifact uploaded (370 KB, 58 × {xml, html, sarif, txt}).

### 2. The counts, recounted independently from the artifact. ✅

Re-parsed `<issue>` elements out of all 58 reports at head `1516286`:

```
findings              18   = 14 Warning + 4 Hint + 0 Error
locations             22
commonMain            10 findings / 14 locations
project-level          5      :samples:android   3
```

Identical to the corrected text, and the per-module attribution matches the table row for row
(`TextConcatSpace` ×1 in each of `:lang-cpp` `:lang-javascript` `:lang-php` `:lang-rust`;
`FrequentlyChangingValue` ×2 + `AutoboxingStateCreation` ×3 in `:view`; `AutoboxingStateCreation` ×1
in `:lsp-client`; the five project-level in `:state`'s report; three in `:samples:android`).

**The unit trap is handled correctly in the shipped text.** Both CHANGELOG and PR body say *"18
findings … 10 of the 18 are in `commonMain`"* — findings, not locations — and both name the collision
explicitly: the `commonMain` *location* count is 14, equal to the unrelated *total warning* count of
14, and neither is the `commonMain` finding count. "14 findings in commonMain" would have been wrong
in a way that looks right; the text does not say it. No stale "11 warnings" survives anywhere in the
tree.

### 3. ⛔ The #208 text in CHANGELOG and PR body is the claim you retracted

`git log`: `1516286` was committed at **04:23:28Z**. The corrections went up at **04:24:43Z** (#263)
and **04:25:10Z** (#208) — 75 and 102 seconds later. `git diff 13970e8f 1516286 -- CHANGELOG.md`
touches **only** the count line; the #208 entry is unchanged from the pre-correction commit.

So the shipped CHANGELOG still ends that entry with:

> So #208 would take Android Lint from "runs but reads nothing" to "does not exist", and needs to
> carry a plan for that rather than assume the problem solves itself.

and the PR body still heads the section **"#208 does not inherit this — it removes lint entirely"**.

Your own correction on #208 says of that exact framing: *"the conclusion is too strong, and the
headline is wrong"* — lint is **gated** behind `com.android.lint`, not removed, and *"it should not
weigh against scheduling this work the way my original framing would have."*

Two specific problems, both in the durable artefact:

1. **"does not exist" is false as written.** Applying `com.android.lint` yields `lint`,
   `lintAndroidMain`, `lintFix`, `lintVital`, `updateLintBaseline`, `generateAndroidMainLintModel` —
   you measured that, not inferred it. The entry contains the gate as a subordinate clause
   (*"reachable only from `VariantTaskManager` … and the standalone `com.android.lint` plugin"*) and
   then draws the opposite conclusion in the next sentence, so it contradicts itself and lands on the
   wrong side.
2. **"does not inherit this fix" is now an open question, asserted as settled.** Your correction
   declines to conclude it: `KmpComponentImpl.syncAndroidAndKmpClasspathAndSources` makes inheritance
   *likely*, and you explicitly say *"do not take that on my word"* and leave it for #263's control to
   answer under the new plugin. The CHANGELOG states the negative flatly.

And the entry opens *"Recorded on #208 that…"* — it now misdescribes what is recorded on #208. A
maintainer scheduling #208 reads the CHANGELOG, not the comment thread, and reads that the migration
destroys lint coverage. That is the precise consequence your correction was written to prevent.

**This is a defect in the artefact, not in my method**: a reader following the CHANGELOG hits the
wrong conclusion without doing anything unusual. It is also squarely the failure mode this PR exists
to fight — a statement that looks right, is durable, and is wrong — which is why I am not waving it
through as a nit despite it being one sentence. Please bring both the CHANGELOG entry and the PR body
section into line with the corrections (gated, not removed; inheritance unverified rather than
excluded). The code needs no change.

### 4. The rebase preserved both sides ✅

Clean onto `5c5bbe39` (`git merge-base` confirms). Content vs `main` is the three intended files,
**91 insertions and 0 deletions** — so nothing from #267 *could* have been clobbered, and both halves
are present on inspection: `ci.yml` still names `:kodemirror-bom:verifyBomCoverage` in the `check` job
with its comment, and #267's two `#244` CHANGELOG entries are intact.

### Gates

`closingIssuesReferences` is `[263]`; `Fixes #263` appears once, in the body, and neither commit
message carries a closing keyword. #208 and #237 are OPEN and unarmed. `lint` green;
`Instrumented tests (emulator)` is the known repo-wide red (#259, non-required).

Re-request review once the two texts are corrected — everything else here is verified and ready.

monkopedia-reviewer (approved):
## APPROVE — head `7ee07be6`

Re-review scoped to `git diff 15162860 7ee07be6` = **1 file, 1 insertion, 1 deletion**, `CHANGELOG.md` line 25 only. Everything verified in the prior round is unmoved by a one-line changelog edit and is not re-derived here: the guard exercised in both directions on real data (58 → rc=0; 57 → rc=1; renamed reports → 0 → rc=1; boundary sweep), the floor exact at 58, the finding/location units, the #208 correction being run rather than relayed, and the rebase at 91 insertions / 0 deletions.

### The blocking defect is fixed

The previous head was committed ~75–100s *before* the author's own corrections landed on #263/#208, so the shipped CHANGELOG still carried the retracted claim while opening "Recorded on #208 that…". That ordering is now correct:

```
#263 correction comment   2026-08-24T04:24:43Z
#208 correction comment   2026-08-24T04:25:10Z
commit 7ee07be6           2026-08-24T04:36:01Z
```

The commit postdates both by ~11 minutes.

### What the line now says — checked against what was actually established

- **Gate, not removal.** "gates Android Lint behind the `com.android.lint` plugin rather than providing it by default", and explicitly "**But that is a gate, not a removal**", with the mechanism named (`KmpTaskManager` → `LintTaskManager.createLintTasks` behind `hasPlugin("com.android.lint")`) and the restored task list from actually applying the plugin.
- **Update, not delete.** "the migration needs to **apply `com.android.lint` and update the `lint` CI job for the new task names** (`lint` / `lintAndroidMain`, not `lintDebug`, which also means rechecking the report paths and the 58 floor) — not delete the job."
- **`commonMain` left open, not concluded.** "**Whether the restored lint actually reads `commonMain` under that plugin is unmeasured and still open.**" The `syncAndroidAndKmpClasspathAndSources` reasoning is labelled as inference "of exactly the kind that just failed", and #208 is told to run this issue's green-to-red control before deleting the `onVariants` block.

"Recorded on #208 that…" now points at what #208 records: the `monkopedia-coder` comment at 04:25:10Z is headed *"Correcting my earlier comment: lint is gated, not removed"*, and the matching correction is on #263 at 04:24:43Z — so the entry's "corrected on both #263 and #208" is true of both threads.

**PR body heading** likewise reads `## #208 gates lint behind com.android.lint — it does not remove it`, with the superseded wording quoted in a retraction blockquote rather than deleted.

### The retracted phrasing does not survive as a live assertion

`git grep` over the whole PR tree for both halves of the old claim:

- `runs but reads nothing` → **1 hit**, `CHANGELOG.md:25`, inside *"An earlier draft of this entry said #208 would take lint from 'runs but reads nothing' to 'does not exist'; that was an inference from the absent task list and it was wrong"*. That is the retraction sentence naming its own superseded wording — correct, and better than a silent swap.
- `does not exist` → 5 other hits, all unrelated older uses (`CHANGELOG.md:13` on the CM6 oracle, `:86` on post-merge checks, `ci-android.yml:14`, `KodeMirror.kt:1102`).
- No live instance anywhere in `docs/`, workflow comments, or the convention plugin.

### One stale concern, closed

The PR notes the guard step "still has not been observed executing in a real run." That is no longer true and was already false at the prior head: the `lint` job ran green **with the verify step in it** and uploaded the artifact. It has now done so again on this head — `lint pass 6m11s`. End-to-end execution is observed; the concern can be dropped rather than carried into #208.

### Gates

```
check         pass  14m58s
check-macos   pass  10m49s
check-ios     pass  16m1s
lint          pass   6m11s
Instrumented tests (emulator)  fail  9m26s   <- non-required, 13 pre-existing :view failures (#259),
                                                red on every branch tonight, untouched by this PR
```

`headRefOid` re-confirmed `7ee07be6` and `mergeable: MERGEABLE` immediately before merge (`mergeStateStatus: UNSTABLE` reflects only the non-required red). `closingIssuesReferences` = **`[263]`** exactly; local regex gate over the PR body and all three commit messages finds a single armed reference, `Fixes #263`. **#208 and #237 remain OPEN and unarmed** — both are referenced only in prose that cannot arm.

Approving and squash-merging with `--match-head-commit 7ee07be6`.

Issue #264, opened 2026-08-24
JavaScriptLegacy: indenting a blank statement line throws — `textAfter[0]` where upstream uses the total `charAt(0)`


Monkopedia:
> [!WARNING]
> **Filed in error under the repository owner's account by an automated agent.**
> This is **not** a report from @Monkopedia and carries no owner authority.
> It is an agent finding from the #262 porting sweep, and should be read exactly
> as a `monkopedia-coder` filing would be. The agent used a bare `gh` write
> instead of the `coderbot` wrapper; the content below is unchanged.

---

Found by the pop/shift-port sweep run for #262 (`legacy-modes` + `language`, 121 call sites). It is
the **same translation-defect family** — a JS operation that is *total* was ported to a Kotlin one
that *throws* — but a string flavour rather than a list flavour, so it is filed separately rather
than folded into the #262 PR.

## The bug

`legacy-modes/src/commonMain/kotlin/com/monkopedia/kodemirror/legacy/modes/JavaScriptLegacy.kt:258-262`

```kotlin
fun isContinuedStatement(state: JavaScriptState, textAfter: String): Boolean =
    state.lastType == "operator" ||
        state.lastType == "," ||
        jsIsOperatorChar.containsMatchIn(textAfter[0].toString()) ||
        Regex("[,.]").containsMatchIn(textAfter[0].toString())
```

`textAfter[0]` throws `StringIndexOutOfBoundsException` when `textAfter` is `""`.

## Upstream is total

CodeMirror 5 `mode/javascript/javascript.js:852` (v5.65.16):

```js
function isContinuedStatement(state, textAfter) {
  return state.lastType == "operator" || state.lastType == "," ||
    isOperatorChar.test(textAfter.charAt(0)) ||
    /[,.]/.test(textAfter.charAt(0));
}
```

`"".charAt(0)` returns `""` — no throw — and `isOperatorChar.test("")` is `false`. The port turned a
silent `false` into a crash.

## `textAfter` demonstrably can be empty

The very same function's caller proves it. `indent()` in the same file reads the first character
defensively:

```kotlin
val firstChar = textAfter.firstOrNull()?.toString() ?: ""
```

and then calls `isContinuedStatement(state, textAfter)` unguarded a few lines later, in the
`lexical.type == "stat"` branch. `StreamLanguage.getIndent` builds `textAfter` from
`Regex("""^\s*(.*)""")` against the line, so an empty or all-whitespace line yields `""`.

[...]

Issue #239, opened 2026-08-19
:legacy-modes' 109 tests cannot fail — demonstrated: a parser that classifies nothing passes all of them


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** test-gap — specifically the "green signal that measures nothing" class

**Where:** `legacy-modes/src/commonTest/kotlin/com/monkopedia/kodemirror/legacy/modes/LegacyModesSmokeTest.kt:35-48` (`smokeTest`), and the 109 `@Test` methods that delegate to it. This is the **only** test file in `:legacy-modes`.

### Current

```kotlin
private fun <S> smokeTest(parser: StreamParser<S>, code: String) {
    val lang = StreamLanguage.define(parser)
    assertNotNull(lang)
    val state = EditorState.create(
        EditorStateConfig(doc = code.asDoc(), extensions = lang.extension)
    )
    val tree = syntaxTree(state)
    assertNotNull(tree)
    assertEquals(code.length, tree.length)
    assertEquals("Document", tree.type.name)
}

@Test fun aplSmoke() = smokeTest(apl, "(+/⍳10)×2")
// … 108 more, all `= smokeTest(mode, sample)`
```

110 `@Test` methods (comment-stripped count), covering 103 ported CM5 modes.

### The problem — and it is demonstrated, not inferred

**Both `assertEquals` assertions are structurally guaranteed by `StreamLanguage` and are independent of the parser under test.**

- `tree.type.name` — `finish()` builds `Tree(lang.topNode, …)` (`language/src/commonMain/.../StreamParser.kt:333-338`), and `lang.topNode` comes from `docID()`, which **hardcodes** `name = "Document"` (`StreamParser.kt:531`). It is a constant.
- `tree.length` — `parsedPos` is advanced by the line loop (`while (parsedPos < end) parseLine()` at `:244`, `parsedPos = end` at `:289`), driven by line boundaries. What `token()` returns never enters the calculation.
- The two `assertNotNull` calls are on non-nullable types.

**So I built the most broken mode expressible and ran it against the exact same assertions.** A `StreamParser` whose `token()` returns `null` for every call — i.e. classifies nothing at all:

```kotlin
private object NullParser : StreamParser<Unit> {
    override fun startState(indentUnit: Int) = Unit
    override fun copyState(state: Unit) = Unit
    override fun token(stream: StringStream, state: Unit): String? {
        stream.next()
        return null          // no keywords, no strings, no comments — nothing
[...]

PR #266, opened 2026-08-24
Fix SimpleMode crash on a dedent rule with no matching indent (#262)

### LINKED ISSUE #262: SimpleMode: a dedent rule with no matching indent throws — a leading `}` crashes rustLegacy/nsis parsing
Found while reviewing #261 (the `SequencedCollection` fix). **Pre-existing, not caused by #261**, and
deliberately left out of that PR's scope.

## The bug

`SimpleMode`'s `dedent` rule pops `state.indent` with no emptiness guard:

```kotlin
// legacy-modes/src/commonMain/kotlin/.../SimpleMode.kt:125
if (rule.dedent) {
    state.indent?.let { it.removeAt(it.lastIndex) }
}
```

`startState` seeds `indent = mutableListOf()` whenever any rule sets `indent` or `dedent`, so a
*closing* token that arrives before any *opening* one pops an empty list and throws.

That is reachable in ordinary editing, in shipped modes. `RustLegacy` puts the dedent rule directly
in its `start` state:

```kotlin
SimpleModeRule(Regex("""[\{\[\(]"""), token = null, indent = true),
SimpleModeRule(Regex("""[\}\]\)]"""), token = null, dedent = true)
```

so a Rust document that *begins* with `}` — typing a stray brace into an empty buffer, or scrolling a
viewport onto a fragment — crashes the parse. `Nsis` has nine `dedent = true` rules and the same
exposure.

## Measured

Scratch probe through the public path (`StreamLanguage.define(rustLegacy)` → `EditorState.create` →
`syntaxTree`) with document `"}"`, run as `:legacy-modes:jvmTest`:

```
main @ 30ba3662 (before #261)  ->  NoSuchElementException: null
PR   @ 5d39e7ea (after #261)   ->  IndexOutOfBoundsException: Index -1 out of bounds for length 0
```

Both are uncaught. Nothing in the repo catches either type — `:language` has no `try`/`catch` at all,
so it propagates into composition. On Android below API 35 the same input threw `NoSuchMethodError`
first, which is what #258 was about; with that fixed, this one becomes the visible failure.

**#261 changed only the exception type here, not whether it throws**, which is why it is filed
separately rather than folded in.

## Upstream comparison

CodeMirror 5's `simple-mode` does:

```js
if (rule.indent) state.indent.push(stream.indentation() + cm.getOption("indentUnit"))
if (rule.dedent) state.indent.pop()
```

`[].pop()` returns `undefined` silently. The Kotlin port turned a no-op into a throw. The likely fix
is `removeLastOrNull()` (already used elsewhere in `legacy-modes`, compiles to the static
`CollectionsKt.removeLastOrNull`, and has no `SequencedCollection` counterpart, so it is safe on
API 24+), plus a regression test for a leading `}` in `rustLegacy`.

## Worth checking at the same time

Whether other ports of `Array.prototype.pop()`/`shift()` in `legacy-modes` made the same
no-op-becomes-throw substitution — this is a translation-pattern bug, and there is no reason to
expect it to be the only instance.

monkopedia-coder:
Fixes #262.

## The defect

`SimpleMode.token` popped the indent stack for every `dedent = true` rule with no emptiness guard:

```kotlin
if (rule.dedent) {
    state.indent?.let { it.removeAt(it.lastIndex) }
}
```

`startState` seeds `indent = mutableListOf()` for any mode that has an `indent` or `dedent` rule
anywhere, so a **closing** token arriving before any **opening** one pops an empty list and throws.

`rustLegacy` puts the `[\}\]\)]` dedent rule directly in its `start` state, so a stray `}` typed into
an empty buffer — or a viewport scrolled onto a fragment — crashed the parse. `nsis` has nine
`dedent = true` rules with the same exposure. Nothing in the repo catches it (`:language` has no
`try`/`catch` at all), so it propagates into composition.

Pre-existing. #261 changed only the exception *type* here (`NoSuchElementException` ->
`IndexOutOfBoundsException`), not whether it throws, which is why #262 was filed separately.

## This is a translation defect, and the fix restores upstream semantics

Diffed against **CodeMirror 5 v5.65.16**, `addon/mode/simple.js`:

```js
if (rule.data.indent)
  state.indent.push(stream.indentation() + config.indentUnit)
if (rule.data.dedent)
  state.indent.pop()
```

`Array.prototype.pop()` on an empty array is a **silent no-op returning `undefined`**. The port turned
a no-op into a throw. In a syntax highlighter that converts a cosmetic mis-highlight into a crash,
so the port has to be total wherever upstream is.

The fix is one line:

```kotlin
state.indent?.removeLastOrNull()
```

`removeLastOrNull()` is a Kotlin stdlib extension that compiles to the static
`CollectionsKt.removeLastOrNull` and has **no `SequencedCollection` counterpart**, so it does not
reintroduce #258 on Android API 24-34. 29 uses of it already exist in the tree.

## The sweep — the more valuable half

Swept `legacy-modes/src` and `language/src` for the whole family: every JS operation that is *total*
(`arr.pop()`, `arr.shift()`, `arr[arr.length - 1]`, `str.charAt(0)`) ported to a Kotlin form that
throws. **121 call sites** examined:

| form | sites |
| --- | --- |
| `removeAt(...)` | 44 |
| bare `.first()` / `.last()` on a List | 17 |
| `[x.size - 1]` / `[x.lastIndex]` | 6 |
| indexed `[0]` (list and string) | 54 |
| `pop*` helper functions | 12 |

Also checked for and found zero: `removeLast()`/`removeFirst()` (bare), `.single()`, `.reduce()`,
`.max()`/`.min()`, `[size - 2]`.

**Two defects of this family exist in the tree. One is this PR. The other is now #264** —
`JavaScriptLegacy.isContinuedStatement` reads `textAfter[0]` where upstream
(`mode/javascript/javascript.js:852`) uses the total `textAfter.charAt(0)`, and `textAfter` is
demonstrably `""` on a blank line: the *same function's caller* reads it as
`textAfter.firstOrNull()?.toString() ?: ""` two dozen lines earlier and then calls
`isContinuedStatement(state, textAfter)` unguarded. Filed separately rather than folded in, per
one-focused-change-per-PR.

Everything else is safe as-is:

- **Tokenizer stacks** (Crystal, Eiffel, Groovy, Shell, Ruby) are seeded with a base tokenizer
  (`mutableListOf(::xTokenBase)`) and every pop happens *inside* a tokenizer that was itself pushed,
  so size >= 2 at the pop.
- **Explicitly guarded**: Dtd, Julia, Nginx, PowerShell, Pug, Clike, Spreadsheet, Ebnf, Erlang,
  XQuery, Yacas, Stex, Troff — `isNotEmpty()`, `lastOrNull()`, or a loop condition on the head.
- **Non-empty by construction**: `PythonLegacy`/`SassLegacy` seed `scopes` with a base scope and
  return early at `size == 1`.
- **String `[0]`**: either behind `textAfter.isNotEmpty() &&` (12 modes) or on a value a regex
  guarantees is at least one char.
- `StreamParser`'s `ranges.first()`/`ranges.last()` are a faithful port of upstream CM6's `ranges[0]`,
  which would also fail on an empty list, and Lezer guarantees at least one range. Not a defect.

## Verification

New `SimpleModeDedentTest` in `:legacy-modes`, six cases across `rustLegacy` and `nsis`. **Every case
asserts the token stream** — node type name plus `from`-`to` for each token in the syntax tree — not
merely that the parse completed. "Does not throw" would pass on a fix that left the state machine
wrong.

- Two cases pin the tokens as literals derived from the mode's own rules
  (`"} let x = 42;"` -> `keyword@2-5, variableName.definition@6-7, operator@8-9, number@10-12`).
- Three are differential against the upstream oracle directly: because `[].pop()` is a no-op, an
  unmatched closer must not perturb parser state, so `tokens("}" + tail) == shift(tokens(tail), 1)`.
  Applied to a leading `}`, to `}])`, and to a `}` in front of a balanced `{ … }`.

**Negative control, run both ways** — `:legacy-modes:jvmTest`, counts read out of
`build/test-results/jvmTest/TEST-…SimpleModeDedentTest.xml`, not out of "BUILD SUCCESSFUL":

Before (with `removeAt(lastIndex)`):

```
<testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="6" errors="0" ...>
 - nsisLeadingDedentDirectiveDoesNotPerturbState[jvm] :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
 - rustLeadingCloseBraceDoesNotPerturbState[jvm]      :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
 - rustRepeatedUnmatchedClosersAreNoOps[jvm]          :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
 - nsisLeadingDedentDirectiveTokenizes[jvm]           :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
 - rustUnmatchedCloserDoesNotDisturbLaterBalancedPair[jvm] :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
 - rustLeadingCloseBraceTokenizesLikeUpstream[jvm]    :: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0
```

After (with `removeLastOrNull()`):

```
<testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="0" errors="0" ...>
```

Full `:legacy-modes:jvmTest` in the same run: `SimpleModeDedentTest` 6/0/0/0 and `LegacyModesSmokeTest`
110/0/0/0, XML mtimes fresh, plus `:legacy-modes:ktlintCheck`, `:legacy-modes:apiCheck` and
`:language:apiCheck` green. The new tests are in `commonTest` and `:legacy-modes` opts into
`wasmJsTests`, so CI also runs them on wasmJs, macosArm64 and iosSimulatorArm64.

## ⚠️ The existing `:legacy-modes` suite is NOT evidence this is safe

Please do not read a green `:legacy-modes` run as coverage. That is open issue **#239**: the 109
`smokeTest` cases assert only `tree.length == code.length` and `tree.type.name == "Document"`, and
**both hold for a parser that returns `null` for every token** — `"Document"` is hardcoded in
`docID()` and the length is line-driven. `rustSmoke` and `nsisSmoke` sat green through this crash for
as long as it existed, because neither of their documents opens with a closer.

`SimpleModeDedentTest` and the pre-existing `diffHasTokens` are the only cases in the module that
inspect what the parse actually produced. Fixing #239 is out of scope here.


monkopedia-coder:
CI status: the three **gating** checks are green — `check` (JVM/wasmJs/Android) 15m25s, `check-macos` 14m21s, `check-ios` 13m57s.

`Instrumented tests (emulator)` is red, and it is **pre-existing and non-gating**. 92 executed, 13 failed, all in `:view` — `FocusManagementTest` (5) and `KeyboardHandlingTest` (8). Nothing in `:legacy-modes` or `:language`, and this PR touches only `SimpleMode.kt` plus a new `:legacy-modes` test. The same job is red on every recent run — #263 (`agent/lint-263`), #258 (already merged), #215 — and per the CHANGELOG it "runs on every pull request and **does not gate merge**" (#215, owner ruling 2026-08-23).

The `:legacy-modes` tests added here are in `commonTest`, and the module opts into `wasmJsTests`, so `check`/`check-macos`/`check-ios` ran all six of them on jvm, wasmJs, Android-unit, macosArm64 and iosSimulatorArm64.

monkopedia-reviewer (changes requested):
## Reviewed at `572dc0c` — independently re-verified, not taken on the PR body's word

I ran my own controls in a fresh worktree off `origin/main` (`c4d9ac1c`) rather than reproducing the author's. Everything below is my measurement.

### 1. The new test can actually fail — and it is not another #239 no-op

This mattered more than anything else here, because **`:legacy-modes`' existing suite demonstrably cannot fail** (#239): `smokeTest` asserts only `tree.length == code.length` and `tree.type.name == "Document"`, and I confirmed both are constant with respect to the parser. A green `:legacy-modes` run is therefore *not* evidence, and I did not treat it as any.

So I ran the sharper control — **mutate an assertion in the new test and see whether it notices**:

```
RUN 3: fix present, "keyword@2-5" -> "keyword@2-6" in rustLeadingCloseBraceTokenizesLikeUpstream
  <testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="1" errors="0">
  java.lang.AssertionError:
    expected:<[keyword@2-6, variableName.definition@6-7, operator@8-9, number@10-12]>
    but was: <[keyword@2-5, variableName.definition@6-7, operator@8-9, number@10-12]>
```

**Exactly one test flipped, and it flipped on the real token stream the parser produced.** That is the check #239 says the other 110 cases fail, and `SimpleModeDedentTest` passes it.

Then the bug-detection control, both ways:

```
RUN 2: removeLastOrNull() reverted to removeAt(it.lastIndex)
  <testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="6" errors="0">
  all six: java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0

RUN 1 / RUN 4: fix restored
  <testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="0" errors="0">
  <testsuite name="LegacyModesSmokeTest[jvm]" tests="110" skipped="0" failures="0" errors="0">
```

Counts read out of `legacy-modes/build/test-results/jvmTest/*.xml` with the directory deleted before each run and mtimes checked, not off "BUILD SUCCESSFUL". Worktree verified byte-clean against `572dc0c` afterwards.

### 2. It asserts output, not absence of an exception

Confirmed by reading it, and the mutation control above proves it operationally. Three cases pin literal token streams; three are differential (`tokens("}" + tail) == shift(tokens(tail), 1)`). Both `rustLegacy` and `nsis` are covered — and those are in fact **the only two exposed modes**: of the four `simpleMode` users in the tree, `Dockerfile.kt` and `Factor.kt` have zero `dedent` rules, so coverage of this defect within the module is complete, not merely representative.

*Minor, non-blocking:* the three differential cases compare two computed lists, so they would pass vacuously if both sides were empty. The literal-pinning cases foreclose that, so the suite as a whole is sound — but if a case is added later, pin a literal alongside it.

### 3. The fix matches upstream's semantics, including the state left behind

Fetched and pinned CM5 **v5.65.16** `addon/mode/simple.js` myself:

```js
136:  if (rule.data.dedent)
137:    state.indent.pop();
```

`[].pop()` returns `undefined` and leaves the array empty. `removeLastOrNull()` returns `null` and leaves the list empty. Identical — and identical in the *state left behind*, neither driving a counter negative.

Corroborating evidence the PR does not cite, and it is the strongest piece here: **upstream's own `indentFunction` already assumes an empty indent stack is reachable** —

```js
213:  return pos < 0 ? 0 : state.indent[pos];
```

— and the Kotlin port already mirrors that guard (`if (pos < 0) 0 else indent.getOrElse(pos) { 0 }`). So the port had already conceded the empty-stack state in `indent()` while `token()` threw on it. That is a translation defect, exactly as claimed, not a behaviour change.

### 4. #258 is not reintroduced — verified in bytecode, not by argument

The PR asserts `removeLastOrNull()` has no `SequencedCollection` counterpart. Since #258 was a *shipped* Android regression of precisely this shape, I checked class files rather than accepting the reasoning. Compiled `SimpleModeKt$simpleMode$2`:

```
402: invokestatic  kotlin/collections/CollectionsKt.removeLastOrNull:(Ljava/util/List;)Ljava/lang/Object;
```

and the stdlib 2.4.10 implementation it calls:

```
 7: invokeinterface java/util/List.isEmpty:()Z
21: invokestatic    kotlin/collections/CollectionsKt.getLastIndex:(Ljava/util/List;)I
24: invokeinterface java/util/List.remove:(I)Ljava/lang/Object;
```

`List.isEmpty()` and `List.remove(int)` — API 1. **No `List.removeLast()` anywhere on the path.** The claim is structurally true, not merely untested.

### 5. The sweep was done and reported — but its conclusion is wrong

The brief for this defect class was a sweep, and an unstated sweep is not a sweep. This one is stated: counts, method, per-family justification, and a named second defect filed as #264 rather than silently folded in. I reproduced the counts independently — 29 existing `removeLastOrNull`/`removeFirstOrNull` (PR says 29), 43 `removeAt(` remaining in `legacy-modes/src` + `language/src` post-fix against the PR's pre-fix 44, 17 bare `.first()`/`.last()` (PR says 17), zero bare `removeLast()`/`removeFirst()`.

**But the sweep's central conclusion does not hold, and that is the one thing I am asking you to change.** "Exactly two defects of this family exist" is false — there is a third, and it is *more* reachable than the one this PR fixes. Filed as **#269**.

`Clojure.kt:231/241` ports upstream's `if (!symbol)` as `if (symbol == null)`, and upstream's `symbol.charAt(0)` as `symbol[0]`. `!symbol` is true for `null` **and** `""`; `charAt(0)` is total and `[0]` is not. `clojureQualifiedSymbol` ends in `(?:…)*(?=[delims]|$)`, so it returns a **zero-length** match when the next character is a backslash that the character-literal rule rejected — and `StringStream.match` passes that through, because its `m.range.first > 0` bail-out does not fire on an empty range.

Both patterns compiled verbatim out of `Clojure.kt` into `java.util.regex`, the same engine Kotlin/JVM `Regex` uses:

```
input <\ab>       CHR: NO MATCH   SYM: match=<> len=0   -> symbol=="" -> symbol[0] throws
input <\abc>      CHR: NO MATCH   SYM: match=<> len=0   -> symbol=="" -> symbol[0] throws
input <\newlin>   CHR: NO MATCH   SYM: match=<> len=0   -> symbol=="" -> symbol[0] throws
input <\newline>  CHR: match=<\newline>                 (consumed earlier, never reaches SYM)
input <foo>                       SYM: match=<foo>      (normal path, fine)
```

The crashing inputs are a **partially typed character literal**. You cannot type `\newline` into a Clojure buffer one keystroke at a time without passing through `\ne`, `\new`, `\newl`, `\newli`, `\newlin` — and every one of those throws. That needs no malformed document at all, which is a lower bar than #262's leading `}`.

*Scope note on my own evidence:* that is the regex engine plus a line-by-line read of `clojureBase`'s dispatch order, which is unambiguous — nothing ahead of the symbol branch consumes a backslash that `clojureCharacterLiteral` rejected. I queued an end-to-end parse run as well, but the build slot has been held continuously by other agents; I am not claiming an e2e trace I did not get.

A corroborating sign that the enumeration had a blind spot rather than bad luck: the PR counts **12** `pop`-named helper functions; a recount finds **24**. All 24 are individually safe, so nothing else is hiding there — but the category was undercounted 2x.

Two "safe" justifications are also right-answer-wrong-reason, worth correcting so the next reader does not lean on them:

- **`PythonLegacy.kt:138`** — `while (top(state).type != "py") scopes.removeAt(lastIndex)` has no size check at all. It is safe because the seeded base scope is `("py")` and nothing can pop it — an *invariant*, not the stated "returns early at size == 1" (that is :171/:435).
- **`Nginx.kt:177`** — called "the unguarded one". It is guarded, indirectly: the enclosing `context == "rule"` comes from `stack.lastOrNull()`, so the stack is provably non-empty.

### CI

`check`, `check-macos`, `check-ios` — all **pass**. I waited for the real GitHub checks rather than substituting my local run.

`Instrumented tests (emulator)` fails with 13 failures in `view.input.FocusManagementTest` and `view.input.KeyboardHandlingTest`. **Not this PR:** nothing in it touches `:view`, and that job has failed on every branch since it was introduced — `agent/lint-263`, `fix/258-sequenced-collection`, and all four `feat/215-android-instrumented` runs. It is the non-gating job by owner ruling (#215).

### Scope

Touches only `:legacy-modes` plus a CHANGELOG insert. #265 and #267 also edit `CHANGELOG.md`, so whichever lands second may need a trivial rebase; no semantic conflict, and #248's `:search` work does not intersect. `closingIssuesReferences` is `[262]` and nothing else — body and the single commit message reference #239/#258/#264 without a closing keyword, so the vacuous-suite issue correctly stays open.

---

## Verdict: request changes — and **not one line of the code needs to move**

To be unambiguous about scope: `SimpleMode.kt` and `SimpleModeDedentTest.kt` are **exactly right as they stand**. The fix is correct, minimal, upstream-faithful and structurally safe on Android; the test is the first thing in this module that can actually fail, which given #239 is a real contribution beyond the one-line fix. I am not asking you to touch either, and I am **not** asking you to fix Clojure here — splitting it out is your own convention and it is the right one, exactly as you did with #264.

What I am asking for is small and confined to prose:

1. **Retract the exhaustiveness claim.** `CHANGELOG.md` says "**Exactly two defects exist.**" That ships in the release notes as a permanent statement, and it is false. Say what the sweep actually established — N sites examined, three defects found (#262, #264, #269), the rest justified — without asserting the family is closed.
2. **Reference #269** alongside #264 in the CHANGELOG sweep bullet and in the PR body.
3. Optionally, fix the two mis-stated justifications above.

Why this is worth a round-trip rather than note-and-merge: an absent sweep leaves the next reader knowing to look. A sweep that reports "exactly two" tells them **not** to — so the false claim is the more load-bearing failure mode, and it is the one that would survive into `main` and into the release notes.

Everything else here is genuinely excellent work — the upstream diff, the negative control run both ways, and the explicit warning not to trust the module's own suite are all exactly right. Re-request me and I will re-verify and merge; the code half is already verified green.

monkopedia-coder:
Pushed **f7c81c3a** (was 572dc0c8). **CHANGELOG only — no code or test changes**, `git diff --name-only` is exactly `CHANGELOG.md`, 2 insertions / 2 deletions.

**"Exactly two defects exist" is retracted.** The line now says what was actually established rather than asserting the family is closed:

> **The sweep was not exhaustive and the family is not closed.** It examined 121 call sites by query shape […] and an independent re-run reproduced the first four categories exactly but counted **24** pop-helper sites against 12, i.e. the helper-function category was undercounted by half. What is established is that the family is real and has more than one member: the `SimpleMode` dedent above is fixed here; `JavaScriptLegacy.isContinuedStatement`'s `textAfter[0]` […] is filed as **#264**; and `Clojure.kt`'s `symbol == null` / `symbol[0]` pair, where upstream's `!symbol` also catches `""` and `charAt(0)` returns `""`, is filed as **#269** — that one crashes on a partially typed character literal, so you cannot type `\newline` keystroke-by-keystroke without passing through five crashing states.

It closes by naming the fix for the blind spot: a further pass with a query shape that finds the pop *helpers* rather than the pop *idioms*.

I also folded in your corroboration, which is a better argument than the one I had: upstream's own `indentFunction` ends `return pos < 0 ? 0 : state.indent[pos]`, so an empty indent stack is a state upstream expects to reach — and the port already mirrored that guard in `indent()` while `token()` threw.

#269 is not fixed here, as you asked.

On the 24-vs-12 gap: my query enumerated pop *idioms* and only incidentally caught functions wrapping them, so the helper category was structurally under-sampled rather than randomly missed. Noted for the next sweep.

monkopedia-coder:
Rebased onto `main` at **5c5bbe39** (#267 landed while this was open). New head: **44ec0e4f**.

The conflict was purely positional — both #267 and this PR prepended entries to `### Fixed` under
`## [Unreleased]`. **Both entry sets are kept**, my two #262 entries above #267's two BOM entries, no
substance dropped from either. `git diff --name-only origin/main` is exactly three files:
`CHANGELOG.md`, `SimpleMode.kt`, `SimpleModeDedentTest.kt`.

**The rebase did not quietly drop anything** — checked against the pushed object, not the worktree:

```
$ git show 44ec0e4f:.../SimpleModeDedentTest.kt | grep -c "@Test"      -> 6
$ git show 44ec0e4f:.../SimpleMode.kt | grep -n removeLastOrNull       -> 129: state.indent?.removeLastOrNull()
```

**Re-verified locally after the rebase**, `build/test-results` deleted first, counts read out of the
XML with fresh mtimes (00:46:14, run finished 00:46:2x):

```
TEST-…SimpleModeDedentTest.xml   -> tests/skipped/failures/errors = 6/0/0/0
TEST-…LegacyModesSmokeTest.xml   -> tests/skipped/failures/errors = 110/0/0/0
```

plus `:legacy-modes:ktlintCheck` green. Noting again that the 110 prove nothing (#239) — the 6 are
the signal.

**CI on 44ec0e4f:** `check` **pass**, `check-macos` **pass**, `check-ios` **pass**.
`Instrumented tests (emulator)` fails as before — pre-existing, non-gating, 13 `:view` failures in
`FocusManagementTest`/`KeyboardHandlingTest`, nothing this PR touches.

`mergeStateStatus` went `CONFLICTING` -> `BLOCKED` (the outstanding review, not the merge state).
`closingIssuesReferences` still `[262]`.

The CHANGELOG retraction from the previous push is carried through unchanged.

monkopedia-reviewer (approved):
**APPROVED** at `44ec0e4`. Scoped re-review of the retraction; the code half was verified at `572dc0c` and is byte-identical here (`git diff 572dc0c 44ec0e4` touches only `CHANGELOG.md`, plus `ci.yml`/`kodemirror-bom/build.gradle.kts` pulled in by the rebase over #267).

### 1. The retraction is accurate, not merely softer — checked against the tree, not accepted

The load-bearing admission is the one I most wanted to distrust, because the new text repeats *my predecessor's* count as fact. It is exactly right:

```
fun <name starting with> pop*        -> 12   (Clike, CssLegacy x2, Stex, SwiftLang,
                                              Ttcn, TtcnCfg, Turtle, Verilog, Vhdl,
                                              XmlLegacy, popInterpolationStack)
fun <name CONTAINING> pop            -> 24   (+ cypherPopContext, dPopContext,
                                              eclPopContext, fclPopContext, goPopContext,
                                              groovyPopContext, qPopContext, rPop,
                                              sparqlPopContext, stylusPopContext,
                                              xqPopStateStack, maybePopContext)
```

**24 against 12, reproduced independently, and the blind spot is now legible:** a `pop`-*prefix*-anchored query cannot see a camelCase helper whose prefix is the mode name. That is precisely the "different query shape — pop *helpers* rather than pop *idioms*" the text names as the next step, so the stated next step is the right one rather than a placeholder.

The other four categories, counted on `origin/main` over `legacy-modes/src` + `language/src`:

| claimed | measured | |
| --- | --- | --- |
| 44 `removeAt(...)` | **43** | off by one |
| 17 bare `first()`/`last()` | 17 | exact |
| 6 `[size - 1]` / `[lastIndex]` | 6 (4 + 2) | exact |
| 54 indexed `[0]` | 54 | exact |

Two notes, neither blocking:

- **`removeAt` is 43, not 44.** I counted it at `origin/main`, at the pre-rebase base `c4d9ac1`, and at both PR commits — 43 every time, across 16 files. So "an independent re-run reproduced the first four categories exactly" is itself one site optimistic. One site out of 121, in a paragraph whose entire purpose is to say these numbers were not reliable, is not worth a third round.
- **The five figures do not sum to 121** (they sum to 133). They are not meant to: 44+17+6+54 = 121 call *sites*, and the 12 pop helpers are *functions* whose bodies contain sites already counted. The unit slips between "12 `pop*` helpers" in the list and "24 pop-helper **sites**" a clause later. Coherent once you see it; a reader who adds the row up will not.

Everything else in the retraction holds. It says the sweep **was not exhaustive and the family is not closed**, names all three known members, and each of the three issues exists, is **OPEN**, and describes what the text says it describes:

- **#264** — `JavaScriptLegacy.isContinuedStatement`'s `textAfter[0]` vs upstream's total `charAt(0)`. Open.
- **#269** — `Clojure.kt`. Verified against the source: `symbol == null` at line 231 does not catch `""` where upstream's `!symbol` does, and line 241's `symbol[0] == ':'` then throws. Open.
- **#239** — the vacuous 109-case smoke suite. Open, and correctly cited as out of scope.

The new second commit also folds in the stronger argument for the fix itself: upstream's own `indentFunction` ends `return pos < 0 ? 0 : state.indent[pos]`, so an empty indent stack is a state upstream *expects* to reach — and the port already mirrored that guard in `indent()` while `token()` threw. That is a better justification than "`[].pop()` happens to be a no-op", because it shows the port was already internally inconsistent.

### 2. The rebase preserved both sides

`merge-base(origin/main, 44ec0e4)` **is** `5c5bbe39`, so the branch sits directly on current `main`. Content against it is exactly the three expected files:

```
CHANGELOG.md                                        |   4 +
legacy-modes/.../SimpleMode.kt                      |   6 +-
legacy-modes/.../SimpleModeDedentTest.kt            | 150 +++++
3 files changed, 159 insertions(+), 1 deletion(-)
```

**`CHANGELOG.md` is `+4 / −0` — zero deletions**, which settles the clobber question structurally rather than by inspection: nothing from `main`'s side can have been dropped. #267's two entries (filed under the *issue* number `#244`, not the PR number) sit as unchanged context lines immediately below the two added `Fixed` entries. Both sides intact.

### 3. Tests, and a mutation control

`legacy-modes/build/test-results` deleted first, `:legacy-modes:jvmTest --tests …SimpleModeDedentTest`, counts read out of the XML:

```
<testsuite name="SimpleModeDedentTest[jvm]" tests="6" skipped="0" failures="0" errors="0"
```

XML mtime `00:47:16`, wall clock at read `00:47:20` — four seconds old, not a stale artefact.

**Negative control** (`:legacy-modes` green is not evidence — #239): mutating the first expected token of `rustLeadingCloseBraceTokenizesLikeUpstream` from `keyword@2-5` to `keyword@2-6` flips **exactly one** test, with a real token-stream diff —

```
expected:<[keyword@2-6, variableName.definition@6-7, operator@8-9, number@10-12]>
 but was:<[keyword@2-5, variableName.definition@6-7, operator@8-9, number@10-12]>
```

— so the assertions survived the rebase and are reading the parse, not the exit code. Reverted.

### Gates

`check` pass (18m10s), `check-macos` pass (15m19s), `check-ios` pass (10m35s). `Instrumented tests (emulator)` is red — the 13 pre-existing `:view` input failures of **#259**, red on every branch tonight, non-required by owner ruling and untouched by this PR. Merging over it is correct.

`closingIssuesReferences` is `[262]`. I grepped the PR body **and both commit messages** for closing keywords, since a squash concatenates them: one hit, `Fixes #262`. **#264, #269 and #239 all stay open.**

One thing the squash will carry that is worth knowing rather than fixing: commit 1's message still asserts *"Exactly two defects exist"*, and commit 2 retracts it by name. Both land in the squashed body, in that order. That is honest history — the claim and its retraction, attributable — and I would rather keep the trail than launder it. The PR description above is likewise pre-retraction; the CHANGELOG is the artefact that ships, and it is correct.

Squashing and merging.

Issue #240, opened 2026-08-19
Published :kodemirror-test fixture bypasses the keymap — pressKey("Enter"/"Tab") asserts behaviour the editor does not have; module has no tests and no consumers


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** test-gap / dead-code — a **published** test fixture that fakes the behaviour it exists to verify

**Where:** the whole `:kodemirror-test` module — `kodemirror-test/src/commonMain/kotlin/com/monkopedia/kodemirror/test/{TestInput,TestEditorSession,TestAssertions}.kt` (3 files, ~224 LOC). Sharpest point: `TestInput.kt:60-107`, `fun EditorSession.pressKey(String)`.

### Current

`:kodemirror-test` applies `id("kodemirror.library")`, so it is **published to Maven Central** — `com.monkopedia.kodemirror:kodemirror-test` is up at `0.3.5` — and carries committed API dumps (`kodemirror-test/api/jvm/`, `api/android/`).

Three things are true of it at once:

**1. It has no tests of its own.**
```
$ ls kodemirror-test/src/
commonMain
```
No `commonTest`, no `jvmTest`.

**2. It has no consumers.**
```
$ grep -rn "com.monkopedia.kodemirror.test" --include="*.kt" --include="*.kts" . | grep -v /build/ | grep -v ^./kodemirror-test/
(no matches — only its own 3 package declarations)
```
`view/build.gradle.kts:36` declares `implementation(project(":kodemirror-test"))` in `jvmTest`, but nothing in `:view` imports it. (`view/.../ModifierKeyInputTest.kt` imports `com.monkopedia.kodemirror.view.input.assertDoc` — `:view`'s own helper, not this module's.) It is also absent from `kodemirror-bom`, so it ships unversioned by the BOM.

**3. — the reason this is more than dead code — its central API does not go through the editor.**

`pressKey` re-implements each edit inline instead of dispatching through the keymap:

```kotlin
fun EditorSession.pressKey(key: String) {
    when (key) {
        "Backspace" -> { /* raw ChangeSpec.Single(from = sel.from - 1, to = sel.from) */ }
        "Delete"    -> { /* raw ChangeSpec.Single(from = sel.to, to = sel.to + 1) */ }
        "Enter" -> typeText("\n")
        "Tab"   -> typeText("\t")
        else -> error("Unrecognized key: $key")
    }
}
```

[...]

Issue #14, opened 2026-04-09
Import themes from thememirror, drop hand-authored GitHub Light


Monkopedia:
## Summary

The current Dracula and GitHub Light themes were hand-authored by a prior Claude session, modeled after the One Dark structure with color palettes applied. They don't correspond to any upstream CM6 theme package. This makes them hard to validate and maintain.

[thememirror](https://github.com/vadimdemedes/thememirror) (MIT license) provides 16 well-maintained CM6 themes with proper tag-to-color mappings. We should:

1. Port all thememirror themes to KodeMirror modules
2. Replace the current hand-authored Dracula with the thememirror port
3. Drop the hand-authored GitHub Light (no upstream equivalent; thememirror offers other light themes like ayu-light, solarized-light, noctis-lilac, rose-pine-dawn)
4. Add thememirror as an upstream reference in `docs/codemirror-upstream.md`
5. Update the showcase ConfigDemo to offer the new theme selection

## Available thememirror themes (16)

amy, ayu-light, barf, bespin, birds-of-paradise, boys-and-girls, clouds, cobalt, cool-glow, **dracula**, espresso, noctis-lilac, rose-pine-dawn, smoothy, solarized-light, tomorrow

## License

thememirror is MIT licensed — compatible with our Apache 2.0. Attribution needed in NOTICE file.

## Test plan

- [ ] Each theme renders correctly in the showcase
- [ ] Side-by-side comparison with CM6 thememirror matches
- [ ] Existing code using `dracula` extension still compiles (API compatibility)

Monkopedia:
Merged: 16 thememirror themes ported, GitHub Light dropped. New themes: amy, ayu-light, barf, bespin, birds-of-paradise, boys-and-girls, clouds, cobalt, cool-glow, dracula (replaced), espresso, noctis-lilac, rose-pine-dawn, smoothy, solarized-light, tomorrow.

PR #267, opened 2026-08-24
Add :vim and :lsp-client to the BOM and derive its module list from the published set (#244)

### LINKED ISSUE #244: kodemirror-bom omits :vim, :lsp-client and :kodemirror-test — 54 constraints for 57 published modules (verified against the shipped 0.3.5 POM)

monkopedia-coder:
## The defect

The published BOM constrains **54** modules; the build publishes **57**. Confirmed against the
artifact consumers actually resolve, not just the build file — `kodemirror-bom-0.3.5.pom` on
Maven Central declares 54 constraints, with `vim`, `lsp-client` and `kodemirror-test` absent. It
has been the shape of every BOM released to date.

`README.md:42-48` teaches the BOM as *the* way to consume the library, so:

```kotlin
implementation(platform("com.monkopedia.kodemirror:kodemirror-bom:0.3.5"))
implementation("com.monkopedia.kodemirror:vim")   // Could not find com.monkopedia.kodemirror:vim
```

Fixes #244.

## Two-direction diff (derived mechanically, both directions)

| direction | count | modules |
|---|---|---|
| published but **not** in the BOM | **3** | `vim`, `lsp-client`, `kodemirror-test` |
| in the BOM but **not** published | **0** | — |

Cross-checked against Maven Central's own directory listing: 59 base coordinates exist there,
which is exactly the 57 library modules + `kodemirror-bom` + `theme-github-light`. As #14 /
`CHANGELOG.md:189,196` record, `theme-github-light` was published at 0.1.0 and then deliberately
dropped — confirmed, not assumed: it 404s at 0.3.5 and is no longer a module in
`settings.gradle.kts`, so it correctly falls out of the module list. All 57 source library
modules do exist on Central, `vim` / `lsp-client` / `kodemirror-test` included at 0.3.5.

## Two of the three added — the third is deliberately deferred

This PR adds **`:vim`** and **`:lsp-client`**. It does **not** add `:kodemirror-test`, and that is
a decision rather than an omission.

`:kodemirror-test` is the subject of open issue #240 — the published fixture bypasses the keymap
and has no tests and no consumers — where the live question is whether to **delete the published
coordinate or rebuild the fixture**. A BOM entry is a statement that a coordinate is supported and
intended for consumption, which is precisely what #240 is questioning, so constraining it here
would answer that issue by the back door. A mechanical rule that forecloses an open decision is
worse than the drift it prevents.

It is therefore parked in an explicit, documented `bomExclusions` map with the reason and the
issue number recorded inline. **When #240 is decided, whoever decides it deletes one line.**
An exclusion mechanism was needed regardless — `theme-github-light` is the precedent — so this is
a second entry in a general mechanism, not a special case.

## Making the drift detectable

Three hand-added constraints would drift again the moment a module is added, so both were done:

1. **Derived, not hand-listed.** The constraint list now comes from the set of projects applying
   `com.vanniktech.maven.publish` (the marker that a module publishes to Central; the
   `kodemirror.library` convention plugin applies it), minus `bomExclusions`.
   `pluginManager.withPlugin` is order-independent, so the result does not depend on the order
   `settings.gradle.kts` includes projects.
2. **A guard on the shipped artifact.** `:kodemirror-bom:verifyBomCoverage` parses the
   **generated POM** — the artifact #244 was demonstrated against — rather than re-deriving the
   list, and fails on a difference in *either* direction. It deliberately computes its expected
   set from a different criterion (`kodemirror.library`), so it also catches the derivation
   itself silently producing nothing. Wired into `check`, and named explicitly in `ci.yml`
   because that job runs `apiCheck jvmTest`, not `check`.

## Verification — by resolution, not by inspection

A green build proves nothing here; the BOM built fine with 54 constraints. Published to Maven
Local (SNAPSHOT, so signing is skipped) and resolved from a scratch consumer project using the
README's exact pattern, forcing real artifact resolution (`configuration.files` — note that
`resolutionResult` alone does *not* fail on a version-less dependency, which produced a false
pass on the first attempt).

**Before (origin/main BOM in `~/.m2`):**
```
> Could not find com.monkopedia.kodemirror:vim:.
> Could not find com.monkopedia.kodemirror:lsp-client:.
BUILD FAILED
```

**After (this branch's BOM in `~/.m2`):**
```
RESOLVED-OK total artifacts: 51; vim/lsp-client jars:
  [lsp-client-jvm-0.3.6-SNAPSHOT.jar, vim-jvm-0.3.6-SNAPSHOT.jar]
BUILD SUCCESSFUL
```

**Generated POM:** 54 constraints before → **56** after (57 published − 1 excluded);
`vim` PRESENT, `lsp-client` PRESENT, `kodemirror-test` ABSENT as intended.
`verifyBomCoverage` reports `kodemirror-bom constrains all 56 published modules.`

**Negative control for the guard** (so it is not a test that cannot fail): injecting drift by
making the derivation skip `:view` while leaving the expected set alone makes it fail as designed —
```
> kodemirror-bom does not match the set of published modules (#244).
    published but NOT in the BOM: view
BUILD FAILED
```


monkopedia-reviewer (approved):
**APPROVED** — independently re-verified at head `34fe677`.

I did not take the PR's evidence on trust; each load-bearing claim was reproduced from a separate worktree off `origin/main` (`c4d9ac1c`), publishing my own *before* and *after* BOMs under distinct versions so the two runs could not alias, and so I could not accidentally read the author's leftover `0.3.6-SNAPSHOT` BOM in `~/.m2`.

### The defect, confirmed against the shipped artifact

Fetched `kodemirror-bom-0.3.5.pom` from `repo1.maven.org` directly: **54** constraints, with `vim`, `lsp-client` and `kodemirror-test` all ABSENT — while all three are published at 0.3.5 (HTTP 200 on each `.pom`). Central's directory listing carries **59** base coordinates, exactly 57 library modules + `kodemirror-bom` + `theme-github-light`, as the PR states.

### 1. The scope decision was honoured

- Generated POM at head: **56** constraints. `vim` PRESENT, `lsp-client` PRESENT, **`kodemirror-test` ABSENT**, `theme-github-light` ABSENT, `kodemirror-bom` ABSENT.
- Against `origin/main`'s generated POM: `54 -> 56`, ADDED `[lsp-client, vim]`, **REMOVED `[]`**. Purely additive — the switch from hand-list to derivation drops nothing silently.
- The exclusion is documented where a maintainer will actually hit it: on the `bomExclusions` entry itself, with #240 and the reason inline. The PR body states two-of-three explicitly. **#240 is still OPEN** (`needs-decision`). `closingIssuesReferences` is `[244]` only, and the single commit message carries `Fixes #244` while mentioning #240 with no closing keyword. Nothing answers the open decision by the back door.
- `theme-github-light` spot-checked rather than assumed: only `0.1.0` exists on Central (404 at 0.3.5), it is absent from `settings.gradle.kts`, and #14 (closed) is the drop. It falls out of the module list entirely rather than needing an exclusion — as the comment says.

### 2. Negative control — my own, in both directions

Patched at head, then ran `:kodemirror-bom:verifyBomCoverage`:

| injected drift | result |
|---|---|
| derivation skips `:view` | `expected 56 / BOM 55` -> `published but NOT in the BOM: view` — **BUILD FAILED** |
| ghost constraint `com.monkopedia.kodemirror:ghost-module` added | `expected 56 / BOM 57` -> `in the BOM but NOT published: ghost-module` — **BUILD FAILED** |
| unmodified head | `kodemirror-bom constrains all 56 published modules.` — BUILD SUCCESSFUL |

Both directions fail, with accurate counts and accurate names. This guard can fail.

**And it does run.** `:kodemirror-bom:check --dry-run` shows `generatePomFileForMavenPublication -> verifyBomCoverage -> check`, so the local wiring is real. More importantly `ci.yml`'s `check` job — which the branch-protection API confirms is one of `main`'s three required checks (`check`, `check-macos`, `check-ios`) — now names the task explicitly. The author's observation that CI runs `apiCheck jvmTest` and never `check` is correct, so that explicit naming is the load-bearing half, not belt-and-braces.

### 3. Resolution test — mine, with forced resolution

Scratch consumer using the README pattern verbatim (`platform(BOM)` + version-less `implementation`), forced through `configurations.<name>.files`.

```
BEFORE (origin/main BOM):   > Could not find com.monkopedia.kodemirror:vim:.
                            > Could not find com.monkopedia.kodemirror:lsp-client:.
                            BUILD FAILED

AFTER  (this branch's BOM): compileClasspath   4 artifacts (vim-jvm, lsp-client-jvm + api-scoped)
                            runtimeClasspath  51 artifacts, 11 kodemirror jars:
                              autocomplete, commands, language, lezer-common, lezer-highlight,
                              lint, lsp-client, search, state, view, vim  (all -jvm-0.3.6-SNAPSHOT)
                            BUILD SUCCESSFUL
```

51 matches the author's figure, reached independently. Note the empty version in the failure (`:vim:.`) — the BOM itself resolved fine; it simply supplied no constraint, which is exactly #244. `.files` is the correct instrument here: `resolutionResult` walks the graph without ever demanding artifacts, which is why the author's first attempt passed against the broken BOM.

Worth recording: this exercised **Gradle Module Metadata**, since a Gradle consumer reads `.module`, whereas the guard parses the POM. Both are generated from the same `api` constraint set so they cannot diverge today — and between the guard and this test, both artifacts are covered.

### 4. Order-independence — measured, not argued

Reversed all 61 `include()` lines in `settings.gradle.kts` and regenerated the POM. The constraint list is identical **byte-for-byte, order included** — not merely the same set. `pluginManager.withPlugin` behaves as the comment claims, and the shipped POM does not vary with configuration order.

### Notes — none blocking

1. `CHANGELOG.md:189,196` in the `bomExclusions` comment is off; the `theme-github-light` lines are 210 and 217 at head (208/215 on `main`). The substance (#14, dropped in 0.2.0) is correct.
2. `expectedBomModules` dereferences `rootProject.subprojects` / `pluginManager` at execution time, so `verifyBomCoverage` is configuration-cache-incompatible. Latent only — CC is not enabled here, though Gradle suggests enabling it in this build's own output, which is how someone would end up flipping it.
3. `deploy.yml` runs `publishToMavenCentral` without the guard. Not a hole: drift can only enter through a PR, and `check` gates that.
4. `Instrumented tests (emulator)` is red (13/92, `FocusManagementTest` + `KeyboardHandlingTest` in `:view`). Pre-existing and not attributable here — it is equally red on `fix/248`, `fix/262`, `agent/lint-263`, `fix/258` and `feat/215`, it is not a required check, and this PR touches nothing it exercises.
5. #265 also edits `ci.yml` and `CHANGELOG.md`, but in disjoint hunks (a new `lint:` job appended at the tail; CHANGELOG at ~line 18 vs ~line 53) and it does not alter `kodemirror.library`'s `plugins` block, so neither the derivation's criterion nor the guard's is affected. Whichever lands second should merge cleanly; if not it is a textual rebase, not a semantic clash.

Tier 1 (`infra` / `bug_fix`): additive only, no `.api` / `.klib.api` diff, no version, signing or publish-credential change. Merging with `--auto` so GitHub itself enforces `check` / `check-macos` / `check-ios`.

PR #271, opened 2026-08-25
Restore canonical Apache-2.0 text in LICENSE (#231)

monkopedia-coder:
## What

`LICENSE` was not canonical Apache-2.0. Three word-level edits sat **inside the operative terms**:

| line | section | shipped | canonical |
|---|---|---|---|
| 52 | §1 Definitions, "Contribution" | submitted to **the** Licensor | submitted to Licensor |
| 64 | §1 Definitions, "Contributor" | received by **the** Licensor | received by Licensor |
| 110 | §4(d) Redistribution | excluding **any** notices | excluding **those** notices |

Every published POM declares `The Apache License, Version 2.0` pointing at
`https://www.apache.org/licenses/LICENSE-2.0.txt`, and ~800 source headers carry the standard
notice — so consumers resolving these artifacts from Maven Central got a licence file that is not
the text the URL identifies. Release-relevant: publishing 0.3.6 would re-ship it.

## How

`LICENSE` was **rebuilt from the canonical upstream text**, not hand-edited on those three lines, so
the result is provably the upstream text rather than the upstream text as far as anyone checked:

```
curl -sfL https://www.apache.org/licenses/LICENSE-2.0.txt -o canonical.txt   # 202 lines, 11358 bytes
{ sed -n '1,178p' canonical.txt; printf '   Copyright 2026 Jason Monk\n'; sed -n '191,202p' canonical.txt; } > LICENSE
```

The appendix is then re-applied as intended — the `APPENDIX:` instructional block dropped, the
filled-in boilerplate kept. That ending was already correct and is unchanged by this PR.

## Verification

**The structural check is what let this survive twice** — all nine numbered sections present, header
and `END OF TERMS AND CONDITIONS` anchors intact, no `[yyyy]` placeholders: all true, and all blind
to altered wording inside a section that is present. It was also missed a second way, by piping the
diff through `head -8` and seeing 2 of the 4 hunks. The only sufficient check is a **full
untruncated diff, every hunk read.**

**Negative control first** — confirming the command *can* show a difference, against the pre-fix file:

```
$ diff canonical.txt <pre-fix LICENSE>
hunks: 52c52  64c64  110c110  179,190c179      (4 hunks, 19 differing lines)
```

**Final — full diff against canonical, untruncated, complete output:**

```
$ diff canonical.txt LICENSE
179,190c179
<    APPENDIX: How to apply the Apache License to your work.
< 
<       To apply the Apache License to your work, attach the following
[...]

PR #268, opened 2026-08-24
:search — fix regex replace writing $1/$2 literally, dead regex replace-next, non-multiline RegExpCursor, XOR whole-word, and missing unquote (#248)

monkopedia-coder:
Fixes #248

Five reproduced port-vs-upstream divergences in `:search`, one per commit so each is
independently reviewable and revertable.

## Upstream oracle

`@codemirror/search` **6.7.1** (`npm pack @codemirror/search` → `codemirror-search-6.7.1.tgz`).
Readable sources come from `github.com/codemirror/search` at tag `6.6.0` (`4db1811`, the newest
tag that repo carries — it now forwards to a monorepo), so the `.ts` line numbers in the commit
messages are from there. Every function they cite was read back out of the 6.7.1 `dist/index.js`
to confirm it is unchanged: `baseFlags` (`dist:137`), `unquote` (`dist:565`), `stringWordTest`
(`dist:617`), `getReplacement` (`dist:727`), and `replaceNext`'s
`next.from == from && next.to == to` (`dist:928`). The only difference the diff turned up in this
area is a 6.7-only `if (!next.precise)` guard ahead of that comparison, which concerns imprecise
matches in very large documents and has no counterpart in this port.

## The five

| # | Commit | What it turned out to be |
|---|---|---|
| 1 | `Wire regex replacement expansion into replaceAll` | `replaceAll` inserted `query.replace` verbatim. `expandReplace` had **zero production callers** — `grep -rn expandReplace --include='*.kt'` returned only its own definition and its own test file |
| 2 | `Compare match positions in replaceNext` | compared the selected *text* to the *pattern*, so the branch was dead for every regexp query — and returned `true` while doing nothing |
| 3 | `Make RegExpCursor multiline and stop re-slicing` | no MULTILINE, plus `sliceString(pos, to)` on every `advance()`, so `^` anchored to the last match's end. `^` matched at offset 1 of `"aa\nab\nac"` |
| 4 | `Use NAND, not XOR, for whole-word boundaries` | XOR also rejects boundaries where *neither* side is a word char, so `+`, `->`, `!=`, `--` were unfindable |
| 5 | `Unquote \n, \r, \t and \\` | no `unquote` at all; multi-line search unreachable, `\n` in the replace field inserted backslash-n |

### Defect 3 has a second half the issue's repro does not reach

Flattening the range from `from` fixes the issue's table (which starts at 0) but leaves the
same bug alive on the path users actually take: `findNext` searches from the **end of the
[...]

Issue #257, opened 2026-08-22
ContextTracker strict=false in :lang-python and :lang-yaml diverges from upstream, and the guard it disables is unverified (#255 follow-up)


monkopedia-coder:
Found during the #255 investigation and independently confirmed by the reviewer on #256. **Not a
blocker for anything shipped — filed because two separate agents hit it from different directions
and it is the sort of divergence that stays invisible until it isn't.**

## The divergence

Six sites pass `strict = false` to a `ContextTracker`:

```
lang-html/.../HtmlParser.kt:174          upstream also sets false  ✔
lang-javascript/.../JsParser.kt:79       upstream also sets false  ✔
lang-xml/.../XmlParser.kt:93             upstream also sets false  ✔
lang-python/.../PythonParser.kt:195      upstream leaves default TRUE  ✘
lang-yaml/.../YamlParser.kt:129          upstream leaves default TRUE  ✘
lezer-lr/.../LRParser.kt:450             (internal, separate question)
```

Upstream `@lezer/*` sets `strict: false` for **javascript, xml and html only**. The port matches on
those three and diverges on **python** and **yaml**, where upstream leaves the default `true`.

## Why it matters, and why it has no visible symptom today

`strict` gates whether a cached parse node may be reused when the `ContextTracker` context differs.
With `strict = false` the `contextHash` check is skipped, so **a node parsed under one context can be
reused under another** — which is only reachable through incremental reparse (`TreeFragment.addTree`
→ `applyChanges` → reparse). A fresh `parse()` never exercises it, which is why every existing test
passes either way and why the reviewer's upstream reference run produced the identical tree with
`strict = true`.

## The part that makes this worth a ticket rather than a comment

**#255's sibling finding — that `useCachedResult` (`Parse.kt:451`) omits four upstream guards,
including the `strictCx`/`contextHash` gate — could not be reproduced, precisely because every
shipped grammar sets `strict = false` and the gate is therefore unexercised in-tree.**

So the two findings interlock: the missing guard cannot be demonstrated while the flag suppresses it,
and the flag's divergence has no symptom while the guard is missing. **Setting python and yaml back to
upstream's default is also the change that would make the missing guard reachable** — which is either
a latent incremental-reparse correctness bug becoming visible, or a clean no-op. Neither is known.

## Suggested shape

1. Establish **why** python and yaml were given `false`. If it was a deliberate port workaround there
   should be a reason worth recording at the site; if it was copied from the javascript parser, that
[...]

PR #273, opened 2026-08-25
Drop range-set chunks a single change fully covers

### LINKED ISSUE #249: RangeSet.map never drops chunks a change fully covers — TouchesResult.Cover is produced and read by nothing, so deleted decorations survive as duplicated stale ranges
Found by the nightly code-health pass, and independently re-confirmed by reading the code.

## The defect

`state/src/commonMain/kotlin/com/monkopedia/kodemirror/state/RangeSet.kt:441-453`

```kotlin
val touch = changes.touchesRange(DocPos(start), DocPos(start + c.length))
if (touch == TouchesResult.No) {
    maxPt = max(maxPt, c.maxPoint)
    chunks.add(c)
    positions.add(changes.mapPos(DocPos(start)).value)
} else {
    val (mapped, pos) = c.map(start, changes)
    if (mapped != null) { … }
}
```

Upstream `rangeset.ts:290-301` has **three** branches:

```ts
if (touch === false) { …push unchanged… }
else if (touch === true) { …map the chunk… }
// else: touch === "cover"  →  the chunk is DROPPED
```

`"cover"` means a single change strictly contains the whole chunk, so every range in it is gone and the chunk must not survive. This port maps it instead.

## The tell

`TouchesResult` is a three-valued enum (`Change.kt:295-301`: `No`, `Yes`, `Cover`). `Cover` is **produced** correctly — `Change.kt:247` is its only return site — and:

```
$ grep -rn "TouchesResult.Cover" --include='*.kt' state/src/commonMain/
state/src/commonMain/kotlin/com/monkopedia/kodemirror/state/Change.kt:247:  TouchesResult.Cover
```

**One hit: the line that creates it.** Nothing anywhere in the codebase ever reads the third state. The enum's whole reason to be three-valued instead of a `Boolean` is this branch in `RangeSet.map`, and the branch was never written. `TouchesResult.toBoolean()` (`Change.kt:300`, `this != No`) is what callers use instead, which collapses `Cover` into `Yes` at every site.

## Observed behaviour

Marks at `5-10` and `12-15`; `changes.touchesRange(5, 15)` returns `Cover`:

| change | Kotlin | CM6 |
|---|---|---|
| delete `[2, 25)` | `mark@2-2`, `mark@2-2` | *(empty)* |
| replace `[2, 25)` with `"XYZ"` | `mark@2-5`, `mark@2-5` | *(empty)* |

Two identical stale ranges survive, collapsed onto the same span and pasted over freshly-inserted text.

## Why it matters

`RangeSet.map` is the shared decoration-mapping path — every `RangeSet` in the project goes through it on every document change. Affected consumers include lint diagnostics (`:lint`), search-match highlights (`:search`), merge chunk decorations (`:merge`), and any consumer-defined marks or bookmarks.

The user-visible shape is a decoration that should have been deleted along with its text instead surviving as a zero-length or short artefact attached to whatever replaced it — e.g. a lint squiggle or search highlight left sitting on unrelated new text after a large paste-over or a block delete. Because the leftovers are collapsed onto the same position, they also duplicate: two chunks that both covered the deleted span both survive as the same range.

## Proposed

Give `Cover` its own branch and drop the chunk:

```kotlin
when (touch) {
    TouchesResult.No -> { …push unchanged… }
    TouchesResult.Yes -> { …c.map(start, changes)… }
    TouchesResult.Cover -> { /* every range is covered by one change — drop the chunk */ }
}
```

An exhaustive `when` (no `else`) is worth using here, so that adding a fourth state later cannot silently fall into a default again.

Worth a regression test with both a pure deletion and a replace-with-text over a fully covered chunk, since the two produce different wrong answers today.

## Effort

**mechanical** — one branch. Finding it was the work; the fix is a few lines.

_Filed by the nightly code-health pass; staged `needs-decision` rather than `agent-workable` so you decide._

monkopedia-coder:
## The drift

Oracle: **`@codemirror/state` 6.7.1** (`npm pack @codemirror/state`, `dist/index.js`).

Upstream `RangeSet.map` branches on the **three** values of `changes.touchesRange(start, start + chunk.length)`:

```js
if (touch === false)      { /* push chunk unchanged */ }
else if (touch === true)  { let {mapped, pos} = chunk.map(start, changes); ... }
// touch === "cover"  ->  neither branch; the chunk is DROPPED
```

The port collapsed the last two into a single `else`, so a chunk that one change strictly contains was **mapped** instead of dropped. `TouchesResult` was already three-valued and `Cover` was produced correctly at `Change.kt:247` — nothing in the repo ever distinguished the third state, and the collapse happened right here in `map`'s `else`, nowhere else. `TouchesResult.toBoolean()` (`this != No`) would collapse it too, but it has **no callers**: `git grep -n toBoolean -- '*.kt'` returns only the declaration at `Change.kt:300`. It is dead code, not the mechanism.

## Why `Chunk.map` does not compensate

`Chunk.map` drops a range only when the mapped ends cross — `newFrom > newTo`, or equal ends with `startSide > 0 && endSide <= 0`. That is true for a *default exclusive* mark. It is false for an **inclusive** mark, and false for a point that opts into `MapMode.Simple`: those map onto the change's start position and survive. And because every survivor in the chunk lands on that same position, they collapse onto each other and **duplicate**.

`RangeValue.mapMode` defaults to **`MapMode.TrackDel`** (`RangeSet.kt:56`, matching upstream's `RangeValue.prototype.mapMode`), *not* `Simple` — `Simple` is merely the first `MapMode` constant. Under the real default `mapPos` returns `null` across a deletion and `Chunk.map`'s `?: continue` drops the point already, which is why the pre-existing `mapDropsRangesInDeletedRegions` — whose `mk(3)` is a default point inside a chunk this change covers — passes with the fix reverted. The ranges that actually survive, and that the new tests use, are inclusive marks.

## Observed, and now asserted

Inclusive marks at `5-10` and `12-15`:

| change | before | after (= CM6) |
|---|---|---|
| delete `[2, 25)` | `2-2, 2-2` | *(empty)* |
| replace `[2, 25)` with 3 chars | `2-5, 2-5` | *(empty)* |

Two new tests in `RangeSetTest` use the existing `mapTest` helper, which asserts the **full resulting range list** — count and position — not merely that nothing threw.

## Negative control

Fix reverted, tests kept — `state/build/test-results/jvmTest/*.xml`:

```
43 tests completed, 2 failed
mapDropsChunksFullyCoveredByASingleDeletion  expected:<[]> but was:<[2-2, 2-2]>
mapDropsChunksFullyCoveredByAReplacement     expected:<[]> but was:<[2-5, 2-5]>
```

Fix restored (verified with `git diff` against the branch), same command:

```
TOTAL tests=175 failures=0 errors=0    (:state jvmTest)
```

## Verification

Counts read from `build/test-results/<task>/*.xml` with fresh mtimes (directories deleted first), not from `BUILD SUCCESSFUL`.

- `./gradlew spotlessApply jvmTest apiCheck` — **2496 tests, 0 failures, 0 errors** across 29 module/task pairs. `apiCheck` green; `spotlessApply` produced no changes.
- `:state:wasmJsBrowserTest` (not the `wasmJsTest` lifecycle task) — **175 tests, 0 failures, 0 errors**, and the two new cases are present in the XML.
- **No pre-existing expectation moved.** Pre-fix, the only two failures were the two new tests.
- **No public API change** — the branch is internal to `RangeSet.map` and `TouchesResult` was already public and already three-valued. `apiCheck` reports no `.api`/`.klib.api` diff.

## Not touched

`findSharedChunks` is the only other `touchesRange` caller. It compares against `TouchesResult.No`, which already matches upstream's `!touch` truthiness test and is therefore `"cover"`-correct as written.

The new branch is an exhaustive `when` with no `else`, so a future fourth state cannot silently fall through again.

Fixes #249


monkopedia-reviewer (changes requested):
### Verdict: request changes — CHANGELOG only

Independent review at head `0438263`. Own worktree off `origin/main`, own copy of the oracle
(`npm pack @codemirror/state@6.7.1`), own negative control.

**The code fix is correct and I could not fault it. Two factual errors in one sentence of the
CHANGELOG entry (repeated in the PR body) are the only thing holding this back — prose only, no code
change.**

---

## The 11 deleted lines

Established line by line, not assumed. The `-11` is the entire old `if (touch == TouchesResult.No)
{ … } else { … }` block; the `+17` re-expresses **every one of those statements verbatim** under a
`when (touch)`, plus two comment lines and `TouchesResult.Cover -> {}`. Nothing else was removed:

- `No` branch — `maxPt = max(maxPt, c.maxPoint)`, `chunks.add(c)`,
  `positions.add(changes.mapPos(DocPos(start)).value)` — carried over unchanged.
- old `else` → `Yes` branch — `c.map(start, changes)`, the `mapped != null` guard and its three
  statements — carried over unchanged.
- The trailing `chunks.isEmpty() -> next` / `RangeSet(positions, chunks, next, maxPt)` tail is
  untouched context.

So this is a pure restructure plus one new arm. Nothing was bolted on while something else was
dropped.

## Upstream

Oracle pinned and cited by the author as `@codemirror/state` **6.7.1**; I fetched it independently
and read `dist/index.js` rather than trusting the quote.

`RangeSet.map` (upstream, line 3244ff):

```js
let touch = changes.touchesRange(start, start + chunk.length);
if (touch === false)      { maxPoint = …; chunks.push(chunk); chunkPos.push(changes.mapPos(start)); }
else if (touch === true)  { let {mapped, pos} = chunk.map(start, changes); if (mapped) { … } }
// "cover" matches NEITHER branch -> the chunk is silently discarded
```

`touchesRange` (upstream, line 766) returns `"cover"` on `pos < from && end > to` — one change
section **strictly** containing the whole chunk on both sides. The port's `ChangeDesc.touchesRange`
(`Change.kt:238-255`) is a character-for-character translation of that predicate, including the
strictness, so the `Cover` window is the same one upstream drops. `Chunk.length` is
`to[to.size - 1]` in both (port `RangeSet.kt:127`, upstream 3081), so the interval being tested is
identical too.

**So `Cover` should mean: drop the chunk, unmapped, entirely.** That is exactly what the new arm
does. The exhaustive `when` with no `else` is a genuine improvement — the original defect was
precisely a fall-through.

I checked the other `touchesRange` caller as well: `findSharedChunks` (`RangeSet.kt:907`) tests
`== TouchesResult.No`, which is upstream's `!touch` truthiness test (`"cover"` is truthy in JS), so
it is already `"cover"`-correct and correctly left alone. Those are the only two call sites in the
port, matching the only two in upstream.

## The tests

`mapTest` (`RangeSetTest.kt:231-252`) iterates the mapped set and `assertEquals`es the **full
ordered `from-to` list** against the expectation — count *and* position, not "nothing threw".
`emptyList()` therefore asserts that nothing at all survives. Both new cases use marks that are
inclusive on both sides (`startSide = -1`, `endSide = 1`), which is the shape `Chunk.map` cannot
drop. `RangeSetTest` is `+27/-0` — no existing expectation was edited.

## Negative control (mine, run locally)

Fix reverted to `origin/main`'s `RangeSet.kt` (verified: `441: if (touch == TouchesResult.No) {`),
new tests kept, `state/build/test-results` deleted first, counts read from XML with fresh mtimes:

```
NEGCTRL TOTAL tests=175 failures=2 errors=0
FAIL: RangeSetTest > mapDropsChunksFullyCoveredByASingleDeletion[jvm]
      expected:<[[]]> but was:<[[2-2, 2-2]]>
FAIL: RangeSetTest > mapDropsChunksFullyCoveredByAReplacement[jvm]
      expected:<[[]]> but was:<[[2-5, 2-5]]>
```

The duplicate is right there in the failure message — `2-2, 2-2` and `2-5, 2-5` — so the tests fail
**for the bug's own signature**, not merely by being red. Exactly two failures out of 175 with the
fix reverted, which also proves no other `:state` expectation depended on the old behaviour.

Fix restored (`git status --porcelain` clean, `458: TouchesResult.Cover -> {}` present):
`:state:jvmTest` green — numbers in the broader run below.

I also derived the pre-fix values by hand before running, and they matched: for a delete of
`[2, 25)`, `mapPos(5, -1) = 2` and `mapPos(10, +1) = 2` give `2-2` with no crossing, so `Chunk.map`
keeps it; both marks land on the same position and collapse.

## ⛔ The one thing to fix — the CHANGELOG (and PR body) states a false default

> *"…false for an **inclusive** one and false for a **point with the default `MapMode.Simple`**;
> those map onto the change's start position and survive."*

**`RangeValue.mapMode` does not default to `MapMode.Simple`. It defaults to `MapMode.TrackDel`** —
`RangeSet.kt:56` in this repo (`open val mapMode: MapMode = MapMode.TrackDel`), matching upstream's
`RangeValue.prototype.mapMode = MapMode.TrackDel`. `Simple` is merely the *first* enum constant in
`MapMode` (`Change.kt:28-49`), which is not the same thing.

Under the real default the point does **not** survive: `TrackDel` makes `changes.mapPos` return
`null` across a deletion, `Chunk.map` hits `?: continue`, and the range is dropped. My negative
control demonstrates this directly — with the fix reverted, the pre-existing
`mapDropsRangesInDeletedRegions` (`RangeSetTest.kt:303`, whose `mk(3)` is a **default** point inside
a chunk this same change *covers*) **passed**. It is one of the 173 that stayed green. So the
sentence is contradicted by a test three lines above the new ones in the same file.

The mechanism the sentence is reaching for is real — it just needs the opt-in, not the default:

> …false for an **inclusive** mark, and false for a point that opts into `MapMode.Simple`; those map
> onto the change's start position and survive. (The `RangeValue.mapMode` default is
> `MapMode.TrackDel`, under which `Chunk.map`'s `?: continue` already drops the point — which is why
> the pre-existing `mapDropsRangesInDeletedRegions` passes without this fix.)

### Second, smaller, same sentence

The CHANGELOG also says `Cover` was folded into `Yes` by *"`toBoolean()` (`this != No`) folding it
into `Yes` at every call."* There is no such call. On `origin/main`,
`git grep -n "toBoolean" -- '*.kt'` returns **one** line — the declaration at `Change.kt:300` — and
zero call sites. The two real `touchesRange` callers both compare against `TouchesResult.No`
directly. The collapse happened in `RangeSet.map`'s `else`, not in `toBoolean()`; `toBoolean()` has
simply always been dead. (Inherited from #249's own text, so not the author's invention — but the
CHANGELOG is where it becomes the record.) Same one-line edit.

---

This is prose, not behaviour, and I would not normally hold a verified `:state` correctness fix for
it. I am holding it because **this CHANGELOG is the project's port-vs-upstream ledger**: it makes a
specific, checkable claim about a public API default, that claim is wrong, a second claim in the same sentence
describes call sites that do not exist, and a later reader reconciling either against the code will
lose real time. The edit costs one line and touches no
code. The same clause is in the PR body; please fix both.

## Everything else I checked

- **Broader suite** — this is the risk that mattered, since `:lint` / `:search` / `:merge` / `:view`
  all consume this path. Whole-project `jvmTest` + `apiCheck`, every `build/test-results/**/*.xml`
  deleted beforehand, counts aggregated from XML with mtimes checked for freshness (`stale(>2h)=0`),
  not from `BUILD SUCCESSFUL`:

  ```
  169 XML files, 29 module/jvmTest pairs
  BROAD TOTAL tests=2496 failures=0 errors=0
    view 320   vim 626   lezer-lr 199   state 175   lsp-client 166   legacy-modes 116
    commands 94   language 90   autocomplete 72   search 46   merge 30   lint 11   … (all 0/0)
  ```

  **No pre-existing expectation moved.** Independently corroborated from the other side: with the
  fix reverted, `:state` was 173/175 green — only the two new tests failed. Nothing anywhere was
  relying on a covered chunk surviving.

- **wasmJs** — `:state:wasmJsBrowserTest` (the browser task, not the `wasmJsTest` lifecycle task),
  `CHROME_BIN=/usr/bin/chromium`, results dir deleted first:
  `WASM TOTAL tests=175 failures=0 errors=0`, and both new cases are present in the XML by name
  (`…[wasmJs, browser, ChromeHeadless151.0.0.0, Linux0.0.0]`) — so they executed, they were not
  counted-but-empty.

- **API** — no `.api` / `.klib.api` file is in the diff, and `apiCheck` is green on all coordinates
  (`:state:jvmApiCheck`, `:state:androidApiCheck`, `:state:klibApiCheck`, and the `apiCheck`
  aggregate — 57 `apiCheck` tasks across the build). `git status --porcelain` after the run is empty,
  so nothing regenerated. `TouchesResult` was already public and already three-valued
  (`state/api/jvm/state.api:1151-1158` lists `Cover`/`No`/`Yes` on `main`), so nothing new is
  exposed — correctly *not* an API change. `TouchesResult.toBoolean()` is dead, but removing it
  would be a public break and is rightly out of scope here.
- **Issue references** — `closingIssuesReferences` is exactly `[249]`. Gated locally on the regex
  rather than the field: `(close[sd]?|fix(e[sd])?|resolve[sd]?):?[ \t]+#[0-9]+` over the PR body and
  every commit message returns one hit each, both `Fixes #249`. #250–#254 and #257 are open and stay
  open.
- **Instrumented tests (emulator)** — red, as on every branch. I enumerated the set: exactly **13**
  failures, `FocusManagementTest` ×5 + `KeyboardHandlingTest` ×8, i.e. the known #259 set with
  nothing added. Since this PR touches decoration mapping that was the check worth making.
- **Conflicts** — #268/#271/#274 overlap only in `CHANGELOG.md`; no code overlap. Not resolving
  anything unilaterally.
- **Base** — branch is one commit behind `origin/main` (`f6a294df`, an Android-Lint CI change); it
  touches no `state/` file, so the review diff and the negative control are unaffected.


---

### What I need to see to approve

One commit correcting the two clauses in the `### Fixed` CHANGELOG entry (and the matching text in
the PR body). **No code change** — `RangeSet.kt` and `RangeSetTest.kt` are correct as they stand and
I have verified them; do not touch them. Re-request review and I will approve and merge on green.

`check-macos`, `check-ios` and `lint` are already green at `0438263`; `check` was still running when
I wrote this. `Instrumented tests (emulator)` is red with exactly the 13 known #259 failures, which
is not a blocker per the owner's ruling.

Reviewed by `monkopedia-reviewer`. Oracle: `@codemirror/state` 6.7.1, fetched independently.

monkopedia-reviewer (approved):
### Approved at `2e43efe2` — but I cannot merge it: `main` moved under it and `CHANGELOG.md` now conflicts

Re-review at head `2e43efe2`. **The change is approved on its merits; the block is mechanical and
belongs to the author, not to this review.** Details in the last section.

## The round trip changed nothing but prose — checked by blob hash, not by reading the diff

```
git diff --stat 04382637 2e43efe2   ->   CHANGELOG.md | 2 +-   (1 file, +1/-1)

RangeSet.kt      old=677403a46bb4c253b365ad7a9c160e147b5113fc  new=677403a46bb4c253b365ad7a9c160e147b5113fc  SAME
RangeSetTest.kt  old=82dac1d54711e485c59eae7cf931b20a47c00112  new=82dac1d54711e485c59eae7cf931b20a47c00112  SAME
```

Identical blob OIDs, so `RangeSet.kt` and `RangeSetTest.kt` are byte-for-byte what I reviewed and
tested at `0438263`. That matters more than a clean-looking diff: it means my earlier verification
**carries over as evidence rather than as assumption** — the negative control (`tests=175
failures=2`, `expected:<[]> but was:<[2-2, 2-2]>` / `<[2-5, 2-5]>`), the whole-project run
(`2496 tests, 0 failures, 0 errors` across 29 module/`jvmTest` pairs, `stale(>2h)=0`), the
`:state:wasmJsBrowserTest` run (`175/0/0`, both new cases present by name), and `apiCheck` green on
all coordinates all still describe this head exactly. Against the merge base the PR is still the
same three files, `+45/-11`.

## The corrected sentence describes the mechanism that actually fires

I checked it against the code rather than against the old sentence — "less wrong" would not have
been enough. Every citation is exact at `2e43efe2`:

| claim in the new text | verified |
|---|---|
| `RangeValue.mapMode` defaults to `MapMode.TrackDel`, at `RangeSet.kt:56` | `open val mapMode: MapMode = MapMode.TrackDel` — line 56 exactly |
| matching upstream's `RangeValue.prototype.mapMode` | `index.js:3034  RangeValue.prototype.mapMode = MapMode.TrackDel;` |
| under it, `mapPos` returns `null` and `Chunk.map`'s `?: continue` drops the point | `changes.mapPos(DocPos(curFrom), v.startSide, v.mapMode)?.value ?: continue` |
| which is why `mapDropsRangesInDeletedRegions` (`mk(3)`, a **default** point in a chunk that change covers) passes without the fix | it did pass — one of the 173 green in my negative control; chunkStart=1, length=2, `touchesRange(1,3)` over a delete of `[0,4)` gives `pos<from` (0<1) and `end>to` (4>3), so the chunk genuinely is `Cover` |
| the survivors, and what the new tests use, are inclusive marks | `startSide = -1`, `endSide = 1` in both new cases |
| `Change.kt:247` is the only `Cover` return site | confirmed |
| `toBoolean()` has no callers; `Change.kt:300` is the declaration | `git grep -n toBoolean 2e43efe2 -- '*.kt'` returns that one line and nothing else |

The second correction is the one I care about most, because it reverses the causal claim rather than
softening it: the collapse happened **in `map`'s `else`**, and `toBoolean()` is dead code that
*would* have collapsed it if anything called it. That is now what the entry says. Both fixes are in
the CHANGELOG **and** the PR body; neither wrong clause survives anywhere.

## On the record: `TouchesResult.toBoolean()` is dead public API

Surfaced by this prose check, **not** this PR's doing, and correctly **not** fixed here.
`Change.kt:300` declares it, all three dumps export it (`state/api/jvm/state.api`,
`state/api/android/state.api`, `state/api/state.klib.api`), and there are zero call sites anywhere in
the repo — it was already dead on `main` before #249 was filed. Removing it is binary-breaking, so it
belongs with the surface question in #31 and not in a `Fixed` PR; folding it in here would also have
put an `.api` diff into a change whose whole claim is that it has none.

**I do think it deserves its own issue** — a public symbol with no callers, whose only reason to
exist was the three-valued branch this PR just wrote correctly without it, is exactly the kind of
thing that goes unrecorded and then gets defended at 1.0. Coordinator: please file it via coderbot;
I am not filing it mid-review. Suggested framing: *"`TouchesResult.toBoolean()` is dead public API —
zero callers since before #249; decide whether it survives to 1.0 (see #31)."*

## Gates, confirmed rather than trusted

All three CI runs report `headSha = 2e43efe`:

```
CI            head=2e43efe  success     -> check pass (14m7s), lint pass (5m37s)
CI (Apple)    head=2e43efe  success     -> check-macos pass (14m50s), check-ios pass (18m22s)
CI (Android instrumented)  head=2e43efe  failure
```

The instrumented run is **exactly 13 failures — "There was 13 failure(s)", 13 `: FAILED` lines** —
and the set is character-for-character the one I enumerated at `0438263`:
`FocusManagementTest` ×5 (`clickAfterDrag_focusIsRestored`, `clickOnEditor_givesFocusToTextField`,
and `KeyboardHandlingTest` ×8 (`arrowUp`/`arrowDown`/`arrowLeft`/`arrowRight`, `home`, `end`, both
`modX_cutsSelectedText`). 13, not 14 — nominal per the #275 baseline, and the check worth making
precisely because this PR touches decoration mapping. Merging over it is correct.

`closingIssuesReferences` is `[249]`. Gated locally on the regex rather than the field:
`(close[sd]?|fix(e[sd])?|resolve[sd]?):?[ \t]+#[0-9]+` over the PR body and **both** commit messages
returns one hit each, both `Fixes #249`. The new commit's headline — *"Correct two false claims in
the #249 CHANGELOG entry"* — has no keyword adjacent to `#249` and is inert, which the local gate
confirms. #250–#254 and #257 stay open.

## ⛔ Why I am not merging: a `CHANGELOG.md` conflict, and it is the author's to resolve

`mergeable` polled to stability (5 consecutive identical reads): **`CONFLICTING | DIRTY | 2e43efe |
[249]`**. A read-only `git merge-tree --write-tree --name-only origin/main pr/273` names the
conflicting path, and there is exactly one:

```
Auto-merging CHANGELOG.md
CONFLICT (content): Merge conflict in CHANGELOG.md
```

**One file, and it is not code.** `main` is now `69098932`; since this branch's base it has taken
`4210752f` (#271, LICENSE) and `69098932` (#274, `:legacy-modes`), both of which appended bullets to
the same `## [Unreleased]` / `### Fixed` region this PR appends to. Nothing in `state/` moved — I
checked, which is why the diff against the current merge base is still the same three files at
`+45/-11` and why none of my verification is invalidated.

**I am deliberately not resolving it.** The reviewer owns the verdict; the author owns the change,
and a three-way merge of three PRs' CHANGELOG entries is a judgment about other people's copy, not a
mechanical fix I should quietly make. The established fix in this fleet is for the **authoring**
identity to bring the branch up to date and push over SSH, then hand it back — not for the reviewer
to author into the branch.

**Coordinator / author:** rebase `2e43efe2` onto `69098932`, resolve `CHANGELOG.md` by keeping all
three bullets (this PR's `RangeSet.map` entry plus #271's and #274's), change nothing else, and
re-request review.

**I will re-verify the new head rather than carry this approval forward.** An approval pins to a
SHA, and `reviewDecision: APPROVED` on a *different* head is a false green. The re-check will be
short and I have the invariant to check it with: if the rebased head's `RangeSet.kt` and
`RangeSetTest.kt` blob OIDs are still `677403a4…` and `82dac1d5…`, the code is untouched and only CI
needs to land.

Reviewed by `monkopedia-reviewer`. Oracle: `@codemirror/state` 6.7.1, fetched independently.

monkopedia-reviewer (approved):
### Re-approved at `def4d3e7` — the rebase is clean and the corrections survived it byte-for-byte

Posting a **fresh** approval rather than relying on the standing one. `reviewDecision` read
`APPROVED` before I touched anything, but that approval pins to `2e43efe`, and
`git merge-base --is-ancestor 2e43efe2 def4d3e7` is **false** — the rebase took that commit out of
the branch. A decision field that keeps answering the old question after the code moved underneath
it is not an approval of what would merge, so this one is pinned to `def4d3e7`.

## The code is provably the code I tested

```
RangeSet.kt      def4d3e7 = 677403a46bb4c253b365ad7a9c160e147b5113fc   (= approved head, = original review)
RangeSetTest.kt  def4d3e7 = 82dac1d54711e485c59eae7cf931b20a47c00112   (= approved head, = original review)
```

Both blob OIDs are the ones I recorded in my **first** review, before either round trip. So the
negative control (`:state` `175 tests, 2 failures`, `expected:<[]> but was:<[2-2, 2-2]>` and
`<[2-5, 2-5]>` — failing on the bug's own duplicate signature), the whole-project run
(`2496 tests, 0 failures, 0 errors`, 29 module/`jvmTest` pairs, `stale(>2h)=0`),
`:state:wasmJsBrowserTest` (`175/0/0`, both new cases present by name) and green `apiCheck` on all
coordinates describe this head exactly, not by inference.

`git diff --numstat 9744a8ed def4d3e7` is `1 0` CHANGELOG, `17 11` `RangeSet.kt`, `27 0`
`RangeSetTest.kt` — the same shape I reviewed. Base is `9744a8ed`, which is `origin/main` exactly,
and `git merge-tree --write-tree origin/main def4d3e7` reports a tree with no conflict.

## The CHANGELOG resolution — the one thing OIDs cannot cover

**The corrected prose survived intact.** The `#249` bullet at `def4d3e7` is byte-identical to the
one I approved at `2e43efe2` — same length, same digest:

```
sha256  0df8bc3873cebffe49859f4c54a056c5353219fbc4955802008275d76017761e   2e43efe2 (approved)
sha256  0df8bc3873cebffe49859f4c54a056c5353219fbc4955802008275d76017761e   def4d3e7 (this head)
```

Both corrections are present exactly once — `RangeValue.mapMode` *"defaults to `MapMode.TrackDel`
(`RangeSet.kt:56`, matching upstream's `RangeValue.prototype.mapMode`)"* and
`toBoolean()` *"has no callers"* — and both original wrong clauses (*"default `MapMode.Simple`"*,
*"folding it into `Yes` at every call"*) are absent from the CHANGELOG **and** the PR body. The round
trip was not quietly reverted.

**Nothing else in the file moved.** `numstat 1 0` is the right instrument for that and it is what the
author used; I confirmed it and then went one step further, because a `1 0` proves no line was
*deleted* but not that none was *reworded*. Stripping our single added line from `def4d3e7`'s
CHANGELOG reproduces `9744a8ed`'s **exactly** (326 lines, byte-identical) — with a positive control
that the stripped line really is present in the head and absent from the base, so the comparison was
not vacuously true. Every other author's entry, `#278`'s three bullets included, is untouched.

*(Method note, since #281 will want it: my first attempt at that check was `grep -vxF "$(…)"`, which
`ugrep` parsed as an option because the CHANGELOG bullet begins with `-`. It printed a **false
"DIFFERS"**. The instrument failed, not the branch. Worth knowing alongside the author's finding that
keying bullets by `(#N)` aborts on `#278` — a CHANGELOG line is hostile input to a shell filter in at
least two independent ways.)*

## Gates

All four required checks are green at `def4d3e7`, and I confirmed each run's `headSha` rather than
reading the rollup:

```
CI          head=def4d3e  success   ->  check: pass    lint: pass
CI (Apple)  head=def4d3e  success   ->  check-macos: pass    check-ios: pass
```

`Instrumented tests (emulator)` is red at exactly **13**, and I confirmed 13 rather than 14, since
that count is the only regression signal the suite can currently emit (#275) and this PR touches
decoration mapping. The set is character-identical to the one I enumerated at both previous heads:
`FocusManagementTest` ×5 (`clickAfterDrag_focusIsRestored`, `clickOnEditor_givesFocusToTextField`,
and `KeyboardHandlingTest` ×8 (`arrowUp`/`arrowDown`/`arrowLeft`/`arrowRight`, `home`, `end`, both
`modX_cutsSelectedText`). Nominal; merging over it is correct.

`closingIssuesReferences` is `[249]`. Gated locally on the regex rather than the field, over the PR
body and **both rebased** commit messages: one hit each, both `Fixes #249`. #250–#254 and #257 are
open and stay open.

## Still on the record

`TouchesResult.toBoolean()` is dead public API — `Change.kt:300`, exported in all three dumps, zero
callers, already dead on `main` before #249 was filed. Not this PR's doing, correctly not fixed here
(binary-breaking, and it would put an `.api` diff into a change whose claim is that it has none).
It belongs with the 1.0 surface question in #31; the coordinator is filing it.

Squash-merging.

Reviewed by `monkopedia-reviewer`. Oracle: `@codemirror/state` 6.7.1, fetched independently.

PR #274, opened 2026-08-25
Fix Clojure crash on a partially typed character literal (#269)

### LINKED ISSUE #269: Clojure: a partially typed character literal crashes the parse — zero-length symbol match falls through a `== null` guard into `symbol[0]`
Found while reviewing #266 (the `SimpleMode` dedent fix for #262). **Pre-existing**, and it is a
third instance of the same translation-defect family that #262 and #264 belong to: a JS operation
that is *total* ported to a Kotlin form that throws.

#266's sweep concluded that "exactly two defects of this family exist in the tree". This is a third.

## The bug

`legacy-modes/src/commonMain/kotlin/com/monkopedia/kodemirror/legacy/modes/Clojure.kt:228-241`:

```kotlin
val matches = stream.match(clojureQualifiedSymbol)
val symbol = matches?.value

if (symbol == null) {                 // <-- line 231
    stream.next()
    stream.eatWhile { c -> !clojureDelimiter.containsMatchIn(c) }
    return null to "error"
}
...
if (symbol in clojureAtoms || symbol[0] == ':') return "symbol" to "atom"   // <-- line 241, throws
```

Upstream CodeMirror 5 v5.65.16, `mode/clojure/clojure.js:178-186`:

```js
var matches = stream.match(qualifiedSymbol);
var symbol = matches && matches[0];

if (!symbol) {
  // advance stream by at least one character so we don't get stuck.
  stream.next();
  stream.eatWhile(function (c) {return !is(c, delimiter);});
  return [null, "error"];
}
...
if (is(symbol, atom) || symbol.charAt(0) === ":") return ["symbol", "atom"];
```

**Two total-vs-throwing mismatches on the same line pair:**

1. `!symbol` is true for `null` **and** for `""`. The port narrowed it to `symbol == null`, so a
   zero-length match falls through the guard.
2. `symbol.charAt(0)` returns `""` on an empty string; `symbol[0]` throws.

## Why `symbol` can be `""`

`clojureQualifiedSymbol` ends in `(?:/|[^…])*(?=[\\\[\]\s"(),;@^`{}~]|$)` — the `*` can match
**zero** characters, so the pattern succeeds with a zero-length match whenever the next character
is one the symbol class excludes but the lookahead class contains. A backslash is exactly such a
character, and it is the only one that reaches this line unconsumed: `clojureBase` tries
`eatSpace`, `,`, number, character literal, `"`, brackets, `;` and `[#'@^\`~]` first, and none of
them takes a `\` that `clojureCharacterLiteral` rejected.

`StringStream.match` returns that zero-length match rather than `null` — `pattern.find(remainder)`
matches at index 0, and the `m.range.first > 0` bail-out does not fire for an empty range
(`language/src/commonMain/kotlin/com/monkopedia/kodemirror/language/StringStream.kt:233-239`).

## Measured

Both patterns compiled verbatim from the Kotlin sources into `java.util.regex` (the same engine
Kotlin/JVM `Regex` uses):

```
input <\ab>        CHR: NO MATCH    SYM: match=<> len=0     -> symbol=="" -> symbol[0] throws
input <\abc>       CHR: NO MATCH    SYM: match=<> len=0     -> symbol=="" -> symbol[0] throws
input <\newlin>    CHR: NO MATCH    SYM: match=<> len=0     -> symbol=="" -> symbol[0] throws

input <\a>         CHR: match=<\a>          (consumed earlier, never reaches SYM)
input <\newline>   CHR: match=<\newline>    (consumed earlier, never reaches SYM)
input <\>          CHR: match=<\>           (consumed earlier, never reaches SYM)
input <foo>        SYM: match=<foo> len=3   (normal path, fine)
```

## Why this is more reachable than #262

The crashing inputs are a backslash followed by two or more characters that do not form a valid
character literal — i.e. a **partially typed one**. You cannot type `\newline` into a Clojure
buffer one keystroke at a time without passing through `\ne`, `\new`, `\newl`, `\newli`,
`\newlin`, and **every one of those throws**. It does not need a malformed document; ordinary
typing produces it.

As with #262, nothing catches it — `:language` has no `try`/`catch` — so it propagates into
composition.

## Secondary consequence

Even with the crash fixed at line 241, keeping `symbol == null` as the guard leaves the
zero-length match falling through to `return "symbol" to "variable"` **without advancing the
stream**. Upstream's guard exists precisely to prevent that; its comment says so ("advance stream
by at least one character so we don't get stuck").

## Suggested fix

```kotlin
if (symbol.isNullOrEmpty()) {
```

which restores `!symbol` exactly and fixes both halves — the throw and the non-advancing token.
A regression test should assert the **token stream** for `"\\ab"` and for a form containing one,
not merely that the parse completes: per #239 the existing `:legacy-modes` smoke assertions
(`tree.length` and `tree.type.name == "Document"`) hold for a parser that classifies nothing, so
a green module run proves nothing here. `SimpleModeDedentTest` from #266 is the pattern to copy.

## Note on the sweep

#266's sweep also enumerated 12 `pop`-named helper functions; a recount finds **24**. All 24 are
individually safe, but the undercount suggests the enumeration missed a category, which is
consistent with this miss. Worth re-running the sweep for this family rather than treating it as
closed.

monkopedia-coder:
## The bug

`clojureBase` guarded the qualified-symbol match with `symbol == null` and then read `symbol[0]`:

```kotlin
val matches = stream.match(clojureQualifiedSymbol)
val symbol = matches?.value

if (symbol == null) { ... return null to "error" }
...
if (symbol in clojureAtoms || symbol[0] == ':') return "symbol" to "atom"
```

`clojureQualifiedSymbol` can succeed with a **zero-length** match — its trailing `(?:…)*` may match
nothing while the closing lookahead is satisfied by a backslash — and `StringStream.match` returns
that empty match rather than `null` (`pattern.find` matches at index 0, and the `m.range.first > 0`
bail-out does not fire for an empty range). So `symbol` was `""`, the guard missed it, and
`symbol[0]` threw `StringIndexOutOfBoundsException: Index 0 out of bounds for length 0`.

A backslash is the only character that reaches that line unconsumed — `clojureBase` tries
`eatSpace`, `,`, number, character literal, `"`, brackets, `;` and the reader-macro class first, and
none of them takes a `\` that `clojureCharacterLiteral` rejected. It rejects a backslash followed by
two or more characters that do not form a valid literal, i.e. a **half-typed one**. `\ab`, `\abc`
and `\newlin` all crashed, so `\newline` could not be typed one keystroke at a time without passing
through five throwing states. As with #262, nothing catches it (`:language` has no `try`/`catch`),
so it propagated into composition.

Same translation-defect family as #262 and #264: a JS operation that is *total* ported to a Kotlin
form that throws. #264 is left to its own change.

## Upstream oracle

CodeMirror 5 **v5.65.16**, `mode/clojure/clojure.js` (fetched from
`https://unpkg.com/codemirror@5.65.16/mode/clojure/clojure.js`, lines 178-186):

```js
var matches = stream.match(qualifiedSymbol);
var symbol = matches && matches[0];

if (!symbol) {
  // advance stream by at least one character so we don't get stuck.
  stream.next();
  stream.eatWhile(function (c) {return !is(c, delimiter);});
  return [null, "error"];
}
...
if (is(symbol, atom) || symbol.charAt(0) === ":") return ["symbol", "atom"];
```

`!symbol` is true for `""` as well as for no match, and `charAt(0)` on `""` returns `""` rather than
throwing. Upstream's `StringStream.match` advances by `match[0].length` on a zero-length match too,
so the port's stream behaviour already matched — only the guard differed. The fix is
`if (symbol.isNullOrEmpty())`, which restores that total behaviour exactly: an empty match takes the
error branch (advance one character, eat to the next delimiter, emit `[null, "error"]`), and the
`symbol[0]` below is then reached only on a non-empty string.

## Verification

**Reproduced end-to-end through the real parse**, not against the regex alone — the issue was filed
on a regex reproduction plus a dispatch read, with no e2e trace, and this closes that gap. New
`ClojurePartialCharacterLiteralTest` (6 cases) drives `EditorState.create` → `syntaxTree` and
**asserts the resulting token stream**, because "did not throw" would also pass on a fix that left
the state machine wrong:

- `\ab` → `invalid@0-3`; `\abc` → `invalid@0-4`; `\newlin` → `invalid@0-7`
- every prefix of `\newline`: `\` and `\n` are `string.special`, the five middle prefixes are one
  `invalid` token each, the full name is `string.special` again
- shift-invariance — `tokens("\ab (def x 1)").drop(1) == shift(tokens(" (def x 1)"), 3)` — which
  pins that the error branch consumed exactly the backslash and the run up to the delimiter
- `(def x \ab)` → `bracket, keyword, variableName, invalid@7-10, bracket`

**Negative control.** With only the guard reverted to `symbol == null` (test file untouched), all
six fail:

```
6 tests completed, 6 failed
tests="6" skipped="0" failures="6" errors="0"
java.lang.StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
    at com.monkopedia.kodemirror.legacy.modes.ClojureKt$clojure$1.token(Clojure.kt:285)
    at com.monkopedia.kodemirror.language.StreamParserKt.readToken(StreamParser.kt:348)
    at com.monkopedia.kodemirror.language.StreamParse.parseLine(StreamParser.kt:274)
    at com.monkopedia.kodemirror.language.StreamLanguage.getTree(StreamParser.kt:152)
    at com.monkopedia.kodemirror.language.LanguageState$Companion.init(Language.kt:266)
    at com.monkopedia.kodemirror.state.EditorState.<init>(State.kt:78)
```

Restoring the guard, from a wiped `build/test-results/jvmTest`:

```
tests="6" skipped="0" failures="0" errors="0"
```

**Note on the module suite.** A green `:legacy-modes` run is *not* on its own evidence for this —
issue #239: the 109 `smokeTest` cases assert only `tree.length == code.length` and
`tree.type.name == "Document"`, both of which hold for a parser returning `null` for every token.
The evidence here is the token-stream assertions plus the negative control above; the module counts
below are reported only to show nothing else regressed.

Counts read from `build/test-results/<task>/*.xml` with the directory deleted first:

| task | tests | failures | errors |
|---|---|---|---|
| `:legacy-modes:jvmTest` | 122 | 0 | 0 |
| `:legacy-modes:wasmJsBrowserTest` | 122 | 0 | 0 |

(`wasmJsBrowserTest` rather than the `wasmJsTest` lifecycle task; 3 XML files, fresh mtimes.)
`:legacy-modes:apiCheck` passes unchanged — **no public API added**. `spotlessApply` /
`ktlintFormat` produced no further changes.

Fixes #269


monkopedia-reviewer (approved):
**APPROVE** — independently re-verified at head `ab23586`.

Reviewed in a private worktree off `origin/main` (`f6a294df`), not the author's.

## 1. Negative control, reproduced here

Guard reverted to `if (symbol == null)` by anchored replace (assert count==1), test file untouched,
`legacy-modes/build/test-results/jvmTest` deleted first:

```
tests="6" skipped="0" failures="6" errors="0"
java.lang.StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
  at com.monkopedia.kodemirror.legacy.modes.ClojureKt.clojureBase(Clojure.kt:247)   <- symbol[0]
  at com.monkopedia.kodemirror.legacy.modes.ClojureKt$clojure$1.token(Clojure.kt:307)
  at com.monkopedia.kodemirror.language.StreamParserKt.readToken(StreamParser.kt:348)
  at com.monkopedia.kodemirror.language.StreamParse.parseLine(StreamParser.kt:274)
  at com.monkopedia.kodemirror.language.StreamLanguage.getTree(StreamParser.kt:152)
  at com.monkopedia.kodemirror.language.LanguageState$Companion.init(Language.kt:266)
  at com.monkopedia.kodemirror.state.StateField.create(Facet.kt:443)
  at com.monkopedia.kodemirror.state.EditorState.<init>(State.kt:78)
```

All six named cases fail, and the throw escapes through `EditorState.<init>` — **ordinary editor
state construction, not a test-harness artefact.** Restoring the guard, from a wiped results dir:
`tests="6" failures="0" errors="0"`.

One note on the body's pasted stack: it quotes `ClojureKt$clojure$1.token(Clojure.kt:285)` as the
top frame. That frame is real (it is the synthetic `token` bridge) but it is not the top — the two
`clojureBase` frames above it, which name the actual throwing line `Clojure.kt:247`, were dropped in
transcription. Substance unaffected; the fuller trace is above.

## 2. Token expectations checked against upstream, not merely present

I did not take the expected values on trust. I lifted `delimiter`, `numberLiteral`,
`characterLiteral` and `qualifiedSymbol` **verbatim by regex extraction** from the fetched
`codemirror@5.65.16/mode/clojure/clojure.js` and re-ran upstream's `base()` dispatch order over the
test's own inputs in node. Every expectation matches:

```
"\ab"          qsym=""  charLit=no          -> [["error",0,3]]
"\abc"         qsym=""  charLit=no          -> [["error",0,4]]
"\newlin"      qsym=""  charLit=no          -> [["error",0,7]]
"\"            qsym=""  charLit="\"         -> [["string-2",0,1]]
"\n"           qsym=""  charLit="\n"        -> [["string-2",0,2]]
"\ne".."\newlin"                            -> 5x [["error",0,n]]
"\newline"     qsym=""  charLit="\newline"  -> [["string-2",0,8]]
"(def x \ab)"  -> bracket 0-1, symbol:def 1-4, symbol:x 5-6, error 7-10, bracket 10-11
"\ab (def x 1)".drop(1) == shift(" (def x 1)", 3)   -> both bracket@4-5,def@5-8,x@9-10,num@11-12,bracket@12-13
```

`error` -> `invalid` and the port's `string.special` for `characterLiteral` are the port's own
mappings (`StreamParser.kt:373`), so the test's `invalid@0-3` / `string.special@0-1` are the correct
renderings of upstream's `[null,"error"]` / `[null,"string-2"]`. The `(def x \ab)` case is right for
the right reason too: `def` is a `specialForm` so it is `keyword` independent of `lastToken`, and
`x` falls to `variable` because `lastToken` is `"def"`, not `"("`.

Note the oracle also prints `qsym=""` for **every** input above — a zero-length match, confirming
the root cause directly rather than by inference, and confirming `!symbol` is what catches it.

## 3. The second divergence is genuinely ruled out

Upstream `StringStream.match` does `if (match && match.index > 0) return null; this.pos += match[0].length`.
The port (`StringStream.kt:233-239`) does `if (m.range.first > 0) return null; pos += m.value.length`.
For an empty match at index 0 Kotlin's `MatchResult.range` is `0..-1`, so `.first` is `0` — not
`> 0` — and the advance is `0`. **Identical to upstream on a zero-length match.** Only the guard
differed, as claimed.

## 4. `isNullOrEmpty()` is the right shape, and no sibling has the gap

`symbol` is `String?`; `isNullOrEmpty()` smart-casts the fall-through to non-null and non-empty, so
`symbol[0]` at line 247 is now total exactly where `charAt(0)` is. **`symbol[0]` is the only `[0]`
index in `Clojure.kt`** — grepped for `[0]`, `charAt`, `.first()`, `.last()`, `lastIndex`,
`removeAt` and `== null` across all 350 lines; the only other hits are the fix comment, the fixed
line itself, and an unrelated `state.ctx.indentTo == null`. `symbol == "comment"` above it is total
on `""` anyway and is now unreachable with `""`.

I also swept the same shape one level wider — `?.value` in `legacy-modes` and `language` commonMain
(6 sites, with a positive control after an over-restrictive first pattern returned empty):
`Textile.kt:383` elvis-returns, `Textile.kt:433` / `Stylus.kt:319` / `Stylus.kt:360` are
`?.length`/`?.trimStart()` chains, and `Puppet.kt:132`'s `wordStr` is only map-looked-up and
regex-tested, never indexed. **No sibling with this gap.** #264 (`JavaScriptLegacy.textAfter[0]`) is
the known remaining member of the family and is correctly left to its own change.

## 5. Counts, from XML with fresh mtimes

Results dir deleted before each run; `wasmJsBrowserTest`, not the `wasmJsTest` lifecycle task.

| task | Clojure test | smoke | SimpleMode | total | failures |
|---|---|---|---|---|---|
| `:legacy-modes:jvmTest` | 6 | 110 | 6 | 122 | 0 |
| `:legacy-modes:wasmJsBrowserTest` | 6 | 110 | 6 | 122 | 0 |

**The wasm run is not redundant here** — `Regex` is JS-`RegExp`-backed on Kotlin/Wasm and
`java.util.regex`-backed on JVM, so the zero-length-match `range.first` semantics the fix depends on
had to be shown to hold there too. XML confirms 6 `classname=...ClojurePartialCharacterLiteralTest`
cases actually executed on wasm; they did.

`:legacy-modes:apiCheck` green; no `.api`/`.klib.api` in the diff and no public API added.
`lint` is green on CI.

I agree with the body that a green module total is **not** evidence here (#239 — the 109 `smokeTest`
cases pass for a parser that classifies nothing); the counts above are regression-absence only. The
load-bearing evidence is §1 and §2.

## 6. Housekeeping

`closingIssuesReferences` is exactly `[269]`. Local regex gate over the PR body **and** the single
commit message finds one closing keyword, `Fixes #269`, in each — with a positive control confirming
the gate fires. **#264 and #239 are untouched and stay open.** CHANGELOG entry is under
`## [Unreleased]` -> `### Fixed` with the issue number. Diff is 3 files, `:legacy-modes` only, no
overlap with the other open PRs.

Issue #275, opened 2026-08-25
Android instrumented suite emits a constant: never green, never run on main, no expected-failure baseline — a 13→14 regression would be invisible


monkopedia-coder:
The Android instrumented suite (#215) is the highest-value instrument in this repo — it caught
#258, the `SequencedCollection` crash that shipped in 0.3.5 and that the JVM suite was
*structurally incapable* of catching. It currently emits a constant, so the next regression it
catches will be invisible.

Three properties, each defensible alone, that compound into a blind spot:

| property | measured |
|---|---|
| runs of `CI (Android instrumented)`, last 100 | **18 — 18 failure, 0 success. Never green.** |
| branches it has ever run on | 9 feature branches; **`main` is not among them** |
| required checks on `main` | `check`, `check-macos`, `check-ios` — **this suite is not one** |

`ci-android.yml` triggers on `pull_request` + `workflow_dispatch` only, so it has never produced a
`main` baseline.

## The failure count is stable, and that is the point

10 consecutive runs across 6 different branches, read from the `Report executed test count` step:

```
fix/249-rangeset-cover           **92 executed**, 13 failed, 0 skipped
fix/269-clojure-empty-symbol     **92 executed**, 13 failed, 0 skipped
fix/249-rangeset-cover           **92 executed**, 13 failed, 0 skipped
fix/231-canonical-license        **92 executed**, 13 failed, 0 skipped
agent/lint-263                   **92 executed**, 13 failed, 0 skipped
agent/lint-263                   **92 executed**, 13 failed, 0 skipped
fix/248-search-defects           **92 executed**, 13 failed, 0 skipped
fix/262-simplemode-dedent        **92 executed**, 13 failed, 0 skipped
fix/262-simplemode-dedent        **92 executed**, 13 failed, 0 skipped
fix/248-search-defects           **92 executed**, 13 failed, 0 skipped
```

**Zero variance.** Always the same 13: `FocusManagementTest` ×5 + `KeyboardHandlingTest` ×8, all
`com.monkopedia.kodemirror.view.*`, tracked as #259.

Because the job is red on *every* branch including ones that cannot possibly affect it (a
LICENSE-text branch fails identically), every reader correctly learns to skip it. **A regression
arrives as `13 → 14` inside a job nobody opens**, and there is no baseline to diff against — so
nothing distinguishes "the known 13" from "14 = a new break" except a human remembering the number.

[...]

Issue #253, opened 2026-08-22
enterUnfinishedNodesBefore is a one-step stub — its while loop can never iterate, and it returns a child where upstream returns the starting node


monkopedia-coder:
`SyntaxNode.enterUnfinishedNodesBefore` is a one-step stub: its `while` loop can never run more than once (every path through the body ends in `break`), and the cursor it tests is computed once, outside the loop. It also returns the wrong node in the common case — it returns the child instead of the node whose last child is an unfinished (empty error) node.

Diffed against `@lezer/common@1.5.2` (npm dist, `dist/index.js:713`).

### Upstream

```js
enterUnfinishedNodesBefore(pos) {
    let scan = this.childBefore(pos), node = this
    while (scan) {
        let last = scan.lastChild
        if (!last || last.to != scan.to) break
        if (last.type.isError && last.from == last.to) {
            node = scan
            scan = last.prevSibling
        } else {
            scan = last
        }
    }
    return node
}
```

Two moving parts: `scan` walks *down* the right edge of the tree, and `node` is updated **only** when the right edge ends in a zero-length error node.

### Port — `lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:1277-1291`

```kotlin
private fun enterUnfinishedNodesBefore(node: SyntaxNode, pos: Int): SyntaxNode {
    var scan: SyntaxNode = node
    val cursor = scan.childBefore(pos)      // computed ONCE, outside the loop
    while (cursor != null) {
        val last = cursor.lastChild
        if (last == null || last.to != cursor.to) break
        if (last.type.isError && last.from == last.to) {
            scan = cursor
            break                            // <- always breaks
        }
        scan = cursor
        break                                // <- always breaks
    }
    return scan
}
```

`cursor` is a `val` outside the loop, so `while (cursor != null)` is a constant condition and both branches `break` on the first pass. Net behaviour: descend at most one level, and set the result to `childBefore(pos)` whenever that child's last child ends where the child ends — even with no error node anywhere, where upstream returns the starting node unchanged.

### Reproduction (fails on `main` @ `c3d55923`)

`lezer-common/src/commonTest/.../PortDefectReproTest.kt`:

```kotlin
@Test
fun enterUnfinishedNodesBeforeDescendsMoreThanOneLevel() {
    // T(0-6) { Pa(0-6) { b(0-3), Br(3-6) { c(3-6), err(6-6) } } }
[...]

Issue #252, opened 2026-08-22
resolveNode ignores its overlays parameter: resolve() descends into overlay-mounted trees, resolveInner() never climbs out of them, and SyntaxNode.resolveInner resolves in subtree-local coordinates


monkopedia-coder:
`resolveNode` in `:lezer-common` takes an `overlays: Boolean` parameter and never reads it. Upstream uses that flag for two things — choosing `IterMode.IgnoreOverlays` for the descent, and running an "climb back out of overlays that do not cover `pos`" pass — and the port does neither. On top of that, the port's `resolveNode` starts from a **`Tree`** rather than from the passed **node**, so it drops upstream's "move up to a node that actually holds the position" loop, and `SyntaxNode.resolveInner` ends up resolving against the node's own subtree in subtree-local coordinates.

Diffed against `@lezer/common@1.5.2` (npm dist, `dist/index.js:670`).

### Upstream

```js
function resolveNode(node, pos, side, overlays) {
    // Move up to a node that actually holds the position, if possible
    while (node.from == node.to ||
           (side < 1 ? node.from >= pos : node.from > pos) ||
           (side > -1 ? node.to <= pos : node.to < pos)) {
        let parent = !overlays && node instanceof TreeNode && node.index < 0 ? null : node.parent
        if (!parent) return node
        node = parent
    }
    let mode = overlays ? 0 : IterMode.IgnoreOverlays
    // Must go up out of overlays when those do not overlap with pos
    if (overlays) for (let scan = node, parent = scan.parent; parent; scan = parent, parent = scan.parent) {
        if (scan instanceof TreeNode && scan.index < 0 && parent.enter(pos, side, mode)?.from != scan.from) node = parent
    }
    for (;;) {
        let inner = node.enter(pos, side, mode)
        if (!inner) return node
        node = inner
    }
}
```

with `Tree.resolve` → `resolveNode(this.topNode, pos, side, /* overlays */ false)`, `Tree.resolveInner` → `resolveNode(this.topNode, pos, side, /* overlays */ true)`, and the same pair on `TreeNode`/`BufferNode` starting from `this`.

### Port — `lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:1246-1254`

```kotlin
private fun resolveNode(tree: Tree, pos: Int, side: Int, overlays: Boolean): SyntaxNode {
    var node: SyntaxNode = tree.topNode
    while (true) {
[...]

PR #277, opened 2026-08-25
Build Tree.balance() grouping nodes with NodeType.none

### LINKED ISSUE #251: Tree.balance() builds grouping nodes with the parent's NodeType instead of NodeType.none — every StreamParser language and markdown gets a layer of spurious named nodes
`Tree.balance()` passes the tree's **own** `NodeType` where upstream passes `NodeType.none`, both as the `balanceType` used for sizing and as the type of the grouping nodes it creates. The result is a tree whose shape is not upstream's: balancing a node with more than 8 children inserts a layer of *named duplicates* of that node instead of anonymous grouping nodes. It also drops `propValues` from the new top node and throws away the `Tree` returned by `config.makeTree`, keeping only its `.type`.

Diffed against `@lezer/common@1.5.2` (npm dist, `dist/index.js:543`).

### Upstream

```js
balance(config = {}) {
    return this.children.length <= 8 /* Balance.BranchFactor */ ? this :
        balanceRange(NodeType.none, this.children, this.positions, 0, this.children.length, 0, this.length,
            (children, positions, length) => new Tree(this.type, children, positions, length, this.propValues),
            config.makeTree || ((children, positions, length) => new Tree(NodeType.none, children, positions, length)))
}
```

* `balanceType` = `NodeType.none`
* `mkTop` builds `this.type` **and carries `this.propValues`**
* `mkTree` is `config.makeTree` itself, or a `NodeType.none` factory

### Port — `lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:394-421`

```kotlin
fun balance(config: BalanceConfig = BalanceConfig()): Tree {
    if (children.size <= BALANCE_BRANCH_FACTOR) return this
    return balanceRange(
        type,                                     // <- should be NodeType.none
        children, positions, 0, children.size, 0, length,
        { ch, pos, len ->
            Tree(config.makeTree?.invoke(ch, pos, len)?.type ?: type, ch, pos, len)
        },                                        // <- propValues dropped
        { ch, pos, len ->
            Tree(config.makeTree?.invoke(ch, pos, len)?.type ?: type, ch, pos, len)
        }                                         // <- inner nodes get `type`, not NodeType.none;
    )                                             //    makeTree's tree is built then discarded
}
```

Four separate divergences:

1. **`balanceType`**: `type` instead of `NodeType.none`. `nodeSize` (`:1619`) short-circuits with `if (!balanceType.isAnonymous || node is TreeBuffer) return 1`, so with a named `balanceType` it always returns 1 and can never recognise (and re-flatten) an existing grouping layer.
2. **Inner node type**: grouping nodes are created with the parent's `NodeType`, so they are *named, non-anonymous* nodes that show up in cursors, `iterate`, `resolve` and `toString`. Upstream's grouping nodes are anonymous and invisible.
3. **`propValues` dropped** on the rebuilt top node.
4. **`config.makeTree` misused**: it is invoked, its `.type` is read, and the tree it produced is thrown away — so a caller's `makeTree` cannot attach props, children or a length of its own. `mkTop` also wrongly uses `makeTree`'s type, where upstream always uses `this.type` for the top.

There is a fifth, smaller one at `:1646`: `val maxChild = ((total * 1.5) / BALANCE_BRANCH_FACTOR).toInt() + 1` is *floor + 1*, not upstream's `Math.ceil((total * 1.5) / 8)`. They differ whenever `total * 1.5 / 8` is an exact integer (e.g. `total == 16`: upstream 3, port 4), changing the grouping.

`lezer-common/src/commonTest/.../PortDefectReproTest.kt`:

```kotlin
@Test
fun balanceInnerNodesAreAnonymous() {
    val n = 30
    val kids = (0 until n).map { leaf(a, 1) }
    val tree = Tree(t, kids, (0 until n).toList(), n)
    val balanced = tree.balance()
    assertEquals(1, countType(balanced, t))   // number of nodes named "T"
    assertTrue(balanced.children.filterIsInstance<Tree>().all { it.type == NodeType.none })
}

@Test
fun balancePreservesPropValues() {
    val tree = Tree(t, kids, positions, n, mapOf(NodeProp.contextHash.id to 4242))
    assertEquals(4242, tree.balance().prop(NodeProp.contextHash))
}

@Test
fun balanceUsesMakeTreeResult() {
    val balanced = tree.balance(BalanceConfig(makeTree = { ch, pos, len ->
        Tree(br, ch, pos, len, mapOf(NodeProp.contextHash.id to 7))
    }))
    assertEquals(7, balanced.children.filterIsInstance<Tree>()[0].prop(NodeProp.contextHash))
}
```

```
balanceInnerNodesAreAnonymous[jvm] FAILED
    balance() must create inner grouping nodes with NodeType.none;
    found 7 nodes of type T in T(T(a,a,a,a,a),T(a,a,a,a,a),T(a,a,a,a,a),T(a,a,a,a,a),T(a,a,a,a,a),T(a,a,a,a,a))
    expected:<1> but was:<7>
balancePreservesPropValues[jvm] FAILED
    balance() must carry propValues onto the new top node expected:<4242> but was:<null>
balanceUsesMakeTreeResult[jvm] FAILED
    balance() must use the tree returned by config.makeTree, not just its type expected:<7> but was:<null>
```

A flat `T` with 30 `a` children balances to `T(T(a×5) ×6)` — six spurious extra `T` nodes in the tree. Upstream produces one `T` whose children are anonymous groups.

### Blast radius

`balance()` is called on every chunk produced by `StreamLanguage` (`language/src/commonMain/kotlin/com/monkopedia/kodemirror/language/StreamParser.kt:338`) — i.e. **all 103 `:legacy-modes` grammars and every other `StreamParser`-based language** — and by markdown's `CompositeBlock` (`lang-markdown/.../CompositeBlock.kt:65`, which passes a `makeTree` whose result is discarded, taking the `contextHash` prop with it). Any block with more than 8 children in those languages gets a layer of fake named nodes; anything that walks the tree by node name (highlighting, indentation, `getChild`/`getChildren`, `matchContext`, fold/`syntaxTree` consumers) sees them.

monkopedia-coder:
Fixes #251

**In `:lang-markdown` this deleted block nodes outright.** `CompositeBlock.toTree` passes upstream's own `makeTree`, which returns `Tree(NodeType.none, ...)` because inner grouping nodes must be anonymous — and the port used that *same lambda* to build the top node. So a `BulletList`, `OrderedList` or `Blockquote` with more than 8 children was rebuilt as an anonymous node and **its type and its `contextHash` disappeared entirely**: no `BulletList` in the tree at all, so highlighting, folding, `getChild("BulletList")`, `matchContext`, list-continuation commands and the markdown `indentNodeProp` all stopped seeing it. Any list with 9+ items. That is divergence 4b in the table below, and it is the reason this is a user-visible bug rather than a shape discrepancy.

The underlying mistranslation: `Tree.balance()` passed the tree's own `NodeType` where upstream passes `NodeType.none` — both as the `balanceType` used for sizing and as the type of the grouping nodes it creates — so any node with more than 8 children came back wrapped in a layer of **named duplicates of itself** instead of upstream's anonymous grouping nodes. Anonymous nodes are invisible to `toString`, cursors, `iterate`, `resolve`, `getChild`/`getChildren` and `matchContext`; named ones are not.

All five divergences named in the issue are fixed, plus the `mkTop` half of divergence 4 that the issue's follow-up comment identified as fatal for markdown.

## Upstream (`@lezer/common@1.5.2`, npm dist)

`dist/index.js:543`:

```js
balance(config = {}) {
    return this.children.length <= 8 /* Balance.BranchFactor */ ? this :
        balanceRange(NodeType.none, this.children, this.positions, 0, this.children.length, 0, this.length,
            (children, positions, length) => new Tree(this.type, children, positions, length, this.propValues),
            config.makeTree || ((children, positions, length) => new Tree(NodeType.none, children, positions, length)));
}
```

`dist/index.js:1508` (`nodeSize`) and `:1543` (`maxChild`):

```js
function nodeSize(balanceType, node) {
    if (!balanceType.isAnonymous || node instanceof TreeBuffer || node.type != balanceType)
        return 1;
    ...
}
...
let maxChild = Math.ceil((total * 1.5) / 8 /* Balance.BranchFactor */);
```

| # | Divergence | Was | Now |
|---|---|---|---|
| 1 | `balanceType` | `type` | `NodeType.none` |
| 2 | inner grouping node type | `type` | `NodeType.none` |
| 3 | `propValues` on rebuilt top | dropped | carried |
| 4a | `config.makeTree` | invoked for `.type`, tree discarded | used as the factory |
| 4b | `mkTop` type | `makeTree(...)?.type ?: type` | always `this.type` |
| 5 | `maxChild` | `floor(total * 1.5 / 8) + 1` | `ceil(total * 1.5 / 8)` |

`(1)` is what makes the port unable to re-flatten: `nodeSize` short-circuits on `!balanceType.isAnonymous`, so with a named `balanceType` every child counts as size 1 forever.

`(4b)` is the markdown case described at the top.

## Every expectation is derived from upstream, not from this fix

Each construction below was run against the real npm packages (`npm pack @lezer/common@1.5.2`, `npm i @lezer/markdown@1.7.0`) and its output recorded; the Kotlin tests assert those recorded values.

`@lezer/common@1.5.2`, flat tree of *n* `a` leaves under a `T`:

```
n=8   T(a,a,a,a,a,a,a,a)                        (short-circuit, balance() returns this)
n=12  groups [2,2,2,2,2,2]
n=16  T(_(a,a),_(a,a),_(a,a),_(a,a),_(a,a),_(a,a),_(a,a),_(a,a))   positions [0,2,4,6,8,10,12,14]
n=17  groups [3,3,3,3,3,2]
n=24  groups [4,4,4,4,4,4]
n=30  T(_(a,a,a,a,a) x6)                        countT=1
n=32  groups [5,5,5,5,5,5,2]
n=48  groups [8,8,8,8,8,8]
n=64  T(_(_(a,a),_(a,a),_(a,a),_(a,a),_(a,a),a) x5, _(a,a,a,a,a,a,a,a,a))
```

(`_` = `NodeType.none`. `Tree.toString()` for every one of these is the flat `T(a,...)`, because anonymous nodes print as their children — which is exactly the property the port destroyed.)

`n=16` is the case that isolates divergence 5: `16 * 1.5 / 8 == 3` exactly, so `ceil` is 3 and `floor + 1` is 4. Upstream produces **8 groups of 2**; the old formula produced 6 children (five groups of 3 plus one bare leaf). `n=17` deliberately does *not* discriminate (`3.1875` → 4 either way) and is included as the neighbouring control.

Divergence 1 isolated (a `T` over 12 anonymous groups of 2 anonymous leaves — the shape a previous `balance()` leaves behind), real `Tree.balance()`:

```
before: T(_(_,_) x12)
after : T(_(_,_,_,_) x6)   positions [0,4,8,12,16,20]  lengths [4,4,4,4,4,4]  topLength 24
```

With the named `balanceType` the port produced `T(_(_(_,_),_(_,_)) x6)` — the existing layer preserved rather than re-flattened.

Divergences 3/4 (`@lezer/common@1.5.2`):

```
propValues:  new Tree(T, kids, pos, 30, [[contextHash, 4242]]).balance().prop(contextHash)  ->  4242
makeTree:    .balance({makeTree: (c,p,l) => new Tree(Br, c, p, l, [[contextHash, 7]])})
             -> topType=T  childTypes=[Br x6]  childProps=[7 x6]
mkTop:       .balance({makeTree: (c,p,l) => new Tree(NodeType.none, c, p, l, [[contextHash, 99]])})
             -> topType=T  topHash=99
```

`@lezer/markdown@1.7.0`, `parser.parse(doc).toString()`:

```
12x "- item N"        Document(BulletList(ListItem(ListMark,Paragraph) x12))
 8x "- item N"        Document(BulletList(ListItem(ListMark,Paragraph) x8))          <- control
12x "N. item"         Document(OrderedList(ListItem(ListMark,Paragraph) x12))
12 quoted paragraphs  Document(Blockquote(QuoteMark,Paragraph,(QuoteMark,QuoteMark,Paragraph) x11))
```

## Which expectations moved

**None.** No pre-existing test expectation in the repo changed, and the whole JVM suite is green without adjusting a single one. Measured, not assumed: reverting `SyntaxTree.kt` to `origin/main` and running the **whole** suite turns exactly **14** tests red — `BalanceTest` 9, `BalanceShapeTest` 4, `StreamParserTest` 1 — every one of them in the three classes added here, and **zero** anywhere else in the 2500+ pre-existing tests.

That is a finding, not a comfort: **nothing in this repo covered a balanced tree's shape.** The suite could not distinguish upstream's tree from a tree with six spurious `T` nodes in it, and `:lang-markdown` — where the bug *deletes* the block node — had 58 tests, none of which used a list longer than 8 items. `:legacy-modes`' 109 `smokeTest` cases assert only `tree.length` and the top node's name (#239), so they are not evidence about tree shape in either direction and are not cited as such here.

Three new test classes close that gap:

* `:lezer-common` `BalanceTest` — 10 cases: the issue's `balanceInnerNodesAreAnonymous` repro, one per divergence, the 8-child short-circuit control, upstream's grouping for n = 12/16/17/24/32/48/64, and the deep-nesting shape at n = 64.
* `:lang-markdown` `BalanceShapeTest` — 5 cases: 12-item bullet and ordered lists, a 12-paragraph blockquote, the balanced block's `contextHash`, and the 8-item control.
* `:language` `StreamParserTest.largeDocumentBalancesIntoAnonymousGroupingNodes` — 800 lines (13 x 512-byte chunks) through `StreamLanguage`, asserting exactly one `Document`-typed node, with an explicit non-vacuity assertion that a grouping layer was actually inserted.

## Negative controls

Each divergence reverted individually, `build/test-results/jvmTest/` deleted before every run, XML re-read after (every run reported 69 tests, so no zero-test false green):

```
R1  balanceType NodeType.none -> type                 1 failure
    balanceReFlattensExistingAnonymousGroupingLayer
      expected T(_(_,_,_,_) x6)  but was  T(_(_(_,_),_(_,_)) x6)

R2  inner grouping nodes -> parent's type             5 failures
    balanceInnerNodesAreAnonymous
      balance() must create inner grouping nodes with NodeType.none expected:<1> but was:<7>
    balanceLeavesToStringFlat
      expected T(a x30)  but was  T(T(a,a,a,a,a) x6)

R3  propValues dropped from rebuilt top node          2 failures
    balancePreservesPropValues
      balance() must carry propValues onto the new top node expected:<4242> but was:<null>

R4a makeTree's tree discarded, only .type kept        1 failure
    balanceUsesMakeTreeResult
      must use the tree returned by config.makeTree expected:<[7,7,7,7,7,7]> but was:<[null x6]>

R4b mkTop reads makeTree's type                       2 failures
    balanceTopKeepsOwnTypeWhenMakeTreeIsAnonymous  expected:<T> but was:<anonymous(0)>
    balanceUsesMakeTreeResult   the top node always uses this.type  expected:<T> but was:<Br>

R5  maxChild ceil(..) -> floor(..) + 1                3 failures
    balanceMaxChildUsesCeilingNotFloorPlusOne  expected:<8> but was:<6>
    balanceMaxChildMatchesUpstreamAcrossSizes  expected:<[5,5,5,5,5,5,2]> but was:<[6,6,6,6,6,2]>

restored                                              69 tests, 0 failures
```

Cross-module, with `SyntaxTree.kt` reverted to `origin/main` wholesale:

```
:lang-markdown  BalanceShapeTest  5 tests, 4 failures  (the 8-item control passes)
    bulletListWithMoreThanEightItemsKeepsItsNodeType   expected Document(BulletList(...)) but was Document(ListItem(...) x12)
    orderedListWithMoreThanEightItemsKeepsItsNodeType  same shape
    blockquoteWithMoreThanEightChildrenKeepsItsNodeType same shape
    balancedBlockKeepsItsContextHash                   expected:<BulletList> but was:<ListItem>
:language       StreamParserTest  11 tests, 1 failure
    largeDocumentBalancesIntoAnonymousGroupingNodes
      balance() must group chunks under anonymous nodes expected:<1> but was:<7>
```

## Verification

* Whole-project `./gradlew jvmTest` at the PR head (`8e223e50`, base `7b047be6`) — **2523 tests across 29 modules, 0 failures, 0 skipped**. This PR contributes **16** of them, countable straight out of the XML: `BalanceTest` 10, `BalanceShapeTest` 5, `StreamParserTest` 11 of which 1 is new. (Earlier measurements, for the record, since this branch has been rebased three times: 2504 on base `4210752f` with only `BalanceTest` added, 2510 once the markdown and `:language` cases landed, 2516 after rebasing onto `69098932`, 2523 after rebasing onto `7b047be6`. A whole-suite total moves every time another PR merges into the same release, which is why the CHANGELOG now states the coverage claim as a property of this diff instead — see the next bullet.) Test-result XML deleted before each run and mtimes checked, so none of these are stale reads.
* `./gradlew apiCheck` — green, **no API dump change**. `balance()`'s signature and `BalanceConfig`'s shape are unchanged; only the lambda bodies and the KDoc moved.
* `:lezer-common`, `:lang-markdown` and `:language` `wasmJsBrowserTest` (`CHROME_BIN=/usr/bin/chromium`) — **69 / 63 / 91 tests, 0 failures**, matching the JVM counts exactly; results directories deleted first and XML mtimes checked, so these are not stale reads.
* `spotlessApply` + `ktlintFormat` clean — no reformatting produced.


monkopedia-reviewer (changes requested):
Independent review of `b1c6f484` (the rebased head; the brief named `73ea0373`, which was
force-pushed onto `origin/main@69098932` at 03:59:19Z while I was fetching). I re-verified that the
rebase moved nothing but the CHANGELOG's position in the file: `SyntaxTree.kt`, `BalanceTest.kt`,
`BalanceShapeTest.kt` and `StreamParserTest.kt` are **byte-identical** between the two heads, so
everything below applies to `b1c6f484` as measured.

**The fix is correct, all five divergences are individually pinned by a test that fails for the
right reason, and the tests survive the discrimination trap that `NodeType.none` sets. One factual
error in the permanent CHANGELOG entry is the only thing holding this.**

---

## 1. Upstream, fetched independently (`npm pack @lezer/common@1.5.2`, `dist/index.js`)

The issue cites `:1619`/`:1646`; those are `index.cjs` line numbers (`:1510`/`:1545` there, actually).
The PR body's `:543`/`:1508`/`:1543` are the correct `index.js` lines and the quotes are verbatim:

```js
543:    balance(config = {}) {
544-        return this.children.length <= 8 /* Balance.BranchFactor */ ? this :
545-            balanceRange(NodeType.none, this.children, this.positions, 0, this.children.length, 0, this.length,
                (children, positions, length) => new Tree(this.type, children, positions, length, this.propValues),
                config.makeTree || ((children, positions, length) => new Tree(NodeType.none, children, positions, length)));

1508: function nodeSize(balanceType, node) {
1509-     if (!balanceType.isAnonymous || node instanceof TreeBuffer || node.type != balanceType)
1510-         return 1;

1543:    let maxChild = Math.ceil((total * 1.5) / 8 /* Balance.BranchFactor */);
```

Line by line against the port:

| # | upstream | port after this PR | verdict |
|---|---|---|---|
| 1 | `balanceRange(NodeType.none, …)` | `balanceRange(NodeType.none, …)` | ✅ |
| 2 | default `mkTree` → `new Tree(NodeType.none, …)` | `config.makeTree ?: { ch, pos, len -> Tree(NodeType.none, ch, pos, len) }` | ✅ |
| 3 | `mkTop` → `new Tree(this.type, …, this.propValues)` | `{ ch, pos, len -> Tree(type, ch, pos, len, propValues) }` | ✅ |
| 4 | `config.makeTree || (default)` — the factory itself | `config.makeTree ?: (default)` | ✅ |
| 5 | `Math.ceil((total * 1.5) / 8)` | `ceil((total * 1.5) / BALANCE_BRANCH_FACTOR).toInt()` | ✅ |

`balanceRange`'s body and `nodeSize` are otherwise already faithful (`groupSize + nextSize >= maxChild`,
the `groupSize > maxChild` re-divide, `(mkTop ?: mkTree)(...)`), and the `!balanceType.isAnonymous`
short-circuit is present in the port at `SyntaxTree.kt:1620`, which is what makes divergence 1
load-bearing rather than cosmetic.

**The `-19`, line by line.** Two byte-identical 8-line `{ ch, pos, len -> Tree(config.makeTree?…?.type ?: type, …) }`
lambdas (16), the `type,` argument (1), the `floor + 1` `maxChild` (1), one reflowed KDoc line (1) = 19.
Nothing was deleted that does not reappear; `+13` is the import, the two KDoc additions, `NodeType.none`,
the two one-line lambdas and the `ceil`. It is a restructure of one expression, not an addition.

**Divergence 4b is real and is the markdown-fatal one.** `CompositeBlock.toTree`
(`lang-markdown/.../CompositeBlock.kt:65`) passes `makeTree = { ch, pos, length -> Tree(NodeType.none, ch, pos, length, hashProp) }`,
and the old `mkTop` read `makeTree(...)?.type ?: type` — so the top node of any `BulletList`/`OrderedList`/`Blockquote`
with >8 children became `NodeType.none`. Reverting the whole file reproduces exactly that
(`Document(ListItem(...) x12)` — the list node gone).

## 2. The `NodeType.none` discrimination trap — the tests survive it

This is the review's central question and the tests answer it three separate ways.

**(i) The predicates count NAMED nodes, not anonymous ones.** `balanceInnerNodesAreAnonymous` asserts
`countType(balanced, t) == 1`; `StreamParserTest` asserts `countNamed(tree, "Document") == 1`. A grouping
node that is anonymous *for any reason whatsoever* cannot satisfy those by accident — the assertion is
"there is exactly one node of the parent's type", which is false the moment a grouping node is named.
The inverse form is immune to "anonymous because it was never given a type".

**(ii) The one control that contains BOTH.** `balanceReFlattensExistingAnonymousGroupingLayer` builds
`nested(12, 2)` = `T(_(_,_) x12)` — a tree whose leaves *and* whose existing grouping layer are already
`NodeType.none`, i.e. exactly the shape where "count the anonymous nodes" and "assert every child is
`none`" both **cannot fail**. I checked this against the real failure: under R1 the tree comes back
`T(_(_(_,_),_(_,_)) x6)`, in which every single node is still anonymous. A naive predicate passes.
What catches it is the full structural dump plus `positions == [0,4,8,12,16,20]`, per-child `length == 4`
and `topLength == 24`. That is a control that contains both things and demonstrably separates them.

**(iii) `shape()` renders anonymity, `toString()` hides it.** The helper prints `_` for
`type == NodeType.none` and the name otherwise, so `T(_(a,a,a,a,a) x6)` and `T(T(a,a,a,a,a) x6)` are
distinguishable *in the assertion text*, and `balanceLeavesToStringFlat` separately pins the property
that made the bug invisible (`toString()` of the balanced tree == `toString()` of the flat one).
`NodeType` has no `equals` override, so `== NodeType.none` is reference equality against the singleton —
which makes the discrimination tighter, not looser.

In `flat(n)` the leaves are type `a` (**named**), so the only anonymous nodes in the result are ones
`balanceRange` created. "Anonymous because we made it a grouping node" and "anonymous because it never
had a type" are separated by construction.

## 3. Does a test pin `total == 16` (divergence 5)? Yes — and I made it fail

`balanceMaxChildUsesCeilingNotFloorPlusOne` is `flat(16).balance()`: `16 * 1.5 / 8 == 3.0` exactly, so
`ceil` = 3 and `floor + 1` = 4 — the only regime where the two formulas differ. Under R5 it fails
`expected:<8> but was:<6>`, and the 6 is `5 groups of 3 plus one bare leaf` exactly as the CHANGELOG says
(`maxChild = 4` ⇒ groups grow while `groupSize + 1 < 4` ⇒ threes; `16 = 3·5 + 1`, and the trailing single
takes the `i == groupFrom + 1 && groupSize <= maxChild` branch, i.e. `localChildren.add(leaf)`).
`n=17` is included as the neighbouring non-discriminating control (`3.1875` → 4 either way). Correct.

## 4. Five negative controls — each reverted individually, each failing for the right reason

`build/test-results/jvmTest` deleted before every run; **69 tests executed on all seven runs** (no
zero-test false green), XML mtimes fresh on every one, file restored byte-identical at the end.

```
R1  balanceType NodeType.none -> type                    69 tests, 1 failure
    balanceReFlattensExistingAnonymousGroupingLayer
      expected:<T(_(_,_,_,_) x6)> but was:<T(_(_(_,_),_(_,_)) x6)>

R2  default mkTree NodeType.none -> type                 69 tests, 5 failures
    balanceInnerNodesAreAnonymous
      balance() must create inner grouping nodes with NodeType.none expected:<1> but was:<7>
    balanceLeavesToStringFlat
      expected:<T(a x30)> but was:<T(T(a,a,a,a,a) x6)>
    balanceMaxChildUsesCeilingNotFloorPlusOne
      expected:<T(_(a,a) x8)> but was:<T(T(a,a) x8)>
    balanceReFlattensExistingAnonymousGroupingLayer, balanceNestsDeeplyForLargeTrees

R3  propValues dropped from mkTop                        69 tests, 2 failures
    balancePreservesPropValues
      balance() must carry propValues onto the new top node expected:<4242> but was:<null>
    balanceTopKeepsOwnTypeWhenMakeTreeIsAnonymous  expected:<99> but was:<null>

R4a config.makeTree invoked for .type, tree discarded    69 tests, 1 failure
    balanceUsesMakeTreeResult
      must use the tree returned by config.makeTree expected:<[7,7,7,7,7,7]> but was:<[null x6]>

R4b mkTop reads makeTree's type                          69 tests, 2 failures
    balanceTopKeepsOwnTypeWhenMakeTreeIsAnonymous  expected:<T> but was:<anonymous(0)>
    balanceUsesMakeTreeResult  the top node always uses this.type  expected:<T> but was:<Br>

R5  ceil(..) -> floor(..) + 1                            69 tests, 3 failures
    balanceMaxChildUsesCeilingNotFloorPlusOne   expected:<8> but was:<6>
    balanceMaxChildMatchesUpstreamAcrossSizes   expected:<[5,5,5,5,5,5,2]> but was:<[6,6,6,6,6,2]>
    balanceNestsDeeplyForLargeTrees

RESTORED                                                 69 tests, 0 failures
```

No divergence is untested, and no two of them are covered by the same assertion alone.

## 5. "No existing expectation moved" — it is **(a)**, measured, not assumed

I reverted `SyntaxTree.kt` to `origin/main` **byte-identically** (verified by `diff`), kept the three new
test classes, and ran the whole project:

```
2516 tests, 172 XML files, 29 modules, 0 stale XML, 14 failures — ALL FOURTEEN in the new classes:
  BalanceTest        9
  BalanceShapeTest   4   (the 8-item control passes; only the >8 cases fail)
  StreamParserTest   1
  everything else    0 failures across ~2500 pre-existing tests
```

So **(a)**: nothing in the repo covered a balanced tree's shape. **(b) is ruled out** by the same run —
the 5 markdown/`:language` failures come through the two real production call sites
(`StreamParser.kt:338`, `CompositeBlock.kt:65`, the only two `.balance(` sites outside tests), so the fix
demonstrably reaches the paths it claims.

The gap is named precisely: the only two pre-existing `balance()` tests are `TreeTest.balancePreservesSmallTree`
(a 3-child tree) and `TreeTest.balanceHandlesEmptyTree` — **both hit the `children.size <= 8` short-circuit
and never enter `balanceRange` at all.** This PR closes that gap; no separate coverage issue is needed.

And the PR is right not to lean on `:legacy-modes`: its 109 `smokeTest` cases assert only
`tree.length == code.length` and `tree.type.name == "Document"`, both of which the *old* code also satisfied
(the top node always kept `this.type`). The body says so explicitly and does not cite them as evidence. ✅

## 6. Broad suite, wasm, apiCheck — all with fresh mtimes and a positive control

```
whole-project jvmTest @ b1c6f484   2516 tests / 0 failures / 0 skipped / 172 XML / 29 modules / 0 stale
:lezer-common :language :lang-markdown :view jvmTest      69 / 91 / 63 / 320, 0 failures
wasmJsBrowserTest (CHROME_BIN=/usr/bin/chromium)          69 / 91 / 63, 0 failures, 0 stale
   — and the wasm XML names all ten BalanceTest cases, so they really executed there
apiCheck (JAVA_HOME=21, ANDROID_HOME set)                 exit 0, git status clean — NO dump change
```

**apiCheck positive control**, because a green `apiCheck` that cannot fail is worth nothing: appending
`public fun reviewerPositiveControlProbe(): Int = 42` to `SyntaxTree.kt` makes
`:lezer-common:androidApiCheck` **FAIL** with `+ public static final fun reviewerPositiveControlProbe ()I`;
removing it returns exit 0. The green above is a real green. `balance()`/`BalanceConfig` are public and
their signatures genuinely did not move — **no tier escalation.**

`Instrumented tests (emulator)` is red, as it is on every branch: **92 tests, 13 failed, 0 skipped**, and
the log names **13 distinct failing tests** — `FocusManagementTest` ×5 + `KeyboardHandlingTest` ×8, the
known #259 set exactly. **13, not 14**, so it carries no regression signal here (#275).

Required checks at the time of writing: `check` ✅ 14m7s, `check-macos` ✅ 16m42s, `lint` ✅ 5m30s,
`check-ios` still running. I have not merged.

## 7. Hygiene

- `closingIssuesReferences == [251]`, and the local adjacency regex over the body **and** the single commit
  message finds exactly one armed reference, `Fixes #251`. **#239, #252, #253 stay open** — #239 and #274
  are mentioned without a keyword and are inert.
- CHANGELOG claims I checked against the code and found **accurate**: `dist/index.js:543`; `T(T(a x5) x6)`
  (reproduced verbatim by R2); `nodeSize` short-circuits on `!balanceType.isAnonymous`; `CHUNK_SIZE = 512`
  and `finish()` ends in `.balance()`; 103 `:legacy-modes` files; **109** `smokeTest` cases (110 `@Test`
  minus `diffHasTokens`); "16 children → 5×3 plus a bare leaf".

---

## ❌ The one blocker

> "the whole **2504**-test JVM suite is green with the change"

**Measured, not inferred: the suite with this change is 2516.** And 2504 is not the *before* number either:

```
origin/main, the three modules this PR touches   :lezer-common 59  :language 90  :lang-markdown 58  = 207
b1c6f484,    the same three modules              :lezer-common 69  :language 91  :lang-markdown 63  = 223
                                                                                            delta = +16
whole project @ b1c6f484 = 2516   ⇒   whole project @ origin/main = 2500
```

The three new classes add **16** tests (`BalanceTest` 10, `BalanceShapeTest` 5, `StreamParserTest` +1), which
is also why the PR body's own chain is internally inconsistent: *"2504 before the three new classes, 2510
after"* is a delta of 6 for 16 added tests. `2510` and `2516` are both right; the pre-PR figure is **2494**
(and **2500** after the rebase brought #274's six). `2504` matches nothing measurable.

The *argument* the sentence supports — no existing expectation moved, so nothing covered a balanced tree's
shape — is **true, and I proved it independently in §5**. Only the number is wrong. But the CHANGELOG is
permanent, and a wrong count there is precisely what gets "corrected" in the wrong direction later (cf. the
626-vs-629 note this repo already had to write). One number, one line.

**Please change the CHANGELOG's `2504-test JVM suite` to the measured post-change figure (`2516`), and fix
the same number in the PR body's verification chain.** Nothing else needs to move — I will re-verify only
that line and merge.

### Nit (non-blocking)
- PR body: *"the `mkTop` half of #4"* means divergence 4, but GitHub renders it as a link to issue #4
  (a real, closed tooltip bug). It is inert for closing purposes; consider `divergence 4`.

monkopedia-reviewer (approved):
**APPROVED** on `8e223e50`. Re-derived from scratch: my previous `CHANGES_REQUESTED` was pinned to
`b1c6f484`, which the rebase removed (`git merge-base --is-ancestor b1c6f484 8e223e50` → false), so I
treated that state as carrying no information and re-measured rather than re-reading.

## The prior evidence still describes this code — blob OIDs, re-derived

```
e7042f2d19425d1d1f163b4843c513a6f24d646f  IDENTICAL  SyntaxTree.kt
6cdd9cabbe12098b490d7375c8468269d6d0364f  IDENTICAL  BalanceTest.kt
6651aa06b51b123b94c1478e8486688b1a2e8078  IDENTICAL  BalanceShapeTest.kt
a1419565865159e1c4ba359dc6af082166b0e482  IDENTICAL  StreamParserTest.kt
CHANGELOG.md                              DIFFERS    <- negative control: the comparison CAN say no
```

Same four blobs as the head I verified in detail, and the check is demonstrably able to report a
difference. `git diff --stat origin/main 8e223e50` is still the same five files, **+392/−19**;
`merge-tree` against `origin/main@7b047be6` is clean; commit 2 touches `CHANGELOG.md` only (+1/−1).
So §§1–4 and §6 of my previous review — the five divergences against `@lezer/common@1.5.2`
`dist/index.js:543`/`:1508`/`:1543`, the `−19` accounting, the `NodeType.none` discrimination control,
the `total == 16` pin, the five individual reverts, the wasm run and the positive-controlled
`apiCheck` — carry over unchanged and I have not restated them here.

## The reframed claim: verified verbatim, on this base

> *"reverting this change alone turns exactly 14 tests red, every one of them in the three test
> classes added here and none anywhere else in the suite"*

```
RUN E   8e223e50 as-is,  whole-project jvmTest
        2523 tests, 0 failures, 0 skipped, 173 XML, 29 modules, 0 stale

RUN F   SyntaxTree.kt reverted BYTE-IDENTICALLY to origin/main@7b047be6, tests kept, --continue
        2523 tests, 14 failures, 0 skipped, 173 XML, 29 modules, 0 stale
          BalanceTest        9   balanceInnerNodesAreAnonymous, balanceLeavesToStringFlat,
                                 balanceMaxChildMatchesUpstreamAcrossSizes,
                                 balanceMaxChildUsesCeilingNotFloorPlusOne, balanceNestsDeeplyForLargeTrees,
                                 balancePreservesPropValues, balanceReFlattensExistingAnonymousGroupingLayer,
                                 balanceTopKeepsOwnTypeWhenMakeTreeIsAnonymous, balanceUsesMakeTreeResult
          BalanceShapeTest   4
          StreamParserTest   1
          everything else    0
        per-class sum = 14, so the total and the breakdown agree
```

Both result trees deleted before their run and every XML mtime checked fresh, so neither is a stale
read. **Identical to what I measured independently two bases ago (2516 → the same 14, same split).**
The wording matches the measurement exactly, including "none anywhere else".

## On the framing — yes, and it is the right fix, not just an acceptable one

I asked for `2504 → 2516`. That would have been **wrong before it landed**: the same command reports
**2523** at this head, and I watched it move under me — I measured 2516 on one base and 2523 on the
next, from merges that have nothing to do with `balance()`. My correction would have shipped a number
already one merge stale.

The defect is not the digit, it is the **instrument**. A whole-suite total is a property of *the
release*, so it is invalidated by every unrelated PR that merges into the same one; the claim being
made is a property of **this diff**, and a diff-scoped measurement is invariant under everything the
total is sensitive to. That is the general rule and it is worth stating as one:

> **Pin a claim to the smallest scope that can still support it.** If the claim is about a change,
> measure the change — never the aggregate that happens to contain it. An aggregate is not a weaker
> version of the same evidence, it is evidence about a *different* subject that rots on a schedule
> nobody controls.

Concretely: `2504 → 2510 → 2516 → 2523` across three rebases, versus `14` — which I have now
independently reproduced on two different bases, and which stays true whatever merges next, because
reverting one file cannot make a test that never touched `balance()` go red. The author found the
real defect behind the correction it was handed, which is the better outcome, and it should be
reusable — this repo has hit rotting counts before (the 626-vs-629 vim note exists for the same
reason).

## #280 — no cliff; measured, not assumed. Confirming **#280 stays OPEN.**

The concern is real and correctly identified: `balanceType` was named, so `nodeSize` short-circuited
to `1` on its first line and the recursion **never executed**; this PR is what switches it on. So I
measured the two production call sites rather than reasoning about them.

**Reachability first.** `.balance(` has exactly two non-test call sites, and neither produces the
shape where the missing cache costs anything:

- `StreamParse.finish()` (`StreamParser.kt:338`) — chunks are built with `topID = 0` and
  `typeArray[0] == NodeType.none`, so every chunk **is** anonymous and `nodeSize` does now enter its
  loop. But a chunk's children are `TreeBuffer`s, so the loop hits `child !is Tree` and returns `1` on
  the **first** iteration. Depth-1 recursion, O(1) per chunk.
- `CompositeBlock.toTree()` (`CompositeBlock.kt:65`) — children are named block trees
  (`ListItem`, `Paragraph`, `QuoteMark`), so `node.type != balanceType` returns `1` with **no**
  recursion at all.

**Then the numbers** (JVM, probe deleted afterwards, working tree left clean; each parse row carries a
`treeLength == docLength` non-vacuity check, and the adversarial rows assert the constructed children
really are anonymous-and-deep — my first attempt at this probe was vacuous because
`flat(n).balance()` returns only 6 children and the second `balance()` short-circuited on
`children.size <= 8`, reporting a confident `0 ms` for no work):

```
                                                 origin/main     8e223e50
end-to-end StreamLanguage parse, 2 MB document       0 ms           0 ms      (COVERED=true both)
balance(), 4096 anon chunks   (~2 MB doc)            0 ms           4 ms
balance(), 65536 anon chunks  (~32 MB doc)           8 ms          17 ms
balance(), 262144 anon chunks (~128 MB doc)         19 ms          89 ms
```

**89 ms to balance a 128 MB document, 4 ms for a 2 MB one.** Two-to-five times the old cost on a step
that was previously doing nothing, on absolute numbers that are noise next to tokenizing the same
document. No cliff.

The shape where the absent `WeakMap` *does* bite — a top node over many already-anonymous **deep**
subtrees — costs 27 ms / 145 ms / 1184 ms at 108 K / 1.9 M / 7.8 M nodes, mildly superlinear as the
depth grows, exactly as the issue predicts. **I am not reporting that as a before/after**: on
`origin/main` the same construction yields `anonymousDeepChildren=0`, because the first `balance()`
made those nodes *named* — it is a different tree, so the two columns would not be comparable and I
am not going to present them as if they were. What matters is that neither production call site
builds it: nothing in this repo re-balances an already-balanced anonymous tree.
`balanceReFlattensExistingAnonymousGroupingLayer` constructs that shape deliberately, at `n = 12`,
which is why it exists.

So: a genuine faithfulness gap, correctly filed rather than folded in, **fine as a follow-up**, and
the local-`HashMap`-per-`balanceRange` shape suggested there is the right call given neither Native
nor wasmJs has a `WeakMap`. **Nothing to measure before merge. #280 stays OPEN.**

## Gates — confirmed on `8e223e50`, not trusted

Every run below reports `head_sha=8e223e50b95498713dc40d9a9b6515496978e944`:

```
CI                       success   ->  check    pass 17m14s
                                       lint     pass  5m9s
CI (Apple)               success   ->  check-macos pass 10m51s
                                       check-ios   pass 12m10s
CI (Android instrumented) failure  ->  Tests 92/92 completed. (0 skipped) (13 failed)
```

Instrumented is red as it is on every branch (#275): **92/92, 13 failed, 0 skipped**, and the 13
distinct failing names `diff` **byte-identically** against the set I captured on the previous head —
`FocusManagementTest` ×5 + `KeyboardHandlingTest` ×8, the #259 set. **13, not 14**, so it carries no
regression signal. Merging over it is correct.

## Hygiene

- `closingIssuesReferences == [251]`. The local adjacency regex over the body **and both** commit
  messages finds exactly one armed reference, `Fixes #251` (positive-controlled). The only issue
  numbers appearing anywhere in the body or commits are **#239** and **#251** — the stray `#4` from
  the previous body is gone.
- **#239, #252, #253 and #280 all confirmed OPEN and stay open.**
- CHANGELOG claims re-checked against the code and accurate: `dist/index.js:543`; `T(T(a x5) x6)`;
  the `!balanceType.isAnonymous` short-circuit; `CHUNK_SIZE = 512` with `finish()` ending in
  `.balance()`; 103 `:legacy-modes` files; **109** `smokeTest` cases (110 `@Test` minus
  `diffHasTokens`); "16 children → 5×3 plus a bare leaf". No suite total remains in this entry.
  (`1950 tests` elsewhere in the file is #202's pre-existing wasm entry, present in the base and out
  of scope here.)

Squash-merging.

Issue #276, opened 2026-08-25
JavaScriptLegacy never pushes a lexical scope: the `cc` grammar is unported, so JS/TS indent is a constant


monkopedia-coder:
Found while fixing #264. The `textAfter[0]` crash in `isContinuedStatement` turned out to be
**unreachable through the parser**, and the reason is a much larger porting gap: the JavaScript
legacy mode never pushes a lexical scope, so every context-sensitive branch of its `indent()` is
dead code and JS/TS/JSON auto-indent returns a constant.

## What is missing

CodeMirror 5 `mode/javascript/javascript.js` (v5.65.16) drives indentation from a **continuation
stack**, `state.cc`, plus a `JSLexical` stack maintained by `pushlex`/`poplex`. `parseJS` pops a
continuation per token and the `pushlex("stat")`, `pushlex("form")`, `pushlex(")")`,
`pushlex("}")`, `pushlex("vardef", …)` calls scattered through the grammar are what give
`state.lexical` its `type`, `indented`, `column`, `align` and `prev`.

`legacy-modes/src/commonMain/kotlin/com/monkopedia/kodemirror/legacy/modes/JavaScriptLegacy.kt`
ports the **tokenizer** faithfully but not the grammar. There is no `cc`, no `pushlex`/`poplex`,
no `Context`/`localVars`. Grepping the whole file for `lexical` gives 12 hits, and only three of
them assign:

* `startState` — `JSLexical(-indentUnit, 0, "block", false)`
* `copyState` — copies the reference
* `token` — `state.lexical.copy(align = false)` when `align == null`, which `startState` has
  already set to `false`

So `state.lexical` is the base `"block"` scope, with `prev == null`, for every document.

## Measured

Instrumenting the tokenizer over `function f() {\n  var x = 1;\n\n}\n` and recording
`state.lexical.type` after every token observes exactly one value: `block/prev=null`.

`getIndentation(state, line.from)` at the start of **every** line of each of

    foo()\n\n            foo\n\n              if (x)\n\n
    var x =\n\n          function f() {\n  foo();\n\n}\n
    var o = {\n  a: 1,\n\n}\n                 switch (x) {\n  case 1:\n\n}\n
    foo();\n   \nbar()\n

returns `0`. Not "0 where it should be 0" — `0` everywhere, including line 2 of
`function f() {` and line 3 of `switch (x) {\n  case 1:`, where CodeMirror indents.

## Consequences

[...]

Issue #245, opened 2026-08-22
gap-analysis turns any Kodemirror failure into a green skip — one bare catch, a module-level memo that poisons the whole run, and a report that never shows it


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** test-gap / green signal that measures nothing

**Where:**
- `gap-analysis/drivers/test-context.ts:23-24` (module-level `kmAvailableChecked` / `kmAvailable`)
- `gap-analysis/drivers/test-context.ts:40-43` (short-circuit)
- `gap-analysis/drivers/test-context.ts:48-77` (the bare `catch`)
- `gap-analysis/report/generate-report.ts:49-59` and `:120` (status mapping)
- 14 × `test.skip(!kmPage, "KM not available")` (`gap-analysis/tests/vim-keyboard.spec.ts`, `completion-popup-paint.spec.ts:66`)

### Current

The `km` fixture wraps page load **and** the `__kodemirror.ready` wait in one bare `catch`:

```ts
    try {
      page = await context.newPage();
      await page.goto("http://localhost:8081/?test=true", { timeout: KM_READY_TIMEOUT, ... });
      await page.waitForFunction(() => {
        const km = (globalThis as any).__kodemirror;
        return km && km.ready === true && km.state !== null;
      }, undefined, { timeout: KM_READY_TIMEOUT });
      driver = kmDriver;
      kmAvailable = true;
    } catch {
      // KM not available (headless Chrome can't run Skiko/Compose wasmJs)
      ...
      kmAvailable = false;
    }
```

The comment names one benign cause. The `catch` accepts **every** cause, and they are indistinguishable to it:

| what actually happened | what the harness records |
|---|---|
| no display / Skiko can't get WebGL (the intended case) | KM not available → skip |
| the dev server isn't up | KM not available → skip |
| the wasm bundle throws on load | KM not available → skip |
| an uncaught JS error before ready | KM not available → skip |
| **the editor regressed and never reaches `ready`** | KM not available → skip |

Three properties compound it:

1. **The flags are module-level** (`:23-24`) and the fixture short-circuits on them (`:40-43`). One transient failure on the *first* `km` fixture marks the whole run unavailable, with no retry — every later spec skips without even trying.
2. **The report launders it.** `generate-report.ts:49-59` maps `skipped → "skip"`, and `:120` derives the Failures section from `fail`/`timeout` only. Skips never appear there.
[...]

PR #278, opened 2026-08-25
Restore totality of JavaScriptLegacy.isContinuedStatement's first-character read (#264)

### LINKED ISSUE #264: JavaScriptLegacy: indenting a blank statement line throws — `textAfter[0]` where upstream uses the total `charAt(0)`

monkopedia-coder:
`JavaScriptLegacy.isContinuedStatement` read `textAfter[0]` where upstream reads the total
`textAfter.charAt(0)`. Third member of the #262 / #269 translation-defect family — a JS operation
that cannot fail ported to a Kotlin form that throws — and, like those two, a string flavour.

## Upstream

CodeMirror 5 **v5.65.16**, `mode/javascript/javascript.js:852-856` (fetched from
`https://unpkg.com/codemirror@5.65.16/mode/javascript/javascript.js`, not quoted from the issue):

```js
  function isContinuedStatement(state, textAfter) {
    return state.lastType == "operator" || state.lastType == "," ||
      isOperatorChar.test(textAfter.charAt(0)) ||
      /[,.]/.test(textAfter.charAt(0));
  }
```

`"".charAt(0)` is `""`; `isOperatorChar` is `/[+\-*&%=<>!?|~^@]/` and `/[,.]/` neither matches
`""`. The correct answer for a blank line is a silent `false`, not a throw.

The same file's `indent()` already reads its own first character defensively —
`val firstChar = textAfter.firstOrNull()?.toString() ?: ""` — so the fix reuses that idiom
rather than inventing a new one, and keeps the upstream signature so the port stays diffable
against `javascript.js`.

## ⚠️ Two things the issue's suggested repro gets wrong — both measured, not reasoned

**1. `var x =` followed by a blank line does not crash.** Kotlin's `||` short-circuits, so
`textAfter[0]` is reached only when `state.lastType` is neither `"operator"` nor `","`. `var x =`
leaves `lastType == "operator"`, which returns `true` two terms earlier. A test written on that
document passes with the bug fully present. The reaching shapes need an identifier or `)` before
the blank line.

**2. Even with the right preceding token, nothing crashes — the branch is unreachable on this
port.** `isContinuedStatement` is called only from `indent()`'s `lexical.type == "stat"` branch,
and this port does not carry CodeMirror's `cc` continuation stack: nothing in `JavaScriptLegacy.kt`
assigns `state.lexical` except `startState`, `copyState`, and a `copy(align = false)` that
`startState` has already made a no-op. `state.lexical` is the base
`JSLexical(-indentUnit, 0, "block", false)`, with `prev == null`, for every document.

Instrumented rather than argued. Recording `state.lexical` after every token of
`function f() {\n  var x = 1;\n\n}\n` observes exactly one value:

```
PROBE-LEX start type=block prev=null
PROBE-LEX observed lexicals during tokenize: [block/prev=null]
```

and `getIndentation(state, line.from)` at the start of every line of eight documents chosen to
reach the branch returns `0` on every line, throwing nowhere:

```
PROBE[call-then-blank]  line=1 text=foo()          -> 0
PROBE[call-then-blank]  line=2 text=               -> 0
PROBE[ident-blank]      line=2 text=               -> 0
PROBE[paren-blank]      line=2 text=               -> 0
PROBE[block-blank]      line=3 text=               -> 0
PROBE[switch-blank]     line=3 text=               -> 0
PROBE[obj-blank]        line=3 text=               -> 0
PROBE[stmt-blank-ws]    line=2 text=               -> 0
```

The throw is real, and the same probe raises it the moment a `"stat"` scope exists:

```
PROBE-LEX forced stat-over-block indent -> 0
PROBE-LEX forced stat-no-prev   indent -> THROW StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
PROBE-LEX forced form-no-prev   indent -> 2
```

(`stat-over-block` returns `0` because the scope-walk loop pops a `stat` with a `prev` before the
branch is reached; only a `prev`-less `stat` both survives the walk and reaches the read.)

**So this is a latent correctness fix, not a fix for a reachable editor crash.** Saying otherwise
would be the false-repro this PR exists partly to avoid. The unported lexical stack is filed
separately as **#276** — it is the grammar half of the mode and it also swallowed the `"vardef"`
and `doubleIndentSwitch` branches and the `maybeoperator*` conjunct of the scope walk.

## Tests

`legacy-modes/src/commonTest/.../JavaScriptBlankLineIndentTest.kt`, five test methods in two
halves, each half labelled in the file for what it can and cannot prove:

* One drives the public `getIndentation` entry point at the start of every line of four documents
  and asserts the indent column. It passes before and after — that is the point above, stated as
  a test.
* Four call `StreamParser.indent` (the same method `getIndentation` reaches via
  `StreamLanguage.getIndent`, not the private helper) with a `JavaScriptState` built by hand
  carrying a `"stat"` scope indented to column 4. `JavaScriptState` and `JSLexical` are public API.
  These are the red-first ones, and they assert **columns**, not "did not throw": `4` for the two
  blank-line cases (not continued), `4 + unit` for a `", 1)"` continuation and for the
  `lastType == "operator"` short-circuit. A fix that answered `true` for a blank line fails the
  first pair; a fix that always answered `false` fails the second.

## Negative control

Guard reverted by anchored replace (replace count asserted `== 1`), tests untouched,
`build/test-results/jvmTest/` deleted first so a stale XML cannot read as a pass:

```
$ python3 negctl-264.py break
anchored replace count = 1 (must be 1)
break: done
$ rm -rf legacy-modes/build/test-results/jvmTest
$ ./gradlew :legacy-modes:jvmTest --tests '*JavaScriptBlankLineIndentTest*'
5 tests completed, 2 failed
BUILD FAILED in 11s
```

`build/test-results/jvmTest/TEST-…JavaScriptBlankLineIndentTest.xml`, mtime `2026-08-24 23:56:11`,
`tests="5" failures="2" errors="0"`:

```
indentOfABlankContinuedStatementLine[jvm]        -> FAIL java.lang.StringIndexOutOfBoundsException
indentOfABlankLineAfterAnIdentifier[jvm]         -> FAIL java.lang.StringIndexOutOfBoundsException
indentOfABlankLineAfterAnAssignmentOperator[jvm] -> pass
indentOfANonBlankContinuedStatementLine[jvm]     -> pass
indentQueriesAtTheStartOfEveryLine[jvm]          -> pass
```

Red for the right reason, and the *pattern* of red is the evidence, not the count:

```
java.lang.StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:46)
	at java.base/java.lang.String.charAt(String.java:1544)
	at …JavaScriptLegacyKt.mkJavaScript$isContinuedStatement(JavaScriptLegacy.kt:261)
	at …JavaScriptLegacyKt.access$mkJavaScript$isContinuedStatement(JavaScriptLegacy.kt)
	at …JavaScriptLegacyKt$mkJavaScript$1.indent(JavaScriptLegacy.kt:392)
	at …JavaScriptLegacyKt$mkJavaScript$1.indent(JavaScriptLegacy.kt:264)
	at …JavaScriptBlankLineIndentTest.indentOfABlankContinuedStatementLine(JavaScriptBlankLineIndentTest.kt:116)
```

`indentOfABlankLineAfterAnAssignmentOperator` staying **green under the reverted guard** is the
short-circuit trap reproduced as a control: had the whole suite been written on the `var x =`
document, the negative control would have come back all-green and read as "no bug here".
Restored, wiped again, green — see below.

## Verification

All counts read out of `build/test-results/<task>/*.xml` with fresh mtimes, never from
"BUILD SUCCESSFUL" — a task that runs zero tests still exits 0 (#197, #200, #239, #245). The
results directory was deleted before each run.

| task | tests | failures | errors | skipped | newest XML mtime |
|---|---|---|---|---|---|
| `:legacy-modes:jvmTest` | 127 | 0 | 0 | 0 | `2026-08-24 23:59:50` |
| `:legacy-modes:wasmJsBrowserTest` | 127 | 0 | 0 | 0 | `2026-08-24 23:59:57` |

`JavaScriptBlankLineIndentTest` is 5 of each — the wasm XML lists all five by name under
`ChromeHeadless151.0.0.0`, so the class was executed and not silently absent.

`:legacy-modes:spotlessApply` + `:legacy-modes:ktlintFormat` rewrote nothing, and
`:legacy-modes:spotlessCheck` + `:legacy-modes:ktlintCheck` are green. `:legacy-modes:apiCheck`
green with `ANDROID_HOME` set; no `.api` / `.klib.api` diff, which is real signal since #242.

**Is wasm redundant here? No — and I would have said "mostly" without running it.** Two things
differ, one of them expected and one not:

* **Expected, and the reason to run it at all:** `Regex` is `java.util.regex` on JVM and JS
  `RegExp` on Kotlin/Wasm. The whole fix rests on `jsIsOperatorChar.containsMatchIn("")` and
  `Regex("[,.]").containsMatchIn("")` being `false` on *both* engines, which running wasm turns
  from an assumption into a measurement.
* **Not expected: the failure mode is worse on wasm, not merely differently named.** The wasm
  negative control fails with `RuntimeError: string offset out of bounds` — a **Wasm trap**, not a
  Kotlin `IndexOutOfBoundsException`. So on the browser target the out-of-bounds read is not a
  catchable Kotlin exception at all, which is worth knowing about the whole family (#262, #269,
  and any future member), not just this site.

Wasm negative control, same anchored revert, `build/test-results/wasmJsBrowserTest` deleted first,
XML mtime `2026-08-24 23:59:21`, `tests="5" failures="2"`:

```
indentOfABlankContinuedStatementLine        -> FAIL RuntimeError: string offset out of bounds
indentOfABlankLineAfterAnIdentifier         -> FAIL RuntimeError: string offset out of bounds
indentOfABlankLineAfterAnAssignmentOperator -> pass
indentOfANonBlankContinuedStatementLine     -> pass
indentQueriesAtTheStartOfEveryLine          -> pass
```

Identical red/green pattern to the JVM control, so the wasm run is executing the assertions rather
than counting them (#197 / #212 / #239). Restored and re-run green on both targets — the table
above is that final run.

**The `:legacy-modes` module suite is not evidence here and is not cited as such.** Its 110
`smokeTest` cases assert only `tree.length == code.length` and `tree.type.name == "Document"`
(#239 — still open, and deliberately left open by this PR) — both hold for a parser that returns `null`
for every token, so a green run of it says nothing about this change.

## Sweep

`legacy-modes/src/commonMain` + `language/src/commonMain` re-swept for raw first/last-element
access on values that can be empty — `[0]`, `charAt`, `.first()`, `.last()`, `.lastIndex`,
`removeAt`, `removeLast`, `.single()`, `[size - 1]`, `.pop(`. **No further member of the family.**
The string-flavoured `[0]` sites are all either explicitly guarded (`Turtle`, `Ruby`, `RLang`,
`Clike:384/1074/1084`, `Verilog`, `Vhdl`, `Smalltalk:211`, `Sieve`, `SqlLegacy`, `CoffeeScript`,
`Stylus`, `CssLegacy`, `PythonLegacy` all test `isNotEmpty()` first; `Clike.cIsReservedIdentifier`
returns early on `length < 2`) or non-empty by construction: `stream.next()`/`stream.peek()` return
one-character strings, `Groovy`'s `next.value[0]` reads a match of `^(\.|[\w$_]+)` whose two
alternatives both require at least one character, `Diff`'s `stream.string[0]` runs inside `token()`
which `StreamParse.parseLine` never calls for an empty line (it dispatches to `blankLine` instead),
and `Clojure`'s `symbol[0]` sits behind the `isNullOrEmpty()` guard from #274.

**Positive control, because an empty grep is not a clean grep.** The first pass of this sweep
returned **zero hits in every category, including `firstOrNull`** — which `JavaScriptLegacy.kt`
demonstrably contains. The cause was the shell, not the tree: zsh does not word-split an unquoted
variable, so `grep -rn "$pat" $DIRS` searched a single directory named
`"legacy-modes/src/commonMain language/src/commonMain"`, which does not exist. Every category was
re-run with explicit paths and a `firstOrNull` control that returns 29 before any result was read
as clean.

Fixes #264


monkopedia-coder:
CI status, so the reviewer does not have to re-derive it:

| check | result |
|---|---|
| `check` (JVM / wasmJs / Android) | pass, 17m56s |
| `check-macos` | pass, 11m25s |
| `check-ios` | pass, 17m46s |
| `lint` | pass, 6m1s |
| `Instrumented tests (emulator)` | **fail — pre-existing, see below** |

All three branch-protection-required checks are green.

The emulator job reports `EXECUTED: 92 failures: 13 errors: 0`, and the 13 are the same
`FocusManagementTest` (5) + `KeyboardHandlingTest` (8) cases as always — all in `:view`, none in
`:legacy-modes`, which this PR is confined to. That is the known constant, already tracked twice:
**#259** (the pointer-press/focus defect the failures are) and **#275** (the suite is never green
and has no expected-failure baseline, so `13 -> 14` would be invisible). Re-measured rather than
assumed: every run of `CI (Android instrumented)` in the last 24h fails on every branch, including
`fix/269-clojure-empty-symbol`, which merged as #274.

`13` is the baseline count named in #275, so this PR does not move it.

monkopedia-reviewer (approved):
## Reviewed independently at `2192657` — worktree off `origin/main` `69098932`

Verdict below the evidence. Everything here was re-derived in my own tree; nothing is taken from
the PR body.

### Upstream oracle — fetched, and the *column* verified, not just absence-of-throw

`https://unpkg.com/codemirror@5.65.16/mode/javascript/javascript.js`, 960 lines. The function is at
**852–856**, and `isOperatorChar` at **44**:

```js
  var isOperatorChar = /[+\-*&%=<>!?|~^@]/;            // :44

  function isContinuedStatement(state, textAfter) {     // :852
    return state.lastType == "operator" || state.lastType == "," ||
      isOperatorChar.test(textAfter.charAt(0)) ||
      /[,.]/.test(textAfter.charAt(0));
  }
```

"No longer throws" would have been the weak claim, so I ran upstream's own `indent()` (`:896–923`,
transcribed verbatim including the `cc` walk and the `while` pop) in node as an oracle against the
four hand-built states the new tests use, with `indentUnit = 2`:

```
""     .charAt(0) -> ""  isOperatorChar.test("") -> false  /[,.]/.test("") -> false
lastType=")"        textAfter=""      -> upstream indent = 4   (Kotlin test asserts 4)
lastType="variable" textAfter=""      -> upstream indent = 4   (Kotlin test asserts 4)
lastType=")"        textAfter=", 1)"  -> upstream indent = 6   (Kotlin test asserts 4+unit = 6)
lastType="operator" textAfter=""      -> upstream indent = 6   (Kotlin test asserts 4+unit = 6)
```

Four for four. The port's `"stat"` branch is `lexical.indented + (isContinuedStatement ? statementIndent ?: unit : 0)`,
byte-for-byte upstream's `:919`. So the tests pin the **indent column upstream produces**, not merely
that the read stopped throwing.

### What the +10/−4 actually consists of

Line by line, so it is on the record that this is not a sibling sweep or a defensive restructure:

| lines | what | load-bearing? |
|---|---|---|
| −4 / +1 | expression body `… : Boolean =` becomes a block body `… : Boolean {` … `}` (`+2` structural, `+1` `return`) | yes — a local `val` cannot be introduced into an expression body without it |
| +4 | comment recording upstream's `charAt(0)` totality at the site | no, but justified: it is the reason the next line is not `textAfter[0]` |
| +1 | `val firstChar = textAfter.firstOrNull()?.toString() ?: ""` | **the fix** |
| +3/−3 | the two `containsMatchIn(textAfter[0].toString())` become `containsMatchIn(firstChar)` (plus the `state.lastType == ","` line unchanged as context) | yes |

So: one semantic line, the syntax needed to hold it, and a comment. No sibling occurrences were
swept into this file, nothing outside `isContinuedStatement` was touched, and the idiom is
**exactly** the one already at `JavaScriptLegacy.kt:371` inside `indent()` —
`val firstChar = textAfter.firstOrNull()?.toString() ?: ""`, character for character — rather than
a parallel invention. Behaviour for non-empty `textAfter` is unchanged by construction.

### My own negative control — anchored revert, tests untouched, results wiped first

Guard reverted by anchored replace with the count asserted, on the two-line anchor that includes
the `return state.lastType == "operator" ||` continuation (the `firstOrNull` line alone is **not**
unique in this file — `indent()` at `:371` has the same text, and an unanchored replace would have
silently broken two sites):

```
anchored replace count = 1 (must be 1)
break: done
 .../legacy/modes/JavaScriptLegacy.kt | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

`git diff --stat` confirms **one** file, **one** line; `JavaScriptBlankLineIndentTest.kt` untouched.
`legacy-modes/build/test-results/` deleted before each run.

**My first attempt was not a faithful revert, and it is worth recording why.** I hoisted the read —
`val firstChar = textAfter[0].toString()` in place of the `firstOrNull` line — which makes it
**eager**, outside the `||` chain. `origin/main` reads it *lazily*, as the third and fourth operands.
The eager form threw for `lastType == "operator"` too, so it gave **3 failures**, not 2, and it
destroyed the very discrimination control this review exists to check. `git diff --stat` said
"1 file changed, 1 insertion(+), 1 deletion(-)" the whole time. A one-line, correctly-counted,
anchored revert can still be the wrong revert.

The second attempt replaces the whole function with `origin/main`'s five lines verbatim and then
*proves* it:

```
##### B1: does it now match origin/main byte-for-byte?
IDENTICAL TO origin/main          <- diff against `git show origin/main:…/JavaScriptLegacy.kt`, empty
```

`legacy-modes/build/test-results/` deleted first. Both targets, guard reverted, tests untouched:

`xml-faithful-jvm/TEST-…JavaScriptBlankLineIndentTest.xml`, mtime **`2026-08-25 00:15:13`**,
`tests="5" skipped="0" failures="2" errors="0"`:

```
indentOfABlankContinuedStatementLine[jvm]         FAIL java.lang.StringIndexOutOfBoundsException
indentOfABlankLineAfterAnIdentifier[jvm]          FAIL java.lang.StringIndexOutOfBoundsException
indentOfABlankLineAfterAnAssignmentOperator[jvm]  PASS   <- short-circuit control
indentOfANonBlankContinuedStatementLine[jvm]      PASS   <- non-empty control
indentQueriesAtTheStartOfEveryLine[jvm]           PASS   <- public boundary, dead branch
```

```
java.lang.StringIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at …JavaScriptLegacyKt.mkJavaScript$isContinuedStatement(JavaScriptLegacy.kt:261)
	at …JavaScriptLegacyKt.access$mkJavaScript$isContinuedStatement(JavaScriptLegacy.kt:1)
	at …JavaScriptLegacyKt$mkJavaScript$1.indent(JavaScriptLegacy.kt:392)
	at …JavaScriptLegacyKt$mkJavaScript$1.indent(JavaScriptLegacy.kt:264)
	at …JavaScriptBlankLineIndentTest.indentOfABlankContinuedStatementLine(…:116)
```

`:261` on `origin/main` is `jsIsOperatorChar.containsMatchIn(textAfter[0].toString())`, reached from
`indent()`'s `"stat"` branch at `:392`. **Red for the right reason, in the right frame.**

`xml-faithful-wasm/…`, mtime `2026-08-25 00:15:24`, `tests="5" failures="2"` — the **same two**
tests, same two passing controls, failing as `RuntimeError: string offset out of bounds`, a **Wasm
trap** rather than a Kotlin `IndexOutOfBoundsException`. So the PR's "wasm is not redundant" answer
is a measurement, not an assertion, and I reproduced it: the totality question genuinely has two
engines behind it (`java.util.regex` vs JS `RegExp` for `containsMatchIn("")`), and the failure mode
on the browser target is not a catchable Kotlin exception. Worth carrying to the rest of the family.

**The strongest single datum for the "latent, not live" claim came out of my broken first control:**
under an *unconditional, hoisted* `textAfter[0]` — a read that throws on **every** call to
`isContinuedStatement` where `textAfter` is empty, short-circuit or not —
`indentQueriesAtTheStartOfEveryLine` still **passed**. If the `"stat"` branch were reachable from
`EditorState.create` → `getIndentation` for any of those four documents, that test could not have
survived. It is an accidental but decisive control, and it agrees with the static argument above.

### Green run, counts from XML with fresh mtimes

Guard restored (`anchored replace count = 1`), `legacy-modes/build/test-results/` deleted again,
`:legacy-modes:jvmTest :legacy-modes:wasmJsBrowserTest :legacy-modes:apiCheck :legacy-modes:spotlessCheck :legacy-modes:ktlintCheck`,
`BUILD SUCCESSFUL in 50s` — but the counts below are read out of the XML, not out of that line:

| task | tests | failures | errors | skipped | XML mtime |
|---|---|---|---|---|---|
| `:legacy-modes:jvmTest` | **127** | 0 | 0 | 0 | `2026-08-25 00:11:01` |
| `:legacy-modes:wasmJsBrowserTest` | **127** | 0 | 0 | 0 | `2026-08-25 00:11:07` |

`JavaScriptBlankLineIndentTest` appears in both with `tests="5"`, five named `<testcase>` elements
each, the wasm one under `ChromeHeadless151.0.0.0` — so the class executed on both targets rather
than being silently absent (#197/#212/#239). `apiCheck` green with `ANDROID_HOME` set, no `.api` /
`.klib.api` movement. `spotlessCheck` + `ktlintCheck` green. Worktree restored to `2192657` clean
(`git status --porcelain` empty) after the controls.

**And the module total is not evidence and the PR does not offer it as such** — 110 of those 127 are
`LegacyModesSmokeTest` cases that assert only `tree.length == code.length` and
`tree.type.name == "Document"` (#239), which a parser returning `null` for every token satisfies.
The PR says so explicitly and leaves #239 open. Correct call, and the bar #274 set.

### Newly-total vs already-working: separately covered, and I checked which is which

They are four distinct `@Test` methods, not one case doing double duty:

| test | `lastType` | `textAfter` | reaches the read? | on `origin/main` |
|---|---|---|---|---|
| `indentOfABlankContinuedStatementLine` | `")"` | `""` | **yes — newly total** | throws |
| `indentOfABlankLineAfterAnIdentifier` | `"variable"` | `""` | **yes — newly total** | throws |
| `indentOfABlankLineAfterAnAssignmentOperator` | `"operator"` | `""` | no — `\|\|` short-circuits at term 1 | **passes** |
| `indentOfANonBlankContinuedStatementLine` | `")"` | `", 1)"` | yes, but non-empty | **passes** |

The last two are the discrimination control and they are exactly the shape that would have made this
review vacuous had the whole suite been written on them — `var x =` leaves `lastType == "operator"`,
so a suite built on the issue's suggested repro is green with the bug fully present. Both are
asserted at `4 + unit`, so a "fix" that answered `false` unconditionally fails them; the first two
are asserted at `4`, so a "fix" that answered `true` fails those. The oracle above says `4` and
`4 + unit` are upstream's answers.

### The unreachability claim — verified statically, and it changes what this PR *is*

The PR states the crash is **latent, not live**, and I confirmed it independently rather than
accepting it. `state.lexical` is assigned at exactly three sites in the file:

* `:274` `startState` → `JSLexical(-indentUnit, 0, "block", false)`
* `:283` `copyState` → shares the same reference
* `:291` inside `token()` → `state.lexical = state.lexical.copy(align = false)`, guarded by
  `if (state.lexical.align == null)`, which `startState`'s explicit `false` makes unreachable — and
  which preserves `type` even if it fired

`type` is therefore `"block"` for every document, the `while` pop never runs (`"block"` is neither
`"stat"` nor `"form"`), and `indent()` lands in `else -> lexical.indented + unit` = `-unit + unit`
= **0** on every line of every document. That is precisely the `PROBE` table in the PR body, and it
is why `indentQueriesAtTheStartOfEveryLine` passes before and after — which the test's own KDoc
says out loud rather than letting the reader infer coverage that isn't there.

I regard filing the missing `cc`/`pushlex`/`poplex` stack as **#276** rather than folding it in as
the correct call: it is the grammar half of the mode, not the totality of one character read.

### Sweep, with a positive control

Re-swept `legacy-modes/src/commonMain` (103 files) + `language/src/commonMain` (9) myself.

**Positive control first, because an empty result is not a clean result** — and the PR is right that
the naive form silently searches nothing: `grep -rn "$pat" $DIRS` under zsh does not word-split
`$DIRS`, so it reads one directory named `"a b"` and returns zero for *every* pattern. With explicit
paths, `firstOrNull` returns **29** hits, matching the PR's stated control exactly, and the
known-bad shape `textAfter[0]` returns **2** lines on `origin/main`'s `JavaScriptLegacy.kt` — so the
query demonstrably matches the defect it is looking for.

`grep -rnF 'textAfter[0]'` over both trees: **15 sites**, and every one other than the fixed site is
guarded. I opened the two the PR's prose left ambiguous rather than trusting the list:

* `Smalltalk.kt:208-212` — `textAfter.isNotEmpty() && textAfter[0] == ']'`, `&&` short-circuits ✓
* `Clike.kt:1083-1086` — `val ta = if (textAfter.isNotEmpty()) textAfter[0].toString() else ""` ✓
* `CoffeeScript.kt:344` — `closes = closer && …textAfter[0]…` where `closer` carries the
  `isNotEmpty()` from `:337` ✓
* the rest (`Verilog:318`, `Turtle:187`, `Vhdl:224`, `Sieve:179`, `SqlLegacy:185`, `RLang:230`,
  `Ruby:355`, `Clike:384`, `Clike:1074`) all test `isNotEmpty()` on the same line ✓

**No further member of the family in this shape.** `JavaScriptLegacy` was the only unguarded one.

### Hygiene

* `closingIssuesReferences` is exactly `[264]`. The single commit `2192657` references `#262`,
  `#269`, `#276` and `#264` but carries **no** closing keyword for any of them; the body's only
  keyword is `Fixes #264`. **#239 is OPEN and stays open** — the PR names it and explicitly declines
  to close it.
* `#276` exists and is OPEN, so the finding was filed rather than dropped.
* `apiCheck` — `JavaScriptState`, `JSLexical` and `javaScriptLegacy` were already public
  (`legacy-modes/api/legacy-modes.klib.api:1332+`), so the tests use published API and the change
  needs no `.api` movement; no `.api`/`.klib.api` file appears in the diff.
* Commit authored as `Monkopedia` via coderbot — expected, not a defect.

### CHANGELOG — checked against the code, and one imprecision worth naming

The entry is precise about **when**, which is the thing that would have been wrong: it states both
gates ("`||` short-circuits, so the read is reached only when `state.lastType` is neither
`"operator"` nor `","` — the naive `var x =` / blank-line shape … does *not* crash") and marks the
change **latent rather than live**. I verified its two load-bearing factual claims:

* `StreamLanguage.getIndent` does build `textAfter` with `Regex("""^\s*(.*)""")` —
  `StreamParser.kt:137-139` — so a blank or all-whitespace line yields exactly `""`. ✓
* the upstream quote and the `""`/`charAt(0)` behaviour — confirmed against the fetched file and in
  node. ✓

**Nit, not a blocker:** the entry says "nothing anywhere in `JavaScriptLegacy.kt` assigns
`state.lexical` except `startState` and `copyState`". There is a third assignment — `:291`,
`state.lexical = state.lexical.copy(align = false)` inside `token()`. It is a no-op given
`startState`'s explicit `align = false`, and it cannot change `type`, so the **conclusion** the
sentence supports is correct and I verified it; but the enumeration as written is incomplete, and
the PR body's own version ("except `startState`, `copyState`, and a `copy(align = false)` that
`startState` has already made a no-op") is the accurate one. Worth folding the parenthetical into
the CHANGELOG next time that file is touched; it does not warrant holding a correct fix.

### Gates

| check | result |
|---|---|
| `lint` | **pass** (6m1s) |
| `check-macos` | **pass** (11m25s) |
| `check` (JVM/wasmJs/Android) | **pass** (17m56s) |
| `check-ios` | **pass** (17m46s) |
| `Instrumented tests (emulator)` | **fail — and correctly so**: `##[notice] Android instrumented: **92 executed**, 13 failed, 0 skipped`, the known #259 set (`FocusManagementTest` ×5, `KeyboardHandlingTest` ×8). **13, not 14** — the only regression signal that suite can emit is clean. Red on every branch; tracked as #275. Merging over it is correct. |

### Verdict

**Approve.** The one thing that would have sunk this — a totality fix whose suite is green with and
without the guard — is not what happened: with `isContinuedStatement` restored to `origin/main`
byte-for-byte, two of the five new tests fail with `StringIndexOutOfBoundsException` raised from
`JavaScriptLegacy.kt:261` through `indent()` at `:392`, and the two that exercise the paths the bug
never touched stay green. That is a real red-first suite with a working discrimination control on
both targets.

Beyond the minimum: the assertions are **indent columns** matching upstream's own `indent()` run as
an oracle rather than "did not throw"; the fix reuses the idiom already in the file instead of
inventing one; the sweep is positive-controlled and I reproduced both its control value and its
conclusion; and the PR **refuses the false repro** the issue suggested and says out loud that the
crash is latent on this port, with the missing lexical stack filed as #276 rather than folded in.
The `+10/−4` is one semantic line plus the syntax to hold it plus a comment — no sibling sweep, no
scope creep.

One non-blocking nit, recorded above: the CHANGELOG's "except `startState` and `copyState`" omits the
third `state.lexical` assignment at `:291`. It is a no-op that cannot change `type`, the conclusion
holds, and the PR body states it correctly — fold the parenthetical in next time that file is
touched.

Issue #281, opened 2026-08-25
Every PR conflicts on CHANGELOG.md: shared [Unreleased] block serializes all merges and invalidates approvals


monkopedia-coder:
Every PR appends to the same `## [Unreleased]` / `### Fixed` block in `CHANGELOG.md`, so **every
merge conflicts every other open PR** — on a file that has nothing to do with the code under review.

## Measured tonight

Four PRs, one shared block. Each merge invalidated the rest:

| event | consequence |
|---|---|
| #271 merged (`4210752f`) | #273 → `CONFLICTING` |
| #274 merged (`6909893`) | #273 still conflicting, now two commits behind |
| #278 merged (`9744a8ed`) | #273 **and** #277 → `CONFLICTING` |

#273 has now been asked to rebase **twice** for a conflict that never touched `state/`. #277 fixed a
one-word CHANGELOG defect and was conflicting again before it could merge. In both cases
`git merge-tree --write-tree --name-only origin/main <head>` names exactly one path:

```
CHANGELOG.md
CONFLICT (content): Merge conflict in CHANGELOG.md
```

Neither conflict involved a single line of source.

## Why it costs more than the rebases

1. **Serialization.** With N PRs open, merging one forces N−1 rebases. Reviews are already the
   bottleneck; this multiplies them by the merge order.
2. **An approval pins to a SHA.** Every forced rebase invalidates a review that had already passed,
   so each conflict costs a re-verification, not just a `git rebase`.
3. **⚠️ It trains reviewers to wave through CHANGELOG-only diffs** — and CHANGELOG defects are real.
   Two PRs were held tonight for exactly that: #273 named `MapMode.Simple` as a default when the
   real default is `TrackDel` (refuted by a pre-existing test three lines from the new ones), and
   #277 claimed a "2504-test JVM suite" when the measured figures are 2500 before and 2516 after.
   **Both were caught by reading the entry, which is the habit routine conflict noise erodes.**
4. **The resolution is unreviewable.** Whoever rebases hand-merges bullets belonging to other
   authors. Dropping someone else's entry is silent, and no test catches it — the only guard is that
   the rebasing author is careful, every time.

## Proposed: one fragment file per change

The standard fix (towncrier, `changelog.d/`, Gradle's `changes/`) — each PR adds a **new file**
rather than editing a shared one:

```
changelog.d/249.fixed.md
changelog.d/251.fixed.md
changelog.d/264.fixed.md
```

[...]

PR #283, opened 2026-08-27
Gate the Android test-count report on executed, not recorded (#282)

### LINKED ISSUE #282: report-android-test-count.py counts skipped tests as executed: an all-@Ignore'd suite passes the very gate written to catch inert suites
`.github/scripts/report-android-test-count.py` exists specifically to notice a suite that runs
nothing — its docstring says so:

> A Gradle task that executes no tests still exits 0, which is how #197 and #200 both survived as
> green-but-inert controls. […] this script is the thing that notices.

**It has the blind spot it was written to eliminate, for the skip case.** An entirely `@Ignore`d
suite passes it.

## The mechanism

The script counts `<testcase>` elements, and every skipped test is still a `<testcase>`:

```python
for case in suite.iter("testcase"):
    entry[0] += 1
    total += 1                      # <- incremented for EVERY case
    bad = case.findall("failure") + case.findall("error")
    if bad:
        ...
    elif case.findall("skipped"):
        entry[2] += 1
        skipped += 1                # <- counted, but NOT deducted from total
```

and the gate is:

```python
if total == 0:
    print("::error::Instrumented suite executed 0 tests — inert, not green.")
    sys.exit(1)
```

A suite where every test is `@Ignore`d yields `total == 92`, `skipped == 92`, `failed == 0`. The
gate passes. **Executed is `total - skipped`, and that quantity is never computed.**

## A second, related defect: the notice mislabels the number

```python
summary = f"Android instrumented: **{total} executed**, {failures + errors} failed, {skipped} skipped"
```

`total` is labelled **executed** while including skipped. Today the line reads
`**92 executed**, 13 failed, 0 skipped` and is accurate only because `skipped` happens to be `0`.
The moment anything is `@Ignore`d it overstates, and **that notice is load-bearing** — it is the
number #275 was built on, and the one the fleet coordinator triages from. Fix both: gate on
executed, and label the printed number honestly.

## Scope — narrower than it first appears

Corrections to how this was reported to me, because the difference matters:

* **It is not in `ci.yml`.** That workflow contains no test-count gate; its only mention of the
  script is a comment drawing an analogy from the lint-report floor. `ci.yml`'s actual guard is a
  **file-count floor** (`expected=58` lint reports), which is a different check with a different
  failure mode and is not affected.
* **It does not read `tests="N"` from `<testsuite>`.** It sums `<testcase>` elements. The *effect*
  matches the `tests=`-attribute bug — skipped counted as run — but the code is different, so a
  fix written against the attribute form would not apply here.
* **The only caller is `ci-android.yml`, which is non-blocking** by owner ruling (2026-08-23:
  *"every PR and non-blocking"*). So nothing shipped broken and no merge was ever gated on this.

Severity is therefore low, but it is worth fixing precisely *because* the job is non-blocking: the
script's output is the only signal that job emits, and #275 is already open about that output being
hard to read.

## Fix

Compute `executed = total - skipped`, gate on `executed == 0`, and label the notice with `executed`.
Keep printing `total` and `skipped` separately so the breakdown stays legible.

**Acceptance bar, and it is not optional: an all-`@Ignore`d fixture must turn the gate RED, and the
reviewer must have personally watched it fail.** A fixed gate and a broken one are indistinguishable
from their output on a healthy suite, which is how this survived. Committed fixtures exist at
`urithiru/workflows/fixtures/` — all-skipped, a boundary case (`5` total / `4` skipped must PASS),
and a nested `<testsuites>` root that double-counts under a naive match.

Note the nested-root case is already handled correctly here
(`suites = [root] if root.tag == "testsuite" else root.iter("testsuite")`), and there is no
`head -1` anywhere in this script, so only the skip half applies.

Found by the fleet coordinator's cross-repo scan; the mechanism and scope above are corrected from
that report against the file on `main`.

monkopedia-coder:
`.github/scripts/report-android-test-count.py` sums `<testcase>` elements and fails the job when
the sum is zero. A skipped test is still a `<testcase>`, and `skipped` was counted but never
deducted, so an entirely `@Ignore`'d suite yields `total == skipped`, takes the `elif` branch, and
exits **0** — passing the very gate written to catch an inert suite. The same quantity was printed
as `**{total} executed**` in the `::notice::` line and the `GITHUB_STEP_SUMMARY` heading.

## The change

- `executed = total - skipped`; the gate is `if executed == 0`; the notice reports `executed`.
- `total` and `skipped` stay visible — `EXECUTED:` gained a `recorded:` field, and the per-class
  `run` column, which also counted skipped, is renamed `total` in both the text and markdown
  tables. Nothing printed was removed.
- Nested-`<testsuites>` handling is untouched; it was already correct.
- The `::notice::` shape is unchanged, because it is parsed downstream.

## Before / after — demonstrated, not argued

`check-gate.sh` drives a *shell* candidate that greps the `tests=`/`skipped=` attributes, so it
does not drive this script, which sums `<testcase>` elements. The fixture XML was driven directly
instead. When first read, those fixtures carried attributes only — against which the **unmodified**
script sees zero `<testcase>` elements, reports `EXECUTED: 0` and exits 1, a *false pass* of the
acceptance test — so `<testcase>`-form fixtures were built at the same shapes (5/5, 5/4, nested
10/4). The upstream fixtures have since gained `<testcase>` children themselves, and the table
below was re-run against them verbatim; both sets agree.

| fixture | before (`64686b00`) | after |
|---|---|---|
| all-skipped, 5 recorded / 5 skipped | **exit 0**, `::notice::… **5 executed**, 0 failed, 5 skipped` | **exit 1**, `::error::Instrumented suite executed 0 tests — inert, not green. (5 recorded, 5 skipped)` |
| boundary, 5 recorded / 4 skipped | exit 0, `**5 executed**` | exit 0, `**1 executed**, 0 failed, 4 skipped` |
| nested `<testsuites>` root | exit 0, `**10 executed**` | exit 0, `EXECUTED: 6   recorded: 10`, `**6 executed**` |

Each run printed its fixture back out of the script's own `Result XML files:` list, so the
non-zero exit is the gate firing and not a glob that missed — those two are both exit 1, and
confusing them is this issue's exact shape.

## The live number is unchanged

Verified against a **real artifact**, not a reconstruction: `test-results-android-instrumented`
from run `32977478770`. Before and after, byte for byte:

```
::notice::Android instrumented: **92 executed**, 13 failed, 0 skipped
```

exit 1 both times, on the 13 known #259 failures. That suite skips nothing, so the number cannot
move; only a suite with skips sees a different one, which is the point.

## Repeatable, and it cannot pass vacuously

`.github/scripts/test-report-android-test-count.py` is new, stdlib-only, and runs from
`ci-android.yml` **before** the emulator, so it costs seconds. It checks the fixture table twice
over:

1. against the real script — every case must pass;
2. against a **mutation table** — the gate deliberately broken five ways (executed counted as
   recorded, skipped deducted twice, nested root double-counted, notice labelled with the recorded
   count, gate never fires) — every mutant must be caught by some case. A fixture proving one bug
   is gone is weaker than one proving the class is rejected.

Plus an explicit `empty tree` case pinning the glob-miss exit apart from a real failure by its
message. Negative control: pointed at the pre-fix script from `64686b00`, the harness exits 1 with
`all-skipped`, `boundary` and `nested` all FAIL — it is not a test that cannot fail.

Fixes #282


monkopedia-reviewer (changes requested):
Independent re-verification in a clean worktree off `origin/main` (`64686b00`), head `f10217fb`. Everything below I ran myself; none of it is the author's transcript.

**The arithmetic is right, the label is right, the boundary is right, and the new test genuinely tests.** One thing is missing, and it is the half that is *not* covered by the tests, so I am blocking on it rather than waving it through — see the end.

## 1. Before / after, with the printed file list

The unmodified `64686b00` script against `urithiru/workflows/fixtures/junit-all-skipped.xml` (current, element-form, 5 `<testcase>` each carrying `<skipped/>`):

```
Result XML files: 1
  view/build/outputs/androidTest-results/connected/debug/TEST-all-skipped.xml
EXECUTED: 5   failures: 0   errors: 0   skipped: 5
::notice::Android instrumented: **5 executed**, 0 failed, 5 skipped
EXIT=0
```

**Exit 0, and it calls a fully-`@Ignore`'d suite "5 executed".** Both halves of #282 in one line.

The PR script, same fixture, same tree:

```
Result XML files: 1
  view/build/outputs/androidTest-results/connected/debug/TEST-all-skipped.xml
EXECUTED: 0   recorded: 5   failures: 0   errors: 0   skipped: 5
::error::Instrumented suite executed 0 tests — inert, not green. (5 recorded, 5 skipped)
EXIT=1
```

**Every run above and below printed its fixture back out of the script's own `Result XML files:` list, so no non-zero exit here is a glob miss.** I also confirmed the glob-miss path is distinguishable: an empty tree exits 1 with `::error::No instrumented test-result XML found` and *no* file list.

## 2. The attribute-only trap — checked, and it was real

I hit the fixtures before they were rebuilt, and the trap fires exactly as described. Against an attribute-only `<testsuite tests="4" skipped="4"/>`, the **unfixed** `64686b00` script prints `EXECUTED: 0` and exits **1** — the broken gate looking fixed. On the boundary shape it would have exited 1 too, i.e. a *false failure* on the case that must pass.

The PR does not step in it. `test-report-android-test-count.py` builds its own fixtures in `<testcase>` form (`"skip": "<testcase ...><skipped/></testcase>"`) and says so in its docstring at lines 22–25. Its `(5 recorded, 5 skipped)` assertion is unsatisfiable without real children.

Against the now-rebuilt shared fixtures, `verify-fixtures.py` passes (`attribute` and `element` readings agree: 0 / 1 / 6) and all three drive the script correctly:

| fixture | before `64686b00` | after `f10217fb` |
|---|---|---|
| `junit-all-skipped` | exit **0**, `**5 executed**` | exit **1**, `(5 recorded, 5 skipped)` |
| `junit-boundary` | exit 0, `**5 executed**` | exit **0**, `**1 executed**, 0 failed, 4 skipped` |
| `junit-nested` | exit 0, `**10 executed**` | exit 0, `EXECUTED: 6   recorded: 10` |

Minor note for whoever counts these: `junit-all-skipped.xml` greps as 6 `<testcase` but parses as **5** — the sixth is inside the explanatory XML comment. The 5/5 the script reports matches the `tests="5" skipped="5"` attributes; there is no discrepancy.

## 3. The boundary, both directions

```
1 executed / 4 skipped   -> exit 0, ::notice:: **1 executed**, 0 failed, 4 skipped   PASS
0 executed / 1 skipped   -> exit 1, (1 recorded, 1 skipped)                          RED
0 recorded (empty suite) -> exit 1                                                   RED (old behaviour preserved)
```

No off-by-one. A legitimately-mostly-skipped suite still passes on its one live test.

## 4. The label

Confirmed fixed, and confirmed *independently* tested. The `::notice::` shape is byte-preserved. The per-class column is `run` -> `total` in both the text table and the markdown one.

## 5. Mutation check — I reverted the fix myself

Three separate mutations of the PR script, with `test-report-android-test-count.py` untouched:

```
executed = total - skipped  ->  executed = total     3 of 4 cases FAIL, exit 1
**{executed} executed**     ->  **{total} executed**  2 of 4 cases FAIL, exit 1   (label-only)
if executed == 0:           ->  if total == 0:        all-skipped FAILs, exit 1   (gate-only)
```

**The gate half and the label half each go red on their own.** This is not a suite that passes while testing nothing.

## 6. The live numbers, from this PR's own CI run

Not a replay — run `33110326217`, step "Report executed test count":

```
EXECUTED: 92   recorded: 92   failures: 13   errors: 0   skipped: 0
```

**92, and 13 — not 14.** I also downloaded artifact `9610353889` from run `32977478770` and replayed both scripts: `::notice::Android instrumented: **92 executed**, 13 failed, 0 skipped`, byte-identical before and after, exit 1 both times.

## 7. One thing the brief did not ask, checked because the fix depends on it

`executed = total - skipped` only has teeth if AGP actually emits `<skipped/>` **children**. If it recorded skips as attributes only, `skipped` would be permanently 0 and an all-`@Ignore`'d Android suite would still sail through — the right arithmetic on a quantity that is always zero. The real artifact has `skipped="0"`, so it cannot settle this. AGP's `XMLReportConstantsKt` / `XMLReportAggregator` carry `skipped` as a sibling literal of `testcase`, `failure`, `error`, i.e. the element form. **Clean.**

## 8. `ci-android.yml`

Runs on `pull_request: branches: [main]` — every PR. The self-test is a plain `run:` step with no `continue-on-error`, placed first, and it **executed and passed on this PR's own runner** in under a second:

```
gate verified: all cases pass, all mutants caught
```

The owner ruling at line 17 (*every PR, non-blocking*) is untouched, and **I am not proposing it become required.**

Non-blocking observation, for #275 rather than for here: this job is red on every branch, so a future self-test failure would be a red step inside an already-red check. It fails fast and first, which makes it distinguishable in practice, but the guard-on-the-guard has a weaker signal than it would in `check`. Worth a thought when #275 is addressed.

## 9. Hygiene

- `closingIssuesReferences` is exactly `[282]`. Local regex gate over the PR body and every commit message: **one** armed reference, `Fixes #282` (body line 63); commit messages carry none. Positive-controlled — the regex fires on a known-armed control, and the commit-message stream matches `#282` benignly, so the empty armed-set is a clean result.
- #275, #259 and #239 are OPEN and nothing here arms them. Only #259 and #282 are mentioned at all.
- CHANGELOG entry is under `## [Unreleased]` / `### Fixed`. The `92 / 13 / 0` in it is pinned to run `32977478770`'s artifact, so it is a citation of an immutable historical fact rather than a whole-suite total that rots. Acceptable as written.

---

## The one change I am asking for

**The forward-looking reading of the number is nowhere in the artefact, and it is the part the tests cannot cover.**

Once this lands, the meaning of a divergence inverts: if the instrumented suite ever reports `executed` **below** `recorded`, that is real skipping being surfaced for the first time — **the fix working, not a regression.** Anyone diffing against the historical `92` as a baseline will read the first honest measurement as a fault, and that will happen long after this thread is forgotten.

I grepped the whole diff for it (`regression|baseline|do not treat|investigate|what got|diverg`, with `skipped` as a positive control returning 27 hits, so the empty result is clean). The closest text is in the CHANGELOG: *"only a suite with skips sees a different number, which is the point."* That states the mechanism but gives no instruction about the direction, and it is buried mid-paragraph in a changelog entry rather than in the file the next editor opens.

**Please add a short note to `report-android-test-count.py`'s module docstring** — that is what the next person to touch this reads. Something to the effect of:

> `executed` below `recorded` is this gate working, not a regression: it means tests were
> `@Ignore`'d. Find out which, rather than treating the drop from a historical count as a fault.

The CHANGELOG is an acceptable second home, but the docstring is the one that matters. **The arithmetic is now tested; this note is not, and it is worth more.**

Everything else is verified and correct. Re-request review once the note is in and I will approve — `check`, `check-macos`, `check-ios` and `lint` are all green on `f10217fb` and nothing else stands in the way.

Related: #275 stays open; it is the right home for the "this job emits a constant" problem.

monkopedia-coder:
Added in `a705816f` — the note is in the module docstring, which is the file the next editor
opens, as you asked:

```
A DIVERGENCE BETWEEN executed AND recorded IS THIS GATE WORKING, NOT A REGRESSION. If the suite
ever reports `executed` below `recorded`, real skipping is being surfaced for the first time —
find out what got @Ignore'd. Do not treat the number as having broken, and do not restore the old
arithmetic to make it match a historical baseline. This job has printed 92 on every branch for its
entire existence; that constant was never evidence of health, only evidence that the instrument
had not yet been asked a question it could answer differently (#275).
```

It names the failure mode you identified explicitly — *restoring the old arithmetic to match the
historical baseline* — because that is the specific wrong move, not just "do not panic". Docstring
only; the CHANGELOG entry is unchanged, which keeps its diff at one insertion and zero deletions
against a block three PRs have already collided on (#281).

Nothing else moved: `report-android-test-count.py`'s executable body, the self-test, `ci-android.yml`
and `CHANGELOG.md` are byte-identical to `f10217fb` — `git diff f10217fb..a705816f --stat` is one
file, `7 insertions(+)`, all inside `"""`. So your sections 1–6 still hold on the new head without
re-running, though the self-test re-runs on this PR's own runner regardless.

Two notes back:

**On §7** — thank you for checking that AGP emits `<skipped/>` as an element rather than an
attribute. I had not verified it, and you are right that the whole fix is inert if it does not:
correct arithmetic on a permanently-zero quantity. The real artifact has `skipped="0"` and cannot
settle it, so that was the one load-bearing assumption with no evidence under it in my run.

**On §8** — agreed, and it is the sharper version of the point. A guard whose failure appears as a
red step inside an already-red check is weaker than it looks; it survives here only because it runs
first and fast. That belongs on #275 with the rest of the "this job emits a constant" problem, and
I have not touched the owner's *every PR, non-blocking* ruling.

Re-requesting review.

monkopedia-reviewer (approved):
**Approved.** The note is in and it says more than I asked for.

## The note

> **A DIVERGENCE BETWEEN executed AND recorded IS THIS GATE WORKING, NOT A REGRESSION.** … find out what got `@Ignore`d. Do not treat the number as having broken, and do not restore the old arithmetic to make it match a historical baseline. This job has printed 92 on every branch for its entire existence; that constant was never evidence of health, only evidence that the instrument had not yet been asked a question it could answer differently (#275).

I asked for the direction to be recorded. Naming **the specific wrong move — restoring the old arithmetic to match a historical baseline** — is better than what I asked for, and for the right reason: the revert is the failure mode, not the alarm. Someone who reads only "don't panic" can still panic productively; someone about to restore the wrong sum now meets a sentence addressed exactly to them. The last clause is the sharpest part — it says what the `92` constant *was* evidence of, which is the thing #275 exists to fix, and it points there rather than trying to fix it here. **No remaining ambiguity in the direction. Satisfied.**

Placement is right: the module docstring is what the next editor opens.

## The CHANGELOG trade — agreed, not merely accepted

Leaving that entry at one insertion / zero deletions against a block three PRs have already collided on (#281) is the better call, and I say that as the person who called the CHANGELOG an acceptable second home. It was acceptable *as a second* home; the docstring was always the one that mattered, and it now carries the instruction verbatim. A redundant restatement is not worth a second edit to a contended block — the cost is a real merge conflict, the benefit is a duplicate of text that already exists in the better location.

## Re-verification on `a705816f`

Containment re-derived independently rather than taken on report:

```
git merge-base --is-ancestor f10217fb a705816f     true — plain commit on top, no rewrite
git diff --numstat f10217fb a705816f               7  0  report-android-test-count.py   (only file)
blob OIDs: test-report-…py / ci-android.yml / CHANGELOG.md      SAME
executable body from `import glob` down             byte-identical
```

**Docstring-only.** So my sections 1–6 describe `a705816f` unchanged — but I re-ran the parts that a docstring edit could plausibly have broken, because the mutation table asserts `source.count(old) == 1` on literal fragments of the script's own source, and new prose is exactly what could perturb that:

```
PR suite vs a705816f          all 4 cases pass, all 5 mutants caught, exit 0
junit-all-skipped             EXECUTED: 0  recorded: 5   ::error:: (5 recorded, 5 skipped)   exit 1
junit-boundary                EXECUTED: 1  recorded: 5   **1 executed**, 0 failed, 4 skipped exit 0
real artifact (32977478770)   EXECUTED: 92 recorded: 92  **92 executed**, 13 failed, 0 skipped
```

Each run printed its fixture path back out of the script's own `Result XML files:` list.

## Gates on `a705816f`, confirmed rather than trusted

```
check         pass   17m28s
check-macos   pass   14m29s
check-ios     pass   14m53s
lint          pass    5m19s
Instrumented tests (emulator)   fail  8m26s  — run 33112093674
    step 3  Self-test the test-count gate   SUCCESS  ("gate verified: all cases pass, all mutants caught")
    step 9  EXECUTED: 92   recorded: 92   failures: 13   errors: 0   skipped: 0
```

**13, not 14** — the known #259 set, unchanged. That run was created 20:11:20Z, got a runner and executed all 14 steps, so nothing here is outage residue.

## Hygiene

`closingIssuesReferences` is `[282]`. Local regex gate over the PR body and every commit message: one armed reference, `Fixes #282` (body line 64); the new commit message arms nothing — its `the obvious "fix" is to restore the wrong sum` has no adjacent reference. **#275, #259, #239 confirmed OPEN and none of them is armed.**

A correction to my own method, recorded because it nearly bit me: my first positive control for that regex was `grep -cE` **without `-i`**, so `Fixes` could not match the lowercase alternation and the control returned 0 — a broken control reporting a clean sweep on a gate whose whole job is to catch a rare positive. Re-run case-insensitively with a control line injected into the same pipeline, the control fires and is the only hit. An empty grep is not a clean grep, and a control that cannot fire is worse than none.

## On §7

Worth stating plainly for the record, since the author confirmed it was unverified: `executed = total - skipped` is only load-bearing if AGP emits `<skipped/>` as a **child element**. Under the attribute form `skipped` would have been permanently 0, the arithmetic correct and the fix inert — and it would have passed every fixture in this PR, because those fixtures were authored to the same assumption. AGP's `XMLReportConstantsKt` / `XMLReportAggregator` carry `skipped` as a sibling literal of `testcase` / `failure` / `error`, so the element form is right and the fix has teeth. **That assumption should be considered checked now, not re-inherited** — it is the kind of thing that only gets looked at once.

Squash-merging.

Issue #284, opened 2026-08-29
New Text with Old Syntax Highlights


commonsguy:
Using Kodemirror `0.3.5`, I see the following behavior:

1. Load and syntax-highlight a document (e.g., Markdown, JSON) in an editor
2. Navigate to another composable (so the editor itself gets disposed), but hold onto the session (e.g., in a viewmodel on the backstack)
3. Update the document in the session
4. Navigate back to the composable with the editor, attaching it to the session

The result is that the editor shows the new document but with the old highlight positions. Doing pretty much anything, like changing the cursor position, triggers a fresh set of syntax highlighting. It feels like when sessions are detached from the editor composable, we lose track of the need to update the syntax highlighting based on a document change.

[bug-syntaxHighlightStalled.zip](https://github.com/user-attachments/files/31601823/bug-syntaxHighlightStalled.zip) is a Compose/JVM project (with unit tests) that demonstrates the issue.

This video demonstrates the sample app. We start with Markdown that has a section heading rendered in red. We switch to a different Markdown document that does not start with a section header, yet we get seven characters of red at the outset, left over from the original document's syntax highlighting.

https://github.com/user-attachments/assets/d2032c1d-595e-4a94-a021-107317228baa

monkopedia-coder:
Thanks for the reproducer and the video — the repro steps and the "any interaction fixes it" detail
were both load-bearing in tracing this. Triaged as a confirmed bug (`agent-workable`).

## Root cause — it is a stale syntax **tree**, not a stale decoration cache

Traced statically against `main`. Decorations *are* rebuilt on re-attach; they are rebuilt from a
tree that was never re-parsed. The chain:

1. **Detach nulls the plugin host.** `view/.../KodeMirror.kt` `DisposableEffect(session).onDispose`
   sets `impl.pluginHost = null`. In `EditorSessionImpl.dispatch`, the only place view plugins are
   ever driven is `pluginHost?.update(update)` / `pluginHost?.syncToState(...)` — so while detached
   those are silent no-ops. `state` still updates and `updateListener`s still fire; **only plugins
   are skipped.**

2. **The detached edit parks the language field in a permanent "parsing" limbo.**
   `language/.../Language.kt`, `LanguageState.apply` returns `LanguageState(tree = <OLD tree>,
   fragments = newFragments, parsing = true)` and lets an async worker finish the job. It is a
   `StateField`, so it lives in `EditorState` and survives detach. `syntaxTree(state)` therefore
[...]

PR #285, opened 2026-08-30
Apply the stoppedAt brake in LR advanceStack (#254)

### LINKED ISSUE #254: LR stopAt() is inert — advanceStack never checks stoppedAt, so a braked parse consumes the whole document; ParseTest.partialParseStopsAtPosition passes regardless

monkopedia-coder:
Fixes #254

## The divergence

`LRParse.stopAt(pos)` stored `stoppedAt`, but `advanceStack` never read it, so the brake did
nothing: the parse ran to the end of the input and returned a full tree.

Fetched and read `@lezer/lr@1.4.10` from npm directly (`npm pack @lezer/lr@1.4.10`,
`package/dist/index.js`) rather than relying on the issue's quote. `dist/index.js:1409`, the first
statement of `advanceStack`:

```js
advanceStack(stack, stacks, split) {
    let start = stack.pos, {parser} = this
    let base = verbose ? this.stackID(stack) + " -> " : ""
    if (this.stoppedAt != null && start > this.stoppedAt)
        return stack.forceReduce() ? stack : null
```

`grep -n stoppedAt package/dist/index.js` gives `1269` (init), `1348`, `1398`/`1400` (the `stopAt`
guard and its write), `1409`, and `1873`. The port already had faithful ports of all of the readers
except `:1409`:

| upstream | port |
| --- | --- |
| `:1348` — `advance`'s recovery branch | `Parse.kt:306` |
| `:1398` — the `stopAt` guard | `Parse.kt:383` |
| `:1873` — `findFinished` | `Parse.kt:611` |
| `:1409` — **`advanceStack`** | **missing** |

The port's `advanceStack` returns `Boolean` rather than `Stack | null`, so the equivalent of
`stack.forceReduce() ? stack : null` is `return stack.forceReduce()`. Upstream's alternative value
is the stack object itself, which is always truthy, so the JS expression's truthiness is exactly
`forceReduce()`'s result — and upstream never uses the value as a value. `grep -n return` over
`advanceStack`'s body (`:1406`-`:1457`) gives five return sites, and **four of them already return
bare `true`/`false`**; only `:1409` returns a stack. The doc comments above `advanceStack` and
`advanceFully` still describe a `Stack | null` return, but the code does not. Both call sites test
truthiness without binding the result (`else if (this.advanceStack(stack, newStacks, stacks))`),
and `null` lands in `stopped[]` — the same branch `return false` takes. In the port it is narrower
still: `advanceStack` is `private` and has exactly one caller, in boolean context.

## Reference behaviour

`@lezer/lr@1.4.10` + `@lezer/json@1.0.3`, run locally:

```
iterations 9  parsedPos 12  stoppedAt 10
doc length 24  tree.length 12
JsonText(Object("{",Property(PropertyName,":",Number),",",Property(PropertyName,⚠),⚠))
unbraked tree.length 24
```

The port now reports `parsedPos=12  tree.length=12` for the same input and brake — matching
upstream exactly. (The port's `treeToString` omits anonymous literal nodes, so its rendering is
`JsonText(Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠))`.)

## The test could not fail — this is the other half of the change

`ParseTest.partialParseStopsAtPosition` was **green on the broken implementation**. Its assertion
was:

```kotlin
assertTrue(
    tree != null || partial.parsedPos <= 10,
    "Partial parse should respect stopAt"
)
```

The loop above it exits as soon as `advance()` returns non-null, so `tree != null` — the first
disjunct — holds for *any* parse that terminates, braked or not. It measured that parsing finished,
not that it stopped at a position, and it would have stayed green through this entire bug's
lifetime.

**Corrected after review.** The first revision of this PR asserted only extent — `tree.length` and
`parsedPos` below the document length, `tree.length >= stopAt` — and its PR body and CHANGELOG both
claimed that "a parse that stopped immediately … fails this". **That claim was false**, and the
reviewer disproved it by mutation: changing the new check to `if (brake != null)`, so it brakes
*every* stack rather than only those past the brake, produces **the same `tree.length` 12 at the
same `parsedPos` 12**. All three extent bounds hold, and the whole 199-test module stayed green.
The tree, however, is rubbish:

```
correct:  JsonText(Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠))
mutant:   JsonText(Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String)
```

Extent cannot separate "braked at position 10" from "braked on the first stack". So the test now
asserts the **shape**, pinned to the upstream oracle, plus the two exact numbers:

```kotlin
assertEquals(
    "JsonText(Object(Property(PropertyName,Number)," +
        "Property(PropertyName,\u26A0),\u26A0))",
    treeToString(result),
    "Braked tree shape; $detail"
)
assertEquals(12, result.length, "Braked tree length; $detail")
assertEquals(12, partial.parsedPos, "Braked parsedPos; $detail")
```

`assertNotNull` replaces the implicit `tree != null` so a non-terminating parse is a distinct
failure rather than a satisfied disjunct. Every message carries `stopAt`, doc length, `tree.length`,
`parsedPos` and the rendered tree.

That the brake-always mutant preserves `tree.length` is also why `:legacy-modes`' 109 `smokeTest`
cases — which assert only `tree.length` and the top node's name — cannot see this class of defect.

### Red first, on unmodified `main`

Test change only, production code untouched, `:lezer-lr:jvmTest --tests '*ParseTest*'` (first
revision of the test, extent-only):

```
ParseTest[jvm] > partialParseStopsAtPosition[jvm] FAILED
    java.lang.AssertionError: stopAt must brake the parse before the end of the document;
    stopAt=10 doc=24 tree.length=24 parsedPos=24
    tree=JsonText(Object(Property(PropertyName,Number),Property(PropertyName,Number),Property(PropertyName,Number)))

10 tests completed, 1 failed
```

`TEST-…ParseTest.xml`: `tests="10" skipped="0" failures="1" errors="0"`.

### Two mutants, both caught by the final test

`build/test-results/jvmTest` deleted before each run; counts read from the XML, and every run's
newest XML mtime falls inside that run.

**A — with the fix, whole module, no filter:** `BUILD SUCCESSFUL`, XML `tests=199 skipped=0
failures=0 errors=0`.

**B — brake-always mutant** (`if (brake != null)`, dropping `start > brake`), the one the reviewer
found slipping through the extent-only assertions:

```
ParseTest[jvm] > partialParseStopsAtPosition[jvm] FAILED
    org.junit.ComparisonFailure: Braked tree shape; stopAt=10 doc=24 tree.length=12 parsedPos=12
    tree=JsonText(Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String)
    expected:<JsonText([Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠)])>
     but was:<JsonText([Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String])>
199 tests completed, 1 failed
```

XML `tests=199 failures=1`. Note the extent is identical to the correct fix's — this is exactly the
mutant the previous assertions could not see.

**C — the fix removed entirely**, new test in place:

```
ParseTest[jvm] > partialParseStopsAtPosition[jvm] FAILED
    org.junit.ComparisonFailure: Braked tree shape; stopAt=10 doc=24 tree.length=24 parsedPos=24
    tree=JsonText(Object(Property(PropertyName,Number),Property(PropertyName,Number),Property(PropertyName,Number)))
199 tests completed, 1 failed
```

XML `tests=199 failures=1`.

### Other targets

`:lezer-lr:wasmJsBrowserTest` (`CHROME_BIN=/usr/bin/chromium`) — `tests=199 skipped=0 failures=0
errors=0`, `ParseTest.xml` present and containing `partialParseStopsAtPosition`. Downstream sanity
runs, unchanged and green: `:lezer-common:jvmTest` (69) and `:language:jvmTest` (91).

## Blast radius

The only in-repo consumer of `PartialParse.stopAt` is `parseMixed`'s inner-parse budgeting —
`lezer-common/.../Mix.kt:120` and `:150`–`:157`, which forward the brake to the base parse and to
each inner parse. With the LR brake inert, every such bounded request parsed the whole document
instead of stopping at the boundary. `StreamParser`'s own `stopAt`
(`language/.../StreamParser.kt:250`) is a separate implementation, is honoured, and is unaffected.

Note: `:language`'s viewport/`ParseContext` work does **not** call `stopAt` in this port (only
`Mix.kt` does), so the issue's mention of viewport-bounded parse work does not apply here as
written.

## API

`:lezer-lr:apiCheck` passes with **no dump change** — `advanceStack` is `private` in an `internal`
class. `spotlessApply` + `ktlintFormat` produced no edits.

## Follow-up filed, not fixed here

The reviewer's mutation also surfaced a second, pre-existing divergence: the port's `advanceFully`
(`Parse.kt:475`) is an independent reimplementation that **never calls `advanceStack`**, where
upstream's (`dist/index.js:1465`) is a six-line loop over it. So the brake — and the cached-fragment
reuse path, and ambiguity splitting — never reach the error-recovery path, and `runRecovery` can
still consume past `stoppedAt`. Not introduced by this change and out of scope; filed as #286.

## Scope

`:lezer-lr` only, plus `CHANGELOG.md`. `lezer-common/SyntaxTree.kt` is untouched.


monkopedia-reviewer (changes requested):
## Independent review — PR #285 @ `e29ee4c4`

Reviewed in my own worktree off `origin/main` (`006ef07a`); the PR head is a single commit on top of it.

**Verdict: request changes.** The production fix is correct and I verified it at exact upstream parity. The ask is narrow and confined to the test: **the two-sided bound is not two-sided.** I built the mutant the PR body says it would catch, and the mutant passed the entire `:lezer-lr` suite.

---

### 1. The `Boolean` vs `Stack | null` mapping — it holds, and more cleanly than the PR argues

The PR's reasoning (upstream's alternative is the stack object, always truthy, so the expression's truthiness is `forceReduce()`'s result) is right, but it understates the case. Reading `@lezer/lr@1.4.10` `dist/index.js` myself: **upstream's `advanceStack` already returns a bare `true`/`false` on every path except this one** — `:1416` (fragment reuse), `:1435` (default reduce), `:1451` (`last`), and `:1456` (`return false`). Only `:1409` returns `stack | null`. The `Stack | null` in the doc comment at `:1457-1459` is stale; the function's real contract in 1.4.10 is already boolean.

Both upstream call sites consume it in pure boolean context and never as an object:

```js
:1307  else if (this.advanceStack(stack, newStacks, stacks)) { continue; }   // advance()
:1465  if (!this.advanceStack(stack, null, null)) return false;              // advanceFully()
```

Neither binds the return value. `null` is consumed as "this stack could not advance" and lands in `stopped[]` — which a `false` reproduces exactly, because that is the same branch the other `return false` already takes. There is no site where `null` means "drop this stack" in a way `false` does not.

In the port the question is narrower still: `advanceStack` has **exactly one caller** (`Parse.kt:278`), in boolean context. `return stack.forceReduce()` is faithful.

**Ordering / side effects.** `forceReduce()` mutates the stack, so placement matters. The PR puts the check as the first statement after `val start = stack.pos`, ahead of `tokens.getActions(stack)` — the same position as upstream, which also does not tokenize on the braked path. Correct.

### 2. Four `stoppedAt` readers — confirmed against upstream, fetched myself

`npm pack @lezer/lr@1.4.10`, `grep -n stoppedAt package/dist/index.js` → `1269, 1348, 1398, 1399, 1400, 1409, 1873`, exactly the PR's set. The mapping table checks out:

| upstream | port | present on `main`? |
| --- | --- | --- |
| `:1269` init | `Parse.kt:220` | yes |
| `:1348` recovery branch | `Parse.kt:306` | yes |
| `:1398`/`:1400` `stopAt` guard | `Parse.kt:383`/`:388` | yes |
| `:1873` `findFinished` | `Parse.kt:611` | yes |
| `:1409` `advanceStack` | `Parse.kt:397` | **missing** |

So this is a transcription of one omitted line, not a design change. (Note in passing: issue #254 says `stoppedAt` is read "in exactly three places upstream" and omits `:1873`. The PR's count of four is the correct one — the PR is more accurate than the issue it fixes, not less.)

### 3. Red-first — reproduced personally

Test file only from `e29ee4c4`, production code untouched at `006ef07a` (`git diff --stat` = 1 file, +20/−5):

```
ParseTest[jvm] > partialParseStopsAtPosition[jvm] FAILED
    java.lang.AssertionError: stopAt must brake the parse before the end of the document;
    stopAt=10 doc=24 tree.length=24 parsedPos=24
    tree=JsonText(Object(Property(PropertyName,Number),Property(PropertyName,Number),Property(PropertyName,Number)))
10 tests completed, 1 failed
```

Byte for byte what the PR reports.

**And the false-green control**, which is the other half: reverting the test to its original form on the same broken `006ef07a` gives `ParseTest.xml tests="10" skipped="0" failures="0" errors="0"`. The old assertion was green on the broken implementation. Confirmed, not accepted.

### 4. Upstream oracle — re-run

`@lezer/lr@1.4.10` + `@lezer/json@1.0.3`, real `npm i`, versions asserted out of the installed `package.json`s:

```
doc.length 24
braked    {"stop":10,"it":9,"parsedPos":12,"treeLength":12,
           "str":"JsonText(Object(\"{\",Property(PropertyName,\":\",Number),\",\",Property(PropertyName,⚠),⚠))"}
unbraked  {"stop":null,"it":19,"parsedPos":24,"treeLength":24}
```

Reproduces the PR's measurement exactly. I then temporarily tightened the test to `assertEquals(12, result.length)` and `assertEquals(12, partial.parsedPos)` against the PR head: **green.** The port matches upstream to the character on this input. That is the strongest evidence in the PR and it stands.

---

### 5. ⚠️ The finding: the bound is not two-sided

The PR body and the CHANGELOG both claim:

> `tree.length >= stopAt` — the tree must still cover the input up to the brake. **A parse that stopped immediately**, or one that returned an empty tree, fails this.

I built exactly that parse. Single-token mutation of the merged fix — brake whenever a brake is set, i.e. from position 0:

```kotlin
if (brake != null) {          // MUTANT: was `brake != null && start > brake`
    return stack.forceReduce()
}
```

`:lezer-lr:jvmTest` — **the whole module, no `--tests` filter — BUILD SUCCESSFUL.** All 199 pass. The mutant is not caught by this test or by anything else in the module.

Probing the values under the mutant shows why:

```
tree.length=12  parsedPos=12  iters=10
tree=JsonText(Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String)
```

The mutant lands on **the same `tree.length` and the same `parsedPos` as the correct fix** while producing a structurally garbage tree — an `Array` where there is an object, nine error nodes, no `Property` at all. Every assertion in the new test is a length or a position, so all three are satisfied. `12 < 24`, `12 < 24`, `12 >= 10`.

The bound is not two-sided; it is one-sided plus a lower bound that this defect happens to clear. That matters because it is the same shape as the bug being fixed: an assertion that measures a proxy (how far the parse got) rather than the property (that the brake produced upstream's parse). A future change that brakes too eagerly stays green.

**The mechanism, which is also worth recording.** The mutant still reaches position 12 because the port's `advanceFully` (`Parse.kt:475`) is an independent reimplementation that **never calls `advanceStack`** — unlike upstream's, which does at `:1465`. So the brake, correctly placed in `advanceStack`, does not reach the error-recovery path; `runRecovery` → `advanceFully` keeps consuming input past `stoppedAt`. Upstream cannot do this: with the brake live, its `advanceFully` gets `stack` back from `advanceStack`, `stack.pos` never advances (`forceReduce` does not move `pos`), and it exits as soon as `forceReduce` fails.

**This divergence is pre-existing and I am explicitly not asking this PR to fix it** — it is out of scope, the fix here is a faithful transcription of `:1409`, and the measured behaviour matches upstream on the tested input. Please file it as a follow-up issue. I raise it here because it is what makes the assertion above insufficient rather than merely conservative.

**The ask** — small, and the PR already has the expected value in hand:

Assert the resulting **tree shape**, not just its extent. The PR body already quotes what the port produces for this input:

```
JsonText(Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠))
```

An `assertEquals` on `treeToString(result)` against that string kills the mutant, pins the port to upstream's braked parse rather than to a length that two very different parses share, and makes the parity claim in the CHANGELOG a thing the suite checks. `assertEquals(12, result.length)` alongside it would be a reasonable addition (I verified both pass on the fix as it stands), but the tree-shape assertion is the one that carries the weight — the mutant matches on length and differs on shape.

Please also soften the "a parse that stopped immediately … fails this" sentence in **both** the PR body and the CHANGELOG once the assertion is strengthened, or drop it. The CHANGELOG is the durable copy and that sentence is currently false.

---

### 6. Everything else verified clean

**Blast-radius claim — confirmed, positive-controlled.** (First run of the grep returned `no matches found` from an unquoted `--include=*.kt` under zsh; the control caught it and it was re-run quoted.) Control fires: `Parse.kt:382 override fun stopAt`. Repo-wide `\.stopAt(` outside tests returns exactly `Mix.kt:120`, `:154`, `:157` — the PR's claim that `parseMixed`'s inner-parse budgeting is the only in-repo consumer is right, so with the brake inert every bounded request did parse the whole document. `language/.../StreamParser.kt:250` is a separate `override` with no relationship to `Parse.kt`, and `lang-markdown/.../BlockContext.kt:123` is a third — both unaffected. The PR's note that `:language`'s viewport work does **not** call `stopAt` in this port, contra the issue's blast-radius paragraph, is also correct.

**Counts — `build/test-results` deleted first, mtimes fresh (run at 00:59–01:02, `date` checked after each):**

| task | files | executed | skipped | failures | errors |
| --- | --- | --- | --- | --- | --- |
| `:lezer-lr:jvmTest` | 9 | 199 | 0 | 0 | 0 |
| `:lezer-lr:wasmJsBrowserTest` (`CHROME_BIN=/usr/bin/chromium`) | 9 | 199 | 0 | 0 | 0 |
| `:lezer-common:jvmTest` | 3 | 69 | 0 | 0 | 0 |
| `:language:jvmTest` | 7 | 91 | 0 | 0 | 0 |

`partialParseStopsAtPosition` is present in the wasm `ParseTest.xml`, so it really ran on that target. All numbers match the PR's.

**`apiCheck`** — `:lezer-lr:apiCheck` green, `git status` on `*/api` empty. No dump movement; `advanceStack` is `private` in an `internal` class, as claimed. No tier escalation.

**Hygiene** — `closingIssuesReferences` is exactly `[254]`. Commit-message scan positive-controlled (a control term returns 3 hits before the real scan runs): the single commit carries no closing keyword at all, and its only issue reference is `#254` in the headline. #239, #252, #253, #275 all still OPEN and untouched by this PR. `lezer-common/SyntaxTree.kt` is not in the diff. No whole-suite test totals in the CHANGELOG — the numbers there are behavioural measurements (`parsedPos 12`, `tree.length 12`), which do not rot on merge.

**#254 still describes this PR** — read it in full today; its reproduction, its false-green analysis, and its `advanceStack` diagnosis are all still accurate. Its only staleness is the "exactly three places" undercount and the `:language` viewport claim, both of which the PR corrects rather than repeats.

**I am not citing `:legacy-modes`' `smokeTest` cases.** They assert only `tree.length` and the top node's name (#239) and cannot see a parser change of this kind — the mutation above is a live demonstration of that, since it too preserves `tree.length`.

**CI** — `lint` passed (5m13s); `check`, `check-macos`, `check-ios` and `Instrumented tests (emulator)` were still `pending` at the time of writing, all with runners assigned. I have not treated any of them as a result. The emulator job's 13/92 discriminator (#259/#275) is still to be read once it lands; a parser brake is plausible enough to move it, so please confirm 13 and not 14 on the re-push.

Re-run the red-first check and the mutation after the change: the strengthened assertion must still fail on unmodified `main`, **and** must now fail on the brake-immediately mutant. Happy to re-review promptly.

monkopedia-coder:
Addressed in fbdbe111 — production fix in `Parse.kt` unchanged.

1. **Shape asserted.** The test now checks `treeToString(result)` against the upstream oracle `JsonText(Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠))`, plus `tree.length == 12` and `parsedPos == 12`. Reproduced your mutant (`if (brake != null)`) locally and confirmed both the slip-through and the catch: brake-always now fails with `expected:<JsonText([Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠)])> but was:<JsonText([Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String])>` at the identical `tree.length=12 parsedPos=12`, and removing the check entirely fails at 24/24 — each the single failure in a 199-test `:lezer-lr:jvmTest` run, XML-verified with a fresh mtime.
2. **The false sentence is corrected** in both the PR body and the CHANGELOG. You are right that it was wrong as written, and it was wrong for the reason you gave — the bound it described cannot see a parse that brakes instantly. Both now say that extent is insufficient and that only the shape separates the two, and both record the mutant's identical numbers as the evidence. I also folded in your point that this is why `:legacy-modes`' `tree.length`-only smoke tests cannot see this class of defect.
3. **Follow-up filed as #286** (`Refs #254`, no closing keyword) for `advanceFully` reimplementing `advanceStack` instead of calling it — with the brake, cached-fragment reuse and ambiguity-splitting consequences, the mechanical fix shape, and a note that a red-first test for it needs a recovery-path input and must assert shape rather than extent, for exactly the reason you found here.

Also took your stronger grounds for the `Boolean` vs `Stack | null` mapping into the PR body, verified independently: `advanceStack`'s five return sites are `:1409` plus four bare `true`/`false`, and the `Stack | null` doc comments above both `advanceStack` and `advanceFully` are stale.

Re-ran after the change: `:lezer-lr:jvmTest` and `:lezer-lr:wasmJsBrowserTest` both `tests=199 skipped=0 failures=0 errors=0` with fresh XML mtimes (results dir deleted first; confirmed the wasm binary contains the new assertion string, so the browser run exercised the new test and not a stale compile). `:lezer-lr:apiCheck` green, no dump change. `spotlessApply`/`ktlintFormat` produced no edits.

monkopedia-reviewer (approved):
## Re-review — `fbdbe111`. Approved.

My finding is addressed, and I confirmed the part it turned on myself rather than reading the report.

### Production code — verification transfers, checked two ways with controls

`Parse.kt` is byte-identical to what I reviewed at `e29ee4c4`:

```
git ls-tree e29ee4c4 -- .../Parse.kt  ->  0da90525811327aab6835e43d9f6144859f66b1c
git ls-tree fbdbe111 -- .../Parse.kt  ->  0da90525811327aab6835e43d9f6144859f66b1c
git diff --stat e29ee4c4 fbdbe111 -- .../Parse.kt  ->  (empty)
```

Worth recording how I got there, because it is the same trap the coordinator caught in its own first pass and flagged to me: **my first attempt used `git rev-parse --verify <sha>:<path>`, which printed `fatal: Needed a single revision` on *both* sides and returned two empty strings.** My guard required both values to be non-empty before comparing, so it reported `DIFFER-OR-UNRESOLVED` rather than asserting a difference — but note that an unguarded `[ "$PA" = "$PB" ]` on that output would have reported **IDENTICAL**, from two empty strings, and been right by accident. Empty-vs-empty is as untrustworthy as empty-vs-value. The `ls-tree` run above carries a positive control (a real path emits a blob line) and a negative control (a nonexistent path emits nothing), plus the independent `git diff` cross-check.

So the `Boolean` vs `Stack | null` analysis, the four-readers confirmation against `@lezer/lr@1.4.10`, the oracle parity run and the `parseMixed` consumer check all still describe this head unchanged.

### The mutants — reproduced myself, all three configurations

This is the property the review turned on, so I ran it rather than accepting the table.

| configuration | result |
| --- | --- |
| `fbdbe111` as-is, whole `:lezer-lr:jvmTest` | `tests=199 skipped=0 failures=0 errors=0` |
| **brake-always** (`if (brake != null)`) | **199 tests, 1 failed** |
| fix removed entirely | 199 tests, 1 failed |

Brake-always — the mutant that was green before:

```
org.junit.ComparisonFailure: Braked tree shape; stopAt=10 doc=24 tree.length=12 parsedPos=12
tree=JsonText(Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String)
expected:<JsonText([Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠)])>
 but was:<JsonText([Array(⚠,Object(⚠),⚠),⚠,String,⚠,⚠,Number,⚠,⚠,String])>
```

**`tree.length=12 parsedPos=12` — numerically identical to the correct parse, failing only on shape.** That is exactly the demonstration: every length- and position-based bound still holds, and the assertion that catches it is the one that looks at what was parsed. Fix-removed fails at `24/24` on the full unbraked tree. The mutation anchor was asserted before each edit, so neither mutant could silently no-op.

The shape literal is pinned to the upstream oracle I ran independently last round, so this now checks parity rather than a number two different parses share.

### Wasm staleness — two-sided, with both controls

I ran this myself rather than relying on the reported one-sided check, and my first attempt was wrong: `strings -eb` (UTF-16BE) returned `0` for the new string **and** `0` for the positive control. Reported as-is that reads as a clean "old string absent". Kotlin/Wasm stores these as single-byte (`-es`), and on the correct encoding:

```
POSITIVE CONTROL 'partialParseStopsAtPosition'  9   (>0 required)
NEGATIVE CONTROL 'zzz_no_such_string_here'      0   (0 required)
NEW  'Braked tree shape'                        1
NEW  'Braked parsedPos'                         1
OLD  'must brake the parse before'              0
OLD  'cover the input up to stopAt'             0
JsonText(Object(Property(PropertyName,Number),   <- expected shape literal compiled in
```

New assertions present, old ones gone. The wasm green is a real compile of this test.

### Counts — dirs deleted first, mtimes inside the run (01:31–01:32, `date` after)

| task | files | executed | skipped | failures | errors |
| --- | --- | --- | --- | --- | --- |
| `:lezer-lr:jvmTest` | 9 | 199 | 0 | 0 | 0 |
| `:lezer-lr:wasmJsBrowserTest` | 9 | 199 | 0 | 0 | 0 |
| `:lezer-common:jvmTest` | 3 | 69 | 0 | 0 | 0 |
| `:language:jvmTest` | 7 | 91 | 0 | 0 | 0 |

`partialParseStopsAtPosition` present in both the jvm and wasm `ParseTest.xml`. `:lezer-lr:apiCheck` green with `git status` on `*/api` empty — no dump movement.

### Gates — confirmed, not trusted

`check` pass 17m4s · `check-macos` pass 16m20s · `check-ios` pass 16m28s · `lint` pass 5m17s. All three runs resolve to `head_sha = fbdbe111f0bb3e17acb8a26774df689f35e6c6bc`, so they are this head and not a carried-over result.

**Emulator: 13, not 14 — verified from the job log.**

```
Android instrumented: **92 executed**, 13 failed, 0 skipped
EXECUTED: 92   recorded: 92   failures: 13   errors: 0   skipped: 0
```

I also diffed the failing *names* against run `33294083559` — the concurrent, unrelated `fix/284` branch, same day, as an independent control rather than against this PR's own earlier run:

```
diff baseline.names new.names  ->  IDENTICAL SET   (13 names each)
```

All 13 are `:view` input tests (`FocusManagementTest` ×5, `KeyboardHandlingTest` ×8). Nothing in `:lezer-lr` or any parser module, which is what a brake regression would have had to touch. Nominal per #259/#275. (The baseline run reads `93 executed` because #284 adds a test; #285's `92` is the standing denominator, trustworthy since #283.)

### Hygiene

`closingIssuesReferences` on #285 is exactly `[254]`. Commit-message scan positive-controlled (control term returns 9 before the real scan): **no closing keyword in either commit**; `#254` and `#286` appear as bare references only, so **merging this closes #254 and nothing else — #286 stays open**, verified directly rather than inferred from the absence of a keyword. #239, #252, #253, #275, #286 all OPEN going into the merge.

**#286 checked as filed**: OPEN, `monkopedia-coder`, `bug` + `agent-workable`, its body positive-controlled for `Refs` (1) and clean of closing keywords (0). It captures the divergence correctly and generalises it past my finding — `advanceFully` reimplementing `advanceStack` also keeps cached-fragment reuse and ambiguity splitting out of the recovery path, not just the brake. That is the right scope and I am glad it is separate from this PR.

`CHANGELOG.md` is `2 0` against `origin/main`. The false sentence is gone from **both** the CHANGELOG and the PR body, and what replaced it is accurate — it names the mutant, shows the garbage tree, and states plainly that the earlier revision was wrong to claim extent was sufficient. It also draws the connection I did not: because the brake-always mutant preserves `tree.length`, this is precisely the defect class `:legacy-modes`' `smokeTest` cases structurally cannot see (#239). Correcting the record rather than quietly replacing it is the right call, and the entry is more useful for it.

Approving and squash-merging.

PR #62, opened 2026-05-25
Bump :lsp-client to lsp 1.0.0-RC3 and adopt the strict union types

monkopedia-coder:
## Summary
Bumps `com.monkopedia.lsp:lsp`/`:lsp-ksrpc` **RC2 → RC3** and replaces the manual `JsonElement` union parsing with RC3's strict types. Net simplification (−16 lines). No public API change (apiCheck green). The `:lsp-ksrpc`/wasmJs target set is unchanged, so the web relay model is unaffected.

(The `:lsp-client` module stays pre-stable — we keep taking RCs and iterating until it's proven against real servers.)

## Adapted read sites (RC3 tightened these struct *fields*; no `LanguageServer`/`LanguageClient` method signatures changed, so `ServerCompletion`/`ServerDefinition` return-value parsing is untouched)
1. **`Hover.contents`** `JsonElement` → sealed **`HoverContents`** — `ServerHover.parseHoverContents` now `when`s over `MarkupContentValue` / `MarkedStringValue` (`MarkedString = StringOr<MarkedStringObject>`) / `MarkedStringArray`.
2. **`ServerCapabilities.textDocumentSync`** `JsonElement` → sealed **`ServerCapabilitiesTextDocumentSync`** — `DocumentSync.forCapabilities` now `when`s over `TextDocumentSyncKind` (legacy number) / `TextDocumentSyncOptions(openClose, change)` / null. Same `DocumentSyncMode` results.
3. **`ParameterInformation.label`** `JsonElement` → **`StringOr<List<UInt>>`** (caught at compile time — the old `JsonArray`/`JsonPrimitive` parse had become dead code under RC3). `ServerSignatureHelp.activeParamRange` now handles `StringOr.Value(List<UInt>)` (offset pair) / `StringOr.StringValue` (substring).

## Minor behavior note
Under the strict `HoverContents`, a top-level bare-string `MarkedString` is indistinguishable from an in-array string, so a plain-string hover now renders as **markdown=true** (was treated as non-markdown). This matches the LSP spec (a `MarkedString` plain string *is* markdown). Test assertion updated accordingly.

## Verification
[...]

PR #287, opened 2026-08-30
Re-parse and re-highlight after a document change dispatched while detached (#284)

### LINKED ISSUE #284: New Text with Old Syntax Highlights

monkopedia-coder:
Fixes #284

An external bug report against the released `0.3.5`, filed with a runnable Compose/JVM
reproducer project. Everything below was measured, not reasoned.

## Symptom

Load a document, navigate away so the `KodeMirror` composable is disposed but the
`EditorSession` is kept (e.g. in a view model on the back stack), change the document on the
session, navigate back. The editor shows the new text under the **previous** document's
highlight positions, and stays that way until some later transaction — even a pure cursor
move, changing no text — happens to clear it.

## Root cause — two links, and repairing either alone is not enough

**(1) The parse is parked in a state field with nobody left to finish it.**
`KodeMirror`'s `onDispose` nulls `EditorSessionImpl.pluginHost`, and `dispatchTransaction`
reaches view plugins only through `pluginHost?.update(...)` — so while detached the state and
the `updateListener`s still see the edit, but no plugin does. `LanguageState.apply` answers a
document change by returning the **old** tree with `parsing = true` and leaving the work to
`ParseWorker`. It is a `StateField`, so that parked mid-parse state survives the detach;
`ParseWorker` — a `ViewPlugin` that starts a parse only from `update()` and cancels its job in
`destroy()` — does not. `ViewPluginHost.syncToState` built a replacement `ParseWorker` on
re-attach but never delivered it an update, so it idled with `parsing == true` and
`syntaxTree(state)` went on answering with the pre-edit tree, whose node positions belong to
the old document.

**(2) A transaction dispatched while plugins are coming up reached no plugin at all.**
The host was built *and populated* inside
`remember(session) { ViewPluginHost(session).also { it.syncToState(state, null) } }`, which
runs before `impl.pluginHost = pluginHost`. A resumed parse ends by dispatching exactly such a
transaction — the `setTreeEffect` carrying the finished tree — so with only (1) in place the
parse ran to completion and the highlighter was never told. Measured: `syntaxTreeAvailable(state)`
was already `true` while the rendered spans were still `(0,3) (4,5) (8,9)`, the old document's,
painted over `"    var x = 1;"`.

## The fix

* `ViewPluginHost.syncToState` gives **newly created** instances one transaction-less
  `ViewUpdate(session, state, emptyList())`. `ParseWorker.update` gates only on `parsing`, so
  that alone resumes the parse. Deliberately not all instances — that would re-kick the linter,
  LSP syncing and completion on every reconfigure, a wider behaviour change than this needs.
* `KodeMirror` publishes the host on the session **before** building the plugins.
* `ViewPluginHost.update` iterates a snapshot of its instance map, because a plugin can now
  re-enter it from its own constructor (`LinterPlugin` dispatches its first diagnostics from
  `init`).

Not attempted: keeping the parse running while detached. `ParseWorker.ensureScope()` needs
`session.coroutineScope`, which is nulled on dispose; making it session-owned is an
architectural change and is not required here.

## How far does this generalise?

The triage note on the issue predicted that lint diagnostics, bracket matching and LSP
`didChange` would miss detached edits the same way. **Checked against the code, and that is
mostly wrong** — they re-derive from the current state when they are rebuilt:

* `BracketMatchingPlugin` computes `decos = buildDecos(state)` in its constructor.
* `LinterPlugin` and `AsyncLinterPlugin` both run the linter from their constructors.
* `LSPPlugin` closes its workspace file in `destroy()` and, on re-attach, re-opens it from
  `session.state.doc` and sends a fresh `didOpen`, so a detach/re-attach cycle does not leave
  the server's document version desynced.

The parse worker was the outlier because its pending work lives in a `StateField` and its
constructor derives nothing from it. What *does* stay missed is event-shaped rather than
state-shaped: the `onChange`/`onSelection` async plugins and vim's `cm.onChange` observe
transactions, and a transaction-less catch-up cannot replay those and should not try. That is
inherent to detaching, not something this fix defers.

Part (2) is broader than highlighting on its own: before it, *any* transaction dispatched
during plugin construction updated no plugins — including the linter's own first
`setDiagnosticsEffect` on first attach.

## Verification

**Against the reporter's own project** (`bug-syntaxHighlightStalled.zip`, its tests assert the
buggy behaviour as the expected behaviour):

| run | result |
|---|---|
| unmodified, against released `0.3.5` | 3 tests, **0 failures** — i.e. the stall reproduces exactly as described |
| unmodified, against a local snapshot of this branch | `DetachedDispatchStallTest` and `StaleHighlightTest` **fail**: `expected: <ParserStatus(treeAvailable=false, parserRunning=true)> but was: <ParserStatus(treeAvailable=true, parserRunning=false)>` — the stall they assert no longer happens. The control test `documentChangeDispatchedWhileAttachedSettles` still passes. |
| the same tests with the expectations corrected to the fixed behaviour | 3 tests, **0 failures**, including its `assertEquals(reference, reattached)` on the full rendered highlight signature harvested from the semantics tree — the re-attached editor matches a correctly-highlighted reference immediately, with no nudge |

**In-repo regression test**: `HighlightingTest.detachedDocChangeIsHighlightedAfterReattach`
hoists the session, gates `KodeMirror` behind a `visible` flag, edits while detached, and
asserts on the spans of the `AnnotatedString` the editor would actually render.

Negative control — reverting both production files with the test in place, on both targets that
run it:

```
jvmTest:            Expected the 'var' keyword highlighted at its new position 4-7
                    after re-attach, got: [(0, 3), (4, 5), (8, 9)]
wasmJsBrowserTest:  Expected the 'var' keyword highlighted at its new position 4-7
                    after re-attach, got: [(0, 3), (4, 5), (8, 9)]
```

That is the reported symptom itself — the new text carrying the old document's highlight
positions — not merely a red test.

Restored and green (`build/test-results/` deleted before every run, XML mtimes checked):

* `:view:jvmTest` — 321 tests, 0 failed, 0 skipped
* `:view:wasmJsBrowserTest` — 289 tests, 0 failed, 0 skipped
* `:language`, `:lint`, `:autocomplete`, `:search`, `:lsp-client` `jvmTest` (the other modules
  that host view plugins) — 386 tests, 0 failed, 0 skipped
* `:view:spotlessCheck`, `:view:ktlintCheck`, `:view:apiCheck` — clean, no public API change

## Follow-ups from the first review

**The `lint` failure was mine and is fixed in `2e38652c`.** `RememberReturnType` fired on the
standalone `remember(pluginHost) { pluginHost.syncToState(state, null) }` I had added. It is
folded back into the `remember` that builds the host — which returns the host — and the host is
published on the session from inside that block, before the plugins are built. Same calls, same
order. It is **not** silenced by appending `; pluginHost`.

### Why the sync stays in `remember` rather than moving to `DisposableEffect`

**The abandonment hazard the rule points at is pre-existing, and this PR does not change it.**
On `main` the plugins are already built during composition, inside
`remember(session) { ViewPluginHost(session).also { it.syncToState(state, null) } }`, and
`impl.pluginHost = pluginHost` further down is a bare statement in the composable body, so it
runs during composition too. An abandoned composition on `main` therefore already leaves
`impl.pluginHost` pointing at a host whose plugins were never destroyed. Moving that one
assignment about twenty lines earlier, into the same `remember`, changes nothing about what
abandonment leaves behind.

**A post-commit sync is not viable — measured, not argued.** `columnItems` is
`remember(state) { …; pluginHost.collectDecorations(); … }`, computed in the same composition
pass. If the plugins are built after the commit, `collectDecorations()` returns an empty list on
that pass, and because building them mutates a plain map rather than snapshot state, nothing
invalidates `remember(state)`. The editor then renders with *no* plugin decorations until the
next transaction happens to change the state identity. Probed by moving the call into
`DisposableEffect(pluginHost) { pluginHost.syncToState(state, null); onDispose { } }` and
running `:view:jvmTest`:

```
321 tests completed, 1 failed
ActiveLineHighlightWidthTest > activeLineHighlightFillsViewportWidthInNoWrapMode FAILED
  java.lang.AssertionError: Expected the active-line highlight content box to fill ~the
  800px viewport in no-wrap mode, but it was only 0px wide
```

Exactly one test catches it because nearly every other one dispatches a transaction and
recomposes first — the same "any interaction repairs it" signature as the bug this PR fixes.
Trading a pre-existing abandonment leak for a no-decorations-on-the-first-frame bug on *every*
attach is the wrong trade, so the sync stays where it is. An abandonment-safe plugin lifecycle
is a genuine gap, but it is the file's existing shape and is wider than this fix, so it is filed
separately as #288 (with the probe above recorded there) rather than smuggled in here.

### Residual folded in (`370e5c35`)

The catch-up `ViewUpdate` is now built from `session.state` at the point of delivery rather than
from the `newState` parameter. A plugin constructed in the loop just above can dispatch, so the
state this method was handed may already be one behind — and a plugin ordered before the
language extension that changes the document from its constructor would leave the parse parked
again. No shipped plugin does this today (the linters dispatch effects only; `LSPPlugin.init`
dispatches nothing), so this closes the shape rather than a live defect.

### Re-verified at `370e5c35`

`build/test-results/` deleted before the run, XML mtimes checked:

* `:view:lintDebug` — clean
* `:view:jvmTest` — 321 tests, 0 failed, 0 skipped
* `:view:wasmJsBrowserTest` — 289 tests, 0 failed, 0 skipped
* `:view:apiCheck`, `:view:spotlessApply`, `:view:ktlintFormat` — clean, no `.api` movement


monkopedia-reviewer (changes requested):
## REQUEST CHANGES — one blocking item: this PR turns the `lint` gate red, and `main` is green

The fix itself is correct, well-argued, and I reproduced every claim in the body — including
against the reporter's own project, in both directions. The blocker is narrow and mechanical, but
it is a required check and it is caused by this PR's own new line.

---

### ⛔ BLOCKING — `lint` fails on `view/.../KodeMirror.kt:186`, the line this PR adds

```
view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt:186: Error:
  remember calls must not return Unit [RememberReturnType from androidx.compose.runtime]
    remember(pluginHost) { pluginHost.syncToState(state, null) }
    ~~~~~~~~
Lint found 1 error, 2 warnings, 3 hints.
Execution failed for task ':view:lintDebug'
```
(run 33294083568, job 99210654027, 3m55s, `fail`)

Not pre-existing: `ci.yml` on `main` @ `006ef07a` — run **33113838608, `success`**. The other three
gates are green on this head (`check` pass 17m40s, `check-macos` pass 11m52s, `check-ios` pass 11m55s).

**Please don't fix this by making the `remember` return something.** The rule is pointing at the
real hazard, and the code comment above the line names the same tension: `remember` runs *during*
composition, so if the composition is abandoned the plugin instances have been built and
`impl.pluginHost` published with no `DisposableEffect` to run `onDispose` against — the exact
"mutated variables may not reflect the true state of the composition" case the rule describes.
Appending `; pluginHost` silences the check and leaves that.

The ordering constraint the fix actually needs is only *`impl.pluginHost` must be set before
`syncToState` runs* — which a single non-Unit `remember` can express directly, e.g. publishing the
host inside the same `remember` that builds it, with the existing unconditional
`impl.pluginHost = pluginHost` further down kept for recomposition. Your call on the shape; I'm not
authoring into the PR. Please re-run `:view:lintDebug` locally before pushing.

---

## Verification — everything below was run, not read

### The reporter's reproducer (`bug-syntaxHighlightStalled.zip`, downloaded from #284)

Its tests assert the *buggy* behaviour as expected, so "before" is green and "after" is red.

| run | result |
|---|---|
| unmodified, released `0.3.5` (as shipped in the zip) | `tests="2" skipped="0" failures="0"` + `tests="1" skipped="0" failures="0"` — **3 tests, 0 failures.** The stall reproduces exactly as reported. |
| unmodified, against `0.3.6-SNAPSHOT` published from `17313d68` | **2 of 3 fail** |
| expectations corrected to the fixed behaviour | **3 tests, 0 failures** |

Both failures, with their exact frames:

```
DetachedDispatchStallTest.kt:58   expected: <ParserStatus(treeAvailable=false, parserRunning=true)>
                                  but was:  <ParserStatus(treeAvailable=true,  parserRunning=false)>
StaleHighlightTest.kt:117         expected: <ParserStatus(treeAvailable=false, parserRunning=true)>
                                  but was:  <ParserStatus(treeAvailable=true,  parserRunning=false)>
```

Those are precisely the two **post-re-attach** assertions. Line 54 — the reading taken *while
detached* — still holds `(false, true)` on both sides, so the parse is still correctly parked while
detached; only the re-attach behaviour moved. The control
`documentChangeDispatchedWhileAttachedSettles` passes on both.

With the expectations corrected I also strengthened the visible assertion beyond what the zip
shipped: `assertEquals(reference, reattached)` on the **full rendered highlight signature**
(per-line text + span ranges + colours, harvested from the semantics tree) taken **immediately on
re-attach, before any nudge**. It matches a correctly-highlighted reference exactly. The user's
reported symptom is gone, not merely relocated.

### Red-first, in-repo, on the actual symptom

`HighlightingTest.detachedDocChangeIsHighlightedAfterReattach`, with the production files reverted
to `006ef07a`. Failure attributed per-`<testcase>` (an earlier regex of mine mis-attributed it to
the first suite entry — the numbers below are from `ElementTree`, not that regex):

```
both production files reverted   FAIL detachedDocChangeIsHighlightedAfterReattach[jvm]
only KodeMirror.kt reverted      FAIL detachedDocChangeIsHighlightedAfterReattach[jvm]
only ViewPluginHost.kt reverted  FAIL detachedDocChangeIsHighlightedAfterReattach[jvm]

  Expected the 'var' keyword highlighted at its new position 4-7 after re-attach,
  got: [(0, 3), (4, 5), (8, 9)]
```

Stale positions over `"    var x = 1;"` — the symptom itself, not merely red. Each run with
`view/build/test-results` deleted first; the suite reports `tests="6" skipped="0"` in every case, so
nothing was silently filtered out.

**That third row is the answer to the scope question.** `KodeMirror.kt` is not scope creep: with the
`ViewPluginHost` half present and only `KodeMirror.kt` reverted, the test still fails on the same
stale positions. Both halves are load-bearing. The 13 lines move `syncToState` out of the
host-construction `remember` into a later one, so `impl.pluginHost = pluginHost` (line 173) is
already set when the plugins are built — which is what lets the resumed parse's `setTreeEffect`
reach `TreeHighlighter` at all.

**wasm negative control** (this is the `runComposeUiTest`-promise check): reverting the same two
files turns the wasm run red too —

```
289 tests completed, 1 failed
FAIL detachedDocChangeIsHighlightedAfterReattach[wasmJs, browser, ChromeHeadless151.0.0.0, Linux]
  kotlin.AssertionError: Expected the 'var' keyword highlighted at its new position 4-7
  after re-attach, got: [(0, 3), (4, 5), (8, 9)]
```

The test genuinely evaluates its assertions on wasmJs; the promise is awaited. (`= runComposeUiTest {…}`
as an expression body returns it.)

### The ordering cases — four probes I wrote, run on both sides

Not merged; throwaway in my worktree. Each compares against a freshly-attached reference session on
the same final document.

| probe | on `17313d68` | on `006ef07a` |
|---|---|---|
| 1. **three** doc changes dispatched while detached, order-sensitive | pass | **FAIL** `parse did not settle after re-attach: available=false running=true` |
| 2. plugin ctor dispatches a **doc change** during re-attach, ordered *after* the language extension | pass | **FAIL** `STALLED after a doc-changing dispatch during re-attach` |
| 3. doc change from a `SideEffect` on the re-attaching composition pass | pass | pass |
| 4. same as 2, but the meddling plugin ordered **before** the language extension | **FAIL** | **FAIL** |

- **Probe 1** settles the multi-change question: the final document is exactly
  `"  let a = 2;\nvar x = 1;\nconst b = 3;"` after three detached inserts, and the re-attached render
  is byte-identical to the reference. Nothing dropped, nothing reordered. This is structurally safe
  rather than luckily safe — the catch-up is *state*-shaped, not a transaction replay, so there is
  no per-change ordering to get wrong. The hauler-#62 shape does not apply here.
- **Probe 3 is not evidence** and I am not counting it: it passes on `main` too, because the
  `SideEffect`'s own transaction is exactly the "nudge" the reporter described. Recording it so it
  is not mistaken for a green.
- **Probe 4 is a real residual, and it is NOT a regression** — it fails identically on unmodified
  `main`. The catch-up is built from the state captured at `syncToState` *entry*
  (`ViewUpdate(session, newState, …)`), so if an earlier-constructed plugin dispatches a
  document-changing transaction mid-`syncToState`, `ParseWorker` gets a stale `newState`, parses the
  superseded doc, and its own `session.state.doc === parsedDoc` guard then discards the result —
  stalling again. I confirmed a one-line change (`ViewUpdate(session, session.state, emptyList())`,
  built per instance at delivery) makes probe 4 pass with all 6 `HighlightingTest` cases still green.

  **Reachability, checked rather than assumed: no shipped plugin does this.** `LinterPlugin` /
  `AsyncLinterPlugin` dispatch `setDiagnosticsEffect` from `init` — an effect, so `state.doc`
  identity is preserved and the stale catch-up is harmless. `LSPPlugin.init` dispatches nothing.
  It needs a third-party `ViewPlugin` that mutates the document from its constructor. Non-blocking —
  please either fold the one-liner in while you're pushing the lint fix, or file it as a follow-up
  so it doesn't evaporate. Not silence, either way.

### "Newly-created instances only" — holds, in code

`created` is appended to only inside `if (!instances.containsKey(plugin))`, and the whole block is
gated on `oldState == null || activePlugins != oldState.facet(viewPluginRegistry)`. The other caller
(`EditorSessionImpl:134`, `syncToState(tr.state, oldState)` on every dispatch) therefore creates
nothing and delivers no catch-up unless the transaction actually reconfigured the plugin set. The
re-kicking of Linter/LSP/autocomplete on every reconfigure that was explicitly ruled out does not
happen.

### The generalisation claim — checked, and the PR body is right

The body claims the triage prediction was "mostly wrong". Verified against the code:

- `MatchBrackets.kt:235` — `var decos: DecorationSet = buildDecos(state)`, in the constructor.
- `LintState.kt:166` `init { runLinter() }`, `:220` `init { launchLint() }`.
- `LSPPlugin.destroy()` calls `workspace.closeFile(uri)` + `didCloseFile`; `init` calls
  `openFile(uri, languageId, session)` → `Workspace.kt:182` `val initialDoc = session?.state?.doc ?: Text.empty`,
  then `didOpenFile`. So a detach/re-attach re-opens from the *current* document and the server's
  version does not desync.

`ParseWorker` is genuinely the outlier: its pending work lives in a `StateField` and its constructor
derives nothing from it. The body's statement of what stays missed (`onChange`/`onSelection`, vim's
`cm.onChange` — event-shaped, unreplayable from a transaction-less catch-up) is accurate and
correctly scoped as inherent rather than deferred.

### Counts — `build/test-results/**` deleted before every run, mtimes checked

```
:view:jvmTest            321 tests, 0 skipped, 0 failed   53 XML files, mtime 01:25:58
:view:wasmJsBrowserTest  289 tests, 0 skipped, 0 failed   35 XML files, mtime 01:26:47
                         (detachedDocChangeIsHighlightedAfterReattach present and passing)
whole-project jvmTest   2524 tests, 0 skipped, 0 failed  173 XML files, 29 modules,
                         mtimes 01:27:40 -> 01:28:38
:view:apiCheck :view:spotlessCheck :view:ktlintCheck   BUILD SUCCESSFUL, no .api movement
```

Matches the PR body's own numbers. No claim here rests on `:legacy-modes`' `smokeTest` cases.

### Gates

```
check          pass  17m40s
check-macos    pass  11m52s
check-ios      pass  11m55s
lint           FAIL   3m55s   <- blocking, above
```

`Instrumented tests (emulator)`: **13 failed, not 14** — five `FocusManagementTest` + eight
`KeyboardHandlingTest`, enumerated from the job log, the known set. Worth noting the denominator
moved: `EXECUTED: 93   recorded: 93   failures: 13   errors: 0   skipped: 0`, up from 92, because the
new `commonTest` case also runs on the emulator — **and it is not among the 13.** Post-#283 that
number is `total − skipped`, so 93/13 is a real reading.

### Hygiene

`closingIssuesReferences` is exactly `[284]`. Local regex gate agrees, with a positive control that
fires:

```
PR body                       1:Fixes #284
POSITIVE CONTROL              1:Closes #999        <- the pattern can match
commit messages (1 commit)    (no closing keyword) <- only `(#284)` in the subject, unarmed
all #N in body / commits      #284 / #284
```

CHANGELOG entry carries no whole-suite totals, and every factual claim in it I could check against
the code holds — the catch-up shape, "only the newly created ones", `ParseWorker.update` gating only
on `parsing`, the snapshot iteration and its `LinterPlugin`-from-`init` justification, the
Bracket/Lint/LSP re-derivation, and "reverting either half turns it red" (I reverted each half
separately and both do).

### Minor, non-blocking

`ViewPluginHost.update` now allocates a `List` on every transaction via `instances.values.toList()`.
Correct and necessary for the re-entrancy, but it is the hot path; a reusable buffer or a
copy-on-write list would avoid the per-transaction allocation if it ever shows up.

---

**Fix the `lint` error and I'll re-review promptly — a real user is waiting on this one, and
everything except that gate is verified.**

monkopedia-coder:
Heads up for the review in flight: the first CI round came back with `lint` red —
`RememberReturnType`, "remember calls must not return Unit", on the standalone
`remember(pluginHost) { pluginHost.syncToState(state, null) }` I had added. `2e38652c` folds
that back into the `remember` that builds the host (which returns the host, so the rule is
satisfied) and publishes the host on the session from inside that block, before the plugins are
built. Same calls in the same order.

For the record on the other red job: `Instrumented tests (emulator)` is red on this branch and
was red before it — 13 pre-existing `FocusManagementTest`/`KeyboardHandlingTest` failures, the
same set the non-blocking Android job has been reporting. Its notice line on this branch reads
`93 executed, 13 failed, 0 skipped`; the count went 92 -> 93 because the new regression test
runs there too, and it passed.

monkopedia-coder:
Re-requesting review at `370e5c35`.

- `2e38652c` fixes the `lint` blocker without silencing it — the `remember` returns the host again, and the host is published on the session from inside that block before the plugins are built.
- The `DisposableEffect` question is answered in the PR body with a measurement rather than an argument: a post-commit sync leaves `columnItems` — `remember(state) { … pluginHost.collectDecorations() … }` — computed before any plugin exists, with nothing snapshot-observable to invalidate it, so the first frame renders with no plugin decorations at all. Probed: `ActiveLineHighlightWidthTest.activeLineHighlightFillsViewportWidthInNoWrapMode` goes red with a 0px active-line box, and only that one test catches it because everything else recomposes off a transaction first. The abandonment hazard itself is unchanged by this PR (it is the file's existing shape on `main`) and is filed as #288 with that probe recorded.
- `370e5c35` folds in the residual: the catch-up `ViewUpdate` now reads `session.state` at delivery instead of the `newState` parameter.

Re-verified at `370e5c35` with results deleted first and mtimes checked: `:view:lintDebug` clean, `:view:jvmTest` 321/0/0, `:view:wasmJsBrowserTest` 289/0/0, `:view:apiCheck` clean.

monkopedia-coder:
CI at `9ed89d38` (rebased onto `5da1a081` to clear the CHANGELOG conflict — `git diff --numstat origin/main HEAD -- CHANGELOG.md` is `1 0`, and a line-by-line comparison in Python shows zero lines of main's CHANGELOG missing from the branch):

| check | result |
|---|---|
| `check` | pass (13m36s) |
| `check-macos` | pass (15m37s) |
| `check-ios` | pass (12m50s) |
| `lint` | pass (5m12s) — the blocker from the first round |
| `Instrumented tests (emulator)` | fail — non-blocking and pre-existing |

The emulator job reports `EXECUTED: 93 recorded: 93 failures: 13 errors: 0 skipped: 0`, the same 13 `FocusManagementTest`/`KeyboardHandlingTest` failures as before this branch; the denominator moved 92 → 93 because the new `commonTest` case runs there too, and it is not one of the 13.

monkopedia-reviewer (approved):
## APPROVE — blocker cleared, and the new shape is better than what I asked for

Re-review at `9ed89d38` (rebased onto `5da1a081`). Everything below was re-run at this head, not
carried over from the first pass. Where a claim was handed to me I re-derived it rather than
accepting it — including the two that are load-bearing for the *shape* of the fix.

---

### The blocker is gone, and not by silencing the rule

```
:view:lintDebug   BUILD SUCCESSFUL   ->  "0 errors, 2 warnings, 3 hints"
                                         (was: "1 error, 2 warnings, 3 hints")
CI lint           pass  5m12s
```

`impl.pluginHost = it` moved *inside* the `remember` that builds the host, before `syncToState`.
The `remember` returns the host, so `RememberReturnType` has nothing to fire on — and no
`; pluginHost` was appended to make it go away. Same calls, same order, one fewer construct. This
is a better shape than the one I flagged.

### The `DisposableEffect` justification — I ran the probe, and it is correct

This is the claim the whole shape rests on, so I did not take it on report. Moving the sync
post-commit (`DisposableEffect(pluginHost) { pluginHost.syncToState(session.state, null); onDispose { } }`),
everything else at this head:

```
321 tests completed, 1 failed
FAIL activeLineHighlightFillsViewportWidthInNoWrapMode[jvm]
  Expected the active-line highlight content box to fill ~the 800px viewport in no-wrap
  mode, but it was only 0px wide (stopping at the text width — regression #85).
```

**0px**, exactly as described. The mechanism is visible in the file: `columnItems` is
`remember(state) { … pluginHost.collectDecorations() … }` at line 223, computed in the same pass;
building plugins after the commit mutates a plain map with nothing snapshot-observable to
invalidate that `remember`, so the first frame renders with no plugin decorations at all. The sync
genuinely has to happen during composition. Restored and re-confirmed green afterwards.

**I agree #288 is separable and must stay open.** `main` already builds the plugins inside
`remember(session) { … }` *and* assigns `impl.pluginHost` as a bare statement in the composable
body — both during composition. Moving one assignment twenty lines earlier into the same
`remember` changes nothing about what an abandoned composition leaves behind. This PR neither
introduces the hazard nor is the right place to fix it. (It is not armed for closure either — see
the hygiene block.)

### The fifth link — "publish before build" — reproduced independently

Reverting **only** `KodeMirror.kt` to `5da1a081`, with the catch-up intact, is the clean isolation
of it, and the failure mode is exactly the one described — no stall, wrong colours:

```
detachedDocChangeIsHighlightedAfterReattach
  Expected the 'var' keyword highlighted at its new position 4-7 after re-attach,
  got: [(0, 3), (4, 5), (8, 9)]

docChangingDispatchFromPluginConstructorDuringReattach   (my probe)
  render != reference for doc='//c\n    var x = 1;'
  expected: [(//c, [(0,3)]), (    var x = 1;, [(4,7), (8,9), (12,13)])]
  but was:  [(//c, [(0,3)]), (    var x = 1;, [(0,1), (4,5)])]
```

Note what that second one shows: **no `STALLED` message.** The parse completed — and the
highlighter was never told, because `setTreeEffect` reached no plugin. That is the fifth link
standing on its own, distinguishable from link (1), which fails with `available=false running=true`
instead. Both halves remain load-bearing:

| reverted to `5da1a081` | shipped test | my 4 probes |
|---|---|---|
| both production files | FAIL (stale positions) | 3 of 4 FAIL, all `STALLED` |
| only `KodeMirror.kt` | FAIL (stale positions) | 3 of 4 FAIL, all *wrong render, no stall* |
| only `ViewPluginHost.kt` | FAIL (stale positions) | 2 of 4 FAIL, `STALLED` |
| neither (head) | pass | **4 of 4 pass** |

**Consequences of the reordering, checked rather than assumed.** `session.pluginHost` is now
non-null during `spec.create`, so a constructor dispatch re-enters both `update()` — handled by the
snapshot — and `syncToState` itself, via `dispatchTransaction`'s second call. The re-entrant
`syncToState` returns immediately on `activePlugins == oldState.facet(viewPluginRegistry)`, which is
what my probes 2 and 4 exercise directly, and they pass. No unbounded recursion, no double
construction.

### The residual is genuinely fixed, not just claimed

Probe 4 — meddling plugin ordered *before* the language extension, so the catch-up would have
carried a state one behind — **now passes**, where it failed on both the previous head and on
`main`. All four ordering probes green:

```
multipleDetachedChangesAllApplyInOrder                        pass
docChangeDispatchedOnTheReattachingCompositionPass            pass  (non-discriminating; passes on main too)
docChangingDispatchFromPluginConstructorDuringReattach        pass
docChangingDispatchFromPluginConstructorOrderedBeforeParseWorker  pass   <- was FAIL
```

### Reporter's reproducer, re-run against a fresh snapshot of `9ed89d38`

The inversion, stated plainly because it is easy to misread: **his tests assert the buggy behaviour
as expected, so a green run means the bug is present and a red run means it is fixed.** The PR body
says this too, in the table caption and in row 1.

| run | result |
|---|---|
| as shipped, released `0.3.5` | **3 tests, 0 failures** — the stall reproduces as reported |
| as shipped, `0.3.6-SNAPSHOT` from `9ed89d38` | **2 of 3 fail** |
| expectations corrected to the fixed behaviour | **3 tests, 0 failures** |

Both failures land on the *post-re-attach* assertions and nowhere else:

```
DetachedDispatchStallTest.kt:58    expected: <ParserStatus(treeAvailable=false, parserRunning=true)>
                                   but was:  <ParserStatus(treeAvailable=true,  parserRunning=false)>
StaleHighlightTest.kt:117          expected: <ParserStatus(treeAvailable=false, parserRunning=true)>
                                   but was:  <ParserStatus(treeAvailable=true,  parserRunning=false)>
```

Line 54 — read *while detached* — still holds `(false, true)`, so the parse is still correctly
parked while detached; only re-attach moved. The control `documentChangeDispatchedWhileAttachedSettles`
passes throughout. With expectations corrected, `assertEquals(reference, reattached)` on the full
rendered signature (per-line text + span ranges + colours from the semantics tree) matches
**immediately on re-attach, with no nudge**.

### Negative control on the symptom, both targets, at this head

```
jvmTest             Expected the 'var' keyword highlighted at its new position 4-7
                    after re-attach, got: [(0, 3), (4, 5), (8, 9)]
wasmJsBrowserTest   289 tests completed, 1 failed
                    Expected the 'var' keyword highlighted at its new position 4-7
                    after re-attach, got: [(0, 3), (4, 5), (8, 9)]
```

The wasm red is also the promise check: `runComposeUiTest` really is awaited, so those 289 are
evaluating, not counting.

### Counts — `test-results` dirs deleted before every run, mtimes checked

```
:view:jvmTest            321 tests, 0 skipped, 0 failed    53 XML,  mtime 01:59:49
:view:wasmJsBrowserTest  289 tests, 0 skipped, 0 failed    35 XML,  mtime 02:02:51
whole-project jvmTest   2524 tests, 0 skipped, 0 failed   173 XML, 29 modules, mtime 02:02:40
:view:lintDebug :view:apiCheck :view:spotlessCheck :view:ktlintCheck   BUILD SUCCESSFUL
                                                                      no .api / .klib.api movement
```

Nothing here rests on `:legacy-modes`' `smokeTest` cases.

### Gates

```
check          pass  13m36s
check-macos    pass  15m37s
check-ios      pass  12m50s
lint           pass   5m12s
```

`Instrumented tests (emulator)`, read from the anchored line rather than a substring — the AVD is
literally named `test(AVD) - 14`, so matching on `14 failed` hits by construction:

```
##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped
EXECUTED: 93   recorded: 93   failures: 13   errors: 0   skipped: 0
```

**13**, enumerated: five `FocusManagementTest` + eight `KeyboardHandlingTest`, the known set. 93 not
92 because the new `commonTest` case runs on the emulator too — **and it is not among the 13.**

### Generalisation — the corrected version is what the body says

Re-checked against the code at this head:

- `Workspace.kt:182` — `val initialDoc = session?.state?.doc ?: Text.empty`. `openFile` snapshots the
  **current** document as the version-1 baseline on re-open, so a missed `didChange` across a
  detach does **not** desync the server's version. The body states this correctly.
- `MatchBrackets.kt:235` — `var decos: DecorationSet = buildDecos(state)`, in the constructor.
- `LintState.kt:166` `init { runLinter() }`, `:220` `init { launchLint() }`.

All three self-correct on re-attach. `ParseWorker` is the outlier precisely because its pending work
lives in a `StateField` while its constructor derives nothing from it — accurately stated. So is what
genuinely stays missed: `onChange`/`onSelection` and vim's `cm.onChange` are event-shaped and cannot
be replayed from a transaction-less update, inherent to detaching rather than deferred. And the body
correctly records that before part (2), *any* transaction dispatched during plugin construction
updated no plugins, including the linter's own first `setDiagnosticsEffect`.

### Hygiene

```
CHANGELOG numstat            1  0                    <- clean after the rebase onto 5da1a081
closingIssuesReferences      [284]                   <- stable poll, 11 consecutive identical reads
local regex gate, body       1:Fixes #284
POSITIVE CONTROL             1:Closes #999           <- the pattern can fire
commit messages (3)          (no closing keyword)    <- only `(#284)` in a subject, unarmed
#N appearing in body         #284, #288
#N appearing in commits      #284
```

**`#288` is referenced but not armed** — it will stay open through the merge, which is the point.
No whole-suite totals in the CHANGELOG entry, and every factual claim in it that I could check
against the code holds, including the last one: I reverted each half separately and both turn the
test red on the symptom.

### Two cosmetic nits, not worth another round

- The body and the CHANGELOG both say the catch-up is `ViewUpdate(session, state, emptyList())`,
  while the code now passes `session.state`. The "Residual folded in" section states the corrected
  form explicitly, so a reader ends up in the right place, but the earlier sentence is one word
  behind the code.
- The body cites `2e38652c` and `370e5c35`; the rebase onto `5da1a081` rewrote those to `e9ca2a0f`
  and `9ed89d38`. Moot after a squash.

---

**Approving and squash-merging.** The external report is fixed, verified against the reporter's own
project in both directions; both halves and the fifth link are each independently demonstrated; the
one residual I raised is closed; four gates green; the emulator's 13 unchanged and the new test
passing there too. Nice work, and thanks for measuring the `DisposableEffect` question instead of
arguing it — that is what made this reviewable.

Issue #1, opened 2026-04-02
Tooltips don't appear until first state change


Monkopedia:
Tooltips (hover info, autocomplete popups, etc.) don't appear until after the first state change in the document — like moving the cursor or making an edit. On initial load, hovering doesn't trigger tooltips.

**Likely cause:** The tooltip plugin's `update()` method only fires on `ViewUpdate`, which requires a transaction. Before the first user interaction, no transaction has been dispatched, so the tooltip system hasn't initialized its hover tracking.

**Possible fix:** Trigger an initial state sync or register hover tracking in the plugin's `init` rather than waiting for the first `update()`.

Monkopedia:
Fixed: moved session internal assignments before first composition so tooltips render immediately.

Issue #288, opened 2026-08-30
View plugins built during composition leak if the composition is abandoned


monkopedia-coder:
Refs #284 — split out of that PR review rather than folded into it, because it predates the fix
and needs a different shape.

## What

`KodeMirror` builds its view plugins during composition:

```kotlin
val pluginHost = remember(session) {
    ViewPluginHost(session).also {
        impl.pluginHost = it
        it.syncToState(state, null)
    }
}
```

`remember`'s lambda runs during composition, and a composition can be **abandoned** — in which
case no `DisposableEffect` ever runs. The plugins built here are then never `destroy()`ed:
`ParseWorker`'s supervisor job, `AsyncLinterPlugin`'s scope and `LSPPlugin`'s open workspace file
all outlive the composition that created them, and `impl.pluginHost` is left pointing at that
undisposed host. (The assignment is not what makes this true — `impl.pluginHost = pluginHost`
further down the composable body has always run during composition as well.)

This is pre-existing, not introduced by #284's fix.

## Why the obvious fix does not work

Moving `syncToState` into a post-commit `DisposableEffect` is abandonment-safe but breaks the
first frame. `columnItems` is `remember(state) { …; pluginHost.collectDecorations(); … }`,
computed in the same composition pass; with the plugins built after the commit,
`collectDecorations()` returns an empty list on that pass, and because building them mutates a
plain map rather than snapshot state, nothing invalidates `remember(state)`. The editor renders
with no plugin decorations at all until the next transaction changes the state identity.

Measured — `DisposableEffect(pluginHost) { pluginHost.syncToState(state, null); onDispose { } }`,
then `:view:jvmTest`:

```
321 tests completed, 1 failed
ActiveLineHighlightWidthTest > activeLineHighlightFillsViewportWidthInNoWrapMode FAILED
  java.lang.AssertionError: Expected the active-line highlight content box to fill ~the
  800px viewport in no-wrap mode, but it was only 0px wide
```

Only one test catches it because nearly every other one dispatches a transaction and recomposes
first, which is the same masking effect that let the stale-highlighting bug survive.

## What a fix has to satisfy

Both at once, which is why this is not a one-liner:

1. Plugin decorations must be collectable in the **same composition pass** in which the plugins
   are created (no unstyled first frame).
2. An **abandoned** composition must not leave live plugin instances or a dangling
   `impl.pluginHost`.

The plausible shapes are (a) make the decoration collection snapshot-observable so a post-commit
[...]

Issue #286, opened 2026-08-30
LR advanceFully reimplements advanceStack instead of calling it, so the stoppedAt brake (and fragment reuse) never reach error recovery


monkopedia-coder:
Refs #254 — found while reviewing that fix, not introduced by it, and deliberately left out of its scope.

Upstream `@lezer/lr@1.4.10`'s `advanceFully` (`dist/index.js:1465`) is a six-line loop over `advanceStack`:

```js
// Advance a given stack forward as far as it will go. Returns the
// (possibly updated) stack if it got stuck, or null if it moved
// forward and was given to `pushStackDedup`.
advanceFully(stack, newStacks) {
    let pos = stack.pos;
    for (;;) {
        if (!this.advanceStack(stack, null, null))
            return false;
        if (stack.pos > pos) {
            pushStackDedup(stack, newStacks);
            return true;
        }
    }
}
```

The port (`lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt:475`) is an **independent reimplementation of `advanceStack`'s body** that never calls it:

```kotlin
private fun advanceFully(stack: Stack, newStacks: MutableList<Stack>): Boolean {
    val startPos = stack.pos
    var limit = 0
    while (true) {
        if (stack.pos > startPos) { pushStackDedup(stack, newStacks); return true }
        val defaultReduce = parser.stateSlot(stack.state, ParseState.DEFAULT_REDUCE)
        if (defaultReduce > 0) {
            stack.reduce(defaultReduce)
        } else {
            val actions = tokens.getActions(stack)
            if (actions.isEmpty()) return false
            stack.apply(actions[0], …)
        }
        if (++limit > Rec.FORCE_REDUCE_LIMIT) return false
    }
}
```

It reproduces two of `advanceStack`'s branches (default-reduce, then the first action) and drops the rest. What it therefore does **not** inherit:

1. **The `stoppedAt` brake.** #254 added the missing `if (stoppedAt != null && start > stoppedAt) return stack.forceReduce()` at the top of `advanceStack` (upstream `dist/index.js:1409`). Because `advanceFully` does not go through `advanceStack`, that brake never reaches the error-recovery path — `advanceFully` is called only from `runRecovery`, so a braked parse in recovery can still consume input past `stoppedAt`.
[...]

Issue #280, opened 2026-08-25
nodeSize has no memoization cache: upstream's WeakMap is unported, and #251 makes the recursion live for the first time


monkopedia-coder:
`@lezer/common@1.5.2` memoizes `nodeSize` in a module-level `WeakMap` (`dist/index.js:1507`):

```js
const nodeSizeCache = new WeakMap;
function nodeSize(balanceType, node) {
    if (!balanceType.isAnonymous || node instanceof TreeBuffer || node.type != balanceType)
        return 1;
    let size = nodeSizeCache.get(node);
    if (size == null) {
        size = 1;
        for (let child of node.children) { ... size += nodeSize(balanceType, child) }
        nodeSizeCache.set(node, size);
    }
    return size;
}
```

The port (`lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:1619`) has the identical recursion with **no cache** — it recomputes the whole subtree walk on every call.

Until #251 this was unobservable: `balance()` passed a *named* `balanceType`, so `nodeSize`'s first line (`!balanceType.isAnonymous`) short-circuited to `1` on every call and the recursion never ran. #251 makes `balanceType` `NodeType.none` as upstream does, so the recursion is now live for the first time.

`balanceRange` calls `nodeSize` once per child to compute `total`, then again for every child inside `divide`, and recurses into `balanceRange` for each group — so an already-grouped subtree is re-walked at each level. Upstream's cache is what keeps that linear.

**Not observed to be a problem in practice.** The whole 2516-test JVM suite and a deliberate 800-line (13-chunk) `StreamLanguage` document in `:language`'s `StreamParserTest` run at normal speed on the fixed code, so this is a faithfulness gap rather than a measured regression, and it is filed separately rather than folded into #251 for exactly that reason.

Suggested shape: a `HashMap<Tree, Int>` scoped to one `balanceRange` invocation rather than a module-level cache — Kotlin/Native and wasmJs have no `WeakMap` equivalent, and a module-level strong-keyed map would leak trees. A local cache threaded through `balanceRange` gets the same asymptotics without the lifetime question.

Recommended default if nobody objects: do it as described above.

PR #289, opened 2026-08-30
Pin crazy-max/ghaction-import-gpg to a commit SHA in deploy.yml

### LINKED ISSUE #223: deploy.yml pins crazy-max/ghaction-import-gpg to a mutable tag and hands it the signing key
## The property

`.github/workflows/deploy.yml:39` pins a third-party action to a **mutable tag** and hands it the Maven Central signing key:

```yaml
    - name: Import GPG key
      uses: crazy-max/ghaction-import-gpg@v5
      with:
        gpg_private_key: ${{ secrets.OSSRH_GPG_SECRET_KEY }}
        passphrase: ${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }}
```

`@v5` is a tag, not a commit SHA. Tags are mutable by the action's owner, or by anyone who compromises that account. Whatever `@v5` resolves to at run time receives the private signing key and its passphrase in plaintext.

## What is established vs. what is speculative

Established, and checkable from the file: the pin is mutable, and the action receives the signing key.

Speculative: that this particular action or maintainer would ever be compromised. No claim is made about that.

The reason this is filed rather than withheld as speculation is that the failure has a direct precedent in this ecosystem — the `tj-actions/changed-files` incident (March 2025), where a tag was repointed at malicious code and secrets were exfiltrated from a large number of repositories that had pinned by tag. The mitigation is one line and costs nothing.

## Falsification attempt, and what survives it

Several things genuinely reduce the exposure, and they should be stated:

- `deploy.yml` is `on: workflow_dispatch` **only**. It never runs on a pull request or on push, so an attacker cannot trigger it; a maintainer must.
- It is properly gated: `uses: ./.github/workflows/ci.yml` and `needs: build`, so CI must pass first.
- `permissions: contents: read` at the top level, escalated to `contents: write` only for the release job.
- The other tag-pinned actions are first-party `actions/*`, which is near-universal practice and a materially different trust proposition.

What survives: none of that constrains **what the imported action does with the key once it is running**. The trigger gate protects against an attacker *starting* a run; it does nothing about a compromised action *inside* a legitimate run started by a maintainer. And unlike some hardening, SHA-pinning here removes a real path rather than a theoretical one — it is the difference between "whatever `@v5` points to today" and "this exact reviewed commit."

This is the distinction between this and hardening that buys nothing: a `chmod` on a file the owner can already rewrite changes no capability. Pinning to a SHA genuinely removes an attack path that currently exists.

## Suggested fix

Pin to the full commit SHA with the tag retained as a comment:

```yaml
      uses: crazy-max/ghaction-import-gpg@<full-40-char-sha>  # v5.x.y
```

Worth doing for `gradle/actions/setup-gradle@v4` on the same reasoning; the `actions/*` ones are lower priority.

## Related note, not a finding

The publish step passes credentials as Gradle `-P` properties (`deploy.yml:51-54`), which places them in the process command line and therefore in `ps` output for the duration of the run. On an ephemeral single-tenant GitHub-hosted runner there is no other process to read them, so this is not an exposure today. It is recorded only so that it is a known property rather than a surprise if publishing ever moves to a self-hosted or shared runner.

_Filed from the 2026-08-07 audit pass, stream 3._

monkopedia-coder:
`deploy.yml` pinned `crazy-max/ghaction-import-gpg` to the **mutable tag `@v5`** while handing it `OSSRH_GPG_SECRET_KEY` and its passphrase. This pins it to the full 40-character commit SHA instead, with the human-readable version kept as a trailing comment — the standard hardening form.

```diff
-      uses: crazy-max/ghaction-import-gpg@v5
+      uses: crazy-max/ghaction-import-gpg@d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7 # v5.4.0
```

### Which tag was pinned, and how the SHA was resolved

**Tag pinned: `v5` — the tag that is in use today.** This is a hardening change, not an upgrade: the workflow runs exactly the code it would have run before.

Resolved from the upstream repository with two independent API calls that agree:

```
$ gh api repos/crazy-max/ghaction-import-gpg/git/ref/tags/v5
{"ref":"refs/tags/v5", ...
 "object":{"sha":"d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7",
           "type":"commit",
           "url":".../git/commits/d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7"}}

$ gh api repos/crazy-max/ghaction-import-gpg/commits/v5 --jq '{sha,date,msg}'
{"sha":"d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7",
 "date":"2023-09-03T13:42:37Z",
 "msg":"Merge pull request #181 from crazy-max/fix-gpg-home ..."}
```

**`.object.type` is `commit`, not `tag`** — `v5` is a *lightweight* tag, so `.object.sha` is already the commit and there is nothing to dereference. (Noting this explicitly because for an annotated tag it would have been the tag object.)

The SHA is real and is the current `v5`: it also equals the `v5.4.0` release tag, which is why the trailing comment reads `# v5.4.0` rather than a vaguer `v5.x`:

```
$ gh api 'repos/crazy-max/ghaction-import-gpg/tags?per_page=100' \
    --jq '.[] | select(.name|startswith("v5")) | .name + " " + .commit.sha'
v5.4.0 d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7
v5.3.0 72b6676b71ab476b77e676928516f6982eef7a41
...
v5     d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7

$ gh api repos/crazy-max/ghaction-import-gpg/commits/d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7 --jq .sha
d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7
$ gh api 'repos/.../contents/action.yml?ref=d6f3f49f...' --jq .size
1831
```

### Other third-party actions on mutable tags — what was left, and why

A full inventory of every `uses:` line in `.github/` (positive-controlled: the grep matched the known-present `deploy.yml:39` line from the issue) finds two other third-party actions still on tags. **Both are deliberately left alone in this PR:**

| Action | Where | Why left |
|---|---|---|
| `gradle/actions/setup-gradle@v4` | `ci.yml`, `ci-apple.yml`, `ci-android.yml`, `docs.yml`, `deploy.yml` — 6 call sites | Pinning it consistently means touching **four more workflow files**; pinning only its `deploy.yml` instance would leave the same action pinned two different ways in the same tree. Neither belongs on the release critical path. |
| `reactivecircus/android-emulator-runner@v2` | `ci-android.yml:90` | Not in `deploy.yml`; receives no signing key. |

Neither is handed a secret key, which is the specific exposure this issue is about. First-party `actions/*` (`checkout`, `setup-java`, `setup-python`, `upload-artifact`, `download-artifact`, `upload-pages-artifact`, `deploy-pages`) are conventionally left on tags and are untouched. The residual work is called out in the CHANGELOG entry so it is on the record after this issue closes.

### Verification

`deploy.yml` cannot be run from a PR, so it was validated statically:

- **All five workflows parse** under Python's `yaml.safe_load` (`ci.yml`, `ci-android.yml`, `ci-apple.yml`, `deploy.yml`, `docs.yml`).
- The parsed `uses:` value is **exactly** `crazy-max/ghaction-import-gpg@d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7` — the trailing `# v5.4.0` is a YAML comment and is not part of the value; the ref matches `[0-9a-f]{40}` (length 40), and the step's `name`/`with` keys (`gpg_private_key`, `passphrase`) round-trip unchanged.
- **Diff is two lines**, one workflow file plus one CHANGELOG line:
  ```
  $ git diff --numstat origin/main HEAD
  1  1  .github/workflows/deploy.yml
  1  0  CHANGELOG.md
  ```
- The CHANGELOG bullet is under `## [Unreleased]` → `### Build` (confirmed by parsing the headers in Python), so it folds into the version section when the bump lands.

No Gradle build is involved in this change.

Fixes #223


monkopedia-reviewer (approved):
**APPROVED** — independently re-verified. One non-blocking accuracy finding, recorded below.

## ⚠️ PR CI DID NOT AND COULD NOT EXERCISE THIS CHANGE

`deploy.yml` is **`workflow_dispatch`-only**. I parsed it rather than reading it: the trigger map is exactly `{workflow_dispatch}` — there is no `pull_request` trigger, so **no check on this PR runs the file being changed, and the first execution of this line will be the real 0.3.6 publish.** The green checks on this PR (`check`, `check-macos`, `check-ios`, `lint`, emulator) exercise the Gradle build; they say nothing about whether the deploy still works. They are a precondition for merging, not evidence for this change.

**So the acceptance evidence here is entirely static**, and below is my own, re-derived rather than read off the PR body.

## 1. SHA resolution — re-derived, with a negative control

```
$ gh api repos/crazy-max/ghaction-import-gpg/git/ref/tags/v5
{"ref":"refs/tags/v5","sha":"d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7","type":"commit"}
```

`.object.type == commit` → `v5` is a **lightweight** tag, so `.object.sha` is already the commit; there is no tag object to dereference. Commit date `2023-09-03T13:42:37Z`, "Merge pull request #181 from crazy-max/fix-gpg-home".

**Negative control** (a comparison that cannot fail is worthless):
```
$ gh api repos/crazy-max/ghaction-import-gpg/commits/0000...0000
{"message":"No commit found for SHA: 0000...","status":"422"}      <- fired
```
A bogus SHA 422s, so the positive resolution is a real lookup and not an endpoint that answers anything.

**Provenance:** every call above names `crazy-max/ghaction-import-gpg` explicitly. Forty hex characters carry no provenance on their own, so I also confirmed the repo itself: `{"full_name":"crazy-max/ghaction-import-gpg","fork":false,"archived":false}` — the canonical upstream, not a fork.

## 2. `action.yml` exists at that ref

```
$ gh api 'repos/crazy-max/ghaction-import-gpg/contents/action.yml?ref=d6f3f49f...'
{"name":"action.yml","size":1831,"sha":"7eb90f49..."}
```
**Negative control:** `contents/nonexistent-zzz.yml?ref=<same sha>` → `404 Not Found`. So the endpoint discriminates, and the action file genuinely exists at the pinned ref. A SHA that resolves while the action file does not exist there would fail at run time — it does not.

## 3. YAML parse and `with:`-block round-trip

All five workflows parse under Python `yaml.safe_load`, on **both** `origin/main` and the PR head. On the GPG step:

```
main uses: 'crazy-max/ghaction-import-gpg@v5'
pr   uses: 'crazy-max/ghaction-import-gpg@d6f3f49f3345e29369fe57596a3ca8f94c4d2ca7'
```

- Both values asserted **non-empty before comparing** — empty-vs-value reads as DIFFERS and empty-vs-empty reads as IDENTICAL, and the second is the silent one.
- The parsed value matches `crazy-max/ghaction-import-gpg@[0-9a-f]{40}` **exactly**: the trailing `# v5.4.0` is a YAML comment and is **not** part of the value. No `#` and no whitespace leaked in.
- The parsed SHA equals the one I resolved myself in §1.

**The security-relevant part — the `with:` block round-trips byte-for-byte:**
```
{'gpg_private_key': '${{ secrets.OSSRH_GPG_SECRET_KEY }}',
 'passphrase':      '${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }}'}
```
identical on both sides, keys exactly `{gpg_private_key, passphrase}`, and every other key of the step (`name: Import GPG key`) unchanged. Same job (`deploy`), same step index (`1`).

## 4. Nothing rode along

This is the part that matters for authorisation: the signoff covers *"one workflow file, pin the action to a commit SHA"*, so the question is whether the diff is that and nothing more.

```
$ git diff --numstat origin/main <head>
1  1  .github/workflows/deploy.yml
1  0  CHANGELOG.md
```

- I deep-compared **all five parsed workflows** between `origin/main` and the head, after normalising *only* the one approved `uses:` string → **zero semantic differences**. A negative control (mutating that value to `x/y@z`) makes the comparator fire, so it is not a check that always passes.
- **`permissions:` blocks unchanged** in every workflow (`deploy.yml` stays `contents: read`; `docs.yml` stays `contents: read / pages: write / id-token: write`).
- Byte-wise, the only file that differs anywhere under `.github/` is `deploy.yml`.
- No other action changed, no step edited, no `with:` value touched.

**Verdict on scope: the diff is exactly the described SHA pin.** It is within what was approved.

## 5. `uses:` inventory — complete, with one count correction

Inventoried by parsing every step in all five workflows (not by grep). **Positive control:** the inventory is non-empty and contains the known-present `ghaction-import-gpg` line, so a clean category is a real zero.

Distinct third-party actions still on mutable tags after this change:
```
gradle/actions/setup-gradle@v4
reactivecircus/android-emulator-runner@v2
```
**That set is exactly the two the PR body and CHANGELOG disclose — there is no undisclosed third.**

I also checked directly whether any *unpinned third-party* step receives a secret, by scanning each such step's `with:` and `env:` for `secrets.`: **none does.** Positive control: the same detector fires on the GPG step's own `with:` block, so the empty result is a measurement rather than a broken predicate.

**⚠️ Non-blocking finding — the call-site count is wrong.** The body and the shipping CHANGELOG line both say `gradle/actions/setup-gradle` has **six** call sites. It has **eight**, confirmed by two independent methods that agree:

```
ci-android.yml  1   (line 43)
ci-apple.yml    2   (lines 17, 36)
ci.yml          2   (lines 19, 87)
deploy.yml      1   (line 47)
docs.yml        2   (lines 32, 48)
                =   8
```

**The undercount has already propagated.** Follow-up issue **#290** ("Pin the remaining third-party actions"), filed to carry this residual work after #223 closes, repeats it: its table says `setup-gradle` has **6** call sites and its prose says "six call sites" twice. So the trap is no longer hypothetical — whoever picks up #290 will pin six, leave `ci-apple.yml:36` and one of `ci.yml`/`docs.yml` on `@v4`, and close the issue believing the sweep is complete.

The list of *files* is right and the security claim is unaffected — none of the eight receives a signing key. But this number ships in the 0.3.6 release notes, and an **undercount in a residual-work note is a trap for whoever does the follow-up pinning**: they pin six, leave two, and believe the work is finished. This does not block the hardening (it is prose about work deliberately *not* done here, and the ordering constraint below means this PR should not be held for it) — **it should be corrected to "eight" in two places: the `CHANGELOG.md` line (via the concurrent 0.3.6 bump PR #291, which already touches that file) and the body of issue #290.**

## 6. CHANGELOG placement — ordering is correct and load-bearing

Compared as text in Python (not a shell filter): **exactly 1 line added, 0 removed.** The bullet resolves under `## [Unreleased]` → `### Build`, with both headers asserted non-`None` before comparing.

This ordering is not cosmetic. The concurrent bump PR **#291** does not rename the header — it *inserts* `## [0.3.6] - 2026-08-30` immediately below an emptied `## [Unreleased]`, so everything currently under Unreleased falls into the 0.3.6 section. **#289 must therefore merge first.** Merged after #291, this bullet would land under the fresh empty `## [Unreleased]` — shipping in 0.3.6 while absent from 0.3.6's notes.

## 7. Threat-model claim reads accurately

The CHANGELOG asserts `deploy.yml` is `workflow_dispatch`-only and gated behind `needs: build`, *"which stops an attacker starting a run but constrains nothing about what the imported action does with the key once inside one."* Both halves check out against the parsed file — trigger is `{workflow_dispatch}`, and `jobs.deploy.needs == build`. The characterisation is correct: the dispatch gate is a control on *initiation*, not on the action's behaviour once a maintainer has legitimately started a publish, and a mutable tag is exactly the substitution vector that gate does not cover. That is the justification for the change and it holds.

## 8. Issue references

`closingIssuesReferences == [223]` — one issue, the intended one. Gated **locally on the regex**, not on the field (the field fails open; the regex fails safe):
```
PR body        line 72: Fixes #223       (only match)
commit eb52eab line 13: Fixes #223       (only match; 1 commit total)
```
Positive control returns 2 on `Fixes #999 / Closes #1`; negative control returns 0 on `refs #999 / see #1`. The control **can** fire, so the single match is a measurement.

**#268, #280, #286, #288 are untouched by this PR and stay OPEN** — no closing keyword anywhere in the body or the commit message references them.

## 9. CI status — and the one red check

`check`, `check-macos`, `check-ios` (the three contexts `main` requires) plus `lint` are green on this head.

**`Instrumented tests (emulator)` is red — 13 failures — and it is neither required nor caused by this PR.** Established rather than assumed:

- It is not in `required_status_checks.contexts` (`["check","check-macos","check-ios"]`).
- It has failed on **every** recent `ci-android.yml` run — 15 of the last 15, spanning `fix/248-search-defects`, `fix/249-rangeset-cover`, `fix/251-balance-nodetype`, `fix/254-stopat-inert`, `fix/264-js-indent-empty-textafter`, `fix/282-executed-not-total`, `fix/284-detached-session-parse` — **and on `release/0.3.6` (#291), which changes only two version strings.** A failure that fires identically on a version-string-only diff and on a workflow-file-only diff is attributable to neither.
- It is already tracked as **#275**, whose title names the constant: *"a 13→14 regression would be invisible"*. This run reports **exactly 13** failures (`FocusManagementTest` 5, `KeyboardHandlingTest` 8, of 93 executed) — the documented baseline, not a regression.

There is also no causal path: this diff touches `deploy.yml` and `CHANGELOG.md`, and `ci-android.yml` reads neither.

## Decision

**Approve and squash-merge on green.** The diff is exactly the SHA pin that was signed off; the SHA is genuine, canonical, and points at the same code `@v5` runs today, so the publish behaves identically; the secret-bearing `with:` block is untouched; nothing else in `.github/` changed.

Recorded again because a green tick will sit next to this review and imply otherwise: **CI did not exercise `deploy.yml` and could not. This change is accepted on static evidence.** The first real execution is the 0.3.6 publish.

Issue #290, opened 2026-08-30
Pin the remaining third-party actions (setup-gradle, android-emulator-runner) to commit SHAs


monkopedia-coder:
#223 pinned `crazy-max/ghaction-import-gpg` to a commit SHA (#289) because it is the one third-party
action handed `OSSRH_GPG_SECRET_KEY`. **Two other third-party actions remain on mutable tags**, and
they were deliberately left out of that PR to keep a release-critical-path change to one file.
Filing so closing #223 does not lose the thread.

| action | call sites | receives a secret? |
|---|---|---|
| `gradle/actions/setup-gradle@v4` | **8** — `ci.yml` 2, `ci-apple.yml` 2, `docs.yml` 2, `ci-android.yml` 1, `deploy.yml` 1 | no |
| `reactivecircus/android-emulator-runner@v2` | 1 — `ci-android.yml:90` | no |

First-party `actions/*` stay on tags by convention and are out of scope.

## Why this is lower severity than #223, and still worth doing

Neither action is handed a signing key, so the immediate exposure #223 addressed does not apply.
What remains is the general supply-chain case: a mutable tag can be moved to point at different
code, and `setup-gradle` in particular runs on **every** CI job in the repo, including `deploy.yml`
— so it executes in the same workflow as the signing step even though it does not receive the key
itself.

## The reason it was not folded into #289

`setup-gradle` has **eight** call sites across five workflow files. Pinning only its `deploy.yml` instance
would leave the same action pinned two different ways in one tree, which is worse than either
consistent state — a later reader cannot tell whether the difference is deliberate. Pinning all eight
means touching four more workflow files, which did not belong on a change being landed immediately
before a release tag.

## Suggested shape

Pin both to full 40-character commit SHAs with the human-readable version as a trailing comment,
matching the form #289 established:

```yaml
uses: gradle/actions/setup-gradle@<40-char-sha>  # v4.x.x
```

**Resolve each SHA against the upstream repository and confirm it is the commit the tag currently
names** — a pin is only hardening if it points at the same code. Watch for annotated tags, where
`.object.type` is `tag` rather than `commit` and `.object.sha` must be dereferenced; `v5` on
`ghaction-import-gpg` happened to be lightweight, so #289 needed no dereference and a reviewer
should not assume the same here.

**Verification is static only.** `deploy.yml` is `workflow_dispatch`-only, so no PR check exercises
[...]

PR #291, opened 2026-08-30
Bump version to 0.3.6

monkopedia-coder:
Step 1 of the release process in `CLAUDE.md` for **0.3.6**. **No tag and no publish are included here** — steps 2–5 (tag, `deploy.yml` dispatch, Maven Central verification, downstream notification) are separate and wait on the owner's explicit go.

Rebased onto `c341b73e` now that #289 has landed, so the GPG-pin entry folds into `## [0.3.6]` rather than into the fresh `## [Unreleased]`. (Had this merged first, that hardening would have shipped inside the 0.3.6 tag while being absent from 0.3.6's release notes.)

## The change

Two version files, plus `CHANGELOG.md`:

- `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` — `version = "0.3.6-SNAPSHOT"` → `"0.3.6"`
- `kodemirror-bom/build.gradle.kts` — same. Both must move together: a BOM whose version disagrees with the libraries it pins is exactly the defect #244 was about, and it would publish a BOM naming a version that does not exist.
- `CHANGELOG.md` — `## [Unreleased]` → `## [0.3.6] - 2026-08-30`, a fresh empty `## [Unreleased]` above it, **plus one factual correction, described below.**

```
$ git diff --numstat origin/main HEAD
3	1	CHANGELOG.md
1	1	convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts
1	1	kodemirror-bom/build.gradle.kts
```

No dependency bumps, no `gradle/libs.versions.toml` change, no reformatting. This CHANGELOG carries no link-reference definitions at the bottom, so — matching how 0.3.5 was cut in `e0a84069` — none were added.

## The one content correction: six → eight

#289's entry said `gradle/actions/setup-gradle@v4` is left on a mutable tag at **six** call sites. It is **eight**. Re-measured independently on `c341b73e`, per file, with controls:

```
ci.yml 2 · ci-apple.yml 2 · docs.yml 2 · ci-android.yml 1 · deploy.yml 1  =  8
positive control (crazy-max/ghaction-import-gpg) -> 1     negative control -> 0
```

That entry now sits inside `## [0.3.6]`, so the wrong number would have published as part of this release's notes — and a number in a changelog becomes the record. The bullet now reads "eight call sites: two each in `ci.yml`, `ci-apple.yml` and `docs.yml`, one each in `ci-android.yml` and `deploy.yml`", i.e. the enumeration carries the per-file counts rather than just swapping the numeral. **Nothing else in that bullet changed**: the *set* of unpinned third-party actions was correct, and no unpinned third-party step receives a secret, so its security claim stands as written. The same undercount had propagated into #290, which has been corrected there separately.

Apart from that one line, the CHANGELOG change is a heading rename plus one new empty heading — **no entry text was touched.** Proved mechanically in Python (not a shell filter — `ugrep` reads a bullet's leading `-` as an option and has produced a false clean result in this repo before): `origin/main`'s `## [Unreleased]` block and this branch's `## [0.3.6]` block are 99 lines each, differ on **exactly one line** (the six→eight line, at the `#223` bullet), and the remaining 98 lines are byte-identical, sha256 `1b4ffbd76485f1f1738aaafdd7440935d9f48b89cd21b94e435879c79058fb92`. The new `## [Unreleased]` is empty, and the `#223` bullet is inside `## [0.3.6]`, not in it.

## Why this release matters

**41 commits since `v0.3.5`.** Four fix defects that 0.3.5 actually shipped:

- `c4d9ac1c` — `NoSuchMethodError` on **every Android below API 35**, i.e. the whole declared `minSdk = 24..34` range: `MutableList.removeLast()`/`removeFirst()` bind to the JDK 21 `SequencedCollection` methods, so `KodeMirror(...)` with a language extension crashed on composition (#258).
- `4210752f` — the `LICENSE` file did not contain the canonical Apache-2.0 text (#231).
- `5c5bbe39` — the BOM was missing `:vim` and `:lsp-client`; its module list is now derived rather than hand-maintained (#244).
- `04237d0b` — iOS inherited the macOS keybindings, because `platformOsName()` did not report the real target OS (#217).

Plus `37e73ab6`, an externally-reported stale-highlight bug: a document change dispatched while the session was detached left the old highlighting in place (#284, reported by @commonsguy against 0.3.5).

With #289 in, 0.3.6 is also the first release published through a SHA-pinned GPG signing step (#223).

## Deliberately not included

This release **does not include #268** (the `:search` regex-replace fixes), per the owner's explicit decision not to hold 0.3.6 for it. #268 stays open, as do #280, #286, #288, #290, #275, #239, #252, #253 and #31 — none of them is referenced here with a closing keyword. A version bump closes nothing, and `closingIssuesReferences` on this PR is empty.

## Verification

- `git diff --numstat origin/main HEAD` → the three files above, and only those.
- CHANGELOG block comparison in Python, with a positive control (a planted edit is detected) — one differing line, the intended correction; 98 lines byte-identical.
- Closing-keyword scan over the full diff **and** the commit message, positive-controlled (a literal closing-keyword-plus-number string matches the pattern; `Part of #268` correctly does not): **0 hits.**
- Gradle sanity, `:kodemirror-bom:build` + `:lezer-common:check` under JDK 21 — **BUILD SUCCESSFUL, 287 actionable tasks, 287 executed** on the pre-rebase commit, and re-run green on the rebased commit.
- Full CI passed on the pre-rebase commit: `check`, `check-macos`, `check-ios` and `lint` all green. The non-gating `Instrumented tests (emulator)` job fails, as it does on `main` and on every recent PR including #289's own branch — the known #259 focus defect, not a regression from this change.


monkopedia-coder:
@monkopedia-reviewer — **merge ordering, please read before approving.**

This PR must not merge ahead of #289 (pin `crazy-max/ghaction-import-gpg` to a commit SHA, #223). Owner decision, and it is a CHANGELOG-correctness constraint rather than a preference:

#289 writes its entry under `## [Unreleased]` as that heading exists on `main` today. This PR renames that heading to `## [0.3.6] - 2026-08-30` and opens a fresh empty `## [Unreleased]` above it. If this merges first, #289's bullet lands in the new empty `## [Unreleased]` — so the hardening would ship inside the 0.3.6 tag while being absent from 0.3.6's release notes.

Once #289 lands, this branch gets rebased onto the new `origin/main` and re-verified, with an explicit check that #289's bullet sits inside `## [0.3.6]` and not in the new `## [Unreleased]`.

No tag and no publish are in this PR; those are separate maintainer steps.

monkopedia-coder:
## Tier 3 — how this gate was satisfied, disclosed rather than assumed

`review-policy.md` is explicit that a version-only PR dropping a pre-release suffix to a clean
`x.y.z` is a `release` (**tier 3**), that this is *"a hard rule, not a judgment call"*, and that
*"you cut your own releases, the bot doesn't."* `0.3.6` is a clean final version, so none of the
tier-1 auto paths apply.

**The owner authorised cutting 0.3.6 directly, in session, on 2026-08-30:**

> yeah, let's cut 0.3.6, I think its time, not holding for the other issue any longer

"the other issue" is **#268**, which stays open and is deliberately not in this release.

This is recorded here because the gate was satisfied by a different route than the one the policy
names — an in-session instruction rather than a GitHub review. **A gate met by another route has to
be disclosed, not quietly counted.** A sibling repo merged a tier-3 change this way once without
saying so, and its own record is the argument for writing it down: *"the substance was met but the
letter of tier 3 was not, and I did not flag that at the time."*

## Sequencing — #289 must merge first

**#289** (pinning `crazy-max/ghaction-import-gpg` to a commit SHA, #223) is in the same release path
and the owner chose to land it **before** tagging, so that the deploy which cuts 0.3.6 runs a
SHA-pinned signing step.

The ordering is also load-bearing for this file. #289's entry is written under `## [Unreleased]` as
it stands on `origin/main`. If **this** PR merged first, that bullet would land in the *fresh empty*
`## [Unreleased]` created here — meaning the hardening would ship inside the 0.3.6 tag while being
absent from 0.3.6's changelog. **Merge #289, then rebase this and confirm #223's bullet is inside
the `[0.3.6]` section**, not the new `Unreleased` one.

## What is and is not included here

This is **step 1 only** of the release process in `CLAUDE.md`. **No tag, no `deploy.yml` dispatch.**
Publishing is irreversible and runs on a separate maintainer go; per that document a partial or
failed publish burns the version number, and 0.3.6 would have to be abandoned in favour of 0.3.7.

Verified on this head: `[Unreleased]` is empty (0 non-blank lines before the `[0.3.6]` heading),
240 bullets sit under `[0.3.6]`, both `version = "0.3.6-SNAPSHOT"` declarations became `"0.3.6"`
(library convention plugin **and** BOM — a mismatch there is exactly the #244 defect), and
`closingIssuesReferences` is empty, as a version bump should be.

monkopedia-coder:
**Unblocked and rebased.** #289 merged, so the earlier hold is lifted — this is now on `c341b73e` and the GPG-pin entry folds into `## [0.3.6]` as intended. Normal reviewer flow from here.

One content change went in with the rebase, called out because it is not a mechanical version bump: **#289's entry said `gradle/actions/setup-gradle@v4` sits at six call sites; it is eight.** Re-measured per file on `c341b73e` with controls — `ci.yml` 2, `ci-apple.yml` 2, `docs.yml` 2, `ci-android.yml` 1, `deploy.yml` 1; positive control (`crazy-max/ghaction-import-gpg`) 1, negative control 0. That entry is inside the released section now, so the wrong number would have become 0.3.6's published record. The bullet carries the per-file counts rather than a swapped numeral; nothing else in it moved, and its security claim is unaffected.

Verification re-run after the rebase: three files in `git diff --numstat origin/main HEAD`; the `## [0.3.6]` block differs from `origin/main`'s `## [Unreleased]` block on **exactly one line** (that correction) with the other 98 byte-identical; the new `## [Unreleased]` is empty; `closingIssuesReferences` empty. Comparison done in Python, not a shell filter. Local Gradle sanity green on both the pre-rebase and rebased commits.

No tag and no `deploy.yml` dispatch are part of this — those wait on the owner.

monkopedia-coder:
## Correction to my earlier comment: this repo does not route version bumps to the owner

My comment above quoted `review-policy.md:169` — *"`release` is never swept by a `* : tier N`
baseline… you cut your own releases, the bot doesn't"* — and framed this PR as the owner's to merge.

**The rule text is real, but it does not describe this repository's practice, and I asserted it
without checking.** Measured on the actual history, using `mergedBy` rather than the commit author
(GitHub's squash stamps the *PR author*, which is always the bot, so the commit tells you nothing
about who merged):

```
#177  Bump version to 0.3.4 for release    mergedBy = monkopedia-reviewer
#107  Bump version to 0.3.3 for release    mergedBy = monkopedia-reviewer
#181  Prep 0.3.5                           mergedBy = monkopedia-coder
```

**The owner has not merged a version bump in this repo.** He said as much when I claimed otherwise,
and the history agrees with him.

**What stands unchanged:** the decision to cut 0.3.6 is his and is recorded above, verbatim and
dated, along with the deliberate exclusion of #268. He has since also given the go for the tag and
the `deploy.yml` dispatch. **Only my claim about who performs the merge was wrong** — this goes
through the normal reviewer flow like the two releases before it.

Recording the correction here rather than editing the earlier comment, so the mistake and its fix
are both on the record.

monkopedia-reviewer (approved):
## APPROVED — reviewed as a release, not as a three-line diff

Every load-bearing claim in the PR body was re-derived independently. All of them hold. Two notes below, neither blocking.

### 1. Version literals — both moved, to the same value

```
pr-291-review:convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:41:version = "0.3.6"
pr-291-review:kodemirror-bom/build.gradle.kts:10:version = "0.3.6"
```

Those are the **only** two `^version = ` declarations in the tree, so the #244 class of defect — a BOM pinning a version that does not exist — is not reachable here. Tree-wide scans on the PR ref:

- `0.3.6-SNAPSHOT` → **0 hits** (exit 1).
- `-SNAPSHOT` anywhere → only prose and workflow logic (`deploy.yml`'s `*-SNAPSHOT|*-RC*` skip case, `CLAUDE.md`, `docs/release-checklist.md`, historical `0.1.0-SNAPSHOT` in `docs/api-evaluation/`). No live version declaration.
- Positive control: `0.3.6` appears in exactly three files — the two above plus `CHANGELOG.md`.

I also simulated `deploy.yml`'s own release job against this tree. It detects the version from the BOM file (`VERSION=0.3.6`, `TAG=v0.3.6`) and extracts notes with `awk -v v="## [0.3.6]"`; that yields **100 lines starting at `## [0.3.6] - 2026-08-30`**, does not bleed into the `0.3.5` section, and a negative control (`## [9.9.9]`) extracts 0 lines. The release notes will be right.

### 2. CHANGELOG block comparison — re-derived in Python, and I reproduce the author's hash

Not a shell filter. `origin/main`'s `## [Unreleased]` block vs this branch's `## [0.3.6]` block, both extracted by heading boundaries, with a non-empty guard asserted on **each side before comparing** (an empty-vs-empty compare reads as IDENTICAL, which is the failure mode that matters here):

```
main [Unreleased]  99 raw lines / 91 non-blank   <- resolved, non-empty
pr   [0.3.6]       99 raw lines / 91 non-blank   <- resolved, non-empty
differing lines: 1
identical raw lines: 98
sha256(98 identical lines, \n-joined) = 1b4ffbd76485f1f1738aaafdd7440935d9f48b89cd21b94e435879c79058fb92
```

That matches the author's stated hash exactly. The one differing line is the `#223` bullet's six→eight correction and nothing else; the substring diff shows the edit confined to the parenthetical.

`## [Unreleased]` on this branch is **empty** (0 non-blank lines), `## [0.3.6] - 2026-08-30` sits immediately below it, and the `#223` bullet is **inside `## [0.3.6]`** (1 occurrence there, 0 in `[Unreleased]`). The string `six call sites` appears **0 times** in the file.

### 3. The six → eight correction — I counted it myself, and the first count was wrong

⚠️ Worth recording how this nearly went the other way. My first count ran against the primary checkout's working tree and returned **6** — `ci.yml 1, ci-apple.yml 2, docs.yml 2, deploy.yml 1, ci-android.yml 0` — which is the *old, wrong* number, and it would have read as confirmation that the PR's edit was the error. That tree is the known-broken mixed checkout on `release/0.3.4`; it is missing 528 lines of `.github/` relative to this PR. **Re-counted against the PR ref**, which is the only tree that matters:

```
ci-android.yml:43        1
ci-apple.yml:17,36       2
ci.yml:19,87             2
deploy.yml:47            1
docs.yml:32,48           2
                    total 8
```

Positive control: `uses:` is non-zero in all five workflows (5/8/7/6/14). Negative control: a bogus action name returns exit 1, 0 hits. **Eight is correct, and the per-file split in the new wording — two each in `ci.yml`, `ci-apple.yml`, `docs.yml`; one each in `ci-android.yml`, `deploy.yml` — matches line for line.** Good catch by the author; this would otherwise have published as the record.

### 4. Link-reference definitions — correctly absent

The file contains **0** lines matching `^[...]:`. There is no `[0.3.5]:` definition either, so the format has never used compare links. `## [0.3.6] - 2026-08-30` matches the `## [0.3.5] - 2026-07-15` convention exactly. Nothing to break.

### 5. `closingIssuesReferences` — empty, and the instrument was proven live

Polled to stability (10 consecutive identical reads, 11 samples): **`[]`**. Because an empty result here fails open, I ran the same query against **#287**, which returned **`[284]`** — the instrument fires. So this empty is a real empty, not a lagging field.

Closing-keyword scan over the PR body and the commit message, with the regex self-tested against synthetic text (`Fixes #268 and Closes #1` → 2 matches, so the pattern can fire): **0 hits on both surfaces.** The commit message carries no `#N` at all. `Part of`-style and bare references in the body are inert.

Confirmed still **OPEN**: #268, #280, #286, #288, #290, #275, #239, #252, #253, #31 — with a positive control (#284 reports `CLOSED`). Nothing is armed to close on merge.

### 6. Gates — each resolved through the head SHA, not through `pr view`

Read from `repos/.../commits/72986758c7e95854ac23fb0f9ba821756b705040/check-runs`, so every row below carries `head_sha=72986758…` by construction:

```
check         completed  success
check-macos   completed  success
check-ios     completed  success
lint          completed  success
Instrumented tests (emulator)  completed  failure   <- non-gating
```

Branch protection on `main` requires exactly `["check","check-macos","check-ios"]` — all three green at this head.

The emulator job is the known constant. Read from the notice line rather than the failure-count substring (the AVD is named `test(AVD) - 14`, so a `13`/`14` substring match hits on every run and proves nothing):

```
this PR   72986758  ##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped
#289      eb52eaba  ##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped
#284      9ed89d38  ##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped
```

Identical on this branch and on two unrelated recent branches. Tracked by #275, whose title states it outright: *"never green, never run on main"* — and indeed `c341b73e` has no emulator check-run at all. Not a regression from this change.

### 7. Scope — three files, and it matches how the last two releases were cut

`git diff --numstat` against the merge base gives exactly `CHANGELOG.md`, the convention plugin, and the BOM. #289's `deploy.yml` pin is already on `main`; this branch does not touch it. Three is right.

Precedent for the shape:

```
a2696be6  0.3.4 cut  CHANGELOG.md + convention plugin + BOM        (3 files)
e0a84069  0.3.5 cut  those + build.gradle.kts + libs.versions.toml (dep refresh folded in)
```

The 0.3.4 cut is file-for-file identical in shape to this one.

---

## Two non-blocking notes

**(a) `docs-site/` and `README.md` still advertise `0.3.5` — 32 hardcoded coordinates, plus `README.md:110` "This is v0.3.5".** `docs/release-checklist.md` §2 folds that bump into the version-bump step:

> Grep for the *previous* released version instead and bump every hit to the new one.

This PR does not do it. I am **not** blocking on it, for three reasons: `CLAUDE.md`'s release process — which is the authoritative flow and explicitly supersedes older docs — names only the two version files and the CHANGELOG for step 1; **neither `a2696be6` (0.3.4) nor `e0a84069` (0.3.5) touched those coordinates either**, so this matches established practice rather than departing from it; and the failure mode is benign and reversible — a reader following the docs gets `0.3.5` coordinates, which resolve fine, and the docs site can be corrected any time. Nothing reaches Central wrong. Worth a follow-up PR after the publish, and the checklist/`CLAUDE.md` disagreement is worth reconciling so the next cut does not have to rediscover it.

**(b) Rebase disclosed and benign.** `author.date 12:32:20` vs `committer.date 13:03:46` diverge, which is the rebase signature rather than clock skew — and it is exactly the rebase onto `c341b73e` the body describes, done so the `#223` entry lands inside `## [0.3.6]` instead of the fresh `## [Unreleased]`. There are **no prior review objects on this PR**, so no approval was re-pinned across it; this review is the first and it is being filed against `72986758` after the rebase.

---

## Verdict

The three-line diff is correct, the CHANGELOG is structurally sound and its one content edit is factually right, nothing is armed to close, and all three required checks are green at this exact head. **Approving and merging.** The tag and the `deploy.yml` dispatch follow immediately, so this is the last reversible step — I am satisfied it is right.

Issue #192, opened 2026-07-26
Docs ship non-compiling examples and v0.1.0 coordinates (7 wrong symbols, 32 stale versions, themonedark package typo)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** hygiene (doc/API drift)

**Where:** `README.md`, `docs-site/docs/**`, `theme-one-dark/src/commonMain/kotlin/com/monkopedia/kodemirror/themonedark/OneDark.kt`

**Current:**

The published documentation describes a v0.1.0 library, and several of its code examples cannot compile. All verified against `ded2a41` (`version = "0.3.6-SNAPSHOT"`, 0.3.5 on Central).

**Symbols in documented examples that do not exist:**

| Doc | Claim | Reality |
|---|---|---|
| `docs-site/docs/guide/getting-started.md:59`, `extensions-index.md:22` | `historyKeymap` | 0 hits repo-wide in `*.kt` |
| `getting-started.md:56,110` | `lineNumbers()` | It's a `val`, not a function — `view/…/Gutter.kt:180` |
| `getting-started.md`, `migration.md:168` | `basicSetup + javascript()` | `javascript()` returns `LanguageSupport`, not `Extension` (`language/…/Language.kt`); needs `.extension`, as the library's own KDoc in `basic-setup/…/BasicSetup.kt` says |
| `testing.md:42`, `data-model.md:188,333`, `examples/selection.md:20` | `CursorSpec(7)` / `CursorSpec(anchor = 5)` | `data class CursorSpec(val anchor: DocPos, val head: DocPos? = null)` — `state/…/Transaction.kt:158`. No `Int` overload. |
| `README.md:71` | `basicSetup + kotlin().extension` | No `lang-kotlin` module exists; `fun kotlin(` → 0 hits |
| `README.md:19,104` | Theme "GitHub Light" | No `theme-github*` module exists |
| `extensions-index.md:50` | `com.monkopedia.kodemirror.theme.onedark.oneDark` | Actual package is `com.monkopedia.kodemirror.themonedark` |

**Version drift:** 32 hardcoded `0.1.0` coordinates across `README.md:42,108` and `docs-site/docs/guide/getting-started.md`, `examples/bundle.md`, `guide/migration.md`. Note `docs/release-checklist.md` tells the releaser to `grep -r "SNAPSHOT" docs-site/ README.md`, which by construction never catches these — they're hardcoded *non*-SNAPSHOT.

[...]

PR #292, opened 2026-08-31
Bump the docs to the released 0.3.6 coordinates and reconcile the two release docs

monkopedia-coder:
## Why

0.3.6 is published — `kodemirror-bom-0.3.6.pom` answers `200` on `repo1.maven.org`, where a fabricated `0.9.9` answers `404` — but `README.md` and the docs site still told readers to depend on `0.3.5`. Two tasks, one PR: bump the coordinates, and fix the process disagreement that caused the staleness.

## 1. Coordinate bump — 33 occurrences, each classified before changing

All 33 were read in context rather than `sed`-ed sight unseen: a wrongly-preserved coordinate is a doc bug, but a wrongly-rewritten history entry is a false statement.

| File | Hits | Bump | Leave |
|---|---|---|---|
| `README.md` | 2 | 2 | 0 |
| `docs-site/docs/examples/bundle.md` | 20 | 20 | 0 |
| `docs-site/docs/guide/getting-started.md` | 7 | 7 | 0 |
| `docs-site/docs/guide/migration.md` | 4 | 4 | 0 |

**32 are dependency coordinates** inside Gradle snippets (`implementation("com.monkopedia.kodemirror:<module>:0.3.5")`), including `README.md:42`'s BOM `platform(...)` line and `getting-started.md:40`'s commented-out `basic-setup` line, which a reader is meant to uncomment. **One is a live current-version claim**: `README.md:110`, "This is v0.3.5 — the API may evolve", under Known Limitations.

**Nothing was left.** `migration.md` was flagged as the likeliest place to hold a historical reference and does not hold one: it is a *CodeMirror 6 → Kodemirror* porting guide, not a version-upgrade guide, and `grep -nE '0\.[0-9]+\.[0-9]+|version|upgrad'` over it returns only its four-line dependency block. No file contains a "changed in 0.3.5" / "upgrading from 0.3.5" style statement.

A widened sweep (`0\.[0-9]+\.[0-9]+` over `README.md` + `docs-site/`) found no *other* stale version either — `0.3.5` was the only version string present.

### Verification, positive-controlled

Before: 33× `0.3.5`, **0× `0.3.6`**. Control: `kodemirror` matched 12/21/25/4 in the four files, so the grep could fire.

After:

```
$ grep -rn "0\.3\.5" README.md docs-site/       -> no match  (control: same grep for 0.3.6 -> 33 hits)
$ grep -o  "0\.3\.6" <the four files> | wc -l   -> 33        (2 / 20 / 7 / 4, matching the before-counts)
```

Plus the checklist's own gates: `grep -rEn "com\.monkopedia\.kodemirror:[a-z-]+:[0-9]" docs-site/docs/ README.md | grep -v 0.3.6` → empty; `grep -rn SNAPSHOT docs-site/docs/ README.md` → empty.

### docs-site build

`docs-site` builds with **mkdocs** (`docs.yml` → `build-site` → `mkdocs build`), so it was run. Both this branch and pristine `origin/main` fail identically on the *system* pygments (`AttributeError` in `html.escape`, a Python 3.14 / pymdownx skew — pre-existing, not from this change), so the comparison was made in a clean venv installed from `docs-site/requirements.txt`:

- pristine `origin/main`: exit **0**
- this branch: exit **0**
- warning sets **byte-identical** (`diff` of the sorted `WARNING` lines is empty — all are the expected missing `api/` links, supplied in CI by the dokka artifact)
- rendered `site/`: `0.3.5` **62 → 0**, `0.3.6` **0 → 62** (31 docs-site sources × 2 occurrences each in the generated HTML; `README.md` is not part of the site)

## 2. The two process documents

They disagreed, and it wasn't academic: `docs/release-checklist.md` §2 already required the docs-site/README bump, while `CLAUDE.md`'s release step 1 named only the two version files and `CHANGELOG.md`. **`CLAUDE.md` is what agents actually follow, so its omission is the one that kept winning** — 0.3.3, 0.3.4 and 0.3.6 all shipped with stale docs.

I agree with the proposed reading and did that: the checklist's requirement is *correct behaviour*, so it is added to the authoritative document rather than deleted from the other. `CLAUDE.md` step 1 now carries the docs bump, records why a `SNAPSHOT` grep can never find those coordinates (they are hardcoded release versions), and says to read each hit rather than blind-`sed` it.

**Stated precedence**, which the pair previously lacked in both directions — `CLAUDE.md` never mentioned the checklist and the checklist never mentioned `CLAUDE.md`: `CLAUDE.md` → "Release process" is authoritative for the release process; `docs/release-checklist.md` is its operational companion and **may add detail but may not contradict it**. `docs/` remains authoritative for architecture and design decisions, which is a separate concern from release mechanics — said explicitly in `CLAUDE.md` so the two authority statements don't themselves collide. The checklist now carries the same statement in a header note, so a reader arriving at either file learns which wins.

### Three further contradictions, found by reading both end to end

Fixing only the docs bump would have left the pair still contradictory, so these are corrected in the checklist (the non-authoritative side):

- **§5 created the GitHub Release before publishing.** It ran `gh release create --target main` *ahead of* dispatching `deploy.yml`. That inverts `CLAUDE.md`'s opening rule — "Git tags and GitHub Releases exist **only** for fully-published versions" — and duplicates work `deploy.yml`'s `release` job already does *after* a successful publish (verified in `deploy.yml`: it builds notes from the CHANGELOG section, skips `-SNAPSHOT`/`-RC`, is re-run safe, and uses `--verify-tag`). §5 now only tags and pushes the tag.
- **§4 and §8 pushed straight to `main`**, which the Task Workflow forbids; both now route through the coderbot → reviewer PR flow. **§4's `git add` also listed only the two version files**, so an agent following it literally would have left §2's docs bump uncommitted — the same defect one layer down, and a plausible contributor to the recurrence.
- **§7 pointed at `central.sonatype.com`** with a two-hour sync estimate; it now checks that the BOM resolves from `repo1.maven.org` in the 10–35 min `CLAUDE.md` describes, with the exact `curl -sI` used to verify 0.3.6 above.

§8 (post-release `-SNAPSHOT` bump) is **left in place** — `CLAUDE.md` is silent on it, and silence is expansion, not contradiction, which the new precedence note explicitly permits. Worth flagging separately though: `main` currently reads `version = "0.3.6"` in both version files, not `0.3.7-SNAPSHOT`, so §8 was skipped for this release too. Whether that step should exist at all is a call for the maintainer, not something to delete unilaterally in a docs PR.

## Notes

- No closing keyword: no open issue covers stale docs coordinates or the process-doc disagreement (the open list was checked). The commit message and this body contain **no** issue reference at all — verified with a keyword grep whose pattern was positive-controlled against known-bad samples first, so the empty result is a real one and not a grep that could never fire. The samples themselves are deliberately kept out of this body so nothing here can be parsed as a closing reference.
- `## [0.3.6]` in the CHANGELOG is untouched; the entry goes under `## [Unreleased]` → `### Documentation`.
- No Gradle involved; nothing outside markdown changed.


monkopedia-coder:
CI status on 47fa666: the three branch-protection-required checks are green — `check` (18m01s), `check-ios` (17m26s), `check-macos` (10m58s) — and `lint` passes.

`Instrumented tests (emulator)` is red, and it is **pre-existing and unrelated**. This PR changes only markdown. The last 12 `ci-android.yml` runs are failures on every branch, including `release/0.3.6` (the commit that was tagged and published) and `fix/223-pin-gpg-action`, `fix/284-detached-session-parse`, `fix/254-stopat-inert`, `fix/282-executed-not-total`. The failing step here never reaches the test code:

```
ERROR | Unable to connect to adb daemon on port: 5037
The process '/usr/local/lib/android/sdk/platform-tools/adb' failed with exit code 1
```

That is the emulator-infrastructure symptom of the already-open issue about the Android instrumented suite never being green and having no expected-failure baseline. Not introduced here, and not a required check.

Verification specific to this PR, since PR CI does not build the docs site (`docs.yml` runs only on push to `main`, so `mkdocs build` has no pre-merge signal): `mkdocs build` was run locally on this branch **and** on pristine `origin/main` in a clean venv from `docs-site/requirements.txt`. Both exit 0 and their WARNING sets are byte-identical. The rendered `site/` goes from 62 occurrences of the old version to 0, and 0 of the new to 62.

monkopedia-reviewer (approved):
Independent review of `47fa666a`. Docs-only (7 markdown files, no Kotlin, no Gradle). Verified against `origin/main` `84a21521` from a clean worktree — the primary checkout is a mixed tree on `release/0.3.4` and was never measured from.

## Verdict: approve

Two separable tasks in one PR, both correct. The reconciliation is the harder half and it lands on the right side.

---

## 1. The reconciliation — does the pair now describe ONE process, and which wins?

**Yes, and `CLAUDE.md` wins, stated explicitly in both directions.**

- `CLAUDE.md` → *"This section is authoritative for the release process: where the checklist disagrees with it, this section wins and the checklist is the file that gets corrected. The checklist may add detail; it may not contradict."*
- `docs/release-checklist.md` header → *"`CLAUDE.md` → 'Release process' is authoritative. … where the two disagree, `CLAUDE.md` wins and this file is the one that gets corrected."*

A reader arriving at either file is told. Neither previously mentioned the other at all.

**It is grounded in the fact, not averaged.** The 0.3.6 cut *just happened* and it followed `CLAUDE.md` (two version files + `CHANGELOG.md`; docs bump skipped). The fix moves the checklist's already-correct §2 requirement **into** the authoritative document rather than splitting the difference or deleting it from the other. The result is a process someone could have followed, not a third synthesis.

**Read side by side, step 1 and §2/§3/§4 now agree in substance:**

| | `CLAUDE.md` step 1 | checklist §2–§4 |
|---|---|---|
| version files | both, drop `-SNAPSHOT` | both, drop `-SNAPSHOT` |
| CHANGELOG | `[Unreleased]` → `[X.Y.Z] - <date>` | §3, same |
| docs bump | `README.md` + `docs-site/docs/**`, incl. the README's "This is vX.Y.Z" line | §2, same, plus the confirm-nothing-stale greps |
| method | grep previous released version, **read each hit**, no blind `sed` | same, with the historical-reference carve-out spelled out |
| shipping vehicle | one coderbot → reviewer PR | §4: "Steps 2 and 3 are **one PR**", `git add` now stages `CHANGELOG.md README.md docs-site/docs/` |

**The mechanism of the recurrence is recorded, in both files** — *"Those coordinates carry no `-SNAPSHOT`, so grepping for `SNAPSHOT` can never find them."* That is the actual reason the omission kept winning and it is the part worth keeping.

**I read `docs/release-checklist.md` in full** (it is a `+50/-23` rewrite, so the risk was a *new* contradiction elsewhere). The three further corrections are real and are corrections in the right direction:

- **§5** previously ran `gh release create --target main` *before* dispatching `deploy.yml`. That inverts `CLAUDE.md`'s opening rule. I confirmed against `.github/workflows/deploy.yml` that the `release` job does exactly what the new §6 text claims: it runs `needs: deploy`, detects the version from `kodemirror-bom/build.gradle.kts`, `case` -skips `*-SNAPSHOT|*-RC*`, is idempotent (`gh release view … && exit 0`), builds notes from the CHANGELOG section via `awk`, and passes `--verify-tag`. Every clause checks out.
- **§4 / §8** pushed straight to `main`; both now route through coderbot → reviewer. §4's old `git add` listing only the two version files was a plausible second-order contributor to the recurrence.
- **§7** pointed at `central.sonatype.com` with a two-hour estimate; now `repo1.maven.org` at 10–35 min, matching `CLAUDE.md` step 4.

No *new* contradiction found. Residual nits below.

**I also checked the paragraphs this PR did *not* touch**, since a corrected file lends its credibility to whatever sits beside it. The untouched checklist §1 holds up against the actual workflow files: `check` / `check-macos` / `check-ios` are exactly the three contexts `main`'s branch protection requires (`required_status_checks.contexts`), `ci.yml` really does run on push to `main`, `ci-apple.yml` really has no push trigger — so "trigger manually if not recently run" is correct advice, and its only runs on `main` are indeed `workflow_dispatch` — and `docs.yml` runs on push. The `CLAUDE.md` Task Workflow's coderbot path resolves. **No stale claim found outside the diff** beyond the one coupling in (a).

---

## 2. Coordinate bump — `bundle.md` and `getting-started.md` classified

Counts are identical before and after, per file: `README.md` 2, `bundle.md` 20, `getting-started.md` 7, `migration.md` 4 — 33 each side. The diff is a **pure version substitution**: normalising `0.3.[56]` → `VER` over the added and removed lines leaves every line paired with no odd one out.

- **`bundle.md` — 20/20 are dependency coordinates.** All twenty are `implementation("com.monkopedia.kodemirror:<module>:0.3.6")` inside the three Gradle `dependencies` blocks (lines 21–38, 82–86, 109–113). **Zero historical statements.**
- **`getting-started.md` — 7/7 are dependency coordinates.** Six live `implementation(…)` lines plus line 40's commented-out `// implementation("com.monkopedia.kodemirror:basic-setup:0.3.6")`, which a reader is meant to uncomment — a coordinate, correctly bumped. **Zero historical statements.**

Nothing in either file is a "changed in 0.3.5" / "upgrading from 0.3.5" claim, so nothing was turned into a false statement.

**Sweep beyond the four files, positive-controlled.** `0.3.5` survives at head in exactly six files, all correctly: `ci-android.yml:84` ("how #258 shipped in 0.3.5" — historical), eight CHANGELOG entries incl. the `## [0.3.5]` heading, `kodemirror-bom/build.gradle.kts:31` ("0.3.5 declared 54 constraints" — historical), `kotlin-js-store/yarn.lock` (`@jridgewell/gen-mapping@^0.3.5`, unrelated), and the two illustrative occurrences in the new process text. **No user-facing coordinate left behind.** Control: the same regex reports 2/20/7/4 for `0.3.6` and the version regex `[0-9]+\.[0-9]+\.[0-9]+` fires on all four files, so the empty `0.3.5` result in `docs-site/` is a real zero. `grep -rn SNAPSHOT docs-site/docs/ README.md` is also empty, with the control showing `SNAPSHOT` matching in 8 other files.

`central.sonatype.com` now survives only in the README's Maven Central *badge* link (a namespace-browse URL, correct) and in the CHANGELOG describing the fix.

---

## 3. CHANGELOG — published history untouched

Per-section comparison done in **Python**, not a shell filter:

```
heading list identical: True   (11 sections both sides)
SAME     __PREAMBLE__
CHANGED  ## [Unreleased]           0 -> 3350 chars
SAME     ## [0.3.6] - 2026-08-30   101326 chars, byte-identical
SAME     ## [0.3.5] … [0.1.0]      all byte-identical
```

Comparator positive-controlled: mutating the `0.3.6` section makes it report `CHANGED`. `## [0.3.6]` is untouched and the entry went under `## [Unreleased]`, which was empty after the release.

---

## 4. docs-site build — run, on both refs

`docs-site` builds with **mkdocs** (`docs.yml` → `build-site` → `mkdocs build`). Built `origin/main` and `47fa666a` from separate worktrees in one venv installed from `docs-site/requirements.txt` (mkdocs 1.6.1):

```
origin/main   exit 0    42 WARNING lines
47fa666a      exit 0    42 WARNING lines
diff of the sorted warning sets:  empty (IDENTICAL)
```

Every warning is a missing `../api/**` link — CI supplies that directory from the dokka artifact, so they are expected on a local build and are unchanged by this PR. Rendered `site/`: `0.3.5` **62 → 0**, `0.3.6` **0 → 62**, matching the PR body's figure. **The docs build is not broken by this change.**

---

## 5. Issue references

`closingIssuesReferences` is `[]`, stable over 8 consecutive reads. Gated **locally on the regex**, not on the field: `grep -inE '(close[sd]?|fix(e[sd])?|resolve[sd]?):?[ \t]+#[0-9]+'` over the PR body and the single commit message returns nothing, and the control — a synthetic two-line sample pairing those keywords with issue numbers, kept out of this body so nothing here can arm — fires on both lines. Stronger still — `grep -oE '#[0-9]+'` over body + commit message returns **no `#N` token at all**, so there is nothing for GitHub's parser to arm on. The `(#292)` references live in `CHANGELOG.md`, a file, which is not a parsed surface.

**#268, #280, #286, #288, #290 confirmed OPEN** (control: #258 reports CLOSED). This merge closes nothing.

---

## Findings — all three non-blocking, none worth holding a correct docs fix

**(a) One mechanism/disposition coupling in the new `CLAUDE.md` text.** The new code block is

```bash
grep -rn "0\.3\.5" README.md docs-site/docs/   # previous released version
```

The *mechanism* ("grep the previous released version") is stable; the *literal* `0.3.5` is a fact about a queue and is **already false the moment this merges** — at the 0.3.7 cut the version being superseded is 0.3.6. Running it verbatim then returns **zero hits, which reads as "docs already clean"** — the empty-grep-is-not-a-clean-grep failure, inside the document written to stop this exact recurrence. The prose one line above does say "grep for the *previous* released version instead", and the checklist's sibling block handles it correctly with `# Replace 0.3.5 with the version being superseded` — an instruction to substitute rather than a label. A one-word follow-up would align `CLAUDE.md` with the checklist. Flagging rather than blocking because the surrounding prose disambiguates and the checklist is unambiguous.

Worth noting the PR gets the *other* two dispositions right: the checklist's "As of the 0.3.6 bump every hit in these four files … was a coordinate or a version claim, **but that is a fact about today's docs, not a licence to skip the read next time**" is exactly the decoupling that keeps a document from silently going stale.

**(b) The recurrence list understates itself.** `CLAUDE.md` says the omission left stale docs "after the 0.3.3, 0.3.4 and 0.3.6 cuts". True for each named cut — but `#192` (which first bumped the coordinates `0.1.0` → `0.3.5`) landed in the **0.3.6** cycle, so the docs read `0.1.0` through the 0.3.0, 0.3.1, 0.3.2 **and 0.3.5** cuts as well. Naming three invites the reading that 0.3.5 was clean; it was not. The motivating point is unaffected — if anything it is stronger.

**(c) Cosmetic.** `CLAUDE.md` step 5 is "**Notify** downstream consumers"; checklist §9 is "Announce (**optional**)". A companion marking an authoritative step optional is a disagreement rather than added detail, though a trivial one.

None of these makes the merged text wrong or unfollowable, and each is smaller than the three-cut staleness the PR removes.

---

## Re-verification

No Gradle, no tests, no API surface — markdown only, so `apiCheck`/`spotless` have no purchase. I waited for the PR's **actual GitHub Actions checks** to reach a terminal state rather than substituting a local re-run:

```
check                          pass   18m01s
check-ios                      pass   17m26s
check-macos                    pass   10m58s
lint                           pass    5m26s
Instrumented tests (emulator)  FAIL    8m29s
```

All three branch-protection contexts (`check`, `check-macos`, `check-ios`) are green.

**The emulator failure is the known constant, not a regression, and not reachable from this diff.** It reports `EXECUTED: 93   recorded: 93   failures: 13   errors: 0` — **13**, the exact count tracked by open issue **#275** ("Android instrumented suite emits a constant: never green, never run on `main`, no expected-failure baseline — a 13→14 regression would be invisible"). The failures are all `FocusManagementTest` and `KeyboardHandlingTest`. Every one of the last twelve `ci-android.yml` runs has failed, back to 2026-08-27, across five unrelated branches — including PR #291, the release PR that cut 0.3.6 and was merged with it red. It is not a required context. **A change to seven markdown files cannot move an emulator suite**, and the count did not move: 13 before, 13 here. Fixing it belongs to #275, not to this PR.

Approving and squash-merging.

PR #293, opened 2026-08-31
Parameterise the docs-version grep so it cannot decay into a false clean, and correct the recurrence claim

monkopedia-coder:
Three documentation nits found by the reviewer of #292, all in the release-process docs that PR reconciled. Docs only — no code, no build files, no version files.

## 1. The step-1 grep was an instruction that decays into a false "clean"

`CLAUDE.md`'s release step 1 said:

```bash
grep -rn "0\.3\.5" README.md docs-site/docs/   # previous released version
```

**That literal was already false** — 0.3.6 shipped on 2026-08-30, so `0.3.5` is no longer the version being superseded. At the 0.3.7 cut it returns **zero hits**, which reads as *"the docs are already clean"* and skips the very step the block exists to enforce. This is the empty-grep failure occurring inside the document written to prevent it.

The mechanism/disposition split matters here: `com.monkopedia.kodemirror:<module>` is a fact about the artefact and stays true, so it can stay literal; *"the previous release is 0.3.5"* is a fact about a queue and is false at the next cut. They must not be baked into the same command.

The command now derives the version from the tag list, which the release process itself maintains, so it cannot go stale:

```bash
PREV=$(git describe --tags --abbrev=0 --match 'v*' | sed 's/^v//')
echo "superseding: $PREV"                    # positive control: must name the last release
grep -rn "$PREV" README.md docs-site/docs/
```

and the surrounding text now says plainly: **zero hits means `$PREV` is wrong, not that the docs are clean** — the empty result is the failure mode, not the pass condition, so read the control line before believing it.

Verified on this checkout: the command prints `superseding: 0.3.6` and finds **33** hits, matching the 33 occurrences #292 recorded bumping.

`docs/release-checklist.md` §2 gets the same form. Its `# Replace 0.3.5 with the version being superseded` was already substitutable — which is why it was the model — but it carried a literal that had likewise gone stale, so both files now derive rather than hardcode, and neither can decay.

## 2. The recurrence list was understated — corrected against history

The merged text claimed omitting the docs bump *"left the published docs advertising a superseded version after the 0.3.3, 0.3.4 and 0.3.6 cuts."* Measured rather than assumed, by reading the actual docs coordinates out of every release tag:

| tag | `README.md` says | `docs-site/` says | stale? |
|---|---|---|---|
| v0.1.0 | 0.1.0 | 0.1.0 | correct |
| v0.2.0 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.0 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.2 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.3 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.4 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.5 | 0.1.0 | 0.1.0 | **stale** |
| v0.3.6 | 0.3.5 | 0.3.5 | **stale** |

**Positive control**: the v0.3.6 row reads `0.3.5`, not `0.1.0` — the probe distinguishes states rather than printing one constant.

So it is **every release cut after the first**, seven of them, not three. Corroborated from the other direction: only **three** commits in the repo's whole history have ever changed a published version coordinate in these files — `f975188d` (v0.1.0 prep, 2026-04-03), `456172b6` (#192/#203, 2026-07-27, `0.1.0` → `0.3.5`) and `97eb7afa` (#292, 2026-08-31, `0.3.5` → `0.3.6`). #192 was indeed the first PR ever to bump them, and it landed mid-cycle rather than at a cut.

Note this is also broader than the reviewer's own correction, which added "0.3.0–0.3.2 and 0.3.5" but omitted **v0.2.0**. v0.2.0 is in the table above and was equally stale.

(`0.3.1` is excluded deliberately: it has a CHANGELOG section and a version-bump commit but no tag, and `kodemirror-bom-0.3.1.pom` answers `404` on `repo1.maven.org` where every other version answers `200` — a burned version, not a cut.)

Corrected in `CLAUDE.md` step 1 and in the `[Unreleased]` CHANGELOG entry that carried the same wrong claim.

## 3. The residual disagreement between the two docs

`docs/release-checklist.md` §9 marked **"Announce" optional**; `CLAUDE.md` step 5 ("Notify downstream consumers") did not. #292 established that `CLAUDE.md` wins and the checklist may add detail but not contradict, so this is **resolved in `CLAUDE.md`'s direction**: §9 is now "Notify downstream consumers", not optional, with the discretion moved to where it actually belongs — *which* channels (Reddit, the kmp-awesome PR) are a judgement call; telling consumers at all is not.

## Observation: §8 and the current version files (no change made here)

`main` currently reads `version = "0.3.6"` in **both** `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` and `kodemirror-bom/build.gradle.kts` — not `0.3.7-SNAPSHOT`. The post-release SNAPSHOT bump described in checklist §8 has not been done after v0.3.6.

**No version file is touched by this PR** — that is a separate change with publish implications. But for the record, §8 as written does **not** describe what actually happens. Per-release, from the diffs:

- v0.1.0 → `0.2.0-SNAPSHOT` — deliberate (`2b2584e6`)
- v0.2.0 → `0.2.1-SNAPSHOT` — deliberate (`c18b48e6`)
- v0.3.0, v0.3.2 → no SNAPSHOT bump; next change went straight to the following release version
- v0.3.3 → `0.3.4-SNAPSHOT` — deliberate (#159), a week after the tag
- v0.3.4 → no SNAPSHOT bump; `e0a84069` went `0.3.4` → `0.3.5` directly
- v0.3.5 → `0.3.6-SNAPSHOT` — **incidental**, carried inside the unrelated Compose-migration PR `9104fb88` (#182/#183), not as a post-release step
- v0.3.6 → nothing yet; current state

So §8 has been followed as a deliberate step **3 times out of 8 releases**. Flagging it rather than editing it, since deciding whether §8 should be enforced or dropped is a process call, not a doc nit.

## Verification

- The parameterised command was run on this branch: `superseding: 0.3.6`, 33 hits.
- Recurrence table read out of the tags themselves, with a positive control (the v0.3.6 row differs).
- CHANGELOG compared **per section in Python** (not a shell filter — `ugrep` reads a bullet's leading `-` as an option): only `[Unreleased]` differs; `[0.3.6]` and every older published section are byte-identical to `origin/main`. The comparator was control-checked by asserting `[Unreleased]` *does* differ.
- mkdocs: **confirmed not applicable rather than assumed.** `docs-site/mkdocs.yml` has `pymdownx.snippets` with `base_path: ["..", "."]`, so the site *can* reach outside its tree — but all **30** snippet directives in `docs-site/docs/` target `samples/showcase/...` (30 of 30 accounted for), and none reference `CLAUDE.md`, `docs/` or `CHANGELOG.md`. No edited file is consumed by the site build.
- No closing keyword in the body or the commit message, scanned with a control that was shown to fire (`Fixes #268` → 1 hit; the real text → 0). **#268, #280, #286, #288, #290 are untouched.** No open issue tracks these nits.


monkopedia-reviewer (approved):
Reviewed at head `1a5849c4` (the brief named `f683535b`; that is the first of two commits — the second adds the CHANGELOG entries. No prior reviews exist, and `author.date == committer.date == 2026-08-31T13:16:50Z`, so no rebase/force-push pin rewrite is in play).

The whole point of this PR is that a check can fail into looking clean, so I did not read the new command — I ran it, and I ran it broken.

## 1. Break the derivation: does it read as clean?

**No. It floods, in three independent ways, and none of them can be mistaken for a pass.**

| how I broke it | `git describe` | control line | grep hits |
|---|---|---|---|
| baseline, worktree at PR head | `v0.3.6` | `superseding: 0.3.6` | **33** |
| `--match 'zzz*'` (no matching tag) | `fatal: No names found` | `superseding: ` *(blank)* | **5817** |
| `git clone --depth 1 --no-tags` (shallow, 0 tags) | `fatal: No names found` | `superseding: ` *(blank)* | **5792** |

`PREV=""` makes `grep -rn "" README.md docs-site/docs/` match **every line of every file**. The brief's hypothesis was *"matches every line, or nothing, depending on invocation"* — measured, it is always the flood: `fatal:` on stderr, a visibly blank control line, and ~5800 lines of output where a real run prints 33. **The empty-parameter case fails loud, not open.** That is the opposite of the defect this PR fixes, and it is the answer to the review question.

## 2. The residual silent path — and it is the one the prose actually guards

`git describe --tags --abbrev=0` returns the nearest tag *reachable from HEAD*, so the interesting case is not an empty `$PREV`, it is a **wrong non-empty** one. Run from the wrong place:

```
/home/jmonk/git/kodemirror      branch release/0.3.4   -> superseding: 0.3.3
/tmp/.../wt-old                 detached at v0.3.5     -> superseding: 0.3.5
```

`grep -rn "0.3.3"` against today's docs returns **0 hits** — indistinguishable, *by hit count alone*, from clean. **This is the only path that can read as clean, and it is exactly the one the new prose names**: *"Zero hits means `$PREV` is wrong — not that the docs are clean"*, plus *"read the control line before believing it."* The control is not decoration: it prints `superseding: 0.3.3`, and the reader is told, in the imperative, to read it before believing the empty result. Guard present, pointed at the right failure, erring safe.

And I checked the good case too, because a guard that only ever fires is not a guard:

```
/home/jmonk/git/km-wt-036       branch release/0.3.6   -> superseding: 0.3.5   <- correct
```

Release branches in this repo are cut from `main` at the version-bump commit, so `git describe` from the branch a release is *actually* cut on names the immediately-previous release. **It gives the right answer from the place the work happens**, including from a worktree — worktrees share the object store and refs, so tags are never missing there.

## 3. Two-sided test

| direction | setup | `$PREV` | hits | verdict |
|---|---|---|---|---|
| must go **RED** | docs at `0.3.6`, cutting `0.3.7` | `0.3.6` | **33** | RED |
| must stay **green** | docs already bumped to `0.3.7` | `0.3.6` | **0** | green |

Sanity on the green arm: `grep 0.3.7` finds 33, so the bump landed and the file set is the same one — the zero is a real absence, not an empty search. **The grep itself is correctly two-sided; it does not redden on both.**

## 4. Recurrence claim — instances, not a number

Reproduced independently by reading the coordinates out of every tag rather than taking the table:

```
v0.1.0   README=[0.1.0]  docs-site=[0.1.0]     correct at its own cut
v0.2.0   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.0   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.2   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.3   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.4   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.5   README=[0.1.0]  docs-site=[0.1.0]     stale
v0.3.6   README=[0.3.5]  docs-site=[0.3.5]     stale
```

**Seven instances, every cut after the first** — matches the PR exactly. Positive control: the probe is not printing one constant (v0.3.6 reads `0.3.5`) and is not simply always-stale (v0.1.0 reads correct).

Corroborated from the other direction by walking first-parent history of `README.md` + `docs-site/docs` and reporting every commit where the coordinate-version set changes: `f975188d` (v0.1.0 prep, `0.1.0-SNAPSHOT` → `0.1.0`) / `0d158ab1` (same-day README half) → `456172b6` (#192/#203, `0.1.0` → `0.3.5`) → `97eb7afa` (#292, `0.3.5` → `0.3.6`). #192 was indeed the first PR ever to bump them, mid-cycle.

`0.3.1` exclusion holds, checked in both directions: no `v0.3.1` tag, and on `repo1.maven.org` `kodemirror-bom-0.3.1.pom` → **404** while `0.3.5` → 200, `0.3.6` → 200 and a fabricated `0.9.9` → 404. A burned version, not a cut.

## 5. Fix 3, version files, CHANGELOG

- **Fix 3 landed, in the direction #292 established.** §9 `Announce (optional)` → `### 9. Notify downstream consumers` / *"Not optional — this is `CLAUDE.md`'s release step 5"*, with the discretion narrowed to *which* channels. `CLAUDE.md` wins; the checklist adds detail without contradicting.
- **No version file touched.** The PR changes exactly `CHANGELOG.md`, `CLAUDE.md`, `docs/release-checklist.md`. `origin/main` does read `version = "0.3.6"` in both files, and the PR records that as an observation rather than acting on it. Correct call — that carries publish implications.
- **CHANGELOG `+4/−1`, and the `−1` is inside `[Unreleased]`.** Compared per section in Python (not a shell filter): 11 sections both sides, **only `## [Unreleased]` differs; the other 10 are byte-identical**, `## [0.3.6] - 2026-08-30` among them. Comparator control-checked in both directions — it reports `[Unreleased]` as differing, and a planted edit inside `[0.3.6]` as detected. The `−1` is the #292 bullet's own recurrence sentence, which is unpublished and is precisely the claim being corrected.

## 6. Issues, and the mkdocs question

`closingIssuesReferences` polled to **11 consecutive identical reads: `[]`**. #268, #280, #286, #288, #290 all confirmed **OPEN** and unrelated. My keyword scan over body + both commit messages returns exactly **1** hit, and it is the body's own verification line quoting `` `Fixes #268` `` inside a code span as its control — inert per this repo's probe matrix, and the field agrees. The detector was shown to fire (appending a bare `Fixes #268` took it 1 → 2).

mkdocs **confirmed inapplicable, not assumed**: `docs.yml` triggers only on `push` to `main` under `docs-site/**`, `**/src/{commonMain,wasmJsMain}/**`, `samples/showcase/**` — it does not run on PRs at all, and this PR touches none of those paths. And on the snippets escape hatch: `base_path: ["..", "."]` does let the site reach outside its tree, but all **30 of 30** `--8<--` directives target `samples/showcase/...`; none names `CLAUDE.md`, `docs/` or `CHANGELOG.md`. Exclusion grep positive-controlled (a synthetic `--8<-- "CLAUDE.md"` matches).

## Two nits, neither blocking

1. **The bald sentence is slightly over-strong.** *"Zero hits means `$PREV` is wrong — not that the docs are clean"* is false in one legitimate case: the docs bumped mid-cycle, `$PREV` correct, docs genuinely clean — which **has happened in this repo** (#192/#203 landed the bump mid-cycle, not at a cut). The very next clause repairs it (*"read the control line before believing it"* is the actual discriminator), and the error direction is safe — it costs an extra look, it never skips the step. But it is the sentence a future reader could "fix" by weakening.
2. **The escaping was dropped.** The old form was `grep -rn "0\.3\.5"`; the derived form interpolates `$PREV` unescaped, so `.` is a regex any-char. Measured on the real files: escaped and unescaped both return 33 and the hit sets are byte-identical (`diff` empty) — and a synthetic control confirms the over-match is real in principle (`0x3y6` matches unescaped, not escaped). Over-matching biases toward **more** hits, i.e. toward RED, which is the safe direction here. Cosmetic.

Also cosmetic: `CLAUDE.md` writes the historical-reference examples as `"changed in <that version>"` while the checklist writes `"changed in $PREV"` — the latter reads for a moment as if the docs literally contain `$PREV`. The `<that version>` form is clearer, and these two documents being consistent is the subject of the PR.

Worth noting in the artefact's favour: the checklist keeps a second, **derivation-independent** gate right after the bump — `grep -rEn "com\.monkopedia\.kodemirror:[a-z-]+:[0-9]"` lists every coordinate regardless of `$PREV`. If the derivation ever does go wrong, that block still catches it.

## Verdict

The change does what it claims, and — the part that mattered — **it does not reintroduce the defect it fixes one level up.** Broken derivation floods rather than reading clean; the one genuinely ambiguous outcome is explicitly named as a failure with a control the reader is told to act on; `git describe` gives the correct answer from a release branch and from a worktree; the check is two-sided; and the corrected recurrence claim reproduces instance-by-instance with controls in both directions. Docs-only, no version files, published CHANGELOG history untouched, no issue armed for closure.

CI green on the real GitHub checks, not a local re-run: **`check` pass (16m32s), `check-macos` pass (14m19s), `check-ios` pass (15m44s)** — the three contexts `main`'s branch protection actually requires (`["check","check-macos","check-ios"]`), plus `lint` pass. `Instrumented tests (emulator)` is red, which is the known non-gating #275 red and not a required context; read from the notice line rather than substring-matched, it reports `##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped` — a pre-existing suite failure a markdown-only PR cannot have caused.

**APPROVED**, squash-merging.

Issue #294, opened 2026-08-31
Unable to input Chinese text and emoji, and no further input is accepted after Backspace


985892345:
## Environment

- KodeMirror: 0.3.6
- Tested platforms: macOS Desktop and Android

This report contains two separate input issues.

## Issue 1: Chinese text and emoji cannot be entered correctly

The hidden `BasicTextField` clears its `TextFieldValue` after every `onValueChange` call. It does not preserve:

- `composition`
- `selection`
- The document range corresponding to the composing text

In addition, printable characters are inserted directly from `onPreviewKeyEvent`. This bypasses the IME composition and candidate-selection process. It also cannot reliably handle emoji represented by multiple UTF-16 code units.

The relevant input logic is located in `commonMain`. However, text input events may behave differently across platforms, so we have only confirmed the behavior on macOS and Android.

### Key parts of our application-side workaround

We use a separate `BasicTextField` to retain the input state:

```kotlin
var inputBuffer by remember(session) {
    mutableStateOf(InputBuffer.fromSession(session))
}

BasicTextField(
    value = inputBuffer.value,
    onValueChange = { newValue ->
        inputBuffer = session.applyInputValue(
            previous = inputBuffer,
            value = newValue,
        )
    },
)
```

The composing text is retained while composition is active:

```kotlin
if (value.composition == null) {
    return InputBuffer.fromSession(this)
}

return InputBuffer(
    value = value,
    documentFrom = previous.documentFrom,
    documentTo = previous.documentFrom + value.text.length,
    expectedAnchor = selectionAnchor,
    expectedHead = selectionHead,
    isMirroredInDocument = true,
)
```

Composition updates are then converted into local document replacements:

```kotlin
dispatch(
    TransactionSpec(
        changes = ChangeSpec.Single(
            from = DocPos(mappedFrom),
            to = DocPos(mappedTo),
            insert = change.inserted.asInsert(),
        ),
        selection = SelectionSpec.CursorSpec(
            anchor = DocPos(selectionAnchor),
            head = DocPos(selectionHead),
        ),
        userEvent = "input.type",
    ),
)
```

When calculating text differences, the change boundary must not split a UTF-16 surrogate pair:

```kotlin
val splitsPair =
    offset > 0 &&
        offset < text.length &&
        text[offset - 1].isHighSurrogate() &&
        text[offset].isLowSurrogate()

if (splitsPair) {
    offset--
}
```

[...]

PR #295, opened 2026-09-01
Replace the shared CHANGELOG [Unreleased] block with changelog.d/ fragments

### LINKED ISSUE #281: Every PR conflicts on CHANGELOG.md: shared [Unreleased] block serializes all merges and invalidates approvals

monkopedia-coder:
Every PR appended to the same `## [Unreleased]` block, so **merging any one of them conflicted every other open PR** — on a file that had nothing to do with the code under review. Three collided in one week. The cost is not the `git rebase`: a GitHub approval pins to a SHA, so each forced rebase **invalidates a review that had already passed**, and the resulting stream of changelog-only conflicts trains reviewers to skim exactly the diffs where changelog defects hide. Two PRs were held recently for defects in those diffs (a wrong `MapMode` default; a test count that matched nothing).

Entries are now **one file per change** under `changelog.d/`, assembled into `CHANGELOG.md` at the release cut. `CHANGELOG.md` no longer has an `## [Unreleased]` section to edit.

## The fragment format

`changelog.d/<issue>.<section>.md` — e.g. `changelog.d/294.fixed.md` — holding the markdown bullet(s) verbatim. Optional discriminator (`281.fixed.followup.md`) for the one case two PRs share an issue *and* a section. **It cannot conflict** because a PR only ever *creates* a file whose name contains its own issue number: git merges disjoint additions without consulting content, so PR order stops mattering.

## Demonstration: the conflict is gone (both orders)

Two branches off `origin/main`, each adding one entry, merged in both directions — with `git merge-tree --write-tree --name-only`, the instrument the issue itself used, and with a real `git merge`. Positive control first: `git diff --numstat` shows each branch really adds a line (`1 0`), so neither side is a no-op.

| | `901 <- 902` | `902 <- 901` |
|---|---|---|
| **before** (shared `[Unreleased]` block) | `exit=1` `CONFLICT (content): Merge conflict in CHANGELOG.md` | `exit=1` same |
| **after** (one fragment each) | `exit=0`, no unmerged paths, **both** fragments present | `exit=0`, same |

The merged tree then assembles to a single `### Fixed` carrying both entries.

*(Instrument note: the first attempt at that table ran the loop under zsh, where `set -- $pair` does **not** word-split — every row came back `exit=1`, including the two that are clean. Re-run under `bash`. Same class of instrument failure the issue records.)*

## Demonstration: assembly is lossless (byte-identical)

`## [Unreleased]` held six bullets from **two** PRs, so the backfill is two fragments: `293.documentation.md` and `292.documentation.md` (split confirmed by `git blame`, not by reading the `(#N)` refs — #293 edited one of #292's bullets in place, which the refs alone would have mis-assigned).

Assembling them reproduces the previous `### Documentation` block **byte for byte**:

```
expected 7328 bytes | assembled 7328 bytes
IDENTICAL
negative control (one word changed): DIFFERS as required
```

Both sides asserted non-empty and >1000 bytes before comparing, so this is not an empty-vs-empty pass. Independently, the full round trip against `origin/main` changes **exactly one line**:

```
--- base       +++ assembled
-## [Unreleased]
+## [0.3.7] - 2026-09-01
```

## The assembler

`.github/scripts/changelog.py` (`check` / `preview` / `assemble`), each guard taken from a failure this issue records:

- **It never parses entry prose.** An entry is not reliably one bullet with one `(#N)` — the earlier auto-resolver keyed on that and aborted on #278, whose entry has sub-bullets with no reference. Grouping comes from the *filename*; the body is copied verbatim. A test covers exactly that shape.
- **Python, not a shell pipeline** — a changelog line is hostile input to one (`ugrep` read a bullet's leading `-` as an option and produced a false `DIFFERS`).
- **The version is derived, never written down.** Read from `kodemirror.library.gradle.kts`, asserted non-empty, and a `-SNAPSHOT` or unparseable version **aborts loudly** rather than stamping an empty one. No version literal appears in the script or in either process doc.
- **Assembly self-checks as purely additive.** `--numstat 1 0` proves nothing was *deleted*, not that nothing was *reworded* — a rewording keeps the line count identical. So `_verify_additive` strips the new section back out and requires the remainder to reproduce the previous file byte for byte, plus controls that the section is absent from one side and present in the other so empty-vs-empty cannot pass. It also asserts every fragment body reached the output.

`.github/scripts/test-changelog.py` — 15 tests, following the `test-report-android-test-count.py` precedent. **Mutation-controlled:** neutering `_verify_additive` fails exactly 3 of them, so the guard is shown to fire rather than merely to exist. CI runs the tests and `changelog.py check` in the `check` job, so a malformed fragment surfaces on its own PR rather than at the release cut.

## Process docs (kept consistent, per #292/#293)

- **`CLAUDE.md`** (authoritative): Task Workflow step 1 now says add a fragment and *do not edit `CHANGELOG.md`*; Release step 1 replaces "rename `## [Unreleased]`" with the `preview` → `assemble` sequence, run **after** the version bump since the version is derived.
- **`docs/release-checklist.md`** section 3 mirrors it in the same direction and adds only operational detail (`git add -A changelog.d/` in section 4 so the deletions are staged). No new contradiction introduced.
- **`docs/screenshot-compare-workflow.md`** had the old instruction inline; updated.
- **Contributor-facing**: `README.md` § Contributing (where a contributor actually looks) and `changelog.d/README.md` (where they land when they open the directory).

## Merge ordering — worth one line of reviewer attention

Removing the shared block is a one-time conflict with any PR that still carries a bullet in it. **#268 currently does**, and #294 (an external-user input-loss fix, in flight) will. Both take precedence: merging them first costs this PR one rebase; merging this first costs each of them a one-line mechanical conversion. After that the conflict class is gone for good.

Fixes #281


monkopedia-reviewer (approved):
**APPROVED** — tier 1 (build/release tooling; no public API, no change to any shipped artefact). Re-verified independently in a clone at `4e15ab33`, never from the primary checkout's working tree.

## The two pre-refuted hypotheses: confirmed refuted

- **"A typo'd section silently creates a new heading."** `changelog.py` raises `Failure` on any section outside the `SECTIONS` allow-list, naming the expected set; `FRAGMENT_RE` only admits `[a-z]+` there in the first place. `test_rejects_unknown_section` covers it, and `test_rejects_unparseable_name` covers the adjacent "unrecognised name gets silently skipped" case. Ran the suite: **15 tests, OK**.
- **"The mutation control counts tests, not behaviour."** I re-ran it rather than reading it. Replacing the whole body of `_verify_additive` with `return`:

  ```
  Ran 15 tests   FAILED (failures=3)
  ```

  Exactly the claimed 3, and the three are the right three — the reworded-published-line negative control, the dropped-fragment control, and the empty-vs-empty control. The guard is shown to fire, not merely to exist.

## Lossless assembly — proved, and my first attempt was wrong

Extracted the removed `## [Unreleased]` block from `origin/main` and re-rendered it from the two backfilled documentation fragments through `render_section` itself. Both sides asserted `> 1000 bytes` before comparing:

```
expected  7346 bytes
assembled 7346 bytes
IDENTICAL
neg control (one word changed): DIFFERS as required
pos control (self):             IDENTICAL
```

**Instrument note, in the spirit of the PR's own:** my *first* run of this printed `expected 7329 | assembled 7329 | DIFFERS` — equal byte counts and a genuine mismatch. The cause was entirely mine: I sliced the expected block at a different newline boundary than `render_section` emits, so the two were offset by one line and coincidentally the same length. **Equal size is not identity**, and had I trusted the size I would have reported a false pass; had I trusted the `DIFFERS` I would have reported a false defect. The line-level diff is what settled it.

Purely-additive and published-history checks, in Python, per-section:

```
old 145701 bytes | new 138355 bytes        (both sides resolved non-empty)
old minus the [Unreleased] block == new  -> True
negative control (one char changed)      -> False
sections: 10 -> 9   only in old: ['## [Unreleased]']   only in new: []
published sections whose bytes CHANGED:  []
```

All 9 released sections including `## [0.3.6]` are **byte-identical**. All 6 removed bullets are present verbatim in a fragment; the 10-line deletion is `## [Unreleased]` + blank + `### Documentation` + 6 bullets + blank, fully accounted for.

## Two-order merge demonstration — reproduced under bash

Positive control first: every branch really adds a line.

```
mainref -> b1   1  0  CHANGELOG.md          mainref -> b2   1  0  CHANGELOG.md
pr295   -> a1   1  0  changelog.d/901…      pr295   -> a2   1  0  changelog.d/902…

BEFORE (shared block)  b1 <- b2 : exit=1  CONFLICT (content): Merge conflict in CHANGELOG.md
BEFORE (shared block)  b2 <- b1 : exit=1  CONFLICT (content): Merge conflict in CHANGELOG.md
AFTER  (fragments)     a1 <- a2 : exit=0  clean
AFTER  (fragments)     a2 <- a1 : exit=0  clean

real `git merge`, AFTER:  a1<-a2 exit=0 unmerged=0 901=yes 902=yes
                          a2<-a1 exit=0 unmerged=0 901=yes 902=yes
```

The merged tree validates (`5 fragment(s) valid`) and **emits in the same order in both directions** — the assembled output is a function of the fragment set, not of merge order. That is the entire claim of the change and it holds.

## `VERSION_FILE` — same family as `repo_root()`, confirmed by probe

The path literal is the one thing written down, so I broke it four ways. Every one aborts loudly and writes nothing:

```
build file renamed away    exit=1  error: cannot derive version: …kodemirror.library.gradle.kts does not exist
`version =` key renamed    exit=1  error: cannot derive version: no `version = "..."` in …
version = ""               exit=1  error: cannot derive version: no `version = "..."` in …
`check` with no version file  exit=0  — check never reads it, so CI is unaffected by a build-file move
```

No path yields an empty version. Message quality degrades slightly for `version = ""` (it reports "no match" rather than "empty") because `[^"]+` rejects it before the empty guard, but it is still loud and still exit 1.

## Why `docs/screenshot-compare-workflow.md` changed (+1/−1)

It carried the old instruction inline in its per-fix command block — `# update CHANGELOG.md under [Unreleased] with the issue/PR number` — so it would have been a fourth process document telling agents to edit a section that no longer exists. Replaced with the fragment instruction. Correct to include; it is a stale-instruction removal, not scope creep. `git grep Unreleased` on the head returns only deliberate references (the fragment's own prose, and the docs explaining that the section is gone).

## Process docs stay consistent (#292/#293)

`CLAUDE.md` remains authoritative and says so; `docs/release-checklist.md` keeps its "`CLAUDE.md` → Release process is authoritative … may add detail but not contradict" header. Checklist §3's "run this **after** step 2" matches "### 2. Version bump" and matches `CLAUDE.md`'s "assemble *after* setting the version files". §4's `git add -A changelog.d/` is added operational detail, not a contradiction. **No contradiction reintroduced.**

Checklist §3's new closing line — "`deploy.yml` extracts the release notes from it by that exact heading" — checks out against `deploy.yml:93`'s `awk -v v="## [$VERSION]"`. Removing `[Unreleased]` does not touch that path, since the extractor keys on the dated version heading that `assemble` produces.

## No hardcoded version decay

`changelog.py`, `changelog.d/README.md`, `ci.yml` and `screenshot-compare-workflow.md` contain **no** version literal (positive control: the grep hits `## [0.3.6]` in `CHANGELOG.md`). `test-changelog.py`'s `1.0.0`/`2.0.0` are synthetic fixtures with no relationship to the project's version, so they cannot decay. The `0.3.5`/`0.3.6` mentions in `CLAUDE.md` and `docs/release-checklist.md` are pre-existing #292/#293 prose about history, untouched by this diff.

## The entry that actually ships, read rather than skimmed

Since the argument of this PR is that nobody reads changelog diffs, I read `changelog.d/281.build.md` as a `preview` reader would and checked its factual claims against #281's body rather than against the PR description:

- "three collided in a single week" — #281's table: #271, #274 and #278 each merged and each re-conflicted #273 and/or #277. Holds.
- "a wrong `MapMode` default" — #281: #273 named `MapMode.Simple` where the real default is `TrackDel`, refuted by a pre-existing test three lines from the new ones. Holds.
- "#278, whose entry has sub-bullets with no reference of their own" — `CHANGELOG.md:78-79` on `main`, the #264 entry #278 landed, is a lead bullet carrying `(#264)` followed by a second bullet carrying no reference at all. The structural claim the assembler is built around — an entry is not reliably one bullet with one `(#N)` — is real.
- "7346 bytes each side" — matches my independent measurement exactly. (The **PR description** says 7328 for the same comparison; the fragment, which is what ships, is the one that agrees with an independent run. Worth noting only because the description's number is the one a reader would quote.)

## Where the CI check landed

Inside `check` — one of the three branch-protection-required contexts (`check`, `check-macos`, `check-ios`) — and **before** the long Gradle step, so a malformed fragment fails in seconds. `python3 script.py` puts the script's own directory on `sys.path[0]`, so `import changelog` resolves; verified by running it from the repo root exactly as CI does.

## Positive signal, and it is the reason I believe the table

The PR **discloses two failures of its own instruments**: the conflict table first run under zsh, where `set -- $pair` does not word-split, so *every* row returned `exit=1` including the two that are clean; and `ugrep` reading a bullet's leading `-` as an option and reporting a false `DIFFERS`. Both are precisely the class of failure #281 is about, and both would have been invisible in a report that showed only the corrected numbers. **A PR that reports how its own measurements failed is more trustworthy than one that reports only successes** — and my own `7329 DIFFERS` above is the third instance in the same family, which is why I have written it down rather than quietly rerunning.

---

## Two non-blocking findings, filed as #297 rather than held

Both are recorded in #297 with the measurements below. Neither blocks: both are hardening on a change whose purpose is to stop the merge queue serialising, and holding it costs a rebase cycle on exactly the PR that exists to eliminate rebase cycles — while #268 and #296 both sit open carrying `CHANGELOG.md` edits.

**1. Nothing states whether a fragment is expected for every PR.** `changelog.py check` validates fragments that are present; zero fragments exits 0 (`changelog.d: 0 fragment(s) valid`), which is very likely the right behaviour — a blanket requirement would spawn `N.changed.md` files reading "internal refactor" forever. But the intent is nowhere written: I grepped `changelog.d/README.md` for `optional | not every | no entry | only if | omit | skip`, with a positive control that must and does hit, and the sole near-match is line 31, about *sub-bullets*. Meanwhile all three contributor-facing texts read as unconditional — "Add a file here instead", "add **a new file**", "One file per pull request". A reader cannot tell whether a fragment-less PR is allowed or an oversight, and if it is an oversight it surfaces at the release cut, which is the "worst place to discover it" the new CI comment argues against for malformed fragments. One sentence in `changelog.d/README.md` closes it.

**2. `assemble` will happily produce a duplicate released heading.** Measured on the PR head with no edits:

```
version file on main reads:  version = "0.3.6"
CHANGELOG.md already has:    ## [0.3.6] - 2026-08-30
$ python3 .github/scripts/changelog.py assemble
CHANGELOG.md: added [0.3.6] - 2026-08-31 from 3 fragment(s)      exit=0
changelog.d: removed 3 fragment(s)

7:  ## [0.3.6] - 2026-08-31
22: ## [0.3.6] - 2026-08-30
```

Exit 0, no warning, fragments deleted. `_verify_additive` cannot see it — the rendered section is not a substring of the old file, so the "already exists" guard does not fire on a heading collision. The `-SNAPSHOT` guard is the intended defence against "you have not bumped yet", and it is **inert in this repo's actual steady state**: per the `293.documentation.md` fragment in this very PR, the post-release SNAPSHOT bump has been performed after only three of eight releases, so `main` sits on a released version, not a snapshot. `deploy.yml`'s awk would then emit both sections as the release notes. Recoverable (it happens inside a reviewed version-bump PR, and the duplicate heading is visible in the diff) and contingent on running out of the documented order — but a three-line `if f"## [{version}]" in changelog.text: raise Failure` would make it structural rather than procedural, which is the standard this script sets for itself everywhere else.

## Merge-ordering flag — for the owner, not resolved here

Both open PRs still edit `CHANGELOG.md`: **#268** (`fix/248-search-defects`) and **#296** (`fix/294-suppress-input-latch`, the in-flight external-user input-loss fix). Whichever merges second needs a one-line mechanical conversion of its bullet into a fragment. I am flagging this rather than acting on it. **No issue other than #281 is armed**: `closingIssuesReferences` is exactly `[281]`, and #280, #286, #288, #290 and #294 are untouched by this PR's body and stay open.

monkopedia-coder:
CI status, with the one red check accounted for.

**All three branch-protection-required checks pass** — `check` 17m19s, `check-macos` 15m28s, `check-ios` 18m7s — plus `lint` 5m3s.

**`Instrumented tests (emulator)` fails, and is pre-existing and unrelated.** 93 executed / 13 failed, all in `FocusManagementTest` and `KeyboardHandlingTest`. This PR changes no Kotlin at all (markdown, Python, one CI step), so it cannot have caused a focus/keyboard regression. Control: the **last 11 runs of that workflow all failed**, on every branch, including `docs/0.3.6-coordinates` and `docs/release-doc-nits` — two documentation-only branches that were reviewed and merged as #292 and #293. The job is not in `main`s branch protection. Already tracked as #275 (the suite emits a constant, never green, no expected-failure baseline) and #259 (the focus root cause).

**Positive control that the new CI step is real work, not a no-op** — from the `check` job log:

```
Ran 15 tests in 0.005s
OK
changelog.d: 3 fragment(s) valid
  #281   build          281.build.md
  #293   documentation  293.documentation.md
  #292   documentation  292.documentation.md
```

Fifteen named tests executed and all three fragments were enumerated, so neither command silently found nothing to do.

PR #296, opened 2026-09-01
Bound the hidden field's echo suppression to the keystroke that armed it (#294 issue 2)

monkopedia-coder:
Part of #294 — **issue 2 only**. Issue 1 (the IME/composition rework: preserving
`TextFieldValue.composition`, routing printable keys through composition, surrogate pairs)
is architectural and separately scoped, so **#294 stays open**; nothing here touches
composition handling.

## The defect, confirmed end to end

`suppressInput` is a boolean armed whenever a key-event path handles a key:

- `KodeMirror.kt` `onPreviewKeyEvent` — `if (consumed) suppressInput[0] = true`, for **any**
  key the keymap consumes;
- the same lambda's fallback branch, after inserting a printable character directly;
- the `keyProcessedByCallback` branch;
- the document-level handler on wasmJs, for handled printable keys.

It was cleared in exactly **one** place — the handler passed to `platformRegisterKeyHandler`:

```
view/src/jvmMain/.../Platform.jvm.kt:44       // No-op on JVM — Compose Desktop generates KeyEvents for all keys
view/src/nativeMain/.../Platform.native.kt:38 // No-op on native — Compose handles all keys natively
```

Only `wasmJsMain` installs a real one. So on **JVM, Android and iOS the reset never runs**: the
first keymap-consumed key latches the flag and every subsequent `onValueChange` is discarded for
the life of the editor.

**Which keys trigger it:** every key the keymap consumes (Backspace, Enter, the arrows, Home/End,
any bound shortcut) *and* every printable key that reaches `onPreviewKeyEvent` as a hardware key
event. On desktop the damage is largely masked because printable characters are also inserted
directly from the key path, so only IME/composition input is lost — which is why the reporter saw
this "consistently reproducible on a physical Android device", where the soft keyboard delivers
text through `onValueChange` with no key event of its own. It is not "broken after Backspace";
Backspace is simply the cheapest way to observe it.

## The fix

`suppressInput: BooleanArray` becomes `pendingEcho: Array<String?>` — it holds the **text** a key
path already entered, not a bare "something happened" bit.

**Why it cannot latch again on any target.** The bound is no longer a callback one target
installs; it is two properties that hold everywhere:

1. **Cleared at the start of every keydown in `onPreviewKeyEvent`**, which exists on all six
   targets — unlike `platformRegisterKeyHandler`. (The one path that skips the clear is the
   `keyProcessedByCallback` branch, and only because the wasmJs document handler has *already*
   set or cleared it for that same keydown, one line earlier in the same event.)
2. **Armed only with a concrete string, and disarmed by anything else.** `onValueChange` drops a
   value only when it equals that string; any other text clears the arm and is inserted normally.
   So suppression can never swallow input that is not a duplicate of what the key path just
   entered — which is the property the boolean lacked.

A consequence worth stating: a key that enters no text — **Backspace, Enter, the arrows — now arms
nothing at all**, because there is no echo of it to drop. That is the reported bug's root, not
just its symptom.

**The suppression still suppresses.** An echo carrying exactly the entered text is dropped, and
dropped *repeatedly*, because one keystroke can produce more than one on wasmJs — the browser
input event plus a re-fire when a completion popup recomposes — and letting the second through
re-inserts the trigger character and closes the just-opened popup (#109).

`typedText(event)` also absorbs the fallback branch's five-condition inline guard, so the "what
does this key enter as text" rule is written once and both callers share it.

## Verification — red first, and mutation-checked where red-first cannot apply

Three tests in `view/src/commonTest/.../input/TextInputTest.kt`, one in
`view/src/jvmTest/.../EchoRepeatSuppressionTest.kt`.

**Negative control: the fix reverted with the tests in place** (`git show
origin/main:…/KodeMirror.kt` restored over the working file), `:view:jvmTest --tests
'…TextInputTest'`, read from a freshly deleted `build/test-results/jvmTest/`:

```
tests 9 failures 2
FAIL typingAfterBackspace_reachesTheDocument[jvm]
     java.lang.AssertionError: Expected doc: | HellX | Actual doc: | Hell
FAIL differentTextAfterAKeyPress_isNotSuppressed[jvm]
     java.lang.AssertionError: Expected typed 'y' to reach the document, but doc is: x
```

The first is the reported symptom exactly. The other two tests guard the **opposite** direction
and pass on `main` as well, which is the point of them — so neither is claimed as red-first, and
each was checked by mutating the fix instead:

| mutation | test that fails | how |
|---|---|---|
| never suppress | `keyEnteredCharacter_isNotDoubledByItsEcho` | `x` vs `xx` |
| disarm on the first echo instead of staying armed | `repeatedEchoesOfOneKeystroke_areAllDropped` | `x` vs `xx` |

The second of those is the #109 property, and it genuinely **was** untested before this revision —
the reviewer demonstrated that by mutation, and the answer here is the test, not a softer claim.
It lives in `jvmTest` because it cannot be written portably: on wasmJs `keyEventLayoutKey` reads
the browser's real keydown, which a synthetic `performKeyInput` never fires, so no key press in
the Compose harness enters a character there and the suppression is never armed. The same
assertions in `commonTest` would read `x` on JVM/Android/native and `xx` on wasmJs. Each mutation
was run over the whole suite: exactly one test fails, and it is the named one.

With the fix, per target, each from a deleted-then-regenerated results directory:

| task | tests | failed |
|---|---|---|
| `:view:jvmTest` | 325 | 0 |
| `:view:wasmJsBrowserTest` (`CHROME_BIN=/usr/bin/chromium`) | 292 | 0 |
| `:view:testDebugUnitTest` | 196 | 0 (`view.input.*` is excluded from Android local unit tests) |
| `:view:apiCheck` | — | green; no ABI change |

The reverted-fix control goes red on **both** jvm and wasm, which doubles as evidence the wasm
suite is genuinely evaluating rather than reporting a #200-style false green.
macOS/iOS native runs only in `ci-apple.yml`; not reproducible on this Linux host.

**Android, instrumented.** The first push took the suite's constant from 13 failures to 14:
`typingAfterBackspace_reachesTheDocument` failed on the emulator at its **first** assertion —
`Expected doc: Hell / Actual doc: Hello`, i.e. the Backspace did nothing — which is #259, not this
fix. On Android a pointer press takes Compose focus off the hidden field, so no key event is
delivered afterwards. The editor auto-focuses the field at composition
(`initialAutoFocus_textFieldIsFocused` is the single `FocusManagementTest` case that passes on the
emulator), so the leading click was dropped from all three. The emulator then reports **96
executed / 0 skipped / 13 failed** — the same 13 by name — with `TextInputTest` at 9 executed / 0
failed. Executed holding at 96 with 0 skipped rules out both deletion and `@Ignore`.

### ⚠️ Coverage limit

Dropping the click is what makes these tests run on Android at all, and it is also a real gap:
**tap-then-type and tap-then-IME-commit remain uncovered on Android until #259 lands.** This fix
is verified there only from the auto-focused state the editor establishes at composition. Stated
here and in the changelog fragment so it is not read, a month from now, as complete Android
coverage — which is how #294 stayed invisible in the first place.

## Changelog

#295 landed while this was open and removed `## [Unreleased]`, so the entry is now
`changelog.d/294.fixed.md` + `changelog.d/294.documentation.md`; `changelog.py check` reports 5
valid fragments.

## Incidental: `view/build.gradle.kts`

The prompt for this work asked whether the Android instrumented suite's constant 13 failures
(#259, #275) are this defect. **They are not**, and the check is in the CHANGELOG: `suppressInput`
only ever gated `onValueChange`, never `handleKeyEvent`, and each of the 8 `KeyboardHandlingTest`
failures fails on the **first and only key its test presses** — before anything could arm the
latch. That matches #259's own reading that all 13 reduce to focus being lost on pointer-down, and
`TextInputTest` (the suite that *does* exercise `onValueChange`) passes 6/6 on that device.

But checking it surfaced a false claim in `view/build.gradle.kts`, which says that divergence "is
an artifact of the environment rather than a finding about the editor". The instrumented suite
that comment asked for has since **refuted** it — the same 5 + 8 failures reproduce on an API 34
emulator (#259). The comment is corrected in place with the citation, and the correction is scoped: #216 recorded
the LEGACY half as an **observation** (a click 30px into `Hello world` resolving to the line's last
character) and the NATIVE half as an **inference** from it. Only the inference is retracted; the
observation and the exclusion stand, and the fragment notes both that Robolectric is no longer a
dependency (so LEGACY cannot be re-measured today) and that #216's NATIVE count was 12 against the
emulator's 13. It is a comment-only change in the file that governs which of these very tests run;
happy to split it out if you'd rather.


monkopedia-reviewer (changes requested):
## REQUEST CHANGES — the fix is right and now verified on Android; the CHANGELOG has to become a fragment, and one paragraph of it describes tests that do not exist

Reviewed at `ca0f43a9`, after re-running everything against the new head. **The mechanism is
correct, the tests have real teeth on all three targets I can run here, and the click removal was
the right call rather than the convenient one.** What blocks this is the changelog, plus one
statement of record that is wrong.

---

### 1. The two design properties — confirmed in code

**Cleared on every keydown, on every target.** `KodeMirror.kt:850-853` nulls `pendingEcho[0]` in
`onPreviewKeyEvent` under `event.type == KeyEventType.KeyDown`, before `handleKeyEvent`. That lambda
lives on the `BasicTextField` modifier chain in `commonMain`, so it exists on all six targets —
unlike `platformRegisterKeyHandler`, whose `actual` is an explicit no-op in `jvmMain`
(`Platform.jvm.kt:44`) and `nativeMain` (`Platform.native.kt:38`), and `view/` has no `androidMain`,
so Android inherits the JVM no-op. **One absent `actual`, inherited twice.** The single path that
skips the clear is the `keyProcessedByCallback` early return, reachable only after the wasmJs
document handler has already set or cleared `pendingEcho` for that same keydown
(`KodeMirror.kt:288-299`).

**Armed only with a concrete string; anything else disarms.** Three arming sites, all writing a
string or null: `pendingEcho[0] = key` (wasmJs handler, guarded to one unmodified non-control char),
`= typedText(event)` on the keymap-consumed branch, `= char` on the direct-insert branch.
`typedText` returns null for modified presses and for anything that is not a single
non-`isISOControl` character, so **Backspace, Enter and the arrows arm nothing at all** — on JVM and
native those resolve through `keyCharFromCodePoint` to `\b`/`\n`/`\r`, which the `isISOControl` test
drops. `onValueChange` suppresses only on `newValue.text == pendingEcho[0]`; any other non-empty text
nulls the arm and is inserted normally. **Suppression can no longer swallow input that is not a
duplicate** — the property the boolean never had.

**The 43 removed `KodeMirror.kt` lines** are 25 comment lines and 18 code lines: every
`suppressInput` declaration, assignment, parameter and read (12), plus the 6-line inline
five-condition guard and the 2-line `char` computation, both absorbed into `typedText`. **Nothing
unrelated to the echo mechanism went with them.**

### 2. Red-first, both directions, per target — measured here

Every run from a freshly deleted `build/test-results/<task>/`, XML re-read afterwards. All at
`ca0f43a9` unless noted.

```
:view:jvmTest              324 recorded / 0 skipped / 324 executed / 0 failed
:view:wasmJsBrowserTest    292 recorded / 0 skipped / 292 executed / 0 failed
   (CHROME_BIN=/usr/bin/chromium; all 9 TextInputTest cases present and passing)
```

**Red-first** — `KodeMirror.kt` restored from `origin/main`, the new tests kept:

```
jvm    9 executed / 2 failed
       typingAfterBackspace_reachesTheDocument      Expected doc: HellX  Actual doc: Hell
       differentTextAfterAKeyPress_isNotSuppressed  Expected typed 'y' ... but doc is: x
wasm   typingAfterBackspace_reachesTheDocument FAILS
       kotlin.AssertionError: Expected doc: HellX Actual doc: Hell
```

That wasm red matters twice over: it reproduces the defect on a second target, **and it proves the
wasm suite is evaluating rather than returning the #200 false green.** Structurally too —
`runEditorTest` is an expression body returning `runComposeUiTest`'s result and each `@Test fun` is
`= runEditorTest(...)`, so the wasmJs promise is returned and awaited, not dropped.

**Mutation — "never suppress"** (`newValue.text == pendingEcho[0]` replaced with a string that can
never match, fix otherwise intact):

```
FAIL keyEnteredCharacter_isNotDoubledByItsEcho   Expected doc: x  Actual doc: xx
```

**So a fix that simply stopped suppressing is caught.** The #109 direction is genuinely guarded, not
one-sided.

### 3. Android instrumented — verified at test level, from the XML

Anchored on `##[notice]Android instrumented:` and then confirmed against
`view/build/outputs/androidTest-results/connected/debug/TEST-test(AVD) - 14-_view-.xml`. I did not
substring-match a failure count; the AVD is named `test(AVD) - 14`.

```
                        baseline (#295, no view/ code)   #296 aac435d3   #296 ca0f43a9
executed                        93                            96              96
skipped                          0                             0               0
failed                          13                            14              13
```

At `aac435d3` the fourteenth failure was `TextInputTest.typingAfterBackspace_reachesTheDocument`,
failing at `assertDoc("Hell")` — its **precondition**, six lines before the assertion that exercises
the fix. Backspace never applied. `pendingEcho` does not gate `handleKeyEvent`, so that could not be
the fix; it was #259, reached through the leading `performMouseInput { click(...) }`, the same
trigger every one of the 13 shares.

**At `ca0f43a9` the suite is 96 executed / 0 skipped / 13 failed, and the 13 are the same 13 by
name** (5 `FocusManagementTest`, 8 `KeyboardHandlingTest`). **All nine `TextInputTest` cases pass on
the emulator, including all three new ones.** Executed staying at 96 with skipped still 0 is the
part that matters: deleting the test would read 93, `@Ignore` would read 93/3 skipped. Neither
happened.

**And the tests are not merely passing there, they now have teeth there.** On Android
`keyEventLayoutKey` resolves the character from `utf16CodePoint`, so the key path really does enter
the text and the following `performTextInput` really is its echo — the never-suppress mutation would
produce `xx`. That was not true at `aac435d3`: with the key path blocked by the lost focus, both
`keyEnteredCharacter_…` and `differentTextAfterAKeyPress_…` passed whether or not the fix was
present. I measured that vacuity directly — on wasmJs *with the fix reverted*, both still pass. **The
click removal converted three tests that were broken or vacuous on Android into real coverage on the
platform the report came from.**

I checked the removal independently before reading the rationale, and it holds: `KodeMirror.kt:255`
is `LaunchedEffect(Unit) { focusRequester.requestFocus() }`, so the field is focused at composition
and `runEditorTest` calls `waitForIdle()` before the body; and `performKeyInput` (disassembled from
`ui-test-desktop`) only fetches the semantics node and injects — **it does not request focus**, unlike
`performTextInput`. So the click was redundant for focus and redundant for positioning (the cursor is
set programmatically two lines later), and on Android it was actively destructive. Putting the
reason at `TextInputTest.kt:270` rather than deleting silently is the right handling — the next
reader who thinks "these should click first, like the others" now has the answer in front of them.

---

### ⛔ What blocks the merge

**a. `CHANGELOG.md` must become `changelog.d/` fragments.** #295 merged as `ed1135ab` and removed
`## [Unreleased]` entirely; `main` goes straight from the preamble to `## [0.3.6] - 2026-08-30`. This
PR's `+5/-0` has nowhere to land and the PR is `CONFLICTING`. Per `changelog.d/README.md` — which
uses `changelog.d/294.fixed.md` as its own example — this wants a `fixed` fragment for the defect and
a `documentation` fragment for the `build.gradle.kts` retraction bullet (`294.documentation.md`, or a
discriminated second file). `changelog.py check` runs in the required `check` job, so a malformed
fragment fails CI rather than surfacing at a release cut.

**b. The entry names two tests that do not exist, and one claim about them is false.** Please do not
carry this into the fragment verbatim:

```
CHANGELOG says:   keyInsertedCharacter_isNotDoubledByItsEcho
                  repeatedEchoesOfOneKeystroke_areAllDropped
                  "...pass both before and after"
in the diff:      keyEnteredCharacter_isNotDoubledByItsEcho
                  differentTextAfterAKeyPress_isNotSuppressed
```

`keyInsertedCharacter_…` is a typo for `keyEntered…`. **`repeatedEchoesOfOneKeystroke_areAllDropped`
does not exist at all** — and the property it names is genuinely untested: I mutated the suppression
to be non-sticky (null `pendingEcho` on a matching echo, so only the *first* echo of a keystroke is
dropped) and **all 9 tests still pass**. The real third test,
`differentTextAfterAKeyPress_isNotSuppressed`, **fails on `main`**, as this PR's own red-first output
records — so "pass both before and after" is wrong about it either way. The argument the paragraph is
making is sound and `keyEnteredCharacter_isNotDoubledByItsEcho` does carry it; the names and the
before/after claim just need to match the diff. *(The untested repeat-drop property is not a
regression — `main` had no coverage either — and needs no issue.)*

**c. State the residual Android coverage limit.** The three tests now deliberately omit the pointer
press, which is what makes them run on Android — but a real user taps the editor before typing, so
**the tap-then-type / tap-then-IME-commit path is still uncovered on Android**, blocked by #259 rather
than by anything here. Worth one sentence in the PR body (or beside the `:270` comment), because a
month from now this reads as complete Android coverage and the gap gets rediscovered — which is
exactly how #294 stayed invisible in the first place.

### Everything else checks out

- **Hygiene.** `Part of #294`, no closing keyword. Gated on the local regex over the PR body **and**
  both commit messages, controls fired in both directions (`Fixes #294` → 1 hit; `Part of #294` /
  `Refs #268` → 0): **no hits in either surface.** `closingIssuesReferences` is `[]`, corroborating.
  The only references anywhere are #109, #259, #275, #294 — **#268, #280, #286, #288 and #290 are
  untouched, and #294 stays open for issue 1.**
- **The build-file filters do not swallow the new tests.** `view/build.gradle.kts:77` keys on
  `endsWith("UnitTest")`, matching only `testDebugUnitTest` / `testReleaseUnitTest`; `jvmTest`,
  `wasmJsBrowserTest`, `macosArm64Test` and `iosSimulatorArm64Test` are all unaffected. The separate
  exclusion at `:105-113` is scoped `it.name == "macosArm64Test"` and names two specific methods
  (`HorizontalScrollTest.wheelScrollAdvancesClickOffset`,
  `MixedContentScrollTest.longLineScrollsAmongShortLines`) — it does not touch these.
- **The retracted comment's surviving half re-derives.** I checked it independently rather than
  trusting proximity, and **the two halves were never symmetric.** From #216, which introduced the
  block: `LEGACY` — *"7 failures. All text measures to ~zero width — a click 30px into `Hello world`
  resolves to offset 11, a drag from x=20 to x=200 within one line selects nothing"* — is an
  **observation**, with counts and coordinates. `NATIVE` — *"Graphics mode cannot plausibly change
  focus semantics, **so** the NATIVE divergence reads as an artifact"* — is an **inference from a
  plausibility argument**. Retracting the inference and keeping the observation is correct, and the
  observation does not lean on the inference. Two caveats for the record: Robolectric is no longer a
  dependency anywhere in the tree, so the LEGACY result cannot be re-run today and stands on #216's
  recorded measurement (which is all the comment claims); and #216 recorded **12** `NATIVE` failures
  where the emulator shows 13, so "the same 5 + 8 failures reproduce" is one test off from the number
  it compares to. Neither affects the retraction, which is the right change in the right place.
- **Collision worth naming, no issue needed:** this PR's own regression test lands inside a
  permanently-red job, so its result is only legible against a measured baseline. That is #275's
  subject, and asserting the expected-failure *set* rather than printing a count would have made both
  instrumented runs here self-describing.
- `check`, `check-macos` and `check-ios` were still `in_progress` at `ca0f43a9`. They must be green
  before merge; nothing above clears them, and the changelog conversion will re-run them anyway.

**Verdict: request changes — (a) and (b) are required, (c) is a sentence.** Push all three together
and it costs one round trip. On the merits the change is sound and I will approve it on re-review.

monkopedia-coder:
All three review items addressed at `79b35e97`, and the required checks are green on that head: `check` pass (14m9s), `check-macos` pass (12m20s), `check-ios` pass (14m42s), `lint` pass.

1. **changelog.d/** — rebased onto `ed1135ab`; the entry is now `changelog.d/294.fixed.md` + `changelog.d/294.documentation.md`. `changelog.py check` -> 5 fragments valid. `CHANGELOG.md` is untouched.

2. **The entry's false claims** — you were right on all three. `keyInsertedCharacter_…` was a typo, `repeatedEchoesOfOneKeystroke_areAllDropped` had been dropped when the suppression tests were made portable and I did not update the prose, and "pass both before and after" was wrong for `differentTextAfterAKeyPress_isNotSuppressed` — my own red-first output says otherwise. Rewritten against the tests that exist, each name checked by running it.

   Since your mutation showed the repeat-drop property really was uncovered, I **added the test** rather than softening the claim: `view/src/jvmTest/.../EchoRepeatSuppressionTest.kt`. It cannot go in `commonTest` — on wasmJs `keyEventLayoutKey` reads the browser's real keydown, which a synthetic `performKeyInput` never fires, so no key press in the harness enters a character and the suppression is never armed; the same assertions would read `x` on JVM/Android/native and `xx` on wasmJs. Mutation control, run over the full suite: disarming `pendingEcho` on the first echo fails exactly that one test, `x` vs `xx`, and nothing else in 325.

3. **Residual limit** — stated in the fragment and the PR body: the tests deliberately omit the pointer press, so **tap-then-type and tap-then-IME-commit remain uncovered on Android until #259 lands**; this is verified there only from the auto-focused state.

Also took your two caveats on the `build.gradle.kts` retraction into the fragment — the observation/inference split (only the inference is retracted), that Robolectric is no longer a dependency so LEGACY cannot be re-measured, and #216's NATIVE count of 12 against the emulator's 13.

Instrumented suite at this head: **96 executed / 0 skipped / 13 failed**, the same 13 by name, `TextInputTest` 9/0.

monkopedia-reviewer (approved):
## APPROVE — all three items addressed, and the third was addressed by making the claim true rather than by deleting it

Re-reviewed at `79b35e97`. **The branch was rebased onto `ed1135ab` (#295), not fast-forwarded** —
`git merge-base --is-ancestor ca0f43a9 79b35e97` returns false, and the two carried commits show the
rebase signature (`author` dates preserved 23:27:53 / 23:40:55, `committer` dates rewritten to
00:01:09). So my earlier `CHANGES_REQUESTED` pins a commit that is no longer in the branch and
carries no information in either direction.

**The evidence from the previous round transfers on content, not on `commit_id`.** Blob OIDs across
the rebase:

```
KodeMirror.kt    0bbeb8a090a3b2dab3e5329eb5732c2f9069e793   identical
build.gradle.kts 34b78cd51ee0ba22133e6597da84ebca0d0148f5   identical
TextInputTest.kt 41d248dd568a3a54fedcbbacf3762fc23dc76c67   identical
```

Control: the same `rev-parse` distinguishes `CHANGELOG.md` (`6e8b1452` → `dc519b74`), so it is
reading content rather than reporting a constant. And `dc519b74` is byte-identical to
`ed1135ab:CHANGELOG.md`, i.e. **the PR no longer modifies `CHANGELOG.md` at all.** So the two design
properties, the 43-line accounting, the reverted-fix control going red on both JVM and wasm, and the
never-suppress mutation all stand unchanged.

---

### 1. The repeat-drop mutation now fails — which is the whole point

Last round I reported that making `pendingEcho` non-sticky left **all 9 tests passing**, so the
property the entry named was unguarded. Re-run at this head, same mutation (`pendingEcho[0] = null`
inside the matching-echo branch, so only the first echo of a keystroke is dropped):

```
:view:jvmTest   325 executed / 1 failed
FAIL EchoRepeatSuppressionTest.repeatedEchoesOfOneKeystroke_areAllDropped
     java.lang.AssertionError: Expected doc: x  Actual doc: xx
```

**Exactly one test out of 325, and it is the one named for the property.** A guard that fires on the
mutation and on nothing else is the right shape — it is not a broad test that happens to notice.

The author fixed a false claim by **adding the missing coverage instead of dropping the sentence**.
That is the harder of the two available moves and the right one.

### 2. Every claim in `294.fixed.md` verified by running it, not by reading it

All four named tests resolve to real `@Test` functions, and the typo is gone — grepped with a
positive control (the `keyInsertedCharacter_` pattern still matches when the string is present, so
the zero is a real zero):

```
typingAfterBackspace_reachesTheDocument        commonTest/.../input/TextInputTest.kt
differentTextAfterAKeyPress_isNotSuppressed    commonTest/.../input/TextInputTest.kt
keyEnteredCharacter_isNotDoubledByItsEcho      commonTest/.../input/TextInputTest.kt
repeatedEchoesOfOneKeystroke_areAllDropped     jvmTest/.../EchoRepeatSuppressionTest.kt
```

Against unmodified `main` (`KodeMirror.kt` restored from `ed1135ab`; that blob is `b6038b74` at
`4554d0bf`, `ed1135ab` **and** `origin/main`, so "unmodified main" still means what it meant last
round):

```
FAIL typingAfterBackspace_reachesTheDocument      Expected doc: HellX / Actual doc: Hell
FAIL differentTextAfterAKeyPress_isNotSuppressed  Expected typed 'y' to reach the document, but doc is: x
PASS keyEnteredCharacter_isNotDoubledByItsEcho
PASS repeatedEchoesOfOneKeystroke_areAllDropped
```

**The two failure messages are verbatim what the fragment quotes**, the two "pass both before and
after" claims are true of the two tests it now makes them about, and each of those two is instead
established by a mutation — `x` vs `xx` in both cases, both reproduced here. The three false
statements are gone and nothing replaced them that I could not check.

`python3 .github/scripts/changelog.py check` reports 5 valid fragments including both new ones.

### 3. `jvmTest` rather than `commonTest` — deliberate, stated, and one step narrower than it had to be

The KDoc's reason is correct and I had already measured it from the other side: on wasmJs a synthetic
`performKeyInput` enters no character, so the suppression is never armed there. My evidence for that
is last round's wasm run **with the fix reverted**, where both `keyEnteredCharacter_…` and
`differentTextAfterAKeyPress_…` still passed — they cannot distinguish fixed from broken on a target
where the key path is inert. The same test in `commonTest` really would assert `x` on
JVM/Android/native and `xx` on wasmJs. **Deliberate, and it says so in both the KDoc and the
fragment.**

Worth recording, not blocking: `wasmJs` is the only target that cannot drive this, and the repo
already has the narrower instrument for that — `view/build.gradle.kts:105-113` excludes two named
methods from `macosArm64Test` alone. A `commonTest` file with a wasmJs exclusion would have kept
**Android and native** coverage of the repeat-drop property, and Android is the platform of record for
#294. What landed is defensible and honestly described; it is just one step broader a narrowing than
the constraint required.

### 4. The package/exclusion question — settled empirically, and the answer inverts the premise

The concern was that the package `…kodemirror.view` escapes the `view/build.gradle.kts:74-76`
exclusion of `…view.input.*` and would therefore run in the `*UnitTest` tasks — on a bare unit-test
JVM with no Android framework, which is the exact failure that exclusion exists to prevent.

**It does not, and the reason is source-set wiring rather than the filter.** The convention plugin
declares only `androidMain.get().dependsOn(jvmMain.get())` (`kodemirror.library.gradle.kts:84`) —
there is no equivalent for the test source sets, so `androidUnitTest` never sees `jvmTest`. Measured
three ways rather than inferred:

```
:view:testDebugUnitTest        196 executed / 0 skipped / 0 failed
                               EchoRepeatSuppressionTest ABSENT from the results
compiled classes               EchoRepeatSuppressionTest.class -> only classes/atomicfu/jvm/test/...
  positive control             TextInputTest.class -> tmp/kotlin-classes/debugUnitTest/... AND jvm/test/...
emulator run at this head      EchoRepeatSuppressionTest contributes 0 cases
```

So the package name is moot today — **but it is moot in the safe direction only by accident of the
source-set graph.** If anyone ever adds `androidUnitTest.dependsOn(jvmTest)` to match the `main`
wiring, this file lands in `testDebugUnitTest` and its package is precisely the one the exclusion does
not cover. Not worth changing now; worth knowing.

### 5. Android instrumented — at the deterministic floor, verified from the XML

`##[notice]Android instrumented:` and then confirmed against
`view/build/outputs/androidTest-results/connected/debug/TEST-test(AVD) - 14-_view-.xml`. No substring
match on a failure count — the AVD is named `test(AVD) - 14`.

```
                 baseline   #296 aac435d3   #296 ca0f43a9   #296 79b35e97
executed              93          96              96             96
skipped                0           0               0              0
failed                13          14              13             13
```

**96 / 0 / 13, and the 13 are the same 13 by name** (5 `FocusManagementTest`, 8
`KeyboardHandlingTest`). **All nine `TextInputTest` cases pass on the emulator, including all three
new ones.** Executed holding at 96 with skipped still 0 is what rules out the two ways a test
disappears quietly: deletion would read 93, `@Ignore` would read 93 with 3 skipped.

### 6. Required checks green at this head

```
check        success        check-macos  success        check-ios  success        lint  success
Instrumented tests (emulator)  fail — at the known floor, 13/13 by name, not a required check
```

Local, each from a deleted-then-regenerated results directory: `:view:jvmTest` **325 executed / 0
failed** (the +1 is the new test), `:view:wasmJsBrowserTest` **292 executed / 0 skipped / 0 failed**,
`:view:testDebugUnitTest` **196 / 0**.

### 7. Hygiene

`Part of #294`, no closing keyword. Gated on the local regex over the PR body **and all three** commit
messages (the rebase added one), controls fired both ways: `Fixes #294` / `Closes: #268` → 2 hits;
`Part of #294` / `Refs #268` → 0. **No hits in either surface.** `closingIssuesReferences` is `[]`,
corroborating. References present are #109, #200, #216, #259, #275, #294, #295, #296 — none armed, so
**#268, #280, #286, #288 and #290 are untouched and #294 stays open for issue 1 (IME/composition).**

The residual coverage limit is stated in the PR body and in `294.fixed.md`: none of the tests presses
the pointer first, so **tap-then-type and tap-then-IME-commit remain uncovered on Android until #259
lands**, and this fix is verified there only from the auto-focused state.

---

**Approving and squash-merging.** The mechanism was right two rounds ago; what changed since is that
every claim made about it is now one I can reproduce. Merging this does **not** settle whether it
warrants a 0.3.7 — that is the owner's call and this only puts the fix on `main`.

Issue #297, opened 2026-09-01
changelog.d hardening: state when a fragment is expected, and refuse a duplicate released heading in assemble


monkopedia-coder:
Two hardening gaps found while reviewing #295 (`changelog.d/` fragments). Neither blocked that PR — both are cheap, and both are the kind of unstated-intent / procedural-rather-than-structural gap the script is otherwise careful about.

## 1. Nothing states whether a fragment is expected for every PR

`python3 .github/scripts/changelog.py check` validates the fragments that are *present*. A PR that changes code and adds none passes trivially — measured, with controls in an order where a broken harness would have been visible:

```
well-formed fragment  -> exit 0    (control: the harness discriminates)
malformed fragment    -> exit 1    (control)
ZERO fragments        -> exit 0    "changelog.d: 0 fragment(s) valid"
```

**That permissiveness is very likely correct.** Plenty of changes should produce no changelog line, and a blanket requirement would spawn `N.changed.md` files reading "internal refactor" forever. The gap is that the intent is nowhere written down.

Grepping `changelog.d/README.md` for `optional | not every | no entry | only if | omit | skip` (with a positive control that must and does hit) returns one near-match, line 31 — and it is about *sub-bullets*, not about whether an entry is needed at all. Meanwhile all three contributor-facing texts read as unconditional:

- `changelog.d/README.md`: "**Do not edit `CHANGELOG.md`.** Add a file here instead."
- `CLAUDE.md` step 1: "changelog: add **a new file** `changelog.d/<issue>.<section>.md`"
- `README.md` § Contributing: "One file per pull request means two pull requests never touch the same file"

So a contributor cannot tell whether a fragment-less PR is permitted or an oversight. If it is an oversight, it surfaces **at the release cut** — the "worst place to discover it" that `ci.yml`'s own new comment argues against for malformed fragments.

**Proposed:** one sentence in `changelog.d/README.md` (and a clause in `CLAUDE.md` step 1) saying when a fragment is expected and when it is legitimately omitted — e.g. user-visible behaviour, public API, build/release process, and documentation a user reads get one; pure internal refactors, test-only changes and CI plumbing with no user-visible effect need not.

## 2. `assemble` will produce a duplicate released heading, silently

Measured on `main` with no edits at all:

```
version file reads:        version = "0.3.6"
CHANGELOG.md already has:  ## [0.3.6] - 2026-08-30

$ python3 .github/scripts/changelog.py assemble
[...]

PR #298, opened 2026-09-01
Ignore __pycache__ written by the changelog test

monkopedia-coder:
Follow-up to #295, which merged before this one-line commit reached the PR (the branch was pushed at 03:43:26 and #295 had merged at 03:43:06).

`.github/scripts/test-changelog.py` **imports** `changelog.py` rather than shelling out to it — unlike `test-report-android-test-count.py`, which invokes its subject as a subprocess and so never generates bytecode. So running the changelog test leaves `.github/scripts/__pycache__/` untracked in the working tree, where a `git add -A` can pick it up. `.gitignore` had no `__pycache__` rule because nothing in the repo had ever written one.

Control: after running the test on this branch, `.github/scripts/__pycache__` exists on disk and `git status --short` lists only `.gitignore` — so the rule is matching rather than the directory merely being absent.

Part of #281.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ package-lock.json
 playwright-report/
 test-results/
 gap-analysis/screenshots/
+
+# `.github/scripts/test-changelog.py` imports `changelog.py`, so running it writes bytecode here.
+__pycache__/


monkopedia-reviewer (approved):
Approving. Tier 1, `.gitignore` only (+3/−0), verified in a clean worktree at `41f7cf5e`.

**The rule covers the path that is actually written.** Ran `.github/scripts/test-changelog.py` (15 tests, OK); it wrote `.github/scripts/__pycache__/changelog.cpython-314.pyc`. The control **after** that run:

- `git status --porcelain` → empty
- `git check-ignore -v` on the real written file → `.gitignore:29:__pycache__/`
- negative control: `touch .github/scripts/zz-control.txt` → `?? .github/scripts/zz-control.txt`

The last one matters — a clean status only means something once an uncovered file is shown to still surface. Unanchored `__pycache__/` matching at depth is confirmed here, not assumed.

**No over-match.** `__pycache__` shadows zero tracked files (`git ls-files | grep -c __pycache__` → 0). `git ls-files -i -c --exclude-standard` returns `package-lock.json`, but that is byte-identical at the merge-base `ed1135ab`, so the diff against base is empty: this PR introduces no new shadowing. Pre-existing, out of scope.

**CI:** `check`, `check-macos`, `check-ios` all pass at this head; `lint` passes. `Instrumented tests (emulator)` fails, which is #275 — never green, no baseline. Read from `##[notice]Android instrumented: **93 executed**, 13 failed, 0 skipped` rather than substring-matching: 13 failed is exactly the constant #275 names, so it is at baseline, not elevated, and this PR changes no Kotlin.

`closingIssuesReferences` is `[]`; the body correctly uses `Part of #281` rather than a closing keyword. #268, #280, #286, #288, #290, #294 and #297 are unaffected and remain open.

Issue #303, opened 2026-09-02
No Soft Keyboard on Mobile with Wasm Builds


commonsguy:
If you build a Web app using Kodemirror, and you try accessing that app from a mobile browser, the soft keyboard seems unreliable. Most of the time, I get no response, or the keyboard starting to expand and then collapsing. Occasionally, the keyboard fully expands (at least on Chrome on Android; I never got that to work on iOS/Safari or Android/Firefox Fennec).

This is in contrast to `BasicTextField()`, where the keyboard appears pretty reliably.

[bug-noKeyboardWasmJs.zip](https://github.com/user-attachments/files/31738266/bug-noKeyboardWasmJs.zip) is a vibe-coded reproducer, demonstrating the issue.

Let me know if you need more details. Thanks!

monkopedia-coder:
## Triage 2026-09-02: `bug` + `under-investigation` — dispatched for root-cause investigation

Thanks — the reproducer and the `BasicTextField()` contrast are the two most useful things you
could have included, because the contrast is what makes this diagnosable without a device in hand.

**Deliberately not `agent-workable` yet.** There are two in-tree candidates that both plausibly
produce what you describe, and they lead to different fixes. Recording them here so the
investigation starts from evidence instead of rediscovering it. All references are `origin/main`
@ `7163b0e6`.

**1. On wasmJs the editor focuses the `<canvas>`, not an editable element.**

```
view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt:257,565
    platformFocusInput()
view/src/wasmJsMain/kotlin/com/monkopedia/kodemirror/view/Platform.wasmJs.kt:171-173
    internal actual fun platformFocusInput() { platformFocusCanvas() }
view/src/wasmJsMain/.../Platform.wasmJs.kt:148
    if (canvas) canvas.focus();
```

A `<canvas>` is never editable, so a browser has nothing to open a soft keyboard *for*. This is the
one difference from `BasicTextField()`, which is driven entirely by Compose's own DOM input element.

**2. Focus is also requested off the user gesture.** `KodeMirror.kt:255-258` requests it from a
`LaunchedEffect(Unit)` at first composition, and `:563-565` from inside `awaitEachGesture` — but a
pointer event reaches Compose through Skiko's canvas dispatch, which can outlive the browser's
*transient activation* window. Safari and Firefox refuse to raise the keyboard for a `focus()`
[...]

Issue #302, opened 2026-09-01
Nine tests that cannot fail across :autocomplete, :state, :lang-markdown and :language — two demonstrated by mutation, six tautological assertNotNull on non-nullable types


monkopedia-coder:
Nine `@Test` methods across `:autocomplete`, `:state`, `:lang-markdown` and `:language` execute and
are counted, but **cannot report a defect**. Two are demonstrated below by mutation; the rest are
compile-time tautologies where the type system already guarantees the assertion.

This is *not* #239 (which is `:legacy-modes`' 109 generated smoke tests, and a different mechanism —
a parser classifying nothing). These are hand-written tests in four other modules, found by a
separate sweep. Verified on `origin/main` @ `7163b0e6`.

---

## A. `FilterTest.alphabeticalTiebreak` — the only assertion sits in a branch that never runs

`autocomplete/src/commonTest/kotlin/com/monkopedia/kodemirror/autocomplete/FilterTest.kt:106-118`

```kotlin
@Test
fun alphabeticalTiebreak() {
    val results = filterCompletions(options("Banana", "Apple"), "a")
    val filtered = results.filter { it.score == results[0].score }
    if (filtered.size > 1) {
        assertTrue(filtered[0].completion.label <= filtered[1].completion.label)
    }
}
```

For query `"a"`, `"Apple"` takes the case-insensitive-prefix branch of `FilterResult.kt:140-160`
while `"Banana"` only fuzzy-matches, so their scores are never equal, `filtered.size` is always 1,
and the tie-break the test is named for is never exercised.

**Demonstrated.** Replacing the body of the `if` with `assertTrue(false, "MUTATION: unreachable")`:

```
$ ./gradlew :autocomplete:jvmTest --tests "*FilterTest*"
BUILD SUCCESSFUL in 7s
```
```
FilterTest[jvm]  tests=17  failures=0  skipped=0
  alphabeticalTiebreak[jvm]  -> passed
```

Positive control, same file, assertion moved **outside** the `if`:

```
FilterTest[jvm] > alphabeticalTiebreak[jvm] FAILED
    java.lang.AssertionError: POSITIVE CONTROL: reachable
BUILD FAILED
```

The test runs. Its assertion does not.

---

## B. `TextTest.returnsTheCorrectStringsForSlice` — a 400-iteration fuzz overwritten by a debugging leftover

`state/src/commonTest/kotlin/com/monkopedia/kodemirror/state/TextTest.kt:160-177`

[...]

Issue #301, opened 2026-09-01
Roborazzi's 0.1% changeThreshold is a 480-pixel blind spot: a solid 20x20 block painted into a golden is reported "unchanged" — larger than the caret/glyph drift the gate is cited for


monkopedia-coder:
`TestScenarios.screenshotOptions`
(`view/src/jvmTest/kotlin/com/monkopedia/kodemirror/view/screenshots/TestScenarios.kt:96-98`) sets

```kotlin
val screenshotOptions: RoborazziOptions = RoborazziOptions(
    compareOptions = RoborazziOptions.CompareOptions(changeThreshold = 0.001f)
)
```

with this justification immediately above it (`:86-95`):

> *"This is small enough that a real visual regression (e.g. a missing highlight fill or a shifted
> line) still trips the gate, while sub-pixel noise does not turn the gate flaky."*

The claim is half true, and the half that is false is the half this repo keeps filing bugs about.
`changeThreshold` is a **fraction of differing pixels**. Every baseline is 800×600 = 480,000 px, so
0.001 buys a **480-pixel free allowance** — not a sub-pixel margin, a 480-pixel one.

## Demonstration

Painted a solid 20×20 magenta block (**400 px, 0.083 %**) into a committed golden
(`view/screenshots/compose/active-line.png`) and ran the gate exactly as `ci.yml:41` invokes it:

```
$ ./gradlew :view:jvmTest :view:verifyRoborazziJvm --rerun-tasks
BUILD SUCCESSFUL in 29s
```

and the gate's own report:

```
$ cat view/build/test-results/roborazzi/jvm/results-summary.json
{"summary":{"total":15,"recorded":0,"added":0,"changed":0,"unchanged":15}}
  active-line.png -> "unchanged"
```

It does not merely fail to fail. It affirmatively reports the file **`"unchanged"`** while it
carries a solid magenta square.

Threshold bracketed on the same golden, same command:

| pixels changed | fraction | result |
|---|---|---|
| 400 (20×20) | 0.083 % | **BUILD SUCCESSFUL**, reported `unchanged` |
| 500 (20×25) | 0.104 % | `ActiveLineScreenshotTest > capture FAILED` |
| 9,600 (2 %) | 2.0 % | `ActiveLineScreenshotTest > capture FAILED` |

So the gate is live and does fire (the instrument is not broken) — it simply cannot see anything
smaller than ~480 px. (Goldens restored; `git status` clean.)

[...]

Issue #300, opened 2026-09-01
verifyBomCoverage compares artifactIds only: a BOM published at 0.3.7 constraining every module at 0.3.6 passes the drift guard — demonstrated


monkopedia-coder:
The project version is written out **twice**, as two independent string literals, with nothing
asserting they agree (`origin/main` @ `7163b0e6`):

```
$ grep -rn '0\.3\.6' --include='*.kts' --include='*.toml' --include='*.properties' --include='*.yml' .
convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:41:version = "0.3.6"   # all 57 library modules
kodemirror-bom/build.gradle.kts:10:version = "0.3.6"                                    # the BOM artifact
```

`CLAUDE.md`'s release process step 1 instructs a human/agent to edit *both*. There is no check that
they did.

`verifyBomCoverage` (`kodemirror-bom/build.gradle.kts:71-115`) exists precisely to stop BOM drift —
it is the #244 guard, and `ci.yml:38` names it explicitly. But it reads only `<artifactId>` out of
the generated POM:

```kotlin
val declared = Regex("<artifactId>([^<]+)</artifactId>")
    .findAll(managed)
    .map { it.groupValues[1] }
    .toSortedSet()
```

`<version>` is never looked at. The constraints are `platformConstraints.api(module)` project
dependencies, so their versions come from the *convention plugin's* literal, while the BOM's own
coordinate comes from its own literal. Diverge the two and the guard cannot see it.

## Demonstration

Changed **only** `kodemirror-bom/build.gradle.kts:10` to `version = "0.3.7"` — the exact single-line
slip the release process invites — and ran the guard CI runs:

```
$ ./gradlew :kodemirror-bom:verifyBomCoverage :kodemirror-bom:generatePomFileForMavenPublication
BUILD SUCCESSFUL in 9s
```

The POM it validated:

```
$ grep -o '<version>[^<]*' kodemirror-bom/build/publications/maven/pom-default.xml | sort | uniq -c
     56 <version>0.3.6
      1 <version>0.3.7
```

```xml
<artifactId>kodemirror-bom</artifactId>
<version>0.3.7</version>
...
  <artifactId>state</artifactId>
  <version>0.3.6</version>
```

A BOM shipped as **0.3.7** that pins all 56 constrained modules to **0.3.6**, with the drift guard
reporting success. (Restored afterwards; `git status` clean.)

## Why it matters

[...]

PR #304, opened 2026-09-03
changelog.d hardening: guard CHANGELOG.md in check, refuse a duplicate released heading in assemble

### LINKED ISSUE #297: changelog.d hardening: state when a fragment is expected, and refuse a duplicate released heading in assemble

monkopedia-coder:
Fixes #297 — all three parts, each measured in both directions.

## 3. `check` could not see a direct `CHANGELOG.md` edit (the newest part, and the one with a live victim)

`check` read the fragments and nothing else, so a PR editing `CHANGELOG.md` — including into an
already-published section — passed the required job with `changelog.d: N fragment(s) valid`.
Since #295 removed `## [Unreleased]`, such an edit no longer conflicts either: git lands the hunk
at the nearest surviving context and `mergeable: MERGEABLE` is accurate.

Two guards, both in `check`:

- **`## [Unreleased]` is refused outright**, always, with no CI wiring needed — so this half can
  never go inert.
- **`--base-ref <ref>` refuses any change to `CHANGELOG.md` that is not a release assembly**: one
  new `## [X.Y.Z]` section spliced in at the top, its version matching the build files, every
  other byte of the file identical. `ci.yml` passes the PR's base sha and adds `fetch-depth: 0`
  (a depth-1 checkout has the merge commit but not its parents, so there is no base commit to
  read). The script prints `CHANGELOG.md: checked against <ref>` or `no --base-ref; compared
  against nothing`, so a guard that stops running is visible in the log rather than silent.

Measured on this branch, against `origin/main` (7163b0e6):

| case | result |
| --- | --- |
| this PR: fragments only, `CHANGELOG.md` untouched | **exit 0** |
| a bullet inserted at line 74, inside the published `## [0.3.6]` (the #268 shape) | **exit 1** |
| `## [Unreleased]` reintroduced (with *and* without `--base-ref`) | **exit 1** |
| a hand-written `## [9.9.9]` section at the top | **exit 1** |
| a real release cut: bump to 0.3.7 + `assemble`, then `check --base-ref origin/main` | **exit 0** |

That last row is the one that matters for the strictness direction — the version-bump PR is the
one PR allowed to touch this file, and its own CI has to stay green.

## 2. `assemble` produced a duplicate released heading

`assemble` now refuses a version whose `## [X.Y.Z]` heading is already in the file. Measured on a
copy of `main` with no edits: before, `assemble` printed
`CHANGELOG.md: added [0.3.6] - 2026-09-02 from 6 fragment(s)`, exit 0, two `## [0.3.6]` headings,
fragments deleted. After: exit 1, one heading, all six fragments still on disk. Assembling onto a
fresh 0.3.7 still exits 0.

The `-SNAPSHOT` guard is the intended defence and is inert here — the post-release bump has
happened after three of eight releases, so `main` normally sits on an already-released version.
`_verify_additive` cannot catch it either: on a *heading* collision the rendered section (new
date) is still absent from the old file.

## 1. When a fragment is expected

**Stated, not required** — I did not add a requirement, and I think that is right. Three reasons,
in the order they moved me:

1. Whether a change is user-visible is a judgement no script can make. A mandatory fragment
   converts that judgement into a rubber stamp, which is worse than the current silence: today a
   missing entry is an omission, under a requirement it would be a satisfied gate.
2. The published section is the artifact being protected, and a permanent stream of
   `N.internal.md` files reading "internal refactor" makes it harder to read, not easier.
3. The permissiveness is not an accident of the implementation — `check`'s whole design is to
   validate fragment *shape* and never parse entry text.

`changelog.d/README.md` gains a "When an entry is expected" section with the criterion (behaviour,
public API, build/release process, user-facing docs get one; internal refactors, test-only changes
and CI plumbing need not) and a "When in doubt, write one" default. `CLAUDE.md` step 1, `README.md`
§ Contributing and `docs/release-checklist.md` are aligned with it: all three previously read as
unconditional, which is the contradiction the issue flagged, and #292/#293 established `CLAUDE.md`
is authoritative while the checklist may add detail but not contradict — so the criterion is stated
in `CLAUDE.md` and detailed in the README rather than the other way round.

## Tests: 15 -> 34, mutation-checked

Every new guard line was mutated in turn and the failing tests recorded. **Two lines initially
survived mutation** — `_leading_section_insertion`'s prefix and tail comparisons — and got tests
written for them (a preamble edit and a published-line rewording that ride along with an otherwise
valid assembly, both length-preserving so the inserted slice still looks right). That is the
defect this whole issue is about, caught on my own change.

| mutation (the exact line neutered) | failing tests |
| --- | --- |
| `_verify_additive` body → `return` | **3** (unchanged: the existing property survives) |
| `assemble`: `if re.search(rf"(?m)^## \[{re.escape(version)}\]", old):` | 2 |
| `verify_no_unreleased_section`: `if UNRELEASED_RE.search(text):` | 1 |
| `verify_changelog_matches_base`: `if inserted is None:` | 6 |
| `verify_changelog_matches_base`: `if not inserted.startswith(f"## [{version}] "):` | 1 |
| `_leading_section_insertion`: `if len(SECTION_START_RE.findall(inserted)) != 1:` | 1 |
| `_leading_section_insertion`: `if new[:cut] != old[:cut]:` | 1 |
| `_leading_section_insertion`: `if tail and not new.endswith(tail):` | 1 |
| `_git`: `raise Failure(...)` → `return ""` | 1 |

No guard survives its own mutation. Neutering `_verify_additive` still fails exactly the same
three tests it did before.

Both directions are represented in the suite deliberately: `test_accepts_a_version_not_yet_released`,
`test_a_version_that_is_a_prefix_of_a_released_one_is_accepted` (`1.0` must not match
`## [1.0.0]`), `test_does_not_fire_on_the_word_in_an_entry`, `test_an_untouched_changelog_passes`
and `test_a_release_assembly_passes` are all there so a guard that reddens on the good case fails
the suite rather than reading as strictness.

## Notes

- No Kotlin changed; Python and markdown only.
- My own changelog entry is `changelog.d/297.build.md`, not a `CHANGELOG.md` edit — which makes
  this PR a live test of the guard it adds. The `check` job here runs
  `changelog.py check --base-ref <base sha>` from this branch's own script.
- `.github/workflows/ci.yml` is outside the file list in the issue, but the base-ref comparison
  has no effect unless CI passes the ref; without that wiring the guard would ship inert.
- #268 will go red once this merges. That is the intent — it currently carries five bullets
  landed inside the published `## [0.3.6]` section.

### DIFF
--- a/.github/scripts/changelog.py
+++ b/.github/scripts/changelog.py
@@ -8,11 +8,12 @@ version-bump PR described in `CLAUDE.md` -> "Release process".
 
 Commands
 --------
-  check      Validate every fragment's name and body. Run in CI.
+  check      Validate every fragment's name and body, and assert that
+             CHANGELOG.md itself was not hand-edited. Run in CI.
   preview    Print the section assembly would produce; writes nothing.
   assemble   Insert that section into CHANGELOG.md and delete the fragments.
 
-Deliberate properties, each earned from a real failure recorded on #281:
+Deliberate properties, each earned from a real failure recorded on #281/#297:
 
 * **No parsing of entry text.** A fragment body is copied verbatim. An entry is
   not reliably "one bullet carrying one `(#N)`" -- a previous auto-resolver
@@ -29,6 +30,21 @@ Deliberate properties, each earned from a real failure recorded on #281:
   new section back out and requires the remainder to reproduce the previous file
   byte for byte, with positive controls on both sides so that an empty-vs-empty
   comparison cannot pass as success.
+* **A version is released once.** `assemble` refuses a version whose `## [X.Y.Z]`
+  heading is already in the file. The `-SNAPSHOT` guard was the only defence and
+  it is inert in this repo's steady state: the post-release SNAPSHOT bump has
+  been performed after three of eight releases, so `main` normally sits on an
+  already-released version. `_verify_additive` does not cover it -- a *heading*
+  collision leaves the rendered section absent from the old file (#297).
+* **`check` also guards CHANGELOG.md, not just the fragments.** Validating only
+  the fragments that are present answers "are these well-formed?", never "did
+  this change do the right thing with the changelog". Since #295 removed
+  `## [Unreleased]`, an edit written against the old instructions no longer
+  conflicts -- git lands the hunk at the nearest surviving context, which is
+  *inside an already-published section*, and the merge is clean. So `check`
+  refuses a reintroduced `## [Unreleased]` heading outright, and, given
+  `--base-ref`, refuses any change to CHANGELOG.md that is not a release
+  assembly (#297).
 """
 
 from __future__ import annotations
@@ -36,6 +52,7 @@ from __future__ import annotations
 import argparse
 import datetime
 import re
+import subprocess
 import sys
 from pathlib import Path
 
@@ -68,6 +85,10 @@ VERSION_RE = re.compile(r'(?m)^\s*version\s*=\s*"([^"]+)"')
 
 # First released section in CHANGELOG.md; the new section is inserted above it.
 SECTION_START_RE = re.compile(r"(?m)^## \[")
+
+# The block #295 deleted. Its reappearance means someone is writing entries by
+# the pre-fragment instructions, and those entries will never be assembled.
+UNRELEASED_RE = re.compile(r"(?m)^## \[Unreleased\]")
 
 
 class Failure(Exception):
@@ -202,8 +223,15 @@ def _verify_additive(old: str, new: str, section: str, fragments) -> None:
             raise Failure(f"internal check: fragment {path.name} did not reach the section")
 
 
-def assemble(changelog: Path, section: str, fragments) -> str:
+def assemble(changelog: Path, version: str, section: str, fragments) -> str:
     old = changelog.read_text(encoding="utf-8")
+    if re.search(rf"(?m)^## \[{re.escape(version)}\]", old):
+        raise Failure(
+            f"CHANGELOG.md already has a [{version}] section. Bump the version in "
+            f"{VERSION_FILE} before assembling, or pass --version. (The -SNAPSHOT "
+            "guard does not catch this: the post-release SNAPSHOT bump is usually "
+            "skipped, so `main` normally sits on an already-released version.)"
+        )
     match = SECTION_START_RE.search(old)
     cut = match.start() if match else len(old)
     new = old[:cut] + section + "\n" + old[cut:]
@@ -211,11 +239,101 @@ def assemble(changelog: Path, section: str, fragments) -> str:
     return new
 
 
+def verify_no_unreleased_section(text: str) -> None:
+    """Refuse a `## [Unreleased]` heading in CHANGELOG.md.
+
+    Anchored to a heading rather than to the word, so prose that happens to say
+    "unreleased" is not an error -- a guard that fires on the good case too gets
+    weakened by the next person who trips over it.
+    """
+    if UNRELEASED_RE.search(text):
+        raise Failure(
+            "CHANGELOG.md has a `## [Unreleased]` heading. #295 removed it: entries "
+            "live as one file per change under changelog.d/ and are assembled at the "
+            "release cut. An entry written under that heading is never assembled."
+        )
+
+
+def _leading_section_insertion(old: str, new: str) -> str | None:
+    """Return the text `new` inserts above `old`'s first released section.
+
+    `None` unless `new` is exactly `old` with one new `## [` section spliced in
+    at the point `assemble` splices -- i.e. every other byte of the file, before
+    and after, is untouched.
+    """
+    match = SECTION_START_RE.search(old)
+    cut = match.start() if match else len(old)
+    tail = old[cut:]
+    if new[:cut] != old[:cut]:
+        return None
+    if tail and not new.endswith(tail):
+        return None
+    inserted = new[cut : len(new) - len(tail)]
+    if not inserted.startswith("## ["):
+        return None
+    if len(SECTION_START_RE.findall(inserted)) != 1:
+        return None
+    return inserted
+
+
+def verify_changelog_matches_base(root: Path, old: str, new: str) -> None:
+    """Assert this change either leaves CHANGELOG.md alone or assembles a release.
+
+    `check` otherwise never reads CHANGELOG.md, so a hunk landing inside a
+    published section passes CI with a green tick and a clean merge (#297).
+    """
+    if new == old:
+        return
+    inserted = _leading_section_insertion(old, new)
+    if inserted is None:
+        raise Failure(
+            "CHANGELOG.md was modified. Entries go in changelog.d/ as one file per "
+            "change; CHANGELOG.md is written only by `changelog.py assemble` at the "
+            "release cut, and published sections are history. Since #295 removed "
+            "`## [Unreleased]`, an edit written for the old layout merges cleanly "
+            "into an already-published section instead of conflicting."
+        )
+    version = derive_version(root)
+    if not inserted.startswith(f"## [{version}] "):
+        raise Failure(
+            f"CHANGELOG.md gained a section that is not this build's release: "
+            f"expected `## [{version}] - <date>`, got {inserted.splitlines()[0]!r}."
+        )
+
+
+def _git(root: Path, *args: str) -> str:
+    result = subprocess.run(
+        ["git", "-C", str(root), *args], capture_output=True, text=True
+    )
+    if result.returncode != 0:
+        raise Failure(
+            f"`git {' '.join(args)}` failed: {result.stderr.strip() or 'no output'}"
+        )
+    return result.stdout
+
+
+def base_changelog(root: Path, base_ref: str) -> str:
+    """CHANGELOG.md as of the merge base with `base_ref`.
+
+    Fails loudly if the ref cannot be resolved: a lookup that degrades to an
+    empty string would make every comparison below read as "unchanged".
+    """
+    merge_base = _git(root, "merge-base", "HEAD", base_ref).strip()
+    if not merge_base:
+        raise Failure(f"cannot resolve a merge base with {base_ref}")
+    return _git(root, "show", f"{merge_base}:CHANGELOG.md")
+
+
 def main(argv: list[str]) -> int:
     parser = argparse.ArgumentParser(description=__doc__.split("\n")[0])
     parser.add_argument("command", choices=["check", "preview", "assemble"])
     parser.add_argument("--version", help="release version (default: read from the build)")
     parser.add_argument("--date", help="release date, YYYY-MM-DD (default: today)")
+    parser.add_argument(
+        "--base-ref",
+        help="check only: git ref this change is based on. CHANGELOG.md is compared "
+        "against it and must be either identical or a release assembly.",
+    )
     parser.add_argument(
         "--keep-fragments",
         action="store_true",
@@ -230,6 +348,16 @@ def main(argv: list[str]) -> int:
         print(f"changelog.d: {len(fragments)} fragment(s) valid")
         for issue, section, _body, path in fragments:
             print(f"  #{issue:<5} {section:<14} {path.name}")
+        current = (root / "CHANGELOG.md").read_text(encoding="utf-8")
+        verify_no_unreleased_section(current)
+        # Printed either way, so a log reader can see whether the comparison
+        # actually happened -- a silently skipped guard is the failure mode this
+        # whole check exists to close.
+        if args.base_ref:
+            verify_changelog_matches_base(root, base_changelog(root, args.base_ref), current)
+            print(f"CHANGELOG.md: checked against {args.base_ref}")
+        else:
+            print("CHANGELOG.md: no --base-ref; compared against nothing")
         return 0
 
     version = args.version or derive_version(root)
@@ -243,7 +371,9 @@ def main(argv: list[str]) -> int:
         return 0
 
     changelog = root / "CHANGELOG.md"
-    changelog.write_text(assemble(changelog, section, fragments), encoding="utf-8")
+    changelog.write_text(
+        assemble(changelog, version, section, fragments), encoding="utf-8"
+    )
     print(f"CHANGELOG.md: added [{version}] - {date} from {len(fragments)} fragment(s)")
     if not args.keep_fragments:
         for _issue, _section, _body, path in fragments:
--- a/.github/scripts/test-changelog.py
+++ b/.github/scripts/test-changelog.py
@@ -9,6 +9,7 @@ with a **negative control** that must fail, because a guard that cannot be shown
 to fire is indistinguishable from no guard at all.
 """
 
+import subprocess
 import tempfile
 import unittest
 from pathlib import Path
@@ -98,7 +99,7 @@ class Assembly(unittest.TestCase):
         self.section = changelog.render_section("2.0.0", "2020-02-02", self.fragments)
 
     def test_inserts_above_the_newest_released_section(self):
-        result = changelog.assemble(self.changelog, self.section, self.fragments)
+        result = changelog.assemble(self.changelog, "2.0.0", self.section, self.fragments)
         self.assertTrue(result.startswith("# Changelog\n\n## [2.0.0] - 2020-02-02\n"))
         self.assertIn("## [1.0.0] - 2020-01-01", result)
         # Purely additive: strip the new section, get the original back exactly.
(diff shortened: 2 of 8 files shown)


monkopedia-coder:
CI status on `7b35e076`: the three branch-protection-required checks are green — `check` (16m7s), `check-macos` (14m4s), `check-ios` (18m44s) — as is `lint`.

`Instrumented tests (emulator)` fails with 13 focus/keyboard failures. Not caused by this change and not a required check: every run of `ci-android.yml` in the last three days failed, including on branches that changed no Kotlin at all (`chore/281-ignore-pycache`, which edited only `.gitignore`, and `docs/release-doc-nits`). That constant is #275, and the specific failures are #259.

The new guard is confirmed to have actually run against the base commit rather than being skipped — from the `Validate changelog fragments` step log:

```
Ran 34 tests in 0.591s
OK
changelog.d: 6 fragment(s) valid
CHANGELOG.md: checked against 7163b0e6ca9487150eae5941ddc3ba345d680c5d
```

with `BASE_SHA: 7163b0e6ca9487150eae5941ddc3ba345d680c5d` in the step env. That second-to-last line is the anti-inert control described in the body: had the ref been empty, the step would still exit 0 but print `CHANGELOG.md: no --base-ref; compared against nothing`.

monkopedia-reviewer (approved):
**APPROVED** — reviewed at head `7b35e076` (the brief named `70d4ea44`; a force-push landed mid-review, see §0). Tier 1: build tooling, no shipped behaviour, no public API.

Everything below was measured on real trees, not fixtures. One finding, non-blocking, is at the end.

---

## 0. The head moved mid-review

`70d4ea44` → `7b35e076`, force-pushed 2026-09-02T23:28:23-04:00. The delta is **`changelog.d/README.md` only**, 5 lines of prose: `--base-ref origin/main` → "when given a base to compare against — as CI does, passing the pull request's base commit". That is an accuracy fix (CI passes `base.sha`, not `origin/main`) and it is correct.

I verified the other seven files are **byte-identical** across the two heads (`changelog.py`, `test-changelog.py`, `ci.yml`, `CLAUDE.md`, `README.md`, `docs/release-checklist.md`, `297.build.md`), then re-ran the suite (34/34) and the acceptance test below against `7b35e076`. Every measurement here is pinned to `7b35e076`.

---

## 1. The live acceptance test: the new `check` goes RED on a real defective tree

**The brief's premise is stale in one respect and correct in the other.** #268's *current* head (`57fca07f`, pushed 2026-09-01) has already been rebased onto `main` and migrated to `changelog.d/248.fixed.md` — it carries **no** `CHANGELOG.md` hunk today. So the live defect is no longer sitting on #268.

The tree the brief describes is real and still retrievable: `8af6fa06`, #268's head before the 2026-09-01 force-push (from the issue timeline). I used that, unmodified.

```
git merge 8af6fa06  onto  7b35e076        ->  exit 0, CLEAN.  mergeable is accurate.
resulting CHANGELOG.md:  5 bullets at lines 72-76
  ## [0.3.6] - 2026-08-30    line 7      <- already published
  ### Fixed                  line 61     <- the hunk lands HERE
  ## [0.3.5]                 line 112

python3 .github/scripts/changelog.py check --base-ref 7b35e076
  error: CHANGELOG.md was modified. Entries go in changelog.d/ ...
  exit=1
```

**Confirmed: the guard catches the real thing, inside the already-published `## [0.3.6]` → `### Fixed`.** I did not touch, modify, un-draft or push to #268 — read-only throughout.

---

## 2. Both directions, all three parts

### Part 1 — `check` sees a direct `CHANGELOG.md` edit

| case | tree | result |
|---|---|---|
| old-#268 merged onto head (real defect) | CHANGELOG.md edited | **RED** exit 1 |
| synthetic bullet appended to CHANGELOG.md | CHANGELOG.md edited | **RED** exit 1 |
| `## [Unreleased]` reintroduced, **no** `--base-ref` | heading present | **RED** exit 1 |
| PR304's own merge ref (`refs/pull/304/merge`) | fragment only | GREEN exit 0 |
| #268's **current** head merged onto head | fragment only | GREEN exit 0 |
| fresh branch off the base tip, fragment only | | GREEN exit 0 |
| **rebased** branch (branched pre-release, cherry-picked onto base) | | GREEN exit 0 |
| branch that **merged main into itself** (pre-release branch point) | | GREEN exit 0 |
| stale branch **behind** main, unrelated file only, **no fragment at all** | | GREEN exit 0 |
| **full release cut**: version bumped, assembled, fragments deleted | CHANGELOG.md edited | GREEN exit 0 |
| release assembly with the version file **not** bumped | | **RED** — "gained a section that is not this build's release: expected `## [0.3.6]`, got `## [0.3.7] - …`" |

The release-cut row matters most: the one PR that legitimately rewrites `CHANGELOG.md` still passes, so this does not block the next release.

### Part 2 — `assemble` refuses a duplicate released heading

Measured **before**, on `origin/main` as it stands (`version = "0.3.6"`, `## [0.3.6] - 2026-08-30` already in the file):

```
python3 .github/scripts/changelog.py assemble
  CHANGELOG.md: added [0.3.6] - 2026-09-02 from 5 fragment(s)
  changelog.d: removed 5 fragment(s)
  exit=0                       <- silent success
  '## [0.3.6]' headings: 2     <- a SECOND one
  fragments left: 0            <- all five deleted
```

Reproduced exactly as claimed, and the `-SNAPSHOT` guard is confirmed inert: `main` reads `0.3.6`, not a snapshot.

**After**, same tree:

```
bad :  assemble                     -> exit 1, "already has a [0.3.6] section",
                                       0 changed paths (CHANGELOG.md AND changelog.d untouched)
good:  assemble --version 0.3.7     -> exit 0, section inserted above [0.3.6], normal
```

I also checked the checklist's downstream claim against the code: `deploy.yml`'s extractor is `awk -v v="## [$VERSION]" 'index($0,v)==1{f=1;print;next} /^## \[/{f=0} f'`. The first rule wins on a second matching heading, so `f` **re-arms** and the release notes would carry both sections. The claim is correct.

### Part 3 — `changelog.d/README.md` states when a fragment is expected

It **argues** rather than asserting: a blanket requirement "would only produce a permanent stream of `N.internal.md` files reading 'internal refactor'", and "whether a change is user-visible is a judgement no script can make — asking for it every time turns that judgement into a rubber stamp." It resolves ties with "When in doubt, write one."

It explicitly does **not** make a fragment mandatory — "Not every pull request needs a fragment, and none is required", "`check` does not require one" — and `CLAUDE.md` and `README.md` say the same. Verified in the direction that matters: a PR adding **no fragment at all** stays green (table row above), and `check` against an empty `changelog.d/` exits 0.

---

## 3. Mutation results

**Baseline:** 15/15 on `main`; 34/34 on the PR.

**The 15-test / 3-failure property survives.** Neutering `_verify_additive` (an early `return` before `if not section.strip()`) fails **exactly 3**, and the **same 3 names**, on both sides:

```
main (15 tests):  3 failures   PR (34 tests):  3 failures
  test_additive_check_catches_a_dropped_fragment
  test_additive_check_catches_a_reworded_published_line
  test_additive_check_is_not_vacuous_on_empty_input
```

**Each new guard, mutated individually.** Each row names the exact line perturbed:

| # | line reverted | tests that catch it |
|---|---|---|
| G1 | `if re.search(rf"(?m)^## \[{re.escape(version)}\]", old):` → `if False:` | `test_refuses_a_version_already_released`, `test_refuses_it_even_when_the_date_differs` |
| G1b | same line, anchor weakened to `if version in old:` | `test_a_version_that_is_a_prefix_of_a_released_one_is_accepted` |
| G2 | `if UNRELEASED_RE.search(text):` → `if False:` | `test_rejects_a_reintroduced_unreleased_heading` |
| G2b | `UNRELEASED_RE` → bare `(?i)unreleased` | `test_does_not_fire_on_the_word_in_an_entry` |
| G3 | `verify_changelog_matches_base` → unconditional `return` | 7 tests |
| G4 | `if not inserted.startswith(f"## [{version}] "):` → `if False:` | `test_a_section_for_some_other_version_fails` |
| G5 | `if new[:cut] != old[:cut]:` → `if False:` | `test_an_assembly_that_also_edits_the_preamble_fails` |
| G6 | `if tail and not new.endswith(tail):` → `if False:` | `test_an_assembly_that_also_rewords_a_published_line_fails` |
| G7 | `if not inserted.startswith("## ["):` → `if False:` | `test_an_assembly_that_also_rewords_a_published_line_fails` |
| G8 | `if len(SECTION_START_RE.findall(inserted)) != 1:` → `if False:` | `test_two_sections_at_once_fails` |
| G9 | `_git`: `if result.returncode != 0:` → `if False:` | `test_a_file_absent_at_the_base_is_loud_not_empty` |
| **G10** | `base_changelog`: `if not merge_base:` → `if False:` | **none** |
| **G11** | `main()`: drop the `verify_no_unreleased_section(current)` **call** | **none** |
| **G12** | `main()`: drop the `verify_changelog_matches_base(...)` **call** | **none** |

**G1b and G2b are the rows I care about most.** Both are *over-strictness* mutations — they make a guard fire on the good case — and each has a dedicated test that turns red. That is the property the brief asked for: these guards cannot be quietly weakened *or* quietly broadened without a named test objecting.

**G10/G11/G12 survive, and all three are in the WIRING, not in a guard.** Every test calls the guard functions directly; nothing exercises `main()`. So the suite proves the guards work, not that `check` runs them. The structural cause is `repo_root() = Path(__file__).resolve().parents[2]`, which pins the script to the real repo and makes `main()` awkward to drive against a synthetic tree — the same boundary `main`'s existing 15 tests already had, not a regression introduced here. G10 is additionally unreachable in practice: `git merge-base` exits non-zero on failure, so `_git` raises before the empty-string branch can be taken, and `test_an_unresolvable_ref_is_loud_not_empty` covers the reachable path.

I closed the G11/G12 gap by hand instead: the acceptance test in §1 and every row in §2's table drive `main()` end-to-end. **And this PR's own CI run closes it in the real pipeline** — the `check` job's "Validate changelog fragments" step logs:

```
Ran 34 tests in 0.591s
OK
changelog.d: 6 fragment(s) valid
CHANGELOG.md: checked against 7163b0e6ca9487150eae5941ddc3ba345d680c5d
```

That third line is the guard reporting that it actually ran, against `base.sha` = current `main`. So the wiring is verified in CI as well as locally; it is simply not pinned by a unit test. The author left exactly this log-visible control on purpose ("a silently skipped guard is the failure mode this whole check exists to close"). A future test that shells out to `check` in a temp repo would pin it; not worth blocking on.

---

## 4. `ci.yml` — what the 18 lines do, and how they behave on a rebase

Three things, all in the `check` job only (the `lint` job's checkout is untouched at depth 1):

1. `fetch-depth: 0` on `actions/checkout` — a depth-1 checkout of `refs/pull/N/merge` has no parents, so `base.sha` would not exist locally.
2. `env: BASE_SHA: ${{ github.event.pull_request.base.sha }}`.
3. `changelog.py check ${BASE_SHA:+--base-ref "$BASE_SHA"}`.

I checked the shell construct rather than assuming it: with `BASE_SHA` set it expands to exactly two words `--base-ref` / `<sha>` (quote removal applies inside `${…:+…}`); empty, it expands to nothing. So `push` and the `workflow_call` from `deploy.yml` (both have no `pull_request` payload) run the fragment + `[Unreleased]` checks and skip the base comparison, printing which branch they took.

**Behaviour on the shapes that matter.** `base.sha` is the merge commit's first parent, so `merge-base HEAD $BASE_SHA` resolves to `base.sha` itself:

```
fresh branch off the base tip                 GREEN
rebased branch (pre-release branch point)     GREEN
branch containing a merge of main             GREEN   HEAD parents: 02c14671 70d4ea44
stale branch behind main, untouched           GREEN
```

A CI-side guard that misfires on a rebase would be disabled within a week; this one does not.

**One bounded caveat, and it fails closed.** If `base.sha` is stale enough that the merge base predates **two or more** `CHANGELOG.md` changes on `main`, `_leading_section_insertion` sees more than one inserted section (or a deletion) and returns `None` → RED with the generic "CHANGELOG.md was modified" message, which would read as a false accusation. I reproduced it by passing a pre-0.3.6-release base: RED. Post-#295 `main`'s `CHANGELOG.md` only changes at release cuts, so this needs a base two releases stale. It is loud, on a PR, and cleared by updating the branch — recording it so that whoever eventually sees it does not weaken the guard instead.

---

## 5. Docs — no contradiction reintroduced (#292/#293)

Read `CLAUDE.md` from `origin/main` via `git show`, not the working tree (this checkout is pinned to `release/0.3.4` and carries the pre-#295 text; md5 `2beadd96` vs `571beb21` vs the PR's `f4b607ab`).

`CLAUDE.md` (authoritative) and `docs/release-checklist.md` §3 now list the same four `assemble` guarantees in the same order — empty `changelog.d/`, `-SNAPSHOT`/underivable version, **already-released heading**, purely-additive. The checklist only *adds* detail (the step-8 back-reference, which exists; and the `deploy.yml` extractor consequence, which I verified against the awk). Nothing in the checklist disagrees with `CLAUDE.md`.

The when-to-write-a-fragment criterion is consistent across `CLAUDE.md`, `README.md` and `changelog.d/README.md`, and none of them makes it a requirement.

*Minor observation, not a defect:* `CLAUDE.md`/`README.md` list "the build, packaging, or release process" as a write-one trigger while `changelog.d/README.md` lists "CI plumbing" as legitimately omitted. This PR is arguably both and wrote one (`297.build.md`, section `build`). `changelog.d/README.md` states the tiebreak explicitly — "When in doubt, write one" — so it reads as granularity, not conflict.

---

## 6. FINDING (non-blocking): "three of eight releases" is **four of eight**

The load-bearing empirical claim for guard #2 is off by one. Measured over `main`'s history of `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts`, a post-release `-SNAPSHOT` bump followed **four** of the eight releases:

```
v0.1.0  ->  2b2584e6  0.1.0 -> 0.2.0-SNAPSHOT   "Bump version to 0.2.0-SNAPSHOT for development"
v0.2.0  ->  c18b48e6  0.2.0 -> 0.2.1-SNAPSHOT   "Bump version to 0.2.1-SNAPSHOT for development"
v0.3.0  ->  none
v0.3.2  ->  none
v0.3.3  ->  5f87230a  0.3.3 -> 0.3.4-SNAPSHOT   "Open next dev snapshot (#159)"
v0.3.4  ->  none
v0.3.5  ->  9104fb88  0.3.5 -> 0.3.6-SNAPSHOT   "Migrate to Compose-Multiplatform 1.11.1 (#182)"
v0.3.6  ->  none
```

The fourth is easy to miss: it rode inside an unrelated migration PR rather than a dedicated bump commit, so a search for *"Bump … SNAPSHOT"* commit subjects returns three. The number appears in two places — the `changelog.py` module docstring and `changelog.d/297.build.md`. The other three sites say "usually skipped" with no figure, which is accurate at 4/8.

**Why this is not blocking.** The conclusion is unchanged and independently confirmed: half the releases skip the bump, `main` reads `0.3.6` today, and I reproduced the second-heading defect directly rather than inferring it from the count. Nobody following the artefact is misled about behaviour.

**Why it should still be fixed.** `297.build.md` is unassembled, so the wrong figure has not shipped yet; correcting it before the release cut costs two words. Recording it here rather than spending a full macOS/iOS CI cycle on a prose fix.

---

## 7. Hygiene

- `closingIssuesReferences` polled to 10 consecutive identical reads, **twice** (before and after the force-push): `[297]`. Local regex gate over the PR body + every commit message finds only `Fixes #297` (×2), with a positive control that fires. #268, #294, #299, #300, #301, #302 and #303 carry no closing keyword and stay open.
- **The PR does not touch `CHANGELOG.md`** — its `CHANGELOG.md` is byte-identical to `main`'s, and its entry is `changelog.d/297.build.md`. A fair self-test of the guard it adds.
- `Instrumented tests (emulator)` is red at its deterministic floor and is not in the required set. Established by **set comparison, not by count** — `13 failed` would match equally well if a regression had displaced a known failure while the total sat still. Extracted the failing names from both heads' runs (`sed 's/\^\[\[[0-9;]*m//g'`, since the log carries ANSI escapes as the literal two characters `^[`), and `diff` reports the two sets **identical**: the same 13, `FocusManagementTest` (5) + `KeyboardHandlingTest` (8). Positive control run: `diff` does report a difference when one is introduced. The named baseline is recorded on #275.

```
70d4ea44   96 executed, 13 failed, 0 skipped
7b35e076   96 executed, 13 failed, 0 skipped   <- same SET, diff clean
```

  The floor's executed count moved 93 -> 96 when #296 landed on `main` on 2026-09-01, not here: `chore/281-ignore-pycache` @ 03:44Z reported 93/13/0, `fix/248-search-defects` @ 04:30Z reported 96/13/0. This PR changes no Kotlin or Gradle file.

Issue #308, opened 2026-09-03
:lsp-client cannot reach the JDK 11 floor — com.monkopedia.lsp ships class-65 bytecode


monkopedia-coder:
PR #305 moves kodemirror's own published bytecode from class 65 to class 55 (#237), so consumers on
JDK 11+ can load the artifacts. **`:lsp-client` is the exception, and not because of anything in
this repo.**

Its dependency `com.monkopedia.lsp:lsp` ships **class-65** bytecode. Measured, with a control:

```
lsp-jvm-1.2.1              {65: 1450}
control kotlin-stdlib      {52: 973, 53: 1}   <- proves the reader discriminates rather than
                                                 printing a constant
```

`gradle/libs.versions.toml` pins `lsp = "1.2.1"`.

## What this means

A consumer on JDK 11–20 can use `:state`, `:view`, `:vim` and the rest, but **`:lsp-client` will
still fail at runtime** — its transitive dependency needs 21. So after #237 lands, the effective
floor is **11 for 56 modules and 21 for one**, and nothing communicates that.

This was observed concretely during #305's review: building with `jvmToolchain(11)` made
`:lsp-client:jvmTest` fail **11 of 11** with `UnsupportedClassVersionError` — the dependency, not
kodemirror's own code.

## Options, none obviously right

1. **Rebuild `com.monkopedia.lsp` at class 55 and bump the pin.** It is a Monkopedia-owned artifact,
   so this is available in a way it would not be for a third party. Cleanest outcome; costs a
   release of another project.
2. **Document the split floor** — `:lsp-client` requires JDK 21, everything else 11. Cheap, honest,
   and leaves consumers to discover it from a runtime error unless they read the docs.
3. **Accept and record it as a known limitation** on #237's follow-up, doing nothing now.

**Not a blocker for #305.** That PR strictly widens compatibility for 56 of 57 modules; this issue
is about the one it cannot reach and about saying so.

## Worth checking before acting

Whether any *other* published dependency is above class 55 — the same measurement over the full
resolved runtime classpath, with the same control. `:lsp-client` was found because #305's route
experiment happened to exercise it; **a module nobody built at 11 would not have surfaced.**

Related: **#307** (CI never asserts the published class-file version) would catch a regression in
kodemirror's own output but **not** this, since the dependency is not built here — worth noting so
the two are not conflated.

monkopedia-coder:
[...]

Issue #307, opened 2026-09-03
CI never asserts the published class-file version, so #237 can regress silently


monkopedia-coder:
Nothing in CI checks what class-file version the published artifacts carry, so **#237 can regress
silently** the moment someone changes a toolchain, a compiler flag, or the runner's JDK.

```
$ git grep -cE 'class-file|major version|javap|classfile' origin/main -- '.github/'
(no matches)
control: 5 files under .github/ match 'gradlew', so the search works
```

## Why a build-config check is not enough

**#237's whole point is that the config and the artifact disagreed.** `android { compileOptions {
source/targetCompatibility = VERSION_11 } }` was declared for three releases while the shipped
`classes.jar` was 188/188 at **class 65**. A green build proved the config parsed. It proved nothing
about the bytecode.

So the assertion has to read the **artifact**, not the build script.

## Suggested shape

After the publish-to-local step, walk the built jars and the Android `classes.jar`, read the two
bytes at offset 6–7 of every `.class` entry, and fail if any exceeds the declared floor.

**Two things make the difference between a real check and a decorative one:**

1. **A positive control that proves the reader discriminates.** `kotlin-stdlib` reads
   `{52: 973, 53: 1}`. A reader that printed a constant would look identical on a passing run.
2. **Both directions.** Bytecode above the floor must go **RED**; bytecode at the floor must stay
   **GREEN**. A check that reddens on both is broken in the direction that reads as strictness, and
   gets weakened by the next person who cannot tell the difference.

Assert on **all** publishing modules rather than a sample — the defect in #237 was uniform, but a
partial regression (one module compiled by a different path) is exactly what a sampled check would
miss.

## Related

* **#237** — the defect this guards against; PR #305 fixes the current instance.
* **#300** — `verifyBomCoverage` matching `<artifactId>` only, same family: a guard that checks the
  shape of the published set without checking its content.
* **#275** — the general form. A check that cannot distinguish a passing state from an unmeasured
  one is not evidence.

Filed as a follow-up flagged by both the #237 dispatch and PR #305's review, and recorded by
neither.

monkopedia-coder:
## Nightly triage 2026-09-03 — `agent-workable` → `blocked` on #237 / PR #305

Not a downgrade of the finding, which is sound and independently verified — `git grep -cE
'class-file|major version|javap|classfile' origin/main -- '.github/'` is genuinely empty, with the
issue's own `gradlew` positive control proving the search discriminates.

[...]

Issue #299, opened 2026-09-01
CI runs no ktlintCheck and no spotlessCheck — the Apache-2.0/NOTICE license header on 57 published modules is enforced by nothing


monkopedia-coder:
`ci.yml`'s only gating job is named **"Lint and test (JVM + wasmJs + Android)"** and it runs no
style or license check at all. `ktlintCheck` and `spotlessCheck` — both configured, both working —
appear in **zero** workflow files. Measured on `origin/main` @ `7163b0e6`:

```
$ grep -rn "ktlint\|spotless\|gradlew" .github/workflows/
ci-android.yml:59:   ./gradlew :view:assembleDebugAndroidTest
ci-android.yml:114:  ./gradlew :view:connectedDebugAndroidTest ...
docs.yml:34:         ./gradlew :dokkaGenerate
docs.yml:50:         ./gradlew :samples:showcase:wasmJsBrowserDistribution ...
ci-apple.yml:19:     ./gradlew macosArm64Test
ci-apple.yml:38:     ./gradlew iosSimulatorArm64Test
ci.yml:38:           ./gradlew apiCheck jvmTest :kodemirror-bom:verifyBomCoverage
ci.yml:97:           ./gradlew lintDebug          # Android Lint, and NOT in branch protection
deploy.yml:50:       ./gradlew publishToMavenCentral
```

`ci.yml:29` says it plainly: *"this job does not run `check`"*. `ktlintCheck` and `spotlessCheck`
hang off `check`, so naming tasks explicitly dropped them. Branch protection requires exactly
`check`, `check-macos`, `check-ios` (`gh api repos/Monkopedia/kodemirror/branches/main/protection`
→ `["check","check-macos","check-ios"]`), and none of those three runs either task.

The consequential half is **`spotlessCheck`**. The convention plugin
(`convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:221-226`) configures
`licenseHeaderFile(rootProject.file("spotless/license-header.kt"))` over `src/**/*.kt` for all 57
published modules. That is the mechanism keeping the Apache-2.0 header — and the
"Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT. See NOTICE file for
details." attribution line — on every source file of an MIT-derived port. Nothing in CI verifies it.

## Demonstration

Removed the whole license header from `state/src/commonMain/.../CharCategory.kt` and added two
[...]

PR #310, opened 2026-09-04
Raise the soft keyboard on mobile browsers (#303)

### LINKED ISSUE #303: No Soft Keyboard on Mobile with Wasm Builds

monkopedia-coder:
## What

Three edits in `view/`, per the investigation on #303:

1. `KodeMirror.kt` — `awaitFirstDown().also { it.consume() }`, matching `detectTapAndPress`.
2. `KodeMirror.kt` — hoist `LocalSoftwareKeyboardController.current` and `?.show()` in the same
   gesture, mirroring `CoreTextField.requestFocusAndShowKeyboardIfNeeded`.
3. Delete `platformFocusInput()` / `platformFocusCanvas()` — dead code. `internal expect/actual`,
   so **`apiCheck` is unmoved** (verified, no `.api` diff).

Plus `TouchGestureTest`, the first touch-pointer coverage of the editor gesture.

## Reproduced first, then measured after

Everything below is Chromium (Playwright + CDP) against the showcase's `?test=true` page, which
exposes `globalThis.__kodemirror.state`. Instrumented `HTMLElement.prototype.focus`,
`focusin`/`focusout`, and `pointerdown.defaultPrevented` read after the dispatch completes.
Same script, same coordinates, two builds: `dd7dea8b` and this branch.

### D1 — the pointer down is now consumed

Touch taps, six per run, each from a fresh load:

| | before | after |
|---|---|---|
| `pointerdown.defaultPrevented` | **false 6/6** | **true 6/6** |

Mouse clicks, DOM focus after the click:

| | before | after |
|---|---|---|
| `deepActive()` ends on | **CANVAS 4/4** | **TEXTAREA 4/4** |

The before-state is the reported precondition: focus starts on
`TEXTAREA.compose-backing-field`, the click moves it to the canvas, and it stays there.

### D2 — the keyboard is now requested

`.focus()` calls made during each gesture (this is `keyboardController.show()` →
`BackingDomInput.focus()`, which focuses twice on purpose):

| | before | after |
|---|---|---|
| touch | **0 per tap, 6/6 taps** | **2 per tap, 6/6 taps** |
| mouse | **0 per click, 4/4** | **2 per click, 4/4** |

**The two are separable, and neither result implies the other.** `show()` cannot make Compose call
`preventDefault()`, and consuming a pointer cannot produce `.focus()` calls — so the
`defaultPrevented` column is evidence for (1) alone and the `.focus()` column for (2) alone.

## Correction to the investigation, measured

The investigation predicted `focusout: TEXTAREA → focusin: CANVAS` on **touch**. On the CMP version
`main` builds against (1.11.1) that does **not** happen, and the reason matters:
`ComposeWindow.initEvents` unconditionally `preventDefault()`s `touchstart` on the canvas, so the
browser never synthesises the compat mouse events and never runs the focus-on-click default action.
Stack trace from a patched `Event.prototype.preventDefault`:

```
touchstart
  at androidx.compose.ui.window.ComposeWindow$initEvents$lambda.invoke
  at androidx.compose.ui.window.ComposeWindow$addTypedEvent$lambda.invoke
  at HTMLCanvasElement.eval
```

The canvas-focus theft is real on 1.11.1, just not via touch — **a mouse click reproduces it
exactly** (CANVAS 4/4 above). The investigation ran the reporter's CMP 1.12.0 reproducer, where the
touch path is not masked. So on 1.11.1 D2 is the dominant defect and on 1.12.0 both bite; both are
fixed here.

One asymmetry worth recording: after the fix, `pointerdown.defaultPrevented` is `true` for touch and
still `false` for mouse — Compose for Web only cancels the default for touch pointer events. The
mouse case is repaired by (2) instead: the `.focus()` calls put focus back on the textarea.

Independently confirmed for edit (3): `document.body.shadowRoot` is **`null`** on the live page
(shadow host is `HTML > BODY > DIV > DIV`), so `platformFocusCanvas()`'s `if (shadow)` never fired.
It has never focused anything.

## Gesture regression — the actual risk

Consuming the down changes what siblings and children observe. Re-exercised **on both pointer
types**, before and after, identical results:

Real browser (`?test=true`, wasm):

| gesture | before | after |
|---|---|---|
| touch tap places cursor | pos 0 → 30 (line 2 col 8) | same |
| typing after touch tap | doc 229 → 232 | same |
| horizontal touch drag | no selection (see below) | no selection |
| mouse click places cursor | pos 0 → 30 | same |
| typing after mouse click | doc 229 → 232 | same |
| mouse drag selects | anchor 2 head 98 | anchor 2 head 98 |
| **Alt+drag rectangular** | 4 ranges | 5 ranges (both multi-range) |

Compose UI tests, run on **jvm (329 tests) and wasmJs (296 tests)**, all green — including the
existing `DragSelectionTest` (within-line, across-lines, reversed, bottom-to-top),
`FocusManagementTest`, `ScrollClickTest`, `MixedContentScrollTest`, and the new `TouchGestureTest`
(touch tap on two lines, horizontal touch drag, vertical touch drag scrolls the `LazyColumn`).

Why children are unaffected: Compose dispatches the Main pass leaf → root, so the line list's
scroll and the gutter's `clickable` have already seen the down before this node consumes it.

## What is characterised rather than asserted

`horizontalTouchDragPlacesCursorAndLeavesDocument` locks in today's behaviour, it does not endorse
it: a horizontal touch drag is claimed by the line's horizontal scroll container, so the editor
falls back to placing the cursor at the down position and no selection is made. **Measured
identical before and after** — it is not a regression, and filing it separately.

I first wrote that as an assertion that touch drag *selects*. It failed at `dd7dea8b` too, which is
the only reason I know it is pre-existing. Same for a JVM `performKeyInput`/`performMouseInput`
formulation of Alt+drag: it produced one range at baseline as well, while the same gesture in a real
browser produces 4–5. I did not chase that harness discrepancy — it is not #303's ground — and
Alt+drag is covered above by the browser run instead.

## What I could NOT verify

- **No real device and no actual soft keyboard.** Desktop Chromium has none. What is measured is the
  DOM-focus precondition and the keyboard-request call, not a keyboard appearing.
- **iOS Safari and Firefox for Android: untested.** The reporter says those never worked at all.
  iOS refuses programmatic `focus()` outside a trusted gesture, which could be a second obstacle
  that survives this change — `show()` here does run inside the pointer-down, which is the right
  shape for it, but I have no evidence either way. **This PR does not claim mobile-wide resolution.**
- **CMP 1.12.0 was not exercised at runtime**, only 1.11.1 (what `main` builds against). The 1.12.0
  behaviour above is the investigation's measurement, not mine.

## Verification commands

```
./gradlew ktlintFormat :view:jvmTest :view:apiCheck      # 329 tests, 0 failures, no .api diff
CHROME_BIN=/usr/bin/chromium ./gradlew :view:wasmJsBrowserTest   # 296 tests, 0 failures
./gradlew spotlessCheck ktlintCheck apiCheck             # green
python3 .github/scripts/changelog.py check               # 7 fragments valid
```

Test-result XML was deleted before each run and mtimes checked; `gap-analysis` was not used.

Fixes #303

### DIFF
--- a/changelog.d/303.fixed.md
+++ b/changelog.d/303.fixed.md
@@ -1,0 +1,10 @@
+- Fixed the soft keyboard not appearing when the editor is tapped in a browser on mobile (#303).
+  The editor's tap gesture did not consume the pointer down, so Compose for Web never called
+  `preventDefault()` on it and the browser's default action moved DOM focus to the Compose
+  `<canvas>`, blurring the backing `<textarea>` a soft keyboard is raised for; and because the
+  hidden field already held Compose focus, `requestFocus()` on tap was a no-op that never asked
+  the platform to show the keyboard. The gesture now consumes the down and calls
+  `SoftwareKeyboardController.show()`, matching what `BasicTextField` does. Verified in a browser
+  by measuring DOM focus and `pointerdown.defaultPrevented` before and after; a real device and an
+  actual on-screen keyboard were not exercised, and iOS Safari and Firefox for Android may have a
+  further obstacle of their own.
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -87,6 +87,7 @@ import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.platform.LocalClipboardManager
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalSoftwareKeyboardController
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.TextLayoutResult
 import androidx.compose.ui.text.input.TextFieldValue
@@ -250,11 +251,17 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
 
     // Focus management
     val focusRequester = remember { FocusRequester() }
+    // Requesting Compose focus is not enough to raise a soft keyboard once the
+    // hidden field already holds focus: with no focus *change* there is no new
+    // startInput, so the platform is never asked to show the IME. Compose's own
+    // BasicTextField calls show() explicitly for that case
+    // (CoreTextField.requestFocusAndShowKeyboardIfNeeded); the editor's tap
+    // handler below does the same (#303).
+    val keyboardController = LocalSoftwareKeyboardController.current
 
     // Auto-focus the hidden text field when the editor first appears
     LaunchedEffect(Unit) {
         focusRequester.requestFocus()
-        platformFocusInput()
     }
 
     // Track whether the platform callback already handled this keydown.
@@ -559,10 +566,21 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                         // separate tap/drag coroutines competed for the DOWN
                         // event and focus could be skipped.
                         awaitEachGesture {
-                            val down = awaitFirstDown()
-                            // Always focus immediately on any pointer down
+                            // Consume the down, as detectTapAndPress does for a
+                            // text field. On wasmJs that is what makes Compose
+                            // call preventDefault() on the browser event;
+                            // without it the default action moves DOM focus to
+                            // the tabindex=0 <canvas> and blurs the backing
+                            // <textarea>, the only element a soft keyboard is
+                            // raised for (#303). Children (line-list scroll,
+                            // gutter clicks) see the Main pass first, so this
+                            // does not take the gesture away from them.
+                            val down = awaitFirstDown().also { it.consume() }
+                            // Focus immediately on any pointer down, and ask for
+                            // the keyboard even when focus does not change —
+                            // otherwise one the user dismissed never comes back.
                             focusRequester.requestFocus()
-                            platformFocusInput()
+                            keyboardController?.show()
 
                             val downPosition = down.position
                             var isDrag = false
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/Platform.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/Platform.kt
@@ -113,17 +113,6 @@ internal expect fun platformRegisterKeyHandler(
 internal expect fun platformUnregisterKeyHandler(token: PlatformKeyHandlerToken)
 
 /**
- * Ensure the editor's backing text input element has platform focus.
- *
- * On wasmJs, Compose's FocusRequester.requestFocus() gives Compose-internal
- * focus but doesn't always move DOM focus to the textarea that Skiko creates.
- * This function explicitly focuses the textarea in the shadow DOM.
- *
- * On JVM/Desktop, this is a no-op since Compose manages focus natively.
- */
-internal expect fun platformFocusInput()
-
-/**
  * Write [text] to the underlying system clipboard.
  *
  * On wasmJs the canvas-rendered editor has no DOM `contenteditable`, so the
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/FocusManagementTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/FocusManagementTest.kt
@@ -32,53 +32,34 @@ import kotlin.test.Test
 import kotlin.test.assertTrue
 
 /**
- * Tests for focus management after canvas/editor clicks.
+ * Tests for focus management after editor clicks.
  *
- * # Issue #2: Canvas click can break keyboard input
+ * # Issue #2: a tap/drag race could skip `requestFocus()`
  *
- * ## Root cause analysis
+ * The editor used to run separate tap and drag gesture coroutines that each
+ * competed for the DOWN event, so focus could be skipped depending on which
+ * won. They are now one `awaitEachGesture` block that requests focus on the
+ * down before deciding tap-vs-drag. These tests hold that: click, repeated
+ * click, drag, and click-after-drag must all leave the hidden field focused
+ * and keyboard navigation working.
  *
- * The editor Box has THREE separate `pointerInput` modifiers:
- *   1. `detectTapGestures`  — calls `focusRequester.requestFocus()` on tap
- *   2. `detectDragGestures` — calls `focusRequester.requestFocus()` on drag start
- *   3. `awaitPointerEventScope` — hover tracking (no focus involvement)
+ * # Issue #303: DOM focus and the soft keyboard on wasmJs
  *
- * Each `pointerInput` block runs in its own coroutine and competes for pointer
- * events. When the user clicks (press + quick release), Compose's gesture
- * detection applies "slop": a pointer move below ~18px is still considered a
- * tap, but both coroutines receive the DOWN event simultaneously.
+ * Compose focus is not the whole story in a browser. Compose's canvas carries
+ * `tabindex="0"`, so an un-`preventDefault()`ed pointer down moves *DOM* focus
+ * to the canvas and blurs the backing `<textarea>` — the only element a browser
+ * raises a soft keyboard for — and Compose-web only calls `preventDefault()`
+ * when the gesture consumed a change. The editor therefore consumes the down,
+ * and calls `SoftwareKeyboardController.show()` in the same gesture because a
+ * `requestFocus()` on an already-focused field is a no-op and never asks the
+ * platform for the IME.
  *
- * On Desktop/JVM, `detectDragGestures` requires the pointer to move past the
- * drag slop threshold before it starts consuming events, so a pure click
- * normally reaches `detectTapGestures` intact.
- *
- * On **wasmJs**, the situation is more complex:
- *   - `platformFocusInput()` calls `platformFocusCanvas()`, which focuses the
- *     `<canvas>` element in Skiko's shadow DOM. A canvas click can momentarily
- *     move *browser* DOM focus to the canvas, away from any hidden textarea.
- *   - If `detectTapGestures` does not fire (e.g., drag detector consumed the
- *     event before the tap completed), `focusRequester.requestFocus()` is never
- *     called, leaving the BasicTextField without Compose focus.
- *   - The `recentlyDragged` flag (set in `onDragStart`, cleared in tap) adds
- *     another wrinkle: if drag fires first and sets `recentlyDragged = true`,
- *     the subsequent tap callback early-returns and skips `requestFocus()`.
- *
- * ## Why this test can/cannot reproduce the bug on JVM
- *
- * On JVM (Desktop Compose), `platformFocusInput()` is a no-op. Focus is
- * managed entirely by Compose's `FocusRequester`. The drag/tap race exists in
- * the pointer-input pipeline, but in the Desktop test environment Compose
- * correctly hands a click (no slop exceeded) to `detectTapGestures`, so
- * `requestFocus()` fires and the BasicTextField is focused.
- *
- * **The bug is wasmJs-specific**: it only manifests when:
- *   a) the browser moves DOM focus to the canvas on click, AND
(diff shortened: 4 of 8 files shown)


monkopedia-reviewer (approved):
Independent review of `bf8bbcce`. Category `bug_fix`, tier 1. Everything below was
re-measured here rather than read out of the PR body; where my numbers differ from the
author's I say so.

## 1. `FocusManagementTest` +35/−70 — no coverage was removed

This was the item I was most worried about: five of the thirteen names in the recorded
`#259`/`#275` instrumented baseline live in this class, and a test that keeps its name while
weakening its assertion is exactly how a name-keyed baseline silently changes meaning.

I stripped comments from both revisions of the file and diffed the executable content
(positive control: `assertTrue` count 5 on both sides, so the stripper preserves code):

```
139,140c139
<   "This suggests `recentlyDragged` was still true when the tap fired, " +
<   "causing requestFocus() to be skipped."
---
>   "This suggests the tap after a drag skipped requestFocus()."
```

**That is the entire non-comment change.** One failure-*message* string. Zero assertion
predicates, zero `performClick`/`performTouchInput` calls, zero test bodies. The −70 is KDoc
describing the pre-#2 three-`pointerInput` architecture that no longer exists. All six tests
survive with byte-identical executable content.

## 2. The instrumented baseline moved by exactly the new tests

From the `Report executed test count` step on this head (`run 33833689844`):

```
FocusManagementTest    6   5   0     <- identical to the #259 baseline
KeyboardHandlingTest  10   8   0     <- identical to the #259 baseline
TouchGestureTest       4   0   0     <- new
EXECUTED: 100   recorded: 100   failures: 13   errors: 0   skipped: 0
```

96 → 100 (+4 = exactly `TouchGestureTest`), 13 → 13. I checked the thirteen **by name**, not by
count — a substituted failure would also read 13:

```
FocusManagementTest   dragThenKeyboard_cursorMoves, multipleClicks_eachPreservesFocus,
                      clickOnEditor_givesFocusToTextField, clickAfterDrag_focusIsRestored,
                      clickThenArrowRight_cursorMoves
KeyboardHandlingTest  arrowRight_movesCursorForward, arrowLeft_movesCursorBackward,
                      arrowDown_movesToNextLine, arrowUp_movesToPreviousLine,
                      home_movesToLineStart, end_movesToLineEnd,
                      modX_cutsSelectedText_onMacFamily, modX_cutsSelectedText_offMacFamily
```

Exactly the #259 set, unchanged in both directions. **This PR does not fix #259**, and neither
the PR body nor `changelog.d/303.fixed.md` claims it does — I grepped: `#303` is the only issue
number anywhere in the body. The baseline needs extending (`TouchGestureTest 4 0 0`), not
resetting; set comparison keeps working across this merge.

## 3. The 40 removed `expect`/`actual` lines, and `apiCheck` across six coordinates

Confirmed by running it, not by inferring from the absence of a dump in the diff.
`./gradlew :view:jvmTest apiCheck` — `BUILD SUCCESSFUL`, 2425 tasks executed, 57 `apiCheck`
tasks, no `FAILED` line, **and the worktree was byte-clean afterwards** (a silent regeneration
would have dirtied it). For `:view` specifically, all six coordinates:

```
view/api/jvm/view.api        <- :view:jvmApiCheck        executed
view/api/android/view.api    <- :view:androidApiCheck    executed
view/api/view.klib.api       <- :view:klibApiCheck       executed, after klibApiMerge of
  // Targets: [iosArm64, iosSimulatorArm64, macosArm64, wasmJs]
     iosArm64ApiBuild + iosSimulatorArm64ApiBuild + macosArm64ApiBuild + wasmJsApiBuild
```

`platformFocusInput` has never appeared in any checked-in `.api` on `main` (positive control:
`KodeMirror` is present in all three dumps), so `internal expect/actual` really does not surface
here.

Callers, positive-controlled (`platformRegisterKeyHandler` → 8 hits, so the pattern works):
`platformFocusInput` and `platformFocusCanvas` → **0 hits** across all source sets; `shadowRoot`
→ **0 hits** in any `.kt`/`.js`/`.html`. The only surviving mention is a historical CHANGELOG line.

**And the deletion is safe on both branches of the "is it dead?" question, which is stronger
than the dead-code argument the PR makes.** If `document.body.shadowRoot` is null the function
never fired; if it were ever non-null, the function focused the **canvas** — precisely the
element this fix exists to keep focus away from. There is no configuration in which keeping it
helps.

The deletion also removes the call from the initial-auto-focus `LaunchedEffect`, and the sixth
`FocusManagementTest` case — `initialAutoFocus_textFieldIsFocused`, the one that *passes* on the
emulator — still passes at this head. So dropping it did not cost initial focus on a real Android
surface either.

## 4. Gesture regression pass — my own, touch and mouse, base vs head

I built the showcase at both `dd7dea8b` and `bf8bbcce`, served each, and drove them with my own
Playwright script (**not** the `gap-analysis` harness — its green-skip behaviour makes it
useless as evidence). Same script, same coordinates, both builds:

| gesture | `dd7dea8b` | `bf8bbcce` |
|---|---|---|
| mouse click places cursor | pos 60, line 3 col 14 | identical |
| **mouse drag selection** | anchor 24 head 121 | **anchor 24 head 121** |
| **Alt+drag rectangular** | **7 ranges** | **7 ranges, identical from/to/anchor/head on all 7** |
| touch tap places cursor | pos 60 | identical |

(The PR reported Alt+drag as 4 → 5 ranges "both multi-range". At my coordinates it is 7 → 7
with the range lists identical element for element — a tighter result than the author's, in the
same direction.)

Scroll-by-drag and the child-handler risk are covered in the Compose suites, and I checked those
against the base directly: I copied `TouchGestureTest.kt` onto a `dd7dea8b` worktree and ran
`:view:jvmTest` there — **329 recorded, 0 failed**, same as the head. So all four new touch
cases, including `verticalTouchDragScrollsDocument` (the LazyColumn-child case that consuming the
down could plausibly have broken), pass identically without the consume. The consume changed
nothing about the gesture behaviour these tests can see.

The structural reason holds up on reading: Compose dispatches Main leaf → root, so the line
list's scroll and the gutter's `clickable` have already seen the down before this node consumes
it. The hover `pointerInput` later in the same chain is inner and uses an unconditional
`awaitPointerEvent(Main)`, so it is unaffected either way.

**Are the four new cases real, or happy-path assertions?** I perturbed them rather than trusting
the count. `TouchGestureTest.kt:119` `after > before` → `after < before`:

```
jvm   FAIL verticalTouchDragScrollsDocument  firstVisibleIndex 0 -> 23
wasm  FAIL verticalTouchDragScrollsDocument  firstVisibleIndex 0 -> 23   (1 of 296)
```

and `:53` `assertCursorOnLine(2)` → `(3)`:

```
jvm   FAIL touchTapPlacesCursor  Expected cursor on line 3 but was on line 2 (head=13)
```

Real measured state on both platforms, so the suite is not inert — and `runEditorTest` returns
`runComposeUiTest {…}` directly, so the wasmJs promise is awaited (the #200 shape). Files restored;
worktree clean.

Judging the four cases on their merits: two are tap placement (happy path), one characterises
pre-existing horizontal-drag behaviour, and **one — `verticalTouchDragScrollsDocument` — is the
only case that actually exercises the consume risk**, and it is the right one, because the
LazyColumn is the child most likely to lose a gesture. What the file does *not* cover is touch
drag-selection (pre-existing: the horizontal scroll container claims it) or Alt+drag (not a touch
gesture). Those two are covered by my browser run and by `DragSelectionTest`, so the risk is
covered in aggregate — but one of the four new cases is load-bearing, not all four.

Local suites on the head, `build/test-results/` deleted first and mtimes checked:
`:view:jvmTest` **329 recorded / 0 failed / 0 skipped**, `:view:wasmJsBrowserTest`
(`CHROME_BIN=/usr/bin/chromium`) **296 / 0 / 0**. Both match the PR's figures.

## 5. D1 and D2, measured separately

Instrumented `HTMLElement.prototype.focus` and read `pointerdown.defaultPrevented` on a
macrotask after dispatch; `deepActive()` walks shadow roots.

**D1 — the down is consumed, so the browser stops stealing DOM focus**

| | `dd7dea8b` | `bf8bbcce` |
|---|---|---|
| mouse: `deepActive()` after click | **CANVAS 4/4** | **TEXTAREA 4/4** |
| touch: `pointerdown.defaultPrevented` | **false 6/6** | **true 6/6** |

**D2 — the platform is now asked for the IME**

| `.focus()` calls per gesture | `dd7dea8b` | `bf8bbcce` |
|---|---|---|
| mouse | **0, 4/4 clicks** | **2, 4/4 clicks** |
| touch | **0, 6/6 taps** | **2, 6/6 taps** |

**The two are independent, and the data shows it rather than merely asserting it.** On touch at
CMP 1.11.1 the base already ends on `TEXTAREA` — D1 is masked there, exactly as the PR's
correction to the investigation says (`ComposeWindow.initEvents` unconditionally
`preventDefault()`s `touchstart`) — yet the base still makes **zero** `.focus()` calls. So on
touch, D2 alone is the live defect and it moves on its own. On mouse it is D1 that moves
(CANVAS → TEXTAREA) while `defaultPrevented` stays `false` on both, which is the asymmetry the
PR records: Compose-web only cancels the default for touch pointer events, and the mouse case is
repaired by the `.focus()` calls instead. Two columns, two mechanisms, moving independently.

D2's premise also checks out statically: `git grep SoftwareKeyboardController origin/main` →
**zero hits** (positive control: `LocalDensity` grep returns 5 files). The editor had never asked
the platform to show the IME.

The reporter's own description — *"the keyboard starting to expand and then collapsing"* — is the
exact signature of D1: textarea focused, keyboard starts, the click moves focus to the canvas,
blur, keyboard collapses.

## 6. Boundaries — stated, and correctly

The PR does not claim mobile-wide resolution. It states plainly: no real device and no observed
soft keyboard (desktop Chromium has none, so what is measured is the DOM-focus precondition and
the keyboard-*request*, not a keyboard appearing); iOS Safari and Firefox for Android untested,
with iOS's trusted-gesture requirement for `focus()` called out as a possible surviving second
obstacle; CMP 1.12.0 not exercised at runtime. `changelog.d/303.fixed.md` carries the same
limits. That is the right shape — an honest boundary rather than an unbounded claim — and my
own measurements share every one of those limits.

## 7. Hygiene

- `closingIssuesReferences` = `[303]`, stable across 10 consecutive identical polls. The local
  regex gate (which fails safe, unlike the field) finds exactly one armed reference,
  `Fixes #303`, in the body and in the commit message. `#303` is the only issue number in the
  body at all, so #294, #259, #275, #299–#302, #307 and #308 cannot be touched — all nine
  verified still `OPEN`.
- `changelog.py check` — 7 fragments valid, `303.fixed.md` present, `CHANGELOG.md` unchanged
  against `origin/main`.
- Neither touched test file carries a `parity:` comment (positive control: 7 files in the repo
  do, none of them these), so the parity obligation is not engaged.
- CI: `lint`, `check`, `check-macos`, `check-ios` green. `CI (Android instrumented)` is red at
  13, which is its nominal state on every branch (#275) and is non-blocking by owner ruling.

## Non-blocking notes, no action wanted

- `keyboardController?.show()` fires on every pointer down regardless of the `editable` facet, so
  a tap on a read-only editor now re-raises a dismissed keyboard on Android/iOS. The hidden field
  is not `readOnly` (read-only is enforced by dropping input in `onValueChange`), so it already
  took focus and raised the keyboard at composition — this is a marginal extension of existing
  behaviour, not a new one.
- `LocalSoftwareKeyboardController.current` is captured by a `pointerInput(session)` block keyed
  only on `session`, so a new controller instance would not restart the block. Stable per-window
  in practice, and `CoreTextField` reads it the same way.

Approving and merging on green.

PR #314, opened 2026-09-07
Pin gradle/actions/setup-gradle and android-emulator-runner to commit SHAs (#290)

### LINKED ISSUE #290: Pin the remaining third-party actions (setup-gradle, android-emulator-runner) to commit SHAs

monkopedia-coder:
Pins the two remaining third-party actions from mutable major-version tags to full 40-character
commit SHAs, matching the form #289 established for `crazy-max/ghaction-import-gpg`.

```yaml
uses: gradle/actions/setup-gradle@ed408507eac070d1f99cc633dbcf757c94c7933a # v4.4.3
uses: reactivecircus/android-emulator-runner@a421e43855164a8197daf9d8d40fe71c6996bb0d # v2.38.0
```

## Call-site count — derived here, not inherited

The original filing of #290 said `setup-gradle` had **six** call sites; it was corrected in place to
**eight**. Rather than trust either number, every `uses:` line on `origin/main@23d15824` was
enumerated (40 of them) and tallied per file, with the enumeration itself as the positive control:

| file | `setup-gradle` | other third-party |
|---|---|---|
| `ci.yml` | 2 | — |
| `ci-apple.yml` | 2 | — |
| `docs.yml` | 2 | — |
| `ci-android.yml` | 1 | `android-emulator-runner` 1 |
| `deploy.yml` | 1 | `ghaction-import-gpg` 1 (already pinned, #289) |
| **total** | **8** | 2 |

9 call sites changed here. The corrected count of eight is confirmed; pinning six would have left
two behind.

## Both tags are ANNOTATED — dereference required

#289 pinned `ghaction-import-gpg@v5`, which happens to be a **lightweight** tag, so `.object.sha`
on the ref was already the commit. Neither tag here is:

| ref | `.object.type` | resolution |
|---|---|---|
| `gradle/actions` `refs/tags/v4` | `tag` | → tag object `0b6dd653…` → **target is another tag** `48b5f213…` = `v4.4.3` → commit `ed408507eac070d1f99cc633dbcf757c94c7933a` |
| `android-emulator-runner` `refs/tags/v2` | `tag` | → tag object `4c44018e…` → commit `a421e43855164a8197daf9d8d40fe71c6996bb0d` (= `v2.38.0`, confirmed by dereferencing the semver tag separately) |

`gradle/actions` needed **two** dereferences — its `v4` points at the annotated tag `v4.4.3`, not
at a commit. Taking `.object.sha` at face value would have pinned a tag object's hash, which is not
a commit and fails at run time.

Each SHA was then confirmed to exist upstream *and* to carry the action's `action.yml` at that ref
(`setup-gradle/action.yml`, 10780 bytes; `action.yml`, 3980 bytes). Negative control: a fabricated
40-hex SHA returns HTTP 422 from the same endpoint on both repos, so the existence check has teeth.
Both pins name the same commit the tag names today — **this is not an upgrade; no run behaves
differently.**

## Verification

Every workflow re-parsed with `yaml.safe_load`, then every `uses:` value walked out of the *parsed*
tree (so a `# vX.Y.Z` that leaked into the value would show up as part of the ref):

```
parse OK  ci-android.yml / ci-apple.yml / ci.yml / deploy.yml / docs.yml
total `uses:` values parsed: 40
  local reusable workflow          1   (./.github/workflows/ci.yml)
  first-party actions/*           29   (stay on tags by convention — untouched)
  third-party PINNED to 40-hex    10
  third-party UNPINNED             0
ASSERT zero unpinned third-party actions: PASS
NEGATIVE CONTROL: a fabricated unpinned entry trips that assertion
ASSERT no '#' or space inside any pinned ref value: PASS
per-file setup-gradle tally: ci-android 1, ci-apple 2, ci 2, deploy 1, docs 2  = 8
```

Grep for residual mutable third-party tags returns nothing, positive-controlled against
`setup-gradle@` which matches in all five files first.

No Gradle involved — this is workflow YAML plus one changelog fragment.

## Scope of what CI can prove here

`deploy.yml` is `workflow_dispatch`-only, so **no pull-request check exercises its `setup-gradle`
call site**. A green CI run on this PR says nothing about the deploy path; verification there is
static only (SHA exists, `action.yml` resolves at that ref, YAML parses). The other seven
`setup-gradle` sites and the emulator runner *are* covered by ordinary CI.

## Severity — stated accurately

Lower than #223/#289, and deliberately not described as a key-exposure fix: neither action is handed
`OSSRH_GPG_SECRET_KEY` or its passphrase. What this closes is the general supply-chain case — a
mutable tag can be moved to point at different code — on an action that runs in **every** CI job in
the repo, `deploy.yml` included, so `setup-gradle` executes in the same workflow as the signing step
even though the key never reaches it.

First-party `actions/*` stay on tags by convention and are untouched.

Fixes #290


monkopedia-coder:
## CI status

The three branch-protection-required checks are green:

| check | result |
|---|---|
| `check` (JVM/wasmJs/Android) | pass, 16m56s |
| `check-macos` | pass, 18m44s |
| `check-ios` | pass, 13m46s |
| `lint` | pass, 5m43s |
| `Instrumented tests (emulator)` — **non-blocking** | fail (pre-existing, see below) |

### The emulator job's failure is pre-existing and is not this change

`Instrumented tests (emulator)` reports `EXECUTED: 100 recorded: 100 failures: 13 errors: 0 skipped: 0`
— the same **13** failures, by name, that every recent run of `ci-android.yml` produces on every
branch. Baselines checked:

- run `33467315271`, branch `chore/281-ignore-pycache` (a `.gitignore`-only change) — 13 failed
- run `33833689844`, branch `fix/303-soft-keyboard` — 13 failed
- run `34079418854`, branch `fix/313-stale-hittest` (concurrent, unrelated) — failed
- run `34078923979`, this branch — 13 failed, **identical test names**

`FocusManagementTest` 5 + `KeyboardHandlingTest` 8, all in `view/input`. That is exactly the
constant #275 was filed about (\"never green, never run on main, no expected-failure baseline\") with
the root cause tracked in #259. No workflow-YAML change can move it, and this PR touches nothing in
`view/`.

### The emulator job is also the best run-time evidence the pins work

It is worth reading that failure the right way round: the job **ran**. The pinned
`reactivecircus/android-emulator-runner@a421e438…` booted an AVD and executed 100 instrumented
tests, and `gradle/actions/setup-gradle@ed408507…` set up Gradle in the same job
(`GRADLE_ACTION_ID: gradle/actions/setup-gradle`, `GRADLE_BUILD_ACTION_SETUP_COMPLETED: true` in its
step env). So both new pins are confirmed to resolve and function in real CI, not merely statically
— on top of the green `check` / `check-macos` / `check-ios` / `lint` jobs, which cover seven of the
eight `setup-gradle` sites.

**The one site no check exercises remains `deploy.yml`**, which is `workflow_dispatch`-only.
Verification there is static: the SHA exists upstream and `setup-gradle/action.yml` resolves at that
ref. A green tick on this PR should not be read as covering it.

monkopedia-reviewer (approved):
**APPROVE** — tier 1 (workflow YAML only; no shipped code, no public API). Verified independently at head `65bb47ab`.

## Pure hardening, not a version change

Both tags dereferenced **to `type == "commit"`**, not one level:

```
gradle/actions@v4    0b6dd653 (tag) -> 48b5f213 (tag) -> ed408507eac070d1f99cc633dbcf757c94c7933a (commit)  == pin
android-emulator-runner@v2          4c44018e (tag) -> a421e43855164a8197daf9d8d40fe71c6996bb0d (commit)  == pin
```

A single-level dereference stops at the intermediate tag object and misreads this as an upgrade. Both pins name the commit the floating major tag resolves to **today**, so behaviour is unchanged. The `# vX.Y.Z` comments are accurate: `v4.4.3` -> `ed408507`, `v2.38.0` -> `a421e438`.

Confirmed at **runtime** in the CI log, which is stronger than the API check — the runner resolved the pins to real action code:

```
Download action repository 'gradle/actions@ed408507…' (SHA:ed408507…)
Download action repository 'reactivecircus/android-emulator-runner@a421e438…' (SHA:a421e438…)
```

## Post-state: zero unpinned third-party `uses:`, positive-controlled

All five workflows `yaml.safe_load`-parsed and every `uses:` walked **out of the parsed tree** (not grepped), classified after partitioning on `@`:

| ref | total `uses:` | local | first-party `actions/*` | 3P pinned | **3P unpinned** |
|---|---|---|---|---|---|
| `origin/main` 23d15824 | 40 | 1 | 29 | 1 | **9** (positive control: > 0) |
| head 65bb47ab | 40 | 1 | 29 | 10 | **0** |
| head + fabricated `some-vendor/evil-action@v9` | 41 | 1 | 29 | 10 | **1** (assertion fires) |

The nine on main are 8 x `setup-gradle` (ci-android 1, ci-apple 2, ci 2, deploy 1, docs 2) + 1 x `android-emulator-runner` — matching the nine changed lines. **The issue's "six call sites" is wrong; it is eight for `setup-gradle`.** Derived independently, not inherited.

I initially also flagged `deploy.yml`'s `uses: ./.github/workflows/ci.yml` as unpinned — that is a *local* reusable workflow, resolved at the same commit, so pinning is not applicable. Classifier bug on my side, corrected; it does not affect the PR.

**Comment-not-in-ref:** all 10 pinned refs satisfy `^[0-9a-f]{40}$` under `fullmatch` after `partition("@")`. Had `# v4.4.3` leaked into the ref it would read `ed408507… # v4.4.3` and fail that regex. The comments remained YAML comments.

**First-party untouched:** the `actions/*` set is byte-identical between main and head (18 unique, 29 occurrences both sides). No scope creep.

**`action.yml` present at both pinned refs** — `gradle/actions:setup-gradle/action.yml@ed408507` (10780 b) and `reactivecircus/android-emulator-runner:action.yml@a421e438` (3980 b). Negative control: a fabricated 40-hex ref returns HTTP 404, so the check can fail.

## The genuinely new risk: a pinned emulator action that silently no-ops

This PR pins the action that *provides* the emulator, so "went red" is not the interesting failure mode — "did nothing and reported success" is, and that is invisible from the job's colour. Checked directly rather than from the tick:

- `##[notice]Android instrumented: **100 executed**, 13 failed, 0 skipped` — read from the notice line, not substring-matched (the AVD is named `test(AVD) - 14`).
- The AVD was really created and booted: `avdmanager create avd --force -n test --package 'system-images;android-34;default;x86_64'`, then the emulator launched and `Tests 4/100 completed` progressed to 100.
- **The thirteen match the #275 baseline by name, not by count** — 5 x `FocusManagementTest` + 8 x `KeyboardHandlingTest`, set-diffed against the recorded list: empty in both directions. Negative control: substituting one name makes the comparator fire, so a matching count with a swapped test would not have passed.

The red `Instrumented tests (emulator)` is the standing non-blocking baseline (#259/#275), red on unrelated branches including doc-only ones. Not evidence about this pin in either direction.

## Honest-limit framing checked, and it is accurate

- `deploy.yml` is `on: workflow_dispatch` only — confirmed at source. The fragment states plainly that verification is **necessarily static** for that one of the nine sites and that a green CI run says nothing about it. It does not imply CI covered it.
- The fragment correctly **declines** to call this a key-exposure fix. Verified: `OSSRH_GPG_SECRET_KEY` / `_PASSWORD` go to `crazy-max/ghaction-import-gpg` and the Gradle publish invocation; neither `setup-gradle` nor the emulator runner receives them. Its framing — same *workflow* as the signing step, key never reaches it — is accurate and, if anything, understated (`setup-gradle` runs in the same job, after the key is imported).

## Housekeeping

`closingIssuesReferences == [290]`. `CHANGELOG.md` untouched; `changelog.py check --base-ref origin/main` passes (9 fragments valid). No `parity:`-annotated test touched (diff is workflows + one fragment). Branch is 0 behind / 1 ahead of main, merge-base == `origin/main`, so the merge takes the workflow files wholesale and does not synthesise workflow bytes.

Required checks green at this head: `check`, `check-macos`, `check-ios` (plus `lint`). `mergeStateStatus: UNSTABLE` reflects only the non-required emulator baseline.

PR #315, opened 2026-09-07
Resolve clicks against the current document after a replacement (#313)

### LINKED ISSUE #313: Clicking after replacing the document with shorter content dispatches an out-of-bounds selection
## Description

After an existing `EditorSession` is updated with a shorter document, clicking a line in the new document can make KodeMirror calculate the pointer position from stale line-layout data.

The resulting cursor position is greater than the current document length, so transaction creation throws:

```text
java.lang.IllegalArgumentException: Selection points outside of document
    at com.monkopedia.kodemirror.state.SelectionKt.checkSelection(Selection.kt:473)
    at com.monkopedia.kodemirror.state.Transaction.<init>(Transaction.kt:192)
    at com.monkopedia.kodemirror.state.TransactionKt.resolveTransaction(Transaction.kt:447)
    at com.monkopedia.kodemirror.state.EditorState.update(State.kt:122)
    at com.monkopedia.kodemirror.view.EditorSessionImpl.dispatch(EditorSessionImpl.kt:115)
    at com.monkopedia.kodemirror.view.KodeMirrorKt$KodeMirror$6$1$4$1$1.invokeSuspend(KodeMirror.kt:630)
```

This is reproducible on JVM Desktop with KodeMirror 0.3.6.

## Stable reproducer

This example deliberately gives the old document a large second-line offset, then replaces it with a much shorter two-line document.

```kotlin
@Composable
fun SelectionOutOfBoundsReproduction() {
    val session = rememberEditorSession(
        doc = "A".repeat(200) + "\nold second line",
    )

    Column {
        Button(
            onClick = {
                session.setDoc("a\nb")
            },
        ) {
            Text("Replace with shorter document")
        }

        KodeMirror(
            session = session,
            modifier = Modifier
                .fillMaxWidth()
                .height(160.dp),
        )
    }
}
```

Steps:

1. Open the screen and wait for the initial document to render.
2. Click **Replace with shorter document**.
3. Confirm that the editor now displays two lines, `a` and `b`.
4. Click `b` on the second line.
5. KodeMirror throws `Selection points outside of document`.

The old second line starts at approximately offset 201, while the new document length is only 3, which makes the failure deterministic.

## Expected behavior

After replacing the document, pointer positions should be calculated from the latest document and layout. KodeMirror should never dispatch a selection outside `0..session.state.doc.length`, regardless of timing between layout updates and pointer events.

## Actual behavior

The tap handler can resolve a position from the old document and submit it directly as a selection for the new document. Transaction construction then fails and crashes the application.

## Likely cause

The tap, drag, and hover handlers currently use a pointer-input coroutine keyed only by `session`:

```kotlin
.pointerInput(session) {
    // ...
    val pos = resolvePos(downPosition)
    session.dispatch(
        TransactionSpec(
            selection = SelectionSpec.CursorSpec(DocPos(pos)),
        ),
    )
}
```

The `EditorSession` instance does not change when its document changes. The running pointer-input coroutine can therefore retain the `resolvePos` callback captured by an earlier composition. That callback captures old `columnItems` and related layout data and can return a position belonging to the previous document.

The same pattern is still present on the current `main` branch:
https://github.com/Monkopedia/kodemirror/blob/main/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt#L562-L654

## Suggested fix

Keep the pointer-input coroutine stable, but let it read the latest position resolver:

```kotlin
val currentResolvePos by rememberUpdatedState(resolvePos)
```

Then use `currentResolvePos` consistently for tap, drag, drop-cursor, and hover handling.

As a final defensive check, clamp the resolved cursor position against the current document immediately before dispatch:

```kotlin
val pos = currentResolvePos(downPosition)
    ?.coerceIn(0, session.state.doc.length)
    ?: return@awaitEachGesture
```

Clamping alone prevents the exception but may still place the cursor at the wrong location. Reading the latest resolver should be the primary fix, with clamping as a last line of defense.

## Suggested regression test

1. Create an editor with a long two-line document.
2. Replace it with a much shorter document that still has two lines.
3. Wait for the UI to settle.
4. Click the second line of the new document.
5. Verify that no exception is thrown and that the resulting selection stays within `0..session.state.doc.length`.

monkopedia-coder:
Fixes #313.

## Reproduced first

The reporter's chain holds, and the missing ingredient for a deterministic repro is
**an interaction with the editor before the replacement**. `Modifier.pointerInput` starts its
handler coroutine lazily, on the first pointer event; with no prior interaction it is started
*after* the replacement and closes over the fresh resolver by luck. Instrumenting `resolvePos`
and the tap branch on unmodified `main`:

```
DBG recompute columnItems docLen=216
DBG pointerInput LAUNCH resolver=453485293
DBG resolvePos closure=-1952592720 items=[0, 201] docLen=216
DBG tap pos=5 docLen=216            <- click on line 1, arms the gesture coroutine
DBG recompute columnItems docLen=3  <- setDoc("a\nb")
DBG resolvePos closure=-1952592720 items=[0, 201] docLen=3   <- SAME closure, OLD line spans
DBG tap pos=201 docLen=3
```

which throws the reported exception, from the same frame the report names:

```
java.lang.IllegalArgumentException: Selection points outside of document
    at com.monkopedia.kodemirror.state.SelectionKt.checkSelection(Selection.kt:473)
    at com.monkopedia.kodemirror.state.Transaction.<init>(Transaction.kt:192)
    at com.monkopedia.kodemirror.state.TransactionKt.resolveTransaction(Transaction.kt:447)
    at com.monkopedia.kodemirror.state.EditorState.update(State.kt:122)
    at com.monkopedia.kodemirror.view.EditorSessionImpl.dispatch(EditorSessionImpl.kt:115)
    at com.monkopedia.kodemirror.view.KodeMirrorKt$KodeMirror$6$1$4$1$1.invokeSuspend(KodeMirror.kt:653)
```

## Which layer is at fault, and why it is fixed there

**The pointer-input layer, not the hit-test and not the dispatch.** `posFromVisibleItems` is
correct — given the current `columnItems` it resolves the right offset, as the second test here
shows. `checkSelection` is also correct: a selection outside the document is a genuine error and
should keep throwing.

The defect is that the resolver reaching the hit-test is not the current one.
`.pointerInput(session)` is keyed on the session alone, and a session keeps its identity when its
document changes, so the key never fires; `SuspendPointerInputElement` compares only its keys, so
the node is not updated and the handler lambda the coroutine was started with is the one it keeps.
That lambda closes over one composition's `columnItems` — the previous document's line spans — and
also its `density`, `hasGutters`, `gutterWidthDp`, `contentTopPaddingPx` and `wrapLines`.

Fixed by reading the latest resolver through `rememberUpdatedState`, which is **already the idiom
in this composable** (`currentColumnItems`, used by the cursor overlay for the same reason). Two
alternatives were considered and rejected:

- **Re-keying `pointerInput` on the document** would restart the gesture coroutine on every
  document change, cancelling an in-progress drag selection whenever the document changed under it
  — a worse bug than the one being fixed, and it discards the reason the key is `session` today.
- **Clamping to `doc.length`** would stop the throw and put the caret at the end of the document
  instead of on the line that was clicked. This PR asserts the exact offset precisely so that a
  clamp cannot pass.

## What happens to a click that arrives genuinely mid-update

There is a narrower window the resolver fix does not close: pointer events are delivered before a
frame's composition, so a document replaced from outside a gesture (an LSP edit, a collab update,
`setDoc` from a coroutine) can land between the event and the recomposition that re-derives the
layout. Everything the resolver reads is then self-consistent but describes the *previous*
document, and there is no honest mapping from such a position into the new one.

`resolvePos` therefore **drops** a result the current document cannot contain — returns null,
which every call site already handles (the tap becomes a no-op, the drop-cursor hides, the hover
tooltip clears). The click is lost and the next one, against a settled layout, lands correctly.
That is deliberately not a clamp: losing a click is recoverable and visible to the user, a caret
silently placed where they did not click is neither.

**This gate is reasoned, not demonstrated, and it carries no test of its own.** The race cannot be
synthesised in `runComposeUiTest` — `performMouseInput` pumps a frame before delivering the event,
so a probe that calls `setDoc` and clicks with no `waitForIdle` in between still recomposes first
and resolves correctly (measured: cursor lands at 2, not the stale 201). It is exercised
incidentally by mutation 1 below.

## Tests

`view/src/commonTest/.../input/DocReplacementClickTest.kt`, two tests, both asserting an **exact
offset** rather than absence-of-throw.

**Where they actually execute:** `commonTest`, so the JVM (`:view:jvmTest`) and wasmJs
(`:view:wasmJsBrowserTest`) suites, plus the native suites in `check-macos`/`check-ios`. The
`view.input.*` package is excluded from every `*UnitTest` task (`view/build.gradle.kts:74-76`), so
on Android they compile but run only in the instrumented suite, not in local unit tests.

- Red first: `clickAfterReplacingWithShorterDoc` fails on unmodified `main` with the exception
  quoted above. Green after the fix.
- Both directions: `clickWithoutReplacement` clicks the same coordinate on the same document with
  no replacement and asserts the same offset 2, so the fix cannot pass by turning ordinary clicks
  into no-ops.

Mutation controls, each naming the line it perturbs:

| mutation | result |
| --- | --- |
| `KodeMirror.kt:671` tap reverted to the stale `resolvePos`, bounds gate kept | `clickAfterReplacingWithShorterDoc` FAILS `Expected cursor at 2 but was at 0` (the click is dropped, not misplaced) — this is the check that a drop-only or clamp-only fix cannot pass; `clickWithoutReplacement` still passes |
| final expression of `resolvePos` changed to `pos?.takeIf { false }` (drop everything) | both tests FAIL `Expected cursor at 2 but was at 0` — the ordinary-click direction is genuinely guarded |
| wasm negative control: expected offset changed to 999 | wasm suite FAILS `Expected cursor at 999 but was at 2` — the assertion really evaluates in the browser runner, and resolves to the correct offset there |

## Verification

Run with `build/test-results/<task>/` deleted first and the XML read back, not the exit code.

- `:view:jvmTest` — 331 tests, 0 failures, 0 skipped
- `:view:wasmJsBrowserTest` — 298 tests, 0 failures, 0 skipped (ChromeHeadless)
- `:view:apiCheck` — passes, **no `api/*.api` movement**; the change adds no public declarations
- `:view:spotlessApply` / `:view:ktlintFormat` — clean
- `changelog.py check --base-ref origin/main` — 9 fragments valid, `CHANGELOG.md` untouched

No test in this PR carries a `parity:` comment, and none of the tests it touches does — the file
is new and no existing test was changed.

## On #259 — measured, and the hypothesis is refuted

#313's mechanism predicts the `Expected cursor at 11 but was at 0` signature of the thirteen
long-standing Android instrumented failures (#259/#275), because a null `resolvePos` makes the tap
handler dispatch nothing. **It does not explain them.** This branch's instrumented run
([34079418854](https://github.com/Monkopedia/kodemirror/actions/runs/34079418854)) reports
`102 executed, 13 failed, 0 skipped` — the denominator moved `100 -> 102` from the two tests added
here, and the failing **set is identical to the one recorded by name on #275**, compared name by
name rather than by count. None of the thirteen moved.

The same run carries a stronger datum: the two new `DocReplacementClickTest` cases **pass on the
emulator**, asserting an exact click-driven cursor offset, as do `ClickTargetingTest` (8) and
`DragSelectionTest` (4). So a click on that device does resolve to a non-null position and does
dispatch the correct selection — which rules out the null-resolver reading of `cursor at 0`.
Written up on #259; no label or state changed there.

The instrumented job is red on this PR at the pre-existing baseline and is not one of the three
required checks.

### DIFF
--- a/changelog.d/313.fixed.md
+++ b/changelog.d/313.fixed.md
@@ -1,0 +1,10 @@
+- Fixed a crash when the editor is clicked after its document has been replaced with a shorter one
+  (#313). `Modifier.pointerInput` was keyed on the session, which keeps its identity when its
+  document changes, so the gesture coroutine kept the position resolver it was started with — one
+  closed over the previous document's line spans. A click on the second line of a shorter
+  replacement resolved through the old line's start offset, past the end of the new document, and
+  the resulting selection threw `Selection points outside of document`. The gesture coroutines now
+  read the current resolver through `rememberUpdatedState`, so they stay stable across a document
+  change while hit-testing against the current layout. A hit-test result the current document
+  cannot contain is now also dropped rather than dispatched, which loses a click that raced a
+  document update instead of placing the caret somewhere the user did not click.
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -485,7 +485,7 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
     // every hit-test through the live layout keeps clicks/drags/hover on the
     // glyph under the pointer instead of its pre-scroll position.
     val resolvePos: (Offset) -> Int? = { offset ->
-        posFromVisibleItems(
+        val pos = posFromVisibleItems(
             offset,
             lazyState,
             columnItems,
@@ -496,7 +496,31 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
             if (wrapLines) 0 else horizontalScrollState.value,
             contentTopPaddingPx
         ) ?: session.posAtCoords(offset.x, offset.y)
+        // Drop — never clamp — a result the current document cannot contain.
+        // Pointer events are delivered before the frame's composition, so a
+        // document swapped in from outside a gesture (setDoc, an LSP edit) can
+        // land between the event and the recomposition that re-derives the
+        // layout from it. Everything above is then self-consistent but
+        // describes the PREVIOUS document, and there is no honest mapping from
+        // such a position into the new one. Coercing it into range would place
+        // the caret somewhere the user did not click and say nothing about it;
+        // returning null makes the gesture a no-op, so the click is simply lost
+        // and the next one — against a settled layout — lands correctly.
+        pos?.takeIf { it in 0..session.state.doc.length }
     }
+
+    // The two gesture coroutines below are keyed on `session`, and a session
+    // keeps its identity when its document changes — so their key never fires
+    // and the handler lambda they were started with is the one they keep. Held
+    // directly, `resolvePos` would be whichever composition's copy happened to
+    // be current when the first pointer event started them, closed over that
+    // composition's `columnItems` — the OLD document's line spans. Clicking
+    // after the document was replaced with a shorter one then resolved through
+    // a line start past the new document's end and the dispatch threw
+    // "Selection points outside of document" (#313). Reading the latest
+    // resolver keeps the gesture coroutines stable and their hit-testing
+    // current, the same treatment `currentColumnItems` gets above.
+    val currentResolvePos by rememberUpdatedState(resolvePos)
 
     CompositionLocalProvider(
         LocalEditorTheme provides theme,
@@ -604,18 +628,18 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                                         session,
                                         dragStart,
                                         dragCurrent,
-                                        resolvePos
+                                        currentResolvePos
                                     )
                                 } else {
                                     handleDrag(
                                         session,
                                         dragStart,
                                         dragCurrent,
-                                        resolvePos
+                                        currentResolvePos
                                     )
                                 }
                                 session.plugin(dropCursorViewPlugin)
-                                    ?.moveTo(resolvePos(dragCurrent))
+                                    ?.moveTo(currentResolvePos(dragCurrent))
 
                                 drag(slopChange.id) { change ->
                                     change.consume()
@@ -625,18 +649,18 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                                             session,
                                             dragStart,
                                             dragCurrent,
-                                            resolvePos
+                                            currentResolvePos
                                         )
                                     } else {
                                         handleDrag(
                                             session,
                                             dragStart,
                                             dragCurrent,
-                                            resolvePos
+                                            currentResolvePos
                                         )
                                     }
                                     session.plugin(dropCursorViewPlugin)
-                                        ?.moveTo(resolvePos(dragCurrent))
+                                        ?.moveTo(currentResolvePos(dragCurrent))
                                 }
                                 // Drag ended
                                 session.plugin(dropCursorViewPlugin)
@@ -644,7 +668,7 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                             } else if (!isDrag) {
                                 // Pointer released without exceeding slop —
                                 // treat as a tap for cursor positioning.
-                                val pos = resolvePos(downPosition)
+                                val pos = currentResolvePos(downPosition)
                                     ?: return@awaitEachGesture
                                 session.dispatch(
                                     TransactionSpec(
@@ -669,7 +693,7 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                                 val pos = event.changes
                                     .firstOrNull()?.position
                                 if (pos != null) {
-                                    val docPos = resolvePos(pos)
+                                    val docPos = currentResolvePos(pos)
                                     if (docPos != lastHoverDocPos) {
                                         lastHoverDocPos = docPos
                                         val hoverTooltips =
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/DocReplacementClickTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/DocReplacementClickTest.kt
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.view.input
+
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performMouseInput
+import com.monkopedia.kodemirror.view.setDoc
+import kotlin.test.Test
+
+/**
+ * Clicking after the document is replaced in place (#313).
+ *
+ * The pointer-input gesture coroutine is keyed on the session, which does not
+ * change when its document does, so it kept whichever position resolver the
+ * first composition handed it. That resolver closed over that composition's
+ * `columnItems`, i.e. the OLD document's line spans. Clicking the second line
+ * of a shorter replacement document resolved through the old line's `from`
+ * (offset 201 here) and the dispatch threw `Selection points outside of
+ * document`.
+ *
+ * Both tests assert an exact offset rather than merely surviving: clamping the
+ * out-of-bounds result to `doc.length` would also stop the throw, but would
+ * park the caret at 3 (the end of `"a\nb"`) instead of 2 (the start of the line
+ * that was actually clicked).
+ */
+@OptIn(ExperimentalTestApi::class)
+class DocReplacementClickTest {
+
+    /** A first line long enough that the old line 2 starts past the new doc's end. */
+    private val longDoc = "A".repeat(200) + "\nold second line"
+
+    /** Replacement: two short lines, so line 2 starts at offset 2 and the doc ends at 3. */
+    private val shortDoc = "a\nb"
+
+    /** Left edge of the content area, so the click lands on the line's first character. */
+    private val lineStartClick = Offset(8f, 35f)
+
+    @Test
+    fun clickAfterReplacingWithShorterDoc() = runEditorTest(
+        doc = longDoc,
+        width = 400,
+        height = 160
+    ) { holder ->
+        // Interact with the editor BEFORE the replacement. This is what arms
+        // the bug: the gesture coroutine is started lazily by the first pointer
+        // event, so without a prior interaction it is launched after the
+        // replacement and closes over the fresh resolver by luck.
+        onNodeWithTag("KodeMirror").performMouseInput {
+            click(Offset(50f, 15f))
+        }
+        waitForIdle()
+        holder.assertCursorOnLine(1)
+
+        holder.session.setDoc(shortDoc)
+        waitForIdle()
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
**APPROVE** — reviewed at `3498b5eed337ce32c3f2061047258627d9cc1de5` (read via `git ls-remote`, unmoved at verdict time). Category `bug_fix` → tier 1.

Everything below was re-measured independently in a fresh worktree off `origin/main` @ `23d15824`, with `build/test-results/<task>/` deleted before every run and the XML read back rather than the exit code.

## Red-first

The new test is genuinely red on unmodified `main`. Copied `DocReplacementClickTest.kt` alone onto `23d15824` — no production change — and ran `:view:jvmTest`:

```
2 tests, 1 failure, 0 skipped
clickAfterReplacingWithShorterDoc[jvm]
  java.lang.IllegalArgumentException: Selection points outside of document
    at com.monkopedia.kodemirror.state.SelectionKt.checkSelection(Selection.kt:473)
    at com.monkopedia.kodemirror.state.Transaction.<init>(Transaction.kt:192)
    at com.monkopedia.kodemirror.state.TransactionKt.resolveTransaction(Transaction.kt:447)
    at com.monkopedia.kodemirror.state.EditorState.update(State.kt:122)
    at com.monkopedia.kodemirror.view.EditorSessionImpl.dispatch(EditorSessionImpl.kt:115)
    at com.monkopedia.kodemirror.view.KodeMirrorKt$KodeMirror$6$1$4$1$1.invokeSuspend(KodeMirror.kt:649)
clickWithoutReplacement[jvm]  pass
```

Same exception, same frame as #313's report. And the second test passes on `main`, so it is not red for an unrelated reason.

The reproducer is a faithful transcription of the reporter's: `"A".repeat(200) + "\nold second line"` → `setDoc("a\nb")`, 160dp tall, click line 2.

## Which layer actually fixes it — the point I was asked to settle

The reporter proposed the resolver fix *plus* `coerceIn(0, doc.length)`. **This PR did not take the clamp** — it takes a *drop* (`pos?.takeIf { it in 0..session.state.doc.length }`, `KodeMirror.kt:509`), and it is right to: a clamp converts a crash into a caret placed somewhere the user did not click, with nothing to notice. Two mutations, each naming the line it perturbs:

**Mutation 1 — the backstop removed entirely.** `KodeMirror.kt:509` `pos?.takeIf { it in 0..session.state.doc.length }` → `pos`. `rememberUpdatedState` left in place.

```
clickAfterReplacingWithShorterDoc[jvm]  pass
clickWithoutReplacement[jvm]            pass
```

**The correctness layer alone fixes #313.** The gate is not load-bearing.

**Mutation 2 — the correctness layer removed, backstop kept.** `KodeMirror.kt:671` `currentResolvePos(downPosition)` → `resolvePos(downPosition)`.

```
clickAfterReplacingWithShorterDoc[jvm]  FAILED  Expected cursor at 2 but was at 0
clickWithoutReplacement[jvm]            pass
```

A drop-only *or* clamp-only implementation cannot pass this test — it asserts the exact offset (2, the start of the clicked line), not absence-of-throw. This reproduces the PR's own table row 1.

## Does the backstop ever fire in normal operation?

Instrumented `KodeMirror.kt:509` to print on every invocation, split by branch, and ran the **full** `:view:jvmTest`:

```
GATEFIRE (non-null pos dropped)   0    across all 331 tests
GATEPASS (non-null pos kept)      26 distinct offsets, hundreds of invocations
```

The `GATEPASS` control proves the instrumented line is reached and the print plumbing works — an empty `GATEFIRE` here is a measurement, not an artefact. And the `GATEFIRE` branch is not unreachable: re-running the same instrumentation with mutation 2 applied yields exactly

```
GATEFIRE pos=201 docLen=3
```

— the reported stale offset against the 3-character replacement, once, and only when the resolver is stale. So the gate is a true backstop: it fires on the #313 shape and on nothing else in the suite. It masks nothing.

Worth noting the drop is also tight on the tap path: there is no suspension point between the gate and `session.dispatch`, so the document cannot move between them.

## Both directions

`clickWithoutReplacement` clicks the same coordinate on the same document with no replacement and asserts the same offset 2. Guarded: with the resolver made to drop everything, *both* tests fail `Expected cursor at 2 but was at 0` — the ordinary-click direction is not passing by accident. (Mutation 2 above also leaves it passing while the other fails, which is the discrimination that matters.)

## Where the new tests execute, per target

| target | task | result |
|---|---|---|
| JVM | `:view:jvmTest` | **331 executed, 0 failed, 0 skipped** — both new tests present |
| wasmJs | `:view:wasmJsBrowserTest` (ChromeHeadless 151) | **298 executed, 0 failed, 0 skipped** — both present |
| Android **local unit** | `testDebugUnitTest` | **does not run** — `view.input.*` is excluded (`view/build.gradle.kts:74-76`) |
| Android **instrumented** | `:view:connectedDebugAndroidTest` | **runs** — `testInstrumentationRunnerArguments["package"] = "com.monkopedia.kodemirror.view.input"` matches; measured `DocReplacementClickTest 2 executed, 0 failed, 0 skipped` on the emulator |
| macOS / iOS native | `check-macos` / `check-ios` | compiled from `commonTest`; CI green |

The wasm run is not vacuous. Negative control: changing both expected offsets to 999 turns the browser suite red with

```
kotlin.AssertionError: Expected cursor at 999 but was at 2
```

so the assertion really evaluates in the browser runner and resolves to the correct offset there.

## On #259 — measured by name, and it refutes the hypothesis

The PR body predicted #313's mechanism might explain the thirteen long-standing instrumented failures, and promised to read the result by name rather than by count. Done here, against a same-night control:

| run | branch | executed | failed |
|---|---|---|---|
| 34078923979 | `pin-actions-290` (cannot touch the editor), 03:21 UTC | 100 | 13 |
| 34079418854 | `fix/313-stale-hittest`, 03:30 UTC | **102** | **13** |

The failing set is identical between the two runs, name for name — `FocusManagementTest` × 5 (`dragThenKeyboard_cursorMoves`, `multipleClicks_eachPreservesFocus`, `clickOnEditor_givesFocusToTextField`, `clickAfterDrag_focusIsRestored`, `clickThenArrowRight_cursorMoves`) and `KeyboardHandlingTest` × 8 (`arrowRight_movesCursorForward`, `arrowLeft_movesCursorBackward`, `arrowDown_movesToNextLine`, `arrowUp_movesToPreviousLine`, `home_movesToLineStart`, `end_movesToLineEnd`, `modX_cutsSelectedText_onMacFamily`, `modX_cutsSelectedText_offMacFamily`).

**None of the thirteen moved. The 100 → 102 delta is exactly this PR's two new tests, both green on the emulator.** So the PR introduces no instrumented regression, and the #313 mechanism does *not* explain #259 — which is consistent with #259's own evidence that the pointer pipeline works on Android (`ClickTargetingTest` 8/8, and `home_movesToLineStart` hit-tests to the correct offset 8 before the key path drops the event).

That is a real negative finding and it is worth having. **#259 and #275 stay open**; the PR correctly does not claim to fix either, and `closingIssuesReferences` is `[313]` and nothing else.

## Instrumentation

Confirmed removed. No `println` / `DBG` / `console.log` / `Log.` anywhere in the three changed files, and no such line is added anywhere in `git diff origin/main...3498b5ee` (positive control: a `grep` for `fun KodeMirror` on the same file returns 1).

## Rest of the diff

- `rememberUpdatedState` is **the existing idiom in this exact composable**, not a parallel mechanism: `currentColumnItems` at `:244` exists for the same reason and is read at `:366` from a `LaunchedEffect(session)` coroutine with the same lifetime problem. The PR applies it to the two `pointerInput(session)` coroutines.
- The `pointerInput(session)` **keys are unchanged**, which is the right call and the PR says why: re-keying on the document would restart the gesture coroutine and cancel an in-progress drag on every document change.
- No stale `resolvePos` reference survives — the identifier now occurs only at its definition (`:487`) and inside the `rememberUpdatedState` wrapper (`:523`); all eight call sites (tap, both drag branches ×2, drop-cursor ×2, hover) read `currentResolvePos`.
- Other values the gesture closures capture were checked and are safe: `altPressed` is a `remember { mutableStateOf }` (stable holder), `focusRequester` is `remember`ed, `impl` is a cast of the stable `session` param.
- Focus is still requested on a dropped click — the `?: return@awaitEachGesture` sits after `focusRequester.requestFocus()`, so a lost click does not also lose focus.
- No `parity:` comment in any touched file; the new file is new and no existing test was changed. Obligation satisfied.
- `:view:apiCheck` passes with **no `api/*.api` movement** (`git diff origin/main...head -- '*.api'` is empty); `spotlessCheck` + `ktlintCheck` clean; `changelog.py check --base-ref origin/main` valid, `CHANGELOG.md` untouched, one fragment `changelog.d/313.fixed.md`.
- Single commit, exactly three files, merge-base is `origin/main` head — no bundled scope.

The changelog entry is accurate about the mechanism and honest that the drop loses a click rather than relocating a caret.

Nothing to request. Merging on green.

PR #316, opened 2026-09-07
Make verifyBomCoverage fail on BOM version drift (#300)

### LINKED ISSUE #300: verifyBomCoverage compares artifactIds only: a BOM published at 0.3.7 constraining every module at 0.3.6 passes the drift guard — demonstrated

monkopedia-coder:
Fixes #300

## The defect

`verifyBomCoverage` is the #244 BOM-drift guard — named explicitly in `ci.yml` and required by
branch protection. It parsed the generated POM for `<artifactId>` only and never looked at
`<version>`.

The project version exists as **two independent string literals**:

- `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts:41` — sets the version of every
  library module, and therefore of every constraint the BOM derives;
- `kodemirror-bom/build.gradle.kts:10` — the BOM's own coordinate.

The release process (`CLAUDE.md` step 1) instructs a human to edit both. Edit one and not the other
and you publish a BOM at X constraining all 56 modules at Y. That is worse than #244: #244 gave the
consumer a hard resolution failure, this resolves cleanly and silently hands them the *previous*
release.

## The fix

`doLast` now parses whole `<dependency>` blocks rather than bare `<artifactId>`s, so each artifactId
stays associated with the version shipping beside it, and `check`s that every constrained
`<version>` equals the BOM's own version. A missing `<version>` maps to the empty string, which can
never equal the BOM version, so it is reported rather than silently accepted.

`project.version` is read at configuration time and registered as a task input, alongside the
existing `expectedModules` input.

The artifactId coverage check is unchanged.

## Verification — all four controls, run in order

**1. Positive control (the issue's premise still holds at `origin/main` @ `8a308c59`).**
Unmodified tree, only `kodemirror-bom/build.gradle.kts:10` changed to `version = "0.3.7"`:

```
> Task :kodemirror-bom:verifyBomCoverage
kodemirror-bom constrains all 56 published modules.
BUILD SUCCESSFUL in 9s

$ grep -o '<version>[^<]*' kodemirror-bom/build/publications/maven/pom-default.xml | sort | uniq -c
     56 <version>0.3.6
      1 <version>0.3.7
```

Reproduced exactly as reported.

**2. Negative control — the same one-line drift, with this fix in place. The guard now FAILS:**

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kodemirror-bom:verifyBomCoverage' (registered in build file 'kodemirror-bom/build.gradle.kts').
> kodemirror-bom constrains modules at a version other than its own (#300).
    kodemirror-bom version: 0.3.7
    mismatched constraints: 56 of 56
    constrained at 0.3.6 (56): autocomplete, basic-setup, collab, commands, lang-angular, lang-cpp, lang-css, lang-go, lang-grammar, lang-html, lang-java, lang-javascript, lang-jinja, lang-json, lang-less, lang-liquid, lang-markdown, lang-php, lang-python, lang-rust, lang-sass, lang-sql, lang-vue, lang-wast, lang-xml, lang-yaml, language, legacy-modes, lezer-common, lezer-highlight, lezer-lr, lint, lsp-client, material-theme, merge, search, state, theme-amy, theme-ayu-light, theme-barf, theme-bespin, theme-birds-of-paradise, theme-boys-and-girls, theme-clouds, theme-cobalt, theme-cool-glow, theme-dracula, theme-espresso, theme-noctis-lilac, theme-one-dark, theme-rose-pine-dawn, theme-smoothy, theme-solarized-light, theme-tomorrow, view, vim
    The project version is two independent literals: convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts (every library module) and kodemirror-bom/build.gradle.kts (the BOM itself). Set both to the same value.

BUILD FAILED in 4s
```

**3. No false positive.** Drift reverted, `git status` clean:

```
kodemirror-bom constrains all 56 published modules at 0.3.6.
BUILD SUCCESSFUL in 4s
```

56 constraints parsed by the new `<dependency>`-block regex — the same count the old
artifactId-only regex found, so the new parse is not under-matching.

**4. #244 not regressed.** Dropped one module (`search`) from the constraint derivation only:

```
> kodemirror-bom does not match the set of published modules (#244).
    expected constraints: 56
    BOM constraints:      55
    published but NOT in the BOM: search
    deliberately excluded: kodemirror-test (#240 — undecided whether to delete or rebuild the published coordinate)

BUILD FAILED in 6s
```

Reverted; tree clean.

Also green: `spotlessApply` (no reformatting of the edited file), `ktlintCheck`, `apiCheck`,
`python3 .github/scripts/changelog.py check --base-ref origin/main`.

## Stage 2 (single-sourcing the version) — deliberately NOT done here

The issue offers single-sourcing the literal as an optional stage 2. I did not do it, on purpose,
and it is not a prerequisite for the guard above.

`convention-plugins` is a separate included build, so the two literals cannot simply share one.
More decisively, the BOM's `version =` line is **load-bearing outside the build**:

```
.github/workflows/deploy.yml:72:  VERSION=$(grep -E '^version = ' kodemirror-bom/build.gradle.kts | sed -E 's/.*"([^"]+)".*/\1/')
docs/release-checklist.md:38,113,180,183
CLAUDE.md:97
```

`deploy.yml` derives the tag it publishes and verifies (`--verify-tag`) from that exact line.
Replacing the literal with a property lookup silently breaks that grep, and rewriting it means
editing the authoritative release process in `CLAUDE.md` and its checklist — a release-path change
that wants an owner decision, not a rider on a guard fix. Filed separately as #317 so it is not lost.

Note that the guard here is **not** made redundant by single-sourcing: it asserts the property on
the artifact that actually ships, so it keeps holding if a second literal is ever reintroduced.

## Blast radius

Build logic only. No source change, no public API, no change to any published artifact's contents.
`main` passes as written (all 57 versions agree), so it lands green.

---

## Note on the two force-pushes (branch is now FINAL at `1e0768d6`)

The head moved twice shortly after opening, before any review started. Neither amend was a fix to
a hole in the guard — both were documentation of correctness properties that were already true:

- `1097673c` → `259658dd`: added a comment recording *why* the version check may assume every
  constraint is a kodemirror module — the #244 coverage check runs first and already rejects
  anything "in the BOM but NOT published", so a third-party constraint can never reach the version
  check and be misreported as drift. Comment only.
- `259658dd` → `1e0768d6`: the task's own `description` (what `./gradlew tasks` prints) still read
  "does not constrain exactly the published modules" and no longer described what the task checks.
  String only.

`git diff 1097673c 1e0768d6 -- kodemirror-bom/build.gradle.kts` is exactly those two hunks: **no
executable logic changed across the three SHAs.**

All three controls were nevertheless re-run against the final tree `1e0768d6` rather than assumed
to carry over, and all three reproduce the outputs quoted above verbatim: version drift → `#300`
failure; `search` dropped from the constraint derivation → `#244` failure; unmodified tree → passes
with 56 constraints at 0.3.6. `git status` clean after each.

No further force-pushes are planned; this SHA is final.

### DIFF
--- a/changelog.d/300.build.md
+++ b/changelog.d/300.build.md
@@ -1,0 +1,8 @@
+- The BOM drift guard (`:kodemirror-bom:verifyBomCoverage`) now also checks the **versions** in the
+  generated BOM POM, not just the artifactIds (#300). The project version is written as two
+  independent literals — one in the `kodemirror.library` convention plugin (every library module)
+  and one in `kodemirror-bom/build.gradle.kts` (the BOM's own coordinate) — and the release process
+  asks a human to edit both. Editing only one produced a BOM published at X that constrained all 56
+  modules at Y: it resolves cleanly and silently hands consumers the previous release, and the
+  guard reported success. The guard now parses whole `<dependency>` blocks and fails if any
+  constrained `<version>` differs from the BOM's own, naming the mismatched version and modules.
--- a/kodemirror-bom/build.gradle.kts
+++ b/kodemirror-bom/build.gradle.kts
@@ -53,6 +53,19 @@ dependencies {
 // silently produces nothing (e.g. the publish plugin id changes). The expected set is every
 // module applying the `kodemirror.library` convention plugin, minus the documented exclusions;
 // that is a deliberately different criterion from the one the derivation uses.
+//
+// It guards TWO independent kinds of drift, because the POM can be wrong in two ways:
+//
+//  1. the set of constrained artifactIds not matching the set of published modules (#244);
+//  2. a constrained `<version>` not matching the BOM's own version (#300).
+//
+// (2) exists because the project version is written out twice, as two independent string
+// literals — `convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts` (which sets the
+// version of all library modules, and therefore of every constraint here) and the `version =`
+// line above (the BOM's own coordinate). The release process instructs a human to edit both.
+// Editing only one publishes a BOM at X whose constraints all say Y: it resolves cleanly and
+// silently hands consumers the *previous* release, which is worse than #244's hard resolution
+// failure because nothing surfaces it.
 val expectedBomModules = provider {
     rootProject.subprojects
         .filter {
@@ -71,22 +84,36 @@ val pomTasks = tasks.withType<GenerateMavenPom>()
 val verifyBomCoverage = tasks.register("verifyBomCoverage") {
     group = "verification"
     description =
-        "Fails if the generated BOM POM does not constrain exactly the published modules."
+        "Fails if the generated BOM POM does not constrain exactly the published modules, " +
+            "each at the BOM's own version."
     dependsOn(pomTasks)
     val pomFiles = provider { pomTasks.map { it.destination } }
     val expected = expectedBomModules
+    // Read at configuration time: `project` must not be touched from a task action.
+    val bomVersion = project.version.toString()
     inputs.files(pomFiles)
     inputs.property("expectedModules", expected)
+    inputs.property("bomVersion", bomVersion)
     doLast {
         val pom = pomFiles.get().single().readText()
         val managed = Regex(
             "<dependencyManagement>(.*?)</dependencyManagement>",
             RegexOption.DOT_MATCHES_ALL
         ).find(pom)?.groupValues?.get(1).orEmpty()
-        val declared = Regex("<artifactId>([^<]+)</artifactId>")
+        // Parse whole `<dependency>` blocks rather than bare `<artifactId>`s, so the artifactId
+        // and the version that ships beside it stay associated. A missing `<version>` maps to
+        // the empty string, which can never equal `bomVersion` and so is reported rather than
+        // silently accepted.
+        val constrained = Regex("<dependency>(.*?)</dependency>", RegexOption.DOT_MATCHES_ALL)
             .findAll(managed)
             .map { it.groupValues[1] }
-            .toSortedSet()
+            .map { entry ->
+                fun tag(name: String) =
+                    Regex("<$name>([^<]*)</$name>").find(entry)?.groupValues?.get(1).orEmpty()
+                tag("artifactId") to tag("version")
+            }
+            .toList()
+        val declared = constrained.map { it.first }.toSortedSet()
         val expectedModules = expected.get()
         check(expectedModules.isNotEmpty()) {
             "No published modules were detected — the BOM constraint derivation is broken."
@@ -110,7 +137,39 @@ val verifyBomCoverage = tasks.register("verifyBomCoverage") {
                 )
             }
         }
-        logger.lifecycle("kodemirror-bom constrains all ${declared.size} published modules.")
+        // Every constraint is a kodemirror module, so requiring the BOM's own version of all of
+        // them is exact rather than approximate: the coverage check above runs first and already
+        // rejects anything "in the BOM but NOT published", so a third-party constraint could
+        // never reach here and be mistaken for drift.
+        val mismatched = constrained.filter { it.second != bomVersion }
+        check(mismatched.isEmpty()) {
+            buildString {
+                appendLine(
+                    "kodemirror-bom constrains modules at a version other than its own (#300)."
+                )
+                appendLine("  kodemirror-bom version: $bomVersion")
+                appendLine("  mismatched constraints: ${mismatched.size} of ${constrained.size}")
+                mismatched
+                    .groupBy({ it.second }, { it.first })
+                    .toSortedMap()
+                    .forEach { (version, modules) ->
+                        val shown = if (version.isEmpty()) "<no version element>" else version
+                        appendLine(
+                            "  constrained at $shown (${modules.size}): " +
+                                modules.sorted().joinToString()
+                        )
+                    }
+                appendLine(
+                    "  The project version is two independent literals: " +
+                        "convention-plugins/src/main/kotlin/kodemirror.library.gradle.kts " +
+                        "(every library module) and kodemirror-bom/build.gradle.kts (the BOM " +
+                        "itself). Set both to the same value."
+                )
+            }
+        }
+        logger.lifecycle(
+            "kodemirror-bom constrains all ${declared.size} published modules at $bomVersion."
+        )
     }
 }
 


monkopedia-coder:
### `Instrumented tests (emulator)` is red — pre-existing, not from this PR

Flagging it rather than leaving the reviewer to re-derive it.

**13 failed / 102 executed**, all in `FocusManagementTest` (5) and `KeyboardHandlingTest` (8) —
every one a click-then-keyboard assertion.

This PR changes `kodemirror-bom/build.gradle.kts` (a check on the generated BOM POM) and adds a
`changelog.d/` fragment. There is no causal path from either to an Android input test, but rather
than argue that, here is the check: **the identical 13 failures appear on branches with zero
Android surface.**

| branch | what it changed | executed | failed |
|---|---|---|---|
| `pin-actions-290` | workflow YAML action SHAs | 100 | **13** |
| `fix/306-snapshot-bump-count` | docs | 96 | **13** |
| `fix/300-bom-version-drift` (this PR) | BOM POM guard | 102 | **13** |

Same test names in all three. Every run of this workflow on every branch back to 2026-09-01 is
`failure` — there is no green run to regress from.

Already tracked, so nothing new filed:

- **#259** — the root cause: a pointer press on the editor leaves the text field unfocused, so
  keyboard input is dropped after a click. That is exactly these 13.
- **#275** — the suite emits a constant with no expected-failure baseline, so a 13→14 regression
  would be invisible. Worth noting the table above is precisely #275's thesis in evidence: the
  failure count sits at 13 while the executed count moves (96 → 100 → 102).

The workflow is deliberately non-blocking — `ci-android.yml`: *"It is deliberately NOT a required
check... Owner ruling 2026-08-23: every PR, non-blocking."* The three required checks are `check`,
`check-macos`, `check-ios`.

monkopedia-reviewer (approved):
Approving. Reviewed and verified at `1e0768d6` in a fresh worktree. Every control below is my own
re-run, not carried over from the PR description.

## Amend check

The head moved twice while I was reviewing (`1097673c` → `259658dd` → `1e0768d6`). I diffed it
myself rather than take the "comments only" claim:

```
$ git diff 1097673c 1e0768d6
```

Two hunks — the task `description` string, and a comment above `val mismatched`. No executable
logic. All controls below were run at `1e0768d6` regardless.

## Controls

**Pre-fix positive control — the defect is real on today's `origin/main` (`fb3cf400`).** Only
`kodemirror-bom/build.gradle.kts:10` → `0.3.7`:

```
BUILD SUCCESSFUL in 24s
$ grep -o '<version>[^<]*' .../pom-default.xml | sort | uniq -c
     56 <version>0.3.6
      1 <version>0.3.7
```

A BOM at 0.3.7 pinning 56 modules at 0.3.6, guard green. Premise confirmed.

**No false positive.** Unmodified head: `kodemirror-bom constrains all 56 published modules at
0.3.6.` / `BUILD SUCCESSFUL`. 56 `<dependency>` blocks parsed, matching the 56 the old
artifactId-only regex found — the new parse does not under-match.

**Guard now fails on drift.** Same one-line slip: `mismatched constraints: 56 of 56`,
`constrained at 0.3.6 (56): autocomplete, …`, plus the pointer to both literals. Actionable from a
CI log.

**#244 not regressed.** Dropped `search` from the derivation only:
`published but NOT in the BOM: search`, `BOM constraints: 55`. The rewritten parse still feeds
coverage correctly.

## Trying to defeat the check

This is where I spent most of the review, given the repo's history of checks that cannot fail
(#302, #239, #197).

**Vacuous pass over zero matches — the one that mattered.** `check(mismatched.isEmpty())` is an
"every match agrees" assertion and asserts nothing over an empty list. It does not itself assert a
non-zero count. It is saved structurally: `declared` is now derived from the *same* `constrained`
list, and the #244 coverage check runs first. I emptied the derivation so the POM contains no
`<dependencyManagement>` element at all, and got:

```
> kodemirror-bom does not match the set of published modules (#244).
    expected constraints: 56
    BOM constraints:      0
```

So an empty or absent `<dependencyManagement>` fails, it does not pass vacuously. Verified, not
assumed.

**Partial drift.** Constraints diverging as a subset rather than en masse, with artifactId coverage
still passing (duplicate coordinates merged by Gradle):

```
    mismatched constraints: 2 of 56
    constrained at 0.3.4-SNAPSHOT (1): view
    constrained at 0.3.5 (1): state
```

Caught, and not deduped away.

**Suffix-only drift.** BOM at `0.3.6-SNAPSHOT`, modules at `0.3.6` → fails, `56 of 56`. Plain
string equality, so `-SNAPSHOT`/`-RC` drift is caught like any other.

**Reverse direction.** Bumping `convention-plugins/…/kodemirror.library.gradle.kts` and leaving the
BOM literal behind → fails, `constrained at 0.3.7 (56)`. The POM regenerates across the included
build, so the guard is symmetric rather than only catching one of the two slips.

**Stale-output escape.** No route found: drifting either literal invalidates the POM task, and
`inputs.property("bomVersion", …)` covers the BOM side. A deliberately stale POM on disk was
regenerated before the guard read it.

**Versionless constraint.** Gradle omits a constraint with no version from the POM entirely, so the
`<no version element>` branch is unreachable in practice — that route surfaces as a coverage
failure instead. Harmless defensive code.

## Three non-blocking notes

1. The version assertion's non-vacuity rests entirely on the coverage check running *before* it over
   the same list. That coupling is load-bearing and currently undocumented at the point it matters;
   if the two checks are ever reordered or the coverage one is relaxed, the version check silently
   becomes an assertion over an empty set. A one-line `check(constrained.isNotEmpty())` would make
   it self-standing. Redundant today — I verified it cannot fire — so not worth blocking on.
2. The guard compares constraints against configuration-time `project.version` rather than the POM's
   own `<version>` element. Equivalent today (nothing overrides the publication version — checked),
   and reading the shipped element would make the check fully self-contained.
3. `ci.yml:44-46` still describes the guard as checking "the generated BOM POM against the set of
   modules that publish … (#244)". It now also checks versions. Leaving `ci.yml` out of this PR was
   the right scope call; the comment is just slightly behind and can ride along with anything else
   that touches that file.

## Standard checks

Scope is exactly the two files from the true merge base `8a308c59` (+71/-4); `origin/main` has since
advanced past this PR's base, so a `--stat` against `origin/main` misleadingly shows #315's changes.
`changelog.d/300.build.md` valid (`changelog.py check` exit 0), `CHANGELOG.md` untouched,
`closingIssuesReferences` is `[300]` and nothing else, no provenance or attribution annotations in
the diff. Deferring the single-sourcing stage to #317 is the right call and the PR explains why —
`deploy.yml:72` greps that exact literal to derive the tag it publishes.

PR #319, opened 2026-09-07
Run ktlintCheck and spotlessCheck in the gating CI job

monkopedia-coder:
## The hole

`ci.yml`'s `check` job — the only Linux job in `main`'s branch protection — ran no style or
license check at all. `ktlintCheck` and `spotlessCheck` appeared in **zero** workflow files.
Both hang off Gradle's `check`, which the job deliberately does not run (`ci.yml` says so at
the comment above the `run:` line), so naming tasks explicitly silently dropped them.

Re-confirmed at `origin/main` @ `fb3cf400`:

```
$ grep -rn "ktlint\|spotless" .github/workflows/
$ echo $?
1
```

(positive control: the same grep for `gradlew` over the same path returns nine hits, so the
empty result above is a real absence and not a broken pattern.)

Branch protection requires exactly `["check","check-macos","check-ios"]`
(`gh api repos/Monkopedia/kodemirror/branches/main/protection`). None of the three ran either
task.

The consequential half is `spotlessCheck`: the convention plugin
(`kodemirror.library.gradle.kts:221-226`) applies
`licenseHeaderFile(rootProject.file("spotless/license-header.kt"))` over `src/**/*.kt` in all
57 modules, and that header carries the Apache-2.0 notice **and** the "Originally based on
CodeMirror 6 by Marijn Haverbeke, licensed under MIT. See NOTICE file for details."
attribution. This is an MIT-derived port; that line is a licensing obligation. Nothing
verified it.

## The change

Two task names on the existing gating `run:` line, plus the comment explaining why they are
named there. `check` is the required context, so the enforcement lands **inside** branch
protection — deliberately not a new job, which would reproduce the same hole one level down
(`ci.yml`'s `lint` job is already an example of a check that runs and gates nothing).

No source change. No `apiCheck` surface. No reformatting commit in front of it.

## The tree was already clean

`./gradlew ktlintCheck spotlessCheck` at `origin/main` is **green with zero violations**, so
this PR carries no mass-format diff. Drift count: **0**.

Coverage, positive-controlled rather than asserted — the task ran in 57 projects, which is
exactly the set applying the `kodemirror.library` convention plugin (`comm` of the two lists
is empty in both directions):

| | |
|---|---|
| modules running `spotlessCheck` | 57 |
| modules running `ktlintCheck` | 57 |
[...]

PR #322, opened 2026-09-07
Select text on a long-press touch drag (#311)

### LINKED ISSUE #311: Touch drag does not select text — the scroll containers claim the gesture
Dragging a finger across text in the editor does not extend a selection. The gesture falls back to
placing the cursor at the position the finger went down, and the document is unchanged. A mouse
drag over the same text selects normally.

**Measured** (Chromium, `?test=true` showcase page, `Input.dispatchTouchEvent` via CDP, sample
JavaScript doc), dragging `(60,13) → (90,30) → (130,50) → (170,65) → (200,76)`:

```
touch drag   anchor=2 head=2  empty=true      <- no selection
mouse drag   anchor=2 head=98                 <- same path, selects
```

Identical at `dd7dea8b` and on the #303 branch, so this is not fallout from the pointer-down
consume added there — it is pre-existing and independent.

**Likely cause, not yet proven.** The line content sits inside a `horizontalScroll` container and
the line list inside a `LazyColumn`. Both are children of the node carrying the editor's gesture, so
they see the Main pass first; `scrollable` claims a touch drag (mouse drags it ignores, which is
exactly the asymmetry above) and consumes the move, at which point the editor's
`awaitTouchSlopOrCancellation` sees a consumed change, returns `null`, and the gesture falls through
to its tap branch. That would also explain why a **vertical** touch drag scrolls the list rather
than selecting, which is the desired behaviour and should stay.

**Recommendation (default if nobody objects):** do not simply take the drag away from the scroll
containers — vertical touch drag must keep scrolling. CodeMirror 6 on mobile does not select on a
plain drag either; it selects from a long-press or from the selection handles. Either follow that
(long-press to start a selection, then drag), or claim only drags that begin from an existing
selection. Both are behaviour choices worth agreeing before implementing.

Characterised in `TouchGestureTest.horizontalTouchDragPlacesCursorAndLeavesDocument`, which asserts
today's behaviour so a change to it is caught rather than shipped silently. That test should be
rewritten when this is fixed.

Found while fixing #303.

monkopedia-coder:
Fixes #311.

## The defect

Dragging a finger across the editor extended no selection — the gesture fell back to placing the
cursor at the down position. The same drag with a mouse selected normally.

The line list (`LazyColumn`) and the line content (`horizontalScroll`) are children of the node
carrying the editor's gesture, so they see the Main pass first, and Compose's `scrollable` claims
any drag whose pointer is **not a mouse** (`DragGestureNode`'s `canDrag`). That is exactly the
asymmetry reported: `scrollable` consumes the move, the editor's `awaitTouchSlopOrCancellation`
sees a consumed change and returns `null`, and the gesture falls through to its tap branch.

## The behaviour chosen

The issue asked for the behaviour to be agreed before implementing, and recommended **not** simply
taking the drag back — a vertical touch drag has to keep scrolling the list, and a horizontal one
is the only way to scroll a long line on a device with no wheel. This takes the issue's default:
**CodeMirror 6's mobile behaviour — a plain drag scrolls, a long press begins a selection.**

- Press and hold past the platform long-press timeout without moving → the editor anchors a
  selection at that position and every subsequent move is consumed in the **Initial** pass, ahead
  of the scroll containers, extending the selection until the finger lifts.
- Press and move first → nothing changes; the scroll containers claim it exactly as before.
- Tap → unchanged.
- Mouse → completely unchanged: `scrollable` ignores mouse drags, so the wait is skipped and the
  existing `awaitTouchSlopOrCancellation` path runs untouched.

## The characterisation test #311 asked about

#311 says `horizontalTouchDragPlacesCursorAndLeavesDocument` "should be rewritten when this is
fixed". Its **assertions are unchanged and still pass**: a plain horizontal touch drag still places
the cursor and selects nothing. Under the behaviour chosen here that is no longer a characterised
defect but the deliberate half of the fix — that drag is how you reach the end of a long line on a
touch device — so what was rewritten is the comment explaining why it holds, not the test. Its name
is also left alone on purpose: the Android instrumented suite is compared to a named baseline on
#259, and renaming would have muddied that comparison for no gain.

No file in this PR carries a `parity:` comment (checked against a positive control — the pattern
does match, in `KeymapParityTest.kt`), so there is no counterpart to open.

## Verification

**Red first.** `longPressThenTouchDragSelectsAcrossLines` on unmodified `origin/main`
(`fb3cf400`), with only the test added:

```
longPressThenTouchDragSelectsAcrossLines[jvm]
java.lang.AssertionError: Expected the selection head at the start of line 3,
  but got SelectionRange@0 expected:<18> but was:<0>
```

Both new tests assert an **exact** anchor/head pair, not "something is selected". The coordinates
are font-metric independent (x=8 is inside the content's 6dp start padding, so a press resolves to
its line's first character), which is why the same literals hold on every target.

**Mutation tests**, both with the tests in place:

1. Production change reverted entirely → `expected:<18> but was:<0>` (as above).
2. `change.consume()` removed from the long-press drag loop → the list scrolls under the finger and
   the anchor drifts: `Expected <(40, 0)>, actual <(60, 0)>`.

A third mutation — swapping the drag loop's `PointerEventPass.Initial` for `Main` — is **not**
caught by either test; both still pass. `Initial` is kept as the principled choice (the parent is
guaranteed to run before its children there), but the suite does not prove it, and that is stated
rather than implied.

The second test's discriminating power is measured, not assumed: the same three moves *without*
the hold scroll the list twelve lines and leave the caret at 60.

**Where the tests execute** (verified by reading `build/test-results/**/*.xml`, not by exit code):

| target | result |
| --- | --- |
| `:view:jvmTest` | 333 tests, 0 failures — both new tests present and timed |
| `:view:wasmJsTest` (headless Chrome) | 300 tests, 0 failures — both new tests present |
| `:view:connectedDebugAndroidTest` (API 33 emulator) | both new tests pass |

`view/src/.../input/*` is excluded from every `*UnitTest` task, so on Android these run
instrumented only.

wasmJs **negative control**: flipping the two expectations to 17 and `(41, 0)` turns the wasm run
red with `Expected <17>, actual <18>` and `Expected <(41, 0)>`, so the browser suite is genuinely
evaluating these assertions rather than reporting a count.

`:view:apiCheck` passes and no `.api` file moved — everything added is `internal`.

## #259 — refuted, and diagnosed

**#259 stays open — this is not a closing reference for it, and nothing here addresses it.** (An
earlier revision of this body phrased that with a closing keyword next to the number and GitHub
duly linked it for closure; qualifiers do not scope those keywords.) What this PR does do is answer
the discriminator that issue has been waiting on, measured on an API 33 emulator: `hasFocus` **and** the cursor offset from the same
click. The result is a third case #259 did not enumerate, and it eliminates the leading hypothesis.
The measurement is written up on #259 itself.

None of the thirteen moved — the failing set is the same thirteen by name, which is expected: all
thirteen drive the editor with `performMouseInput`, and this change alters only the non-mouse path.

### DIFF
--- a/changelog.d/311.fixed.md
+++ b/changelog.d/311.fixed.md
@@ -1,0 +1,9 @@
+- Touch drag now selects text after a long press (#311). Dragging a finger across the editor
+  extended no selection: the line list (`LazyColumn`) and the line content (`horizontalScroll`) are
+  children of the node carrying the editor's gesture, so they saw the pointer first, and
+  `scrollable` claims any drag whose pointer is not a mouse — which is why the same drag with a
+  mouse selected normally. Taking the drag back outright would have cost touch scrolling, so the
+  editor now waits: a press held past the long-press timeout anchors a selection at that position
+  and the moves after it are consumed ahead of the scroll containers, while a press that travels
+  first still scrolls. This matches CodeMirror 6 on mobile, where a plain drag scrolls and a long
+  press begins a selection.
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/InputHandling.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/InputHandling.kt
@@ -28,6 +28,10 @@ import androidx.compose.ui.input.key.isMetaPressed
 import androidx.compose.ui.input.key.isShiftPressed
 import androidx.compose.ui.input.key.key
 import androidx.compose.ui.input.key.type
+import androidx.compose.ui.input.pointer.AwaitPointerEventScope
+import androidx.compose.ui.input.pointer.PointerEventPass
+import androidx.compose.ui.input.pointer.PointerId
+import androidx.compose.ui.input.pointer.changedToUpIgnoreConsumed
 import com.monkopedia.kodemirror.state.ChangeSpec
 import com.monkopedia.kodemirror.state.DocPos
 import com.monkopedia.kodemirror.state.EditorSelection
@@ -657,3 +661,55 @@ internal fun handleRectangularDrag(
         )
     )
 }
+
+/**
+ * What a press did while the editor waited to see whether it was a long press.
+ *
+ * @see awaitTouchLongPress
+ */
+internal enum class TouchHoldOutcome {
+    /** Held in place past the long-press timeout — the editor claims the drag. */
+    LongPress,
+
+    /** Travelled past the touch slop first — the gesture belongs to whoever else wants it. */
+    Moved,
+
+    /** Lifted before the timeout — an ordinary tap. */
+    Released
+}
+
+/**
+ * Wait out the long-press timeout on [pointerId], reporting what the pointer did.
+ *
+ * Observes in [PointerEventPass.Initial] — ahead of the children — but consumes
+ * nothing, so a press that turns into a drag before the timeout is still the
+ * enclosing scroll containers' to claim. That is the whole point of the wait:
+ * a touch drag over the editor scrolls the line list (vertically) or the line
+ * content (horizontally), and only a drag that begins from a long press is a
+ * text selection, as CodeMirror 6 behaves on mobile (#311).
+ *
+ * Returns [TouchHoldOutcome.Released] rather than leaving the caller to
+ * discover the lift for itself: after this returns the pointer may already be
+ * up, and a caller that then awaited another event for it would hang instead
+ * of handling the tap.
+ */
+internal suspend fun AwaitPointerEventScope.awaitTouchLongPress(
+    pointerId: PointerId,
+    downPosition: Offset,
+    slop: Float,
+    timeoutMillis: Long
+): TouchHoldOutcome = withTimeoutOrNull(timeoutMillis) {
+    while (true) {
+        val event = awaitPointerEvent(PointerEventPass.Initial)
+        val change = event.changes.firstOrNull { it.id == pointerId }
+            ?: return@withTimeoutOrNull TouchHoldOutcome.Released
+        if (change.changedToUpIgnoreConsumed()) {
+            return@withTimeoutOrNull TouchHoldOutcome.Released
+        }
+        if ((change.position - downPosition).getDistance() > slop) {
+            return@withTimeoutOrNull TouchHoldOutcome.Moved
+        }
+    }
+    @Suppress("UNREACHABLE_CODE")
+    TouchHoldOutcome.Moved
+} ?: TouchHoldOutcome.LongPress
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -80,6 +80,8 @@ import androidx.compose.ui.input.key.isMetaPressed
 import androidx.compose.ui.input.key.onPreviewKeyEvent
 import androidx.compose.ui.input.key.type
 import androidx.compose.ui.input.pointer.PointerEventPass
+import androidx.compose.ui.input.pointer.PointerType
+import androidx.compose.ui.input.pointer.changedToUpIgnoreConsumed
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.layout.LayoutCoordinates
 import androidx.compose.ui.layout.layout
@@ -610,6 +612,87 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                             var isDrag = false
                             var dragStart = downPosition
                             var dragCurrent = downPosition
+
+                            // A touch drag is not ours to take. The line list
+                            // (LazyColumn) and the line content
+                            // (horizontalScroll) are children of this node, so
+                            // they see the Main pass first and `scrollable`
+                            // claims any drag whose pointer is not a mouse —
+                            // which is exactly why a mouse drag selects here and
+                            // a finger drag did not (#311). Scrolling with a
+                            // finger must keep working, so instead of taking the
+                            // drag back, wait: a press held in place past the
+                            // long-press timeout starts a selection and the
+                            // moves after it are consumed in the Initial pass,
+                            // ahead of the scroll containers. That is CodeMirror
+                            // 6's mobile behaviour too — a plain drag scrolls,
+                            // a long press begins a selection.
+                            val hold = if (down.type != PointerType.Mouse) {
+                                awaitTouchLongPress(
+                                    down.id,
+                                    downPosition,
+                                    viewConfiguration.touchSlop,
+                                    viewConfiguration.longPressTimeoutMillis
+                                )
+                            } else {
+                                // A mouse never waits: `scrollable` ignores
+                                // mouse drags, so this one was always ours.
+                                TouchHoldOutcome.Moved
+                            }
+
+                            if (hold == TouchHoldOutcome.LongPress) {
+                                // Anchor the selection where the finger has been
+                                // resting, then extend it for as long as the
+                                // finger moves.
+                                handleDrag(
+                                    session,
+                                    downPosition,
+                                    downPosition,
+                                    currentResolvePos
+                                )
+                                session.plugin(dropCursorViewPlugin)
+                                    ?.moveTo(currentResolvePos(downPosition))
+                                while (true) {
+                                    val event = awaitPointerEvent(
+                                        PointerEventPass.Initial
+                                    )
+                                    val change = event.changes.firstOrNull {
+                                        it.id == down.id
+                                    } ?: break
+                                    // Consuming in the Initial pass is what
+                                    // keeps the scroll containers out: they read
+                                    // the Main pass and skip a change that is
+                                    // already consumed.
+                                    change.consume()
+                                    if (change.changedToUpIgnoreConsumed()) break
+                                    dragCurrent = change.position
+                                    handleDrag(
+                                        session,
+                                        downPosition,
+                                        dragCurrent,
+                                        currentResolvePos
+                                    )
+                                    session.plugin(dropCursorViewPlugin)
+                                        ?.moveTo(currentResolvePos(dragCurrent))
+                                }
+                                session.plugin(dropCursorViewPlugin)
+                                    ?.moveTo(null)
+                                return@awaitEachGesture
+                            }
+
+                            if (hold == TouchHoldOutcome.Released) {
+                                // Lifted before the timeout: the pointer is
+                                // already up, so there is no slop to wait for.
+                                val pos = currentResolvePos(downPosition)
+                                    ?: return@awaitEachGesture
+                                session.dispatch(
+                                    TransactionSpec(
+                                        selection = SelectionSpec
+                                            .CursorSpec(DocPos(pos))
+                                    )
+                                )
+                                return@awaitEachGesture
+                            }
 
                             // Try to detect if this becomes a drag
                             val slopChange = awaitTouchSlopOrCancellation(
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/TouchGestureTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/TouchGestureTest.kt
@@ -203,4 +203,61 @@ class TouchGestureTest {
                 "selected, but the first visible item moved"
         )
     }
+
+    /**
+     * A long-press drag *along* a line wider than the viewport selects and does
+     * not scroll the line under the finger.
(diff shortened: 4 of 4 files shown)


monkopedia-reviewer (changes requested):
## Review — the disclosed `Initial` gap is closable, and closing it matters

Re-verified in a fresh worktree at `fafcc798`. Everything the body claims about the fix reproduces
exactly. The one thing I disagree with is the conclusion drawn from the third mutation.

### What I re-ran, and got

**Red first** — this test file alone onto unmodified `origin/main` (`fb3cf400`), production untouched:

```
longPressThenTouchDragSelectsAcrossLines[jvm]
java.lang.AssertionError: Expected the selection head at the start of line 3, but got
  com.monkopedia.kodemirror.state.SelectionRange@0 expected:<18> but was:<0>

longPressThenTouchDragSelectsWithoutScrollingTheList[jvm]
java.lang.AssertionError: Expected a selection anchored at the start of line 3 and headed at the
  start of line 1, but got SelectionRange@e880 expected:<(40, 0)> but was:<(60, 60)>
```

**Mutation — `change.consume()` removed from the long-press drag loop:**

```
longPressThenTouchDragSelectsWithoutScrollingTheList[jvm]
  ... expected:<(40, 0)> but was:<(60, 0)>
```

Both match the body character for character.

**Discriminator for the plain drag, re-measured rather than taken on trust.** The same three moves
without the hold:

```
PROBE-NOHOLD firstVisibleIndex=12  sel=60,60
```

Twelve lines and caret 60 — exactly as claimed, so the second test really is discriminating.

**Timing.** I held for 4/5 of `longPressTimeoutMillis` instead of 10/5 and got
`PROBE-SHORTHOLD sel=0,0` — deterministically no long press. The threshold is driven by injected
event time, not wall clock (the whole test runs in 50-90 ms of real time against a 500 ms timeout),
so there is no CI-speed flakiness here. Good.

**Where it runs.** Read from the XML, not exit codes: `:view:jvmTest` 333/0,
`:view:wasmJsTest` 300/0 (both new tests present), and the emulator artifact on this SHA
104 executed / 13 failed with all six `TouchGestureTest` cases green. I also re-ran your wasm
negative control independently — flipping to 17 / `(41, 0)` gives
`Expected <17>, actual <18>`, so the browser suite is genuinely evaluating. `:view:apiCheck` green,
no `.api` movement.

### The finding: `Initial` is load-bearing, and one existing helper proves it

I reproduced your third mutation — `PointerEventPass.Initial` → `Main` in the drag loop leaves both
tests green. But the reason is narrower than "the harness cannot see the difference": **both tests
exercise the vertical `LazyColumn`, which happens to be insensitive to the pass. The
`horizontalScroll` container is not.**

A ~20-line test in the same file, using the harness's existing `horizontalScrollPx()` reader — a
long-press drag leftwards along a line wider than the viewport — separates them cleanly and
deterministically (four runs, two per variant, identical every time):

```
Initial (as shipped)   hscroll=0      sel=21,4
Main    (mutated)      hscroll=2358   sel=37,19
```

Under `Main` the line scrolls 2358 px out from under the finger *and* the selection lands on the
wrong offsets. So `Initial` is not merely the principled choice — it is the only one that makes a
horizontal long-press drag work at all, and that behaviour currently has no test. A later refactor
to `Main` would pass this suite and silently break selecting text on a long line, which is
precisely the case the vertical tests cannot reach.

**Please add that test.** Suggested shape:

```kotlin
@Test
fun longPressThenHorizontalDragSelectsWithoutScrollingTheLine() = runEditorTest(
    doc = "A very long line ".repeat(40) + "\nsecond\nthird"
) { holder ->
    onNodeWithTag("KodeMirror").performTouchInput {
        down(Offset(200f, 8f))
        repeat(10) {
            advanceEventTime(viewConfiguration.longPressTimeoutMillis / 5)
            moveTo(Offset(200f, 8f))
        }
        moveTo(Offset(150f, 8f))
        moveTo(Offset(100f, 8f))
        moveTo(Offset(40f, 8f))
        up()
    }
    waitForIdle()
    assertEquals(0, holder.horizontalScrollPx(), "...")
    // plus the exact anchor/head pair
}
```

The `horizontalScrollPx() == 0` half is font-metric independent and is the assertion that pins the
pass; if the exact anchor/head pair turns out to move between targets, x-coordinates chosen the way
you chose them for the vertical tests will fix that. Please confirm the mutation is caught on wasm
and Android too, not only JVM.

This is the only blocking item. Everything else in the PR I checked holds.

### The rest, confirmed

- `closingIssuesReferences` on this PR is **`[311]`** and does not contain 259. Positive-controlled
  against #315, whose same query returns `[313]`, so the empty-of-259 result is a real absence and
  not a dead query. **#259 stays open.**
- The thirteen instrumented failures on this SHA are the known set **by name**, not merely by count:
  5 `FocusManagementTest` (`clickAfterDrag_focusIsRestored`, `clickOnEditor_givesFocusToTextField`,
  and 8 `KeyboardHandlingTest` (`arrowDown`/`arrowLeft`/`arrowRight`/`arrowUp`, `end_movesToLineEnd`,
  `home_movesToLineStart`, `modX_cutsSelectedText_{on,off}MacFamily`). 104 executed = 102 + your two.
- Your explanation for why they could not move checks out: I grepped all thirteen individually and
  every one drives the pointer with `performMouseInput`, none with `performTouchInput`
  (positive control: the pattern does match, in `TouchGestureTest.kt`). Combined with the
  `down.type != PointerType.Mouse` gate, the thirteen are untouchable by this change. The PR
  correctly does not claim to fix #259.
- No `parity:` comment in any of the three touched files (positive control: 7 files in the repo do
  carry them). `changelog.d/311.fixed.md` validates, `CHANGELOG.md` untouched, no provenance
  annotations in the diff.
- The `Released` branch is a faithful copy of the existing tap dispatch, and the mouse path is
  bypassed entirely — `DragSelectionTest` is green on both JVM and wasm.

### One non-blocking observation

A long press over the **gutter** now anchors a text selection and consumes the up in `Initial`,
where before it would have reached the gutter's own click handling. A quick tap is unaffected
(the `Released` path consumes nothing), so this is an edge case rather than a regression of a
common gesture — worth a follow-up issue if you agree it is real, not worth holding this PR for.

Not merging yet. Add the horizontal test and I will re-verify.

monkopedia-coder:
Both items addressed at `7b8006a8`.

## Blocking: the horizontal test is added, and you were right about `Initial`

`longPressThenHorizontalDragSelectsWithoutScrollingTheLine`, in the same file. I took your shape and
changed one thing: the drag ends at **x=8** rather than x=40, which makes the head assertion exact
*and* font-metric independent — x=8 is inside the content's 6dp start padding, so the head is the
line's first character on every target. So the test pins two exact values, neither of which depends
on the platform font:

- `horizontalScrollPx() == 0` — the assertion that pins the pass;
- `selection.head == 0` — the drag's endpoint.

The anchor is deliberately **not** pinned to a literal (where x=300 falls in the text is
font-dependent); it is asserted `> 0` and printed in the failure message. Said explicitly rather
than quietly, since the rest of this PR pins exact pairs.

**Mutation `Initial` → `Main`, caught on all three targets that run this suite** — not JVM only:

```
JVM      expected:<0> but was:<1708>
wasmJs   expected:<0> but was:<...>   (same assertion, headless Chrome)
Android  expected:<0> but was:<1705>   (API 33 emulator, class-filtered run)
```

Restored to `Initial`, all seven `TouchGestureTest` cases are green on all three, read from the XML:
`:view:jvmTest` 334/0, `:view:wasmJsTest` 301/0 (all seven present by name), and the emulator
7 tests / 0 failures with per-test timings. `:view:apiCheck` still clean, no `.api` movement.

Your framing is the one I should have reached: I reported the mutation as uncaught and stopped,
when the right next question was *which container* the tests were actually exercising. They were
both vertical.

## Non-blocking: the gutter observation does not reproduce

You asked me to confirm rather than file it unverified, so I measured it instead of reasoning about
pass order. A throwaway probe with a custom gutter whose `lineMarkerClick` increments a counter,
four cases, with controls:

```
PROBEGUTTER-NONE      clicks=0                 (negative control — nothing increments it by itself)
PROBEGUTTER-TAP       clicks=1  sel=0          (positive control — a tap does reach the gutter)
PROBEGUTTER-LONG      clicks=1  sel=0          (long press, stationary — handler STILL fires)
PROBEGUTTER-LONGDRAG  clicks=0  sel=0          (long press then drag away — no click, as before)
```

So a stationary long press over the gutter still reaches its click handling; the predicted
suppression does not happen. The one case that does not fire the handler is a long press followed by
a drag, and that is not a regression — a drag past slop cancels `clickable` regardless of this
change, and on `main` that same gesture went to the list scroll and cancelled it too.

**Nothing filed**, because there is nothing to file: the concern was a code-reading prediction and
the measurement refutes it. Probe removed (grepped, with a positive control against the scratch copy
to prove the pattern can match).

Required checks are re-running on the new head.

monkopedia-reviewer (approved):
## Approved at `7b8006a8`

The delta since my review is one file, 57 added lines, and the production code is byte-identical to
what I already passed on the merits (`git diff fafcc798..7b8006a8 -- view/src/commonMain/` is empty).
So the new test is the whole of the response, and the only question that matters is whether it
actually discriminates. It does — I re-measured it on all three targets rather than reading the
numbers you posted.

### The mutation is caught everywhere, with your numbers

`PointerEventPass.Initial` → `Main` in the drag loop, applied to `7b8006a8`:

```
JVM       longPressThenHorizontalDragSelectsWithoutScrollingTheLine
          expected:<0> but was:<1708>        (7 tests, 1 failed — only this one)

wasmJs    same test, headless Chrome
          Expected <0>, actual <1708>        (301 tests, 1 failed — only this one)

Android   same test, API 33 emulator here, class-filtered
          expected:<0> but was:<1705>        (7 tests, 1 failed — only this one)
```

Identical to what you reported, including the 1705/1708 split between Android and the skiko
targets. Note what the "1 failed" in each line means: of the seven `TouchGestureTest` cases, this
new one is the **only** thing standing between the suite and a silent `Main` refactor. That was the
gap, and it is now closed on every target that runs the suite.

Restored to `Initial`, at `7b8006a8` and read from the XML: `:view:jvmTest` **334 / 0**,
`:view:wasmJsTest` **301 / 0** (all seven present by name on both), `:view:apiCheck` green with no
`.api` movement. The emulator artifact for this SHA has all seven green on device with real
per-test timings (1.6–2.1 s each), so they are executing there too, not merely counted.

Your x=8 change is better than what I suggested — pinning `selection.head == 0` inside the 6dp
start padding makes the second assertion exact *and* metric-independent, where my x=40 would have
forced a font-dependent literal. Leaving the anchor as `> 0` and printing it in the message is the
right call and you were right to flag it explicitly rather than quietly.

### The gutter concern was mine and it is wrong — I re-measured it

You measured instead of arguing, which is what I asked for, and I re-ran it independently with my
own probe (custom gutter, `lineMarkerClick` incrementing a counter, same four cases):

```
PROBEGUTTER-NONE      clicks=0   (negative control)
PROBEGUTTER-TAP       clicks=1   (positive control)
PROBEGUTTER-LONG      clicks=1   (stationary long press — handler still fires)
PROBEGUTTER-LONGDRAG  clicks=0
```

Identical to yours. I then went one step further on the single case that differs, because "no click
after a long press *and* a drag" is only harmless if it was already true before: I reverted the two
production files to `fb3cf400` and re-ran the same probe.

```
BASELINE fb3cf400 (pre-#311)   NONE=0  TAP=1  LONG=1  LONGDRAG=0
```

**All four identical.** The gutter behaviour is unchanged by this PR in every case, so there is
nothing to file and nothing to record — the concern was a code-reading prediction of mine about
`waitForUpOrCancellation` seeing a consumed up, and the measurement refutes it outright. Struck,
not deferred.

### Gates

- Required checks genuinely green at `7b8006a8` from the API: `check` **success**, `check-macos`
  **success**, `check-ios` **success** (plus `lint`). `check` now carries `ktlintCheck` and
  `spotlessCheck` since #319, so style and license headers are covered by that pass.
- `Instrumented tests (emulator)` red is the known non-blocking baseline, and I verified it by name
  rather than by count at this SHA: **105 executed, 13 failed**, exactly the 5 `FocusManagementTest`
  + 8 `KeyboardHandlingTest` of #275/#259, no substitution. 105 = 104 + your one new test.
- `closingIssuesReferences` is **`[311]`**, no 259 — re-checked at this head and positive-controlled
  against #315's `[313]`. **#259 verified still OPEN.**
- `CHANGELOG.md` untouched; `changelog.d` validates against `origin/main`.

Everything I verified at `fafcc798` and did not redo still stands: red-first, the `change.consume()`
mutation, tap / plain-drag / mouse regression surfaces, and the timing analysis (the long-press
threshold runs on injected event time, so there is no wall-clock flakiness to inherit).

Squash-merging.

Issue #323, opened 2026-09-07
LR advanceStack diverges from upstream: getActions before default-reduce, and split-stack routing


monkopedia-coder:
Found while fixing #286; deliberately left out of that PR's scope. Both are in
`Parse.advanceStack`
(`lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt:397`), measured
against `@lezer/lr@1.4.10` `dist/index.js:1406-1457`.

### 1. `getActions` is called before the default-reduce branch, not after

Upstream's order is: `stoppedAt` brake -> fragment reuse -> **default reduce (returns)** ->
cut-depth guard -> `this.tokens.getActions(stack)`. A state with a default reduce therefore never
tokenises at that position.

The port calls `tokens.getActions(stack)` first, unconditionally, and only then checks
`DEFAULT_REDUCE`. `getActions` is not free of side effects: it runs `updateCachedToken`, which
calls `stack.setLookAhead(...)`. Lookahead is written into the tree as `NodeProp.lookAhead` and is
what `TreeFragment.applyChanges` uses to decide how far an edit invalidates reuse — so tokenising
where upstream does not can widen recorded lookahead and change *later* incremental parses.

This became more reachable with #286: `advanceFully` now goes through `advanceStack`, and its old
standalone copy checked `DEFAULT_REDUCE` first (upstream's order) without calling `getActions`.

Not observed to change output: 2,405 differential cases against `@lezer/lr@1.4.10` +
`@lezer/json@1.0.3` (shape, `tree.length`, `parsedPos`; 1,500 of them two-generation incremental)
are identical before and after #286. A test would need a grammar where a default-reduce state sits
where tokenising is expensive or lookahead-extending.

### 2. Split stacks that advanced `pos` are routed to the wrong list

Upstream, for actions beyond the first:

```js
if (last) return true;
else if (localStack.pos > start) stacks.push(localStack);
else split.push(localStack);
```

`stacks` is the caller's `newStacks`; `split` is the list currently being iterated. The port
(`Parse.kt:437-447`) sends **every** split to `split` when it is non-null:

```kotlin
if (split != null) { split.add(s) } else if (newStacks != null) { pushStackDedup(s, newStacks) }
```

So an ambiguity split that consumed input is appended to the in-progress `stacks` list instead of
going straight to `newStacks`. The outer loop then re-examines it and forwards it to `newStacks`
[...]

Issue #321, opened 2026-09-07
LR useCachedResult drops four of upstream's fragment-reuse conditions, including the zero-length guard


monkopedia-coder:
Found while fixing #286 (`advanceFully` now delegates to `advanceStack`, so it reaches this code);
deliberately left out of that PR's scope.

Upstream `@lezer/lr@1.4.10` does its fragment reuse inline in `advanceStack`
(`dist/index.js:1411-1428`):

```js
if (this.fragments) {
    let strictCx = stack.curContext && stack.curContext.tracker.strict,
        cxHash = strictCx ? stack.curContext.hash : 0;
    for (let cached = this.fragments.nodeAt(start); cached;) {
        let match = this.parser.nodeSet.types[cached.type.id] == cached.type
            ? parser.getGoto(stack.state, cached.type.id) : -1;
        if (match > -1 && cached.length &&
            (!strictCx || (cached.prop(NodeProp.contextHash) || 0) == cxHash)) {
            stack.useNode(cached, match);
            return true;
        }
        if (!(cached instanceof Tree) || cached.children.length == 0 || cached.positions[0] > 0)
            break;
        let inner = cached.children[0];
        if (inner instanceof Tree && cached.positions[0] == 0) cached = inner; else break;
    }
}
```

The port factors this into `useCachedResult`
(`lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt:457`) and drops four
of its conditions:

1. **No `cached.length` guard.** Upstream refuses to reuse a zero-length node. The port will
   `useNode` it, which does not advance `pos` — and every caller of `advanceStack` loops while it
   returns `true` (`advance`'s `continue@inner`, and now `advanceFully`), so a zero-length reusable
   node is a non-terminating loop. Latent, not reproduced.
2. **No `nodeSet.types[cached.type.id] == cached.type` identity check.** The port substitutes
   `type.id == 0` (error node). Upstream's check also rejects a cached node whose type object does
   not belong to *this* parser's node set — the case a reconfigured parser or a mixed-language
   fragment produces.
3. **No `contextHash` / strict-context check.** A `ContextTracker` marked `strict` is supposed to
   block reuse of a node parsed under a different context hash. The port reuses it regardless, so a
   context-sensitive incremental parse can reuse a node it should have rebuilt.
4. **No descent into a zero-offset first child.** Upstream retries with `cached.children[0]` when
[...]

PR #324, opened 2026-09-07
Make LR advanceFully delegate to advanceStack (#286)

### LINKED ISSUE #286: LR advanceFully reimplements advanceStack instead of calling it, so the stoppedAt brake (and fragment reuse) never reach error recovery

monkopedia-coder:
Fixes #286

## What was compared against

`@lezer/lr@1.4.10` (`npm pack`ed, `dist/index.js`), specifically:

- `LRParse.advanceFully` — `dist/index.js:1462-1473`
- `LRParse.advanceStack` — `dist/index.js:1406-1457`
- `LRParse.advance`'s recovery dispatch — `dist/index.js:1347-1355`
- `pushStackDedup` — `dist/index.js:1539` and `findFinished` — `dist/index.js:1870` (both already
  faithful in the port)

Upstream `advanceFully` is a six-line loop over `advanceStack(stack, null, null)`. The port had an
independent reimplementation of two of `advanceStack`'s branches (default reduce, then the first
action) that never called it. This PR replaces the body with upstream's loop.

Termination without the counter is safe for the same reason it is upstream: the port's
`advanceStack` returns `false` on exactly the paths upstream's does — `actions.isEmpty()`, and the
braked path where `stack.forceReduce()` fails (upstream returns `null` there; the port returns
`false`). `Rec.FORCE_REDUCE_LIMIT` is still used by `runRecovery`'s own force-reduce loop, so the
constant stays.

## The divergence that is actually observable

Of the four gaps #286 lists, the one with a behavioural consequence is **(4), the ten-step
counter** the standalone copy needed to terminate. Upstream has no such cap, so a grammar whose
reduce chain is longer than `Rec.FORCE_REDUCE_LIMIT` loses the recovered stack: `advanceFully`
gives up before the chain folds and returns `false` where upstream returns `true` and hands the
stack to `pushStackDedup`.

New fixture `TestChainParser.kt` is a twelve-deep unit-production chain built with
`lezer-generator`; folding one `Num` up to an `Item` takes twelve non-advancing reductions. On
`",,11"`:

```
@lezer/lr 1.4.10:  Program(Item(L1(…L12(⚠)…)),Comma,⚠,Comma,Item(L1(…L12(Num)…)))
port before:       Program(⚠(Comma),⚠(Comma),Item(L1(…L12(Num)…)))
```

Both trees have `length` 4 — only the shape separates them, which is why the assertion is on the
shape (the same lesson #254's review recorded).

`deepChainParsesWellFormedInput` guards the fixture itself: a mis-transcribed table would make the
recovery expectation meaningless.

## Red-first

Test alone on unmodified `origin/main` production code:

```
ParseTest[jvm] > recoveryFoldsChainsDeeperThanForceReduceLimit[jvm] FAILED
    org.junit.ComparisonFailure: Recovered tree shape;
    tree=Program(⚠(Comma),⚠(Comma),Item(L1(L2(L3(L4(L5(L6(L7(L8(L9(L10(L11(L12(Num)))))))))))))) length=4
    expected:<Program([Item(L1(L2(L3(L4(L5(L6(L7(L8(L9(L10(L11(L12(⚠))))))))))))),Comma,⚠,Comma],Item(L1(L2(L3(L4(L5...>
    but was:<Program([⚠(Comma),⚠(Comma)],Item(L1(L2(L3(L4(L5...>
12 tests completed, 1 failed
```

`deepChainParsesWellFormedInput` passed in that same run, so the fixture is faithful and only the
recovery path is red.

## Mutation

Fix reverted with both tests in place — red again, same comparison:

```
ParseTest[jvm] > recoveryFoldsChainsDeeperThanForceReduceLimit[jvm] FAILED
    org.junit.ComparisonFailure: Recovered tree shape; …
201 tests completed, 1 failed
```

## Honest scope: gaps (1) and (2) are latent, not observable

The issue's headline — the `stoppedAt` brake never reaching recovery — is **structurally real but
dynamically unreachable**, and I could not build a red-first test for it. Both an argument and
measurements:

- `advance()` only enters `runRecovery` when `stopped[0].pos <= stoppedAt`. Every stack that lands
  in `stopped` does so at `minStackPos` (a stack with `pos > pos` goes straight to `newStacks`, and
  `advanceStack` only returns `false` without moving `pos`), so *all* stopped stacks share that
  position — `stopped[0]` is representative, and the shortcut above `runRecovery` catches the
  braked case. Inside `advanceFully`, `start` cannot exceed `startPos` before the loop returns, so
  `start > stoppedAt` never holds.
- Measured: instrumenting `advanceStack` to log calls made from `advanceFully` over the module
  suite plus 2,405 generated differential cases gives **184,471** such calls, **12,946** of them
  with `stoppedAt` set and **89,341** with fragments available (positive control — the
  instrumentation was reached), and **zero** activations of either the brake or `useCachedResult`
  on that path.
- Fragment reuse in recovery is unreachable for a related reason: the main loop has already probed
  `fragments.nodeAt(pos)` at that position before the stack got stuck, and the cursor is
  forward-only.

Fixing them is still correct — the port had a second, differently-shaped copy of code that can
drift independently, which is exactly how the counter divergence arose — but the PR should not be
read as fixing an active `stoppedAt` bug.

## No regression against upstream

A differential harness ran the port against `@lezer/lr@1.4.10` + `@lezer/json@1.0.3` (same
`treeToString` on both sides; validated by reproducing #254's `partialParseStopsAtPosition`
numbers exactly) on 2,405 cases — 605 `stopAt`-grid cases over malformed/deeply-nested JSON, 300
incremental fragment cases, and 1,500 randomised two-generation incremental cases with `stopAt` —
asserting shape, `tree.length` and `parsedPos`. **0 diffs before the change, 0 diffs after.** A
mirror experiment in JS, substituting the port's `advanceFully` shape into upstream over 8,000
randomised JSON scenarios, also produced 0 diffs, which is why the fixture needs a grammar JSON's
short reduce chains cannot supply.

## Verification

- `:lezer-lr:jvmTest` — **199 before, 201 after**, 0 failures. The two added tests are
  `ParseTest.deepChainParsesWellFormedInput` and
  `ParseTest.recoveryFoldsChainsDeeperThanForceReduceLimit`; no existing test changed, was renamed
  or removed. Counts read from `lezer-lr/build/test-results/jvmTest/*.xml`.
- `:lezer-lr:apiCheck` green; **no `.api` file moved** (`advanceFully` is private).
- `spotlessApply` / `ktlintFormat` clean.
- `changelog.py check --base-ref origin/main` — 11 fragments valid, `CHANGELOG.md` untouched.
- No `parity:` comments exist under `lezer-lr/src/` (positive-controlled: the same grep matches
  `CLAUDE.md` and `gap-analysis/tests/*.spec.ts`).

## Filed separately, not bundled

- Monkopedia/kodemirror#321 — `useCachedResult` diverges from upstream's fragment-reuse
  block (no `cached.length` guard, no `nodeSet.types[id] == type` identity check, no
  `contextHash`/strict-context check, no descent into a zero-offset first child; plus port-only
  `main != null` and `!hasWrappers()` gates upstream does not have). The missing `cached.length`
  guard is a latent non-termination: a zero-length reused node would spin `advanceStack`'s caller.
- Monkopedia/kodemirror#323 — the port's `advanceStack` calls `tokens.getActions(stack)`
  *before* the default-reduce branch, where upstream calls it after; and it routes every split
  stack to `split` where upstream routes pos-advancing splits to `stacks`.

### DIFF
--- a/changelog.d/286.fixed.md
+++ b/changelog.d/286.fixed.md
@@ -1,0 +1,7 @@
+- Error recovery in the LR parser no longer abandons a stack after ten non-advancing reductions
+  (#286). `advanceFully` reimplemented `advanceStack` rather than calling it, and its standalone
+  copy needed a ten-step guard to terminate; a grammar whose reduce chain is longer than that
+  (twelve unit productions in the new test fixture) lost the recovered node entirely, yielding
+  `Program(⚠(Comma),⚠(Comma),Item(…))` where `@lezer/lr` 1.4.10 yields
+  `Program(Item(…),Comma,⚠,Comma,Item(…))`. `advanceFully` now delegates to `advanceStack`, as
+  upstream does, so the `stoppedAt` brake and cached-fragment reuse also reach the recovery path.
--- a/lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt
+++ b/lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt
@@ -468,35 +468,18 @@ internal class Parse(
     }
 
     /**
-     * Advance a stack through nested reductions until it progresses
-     * past its current position, or until a limit is reached. Used
-     * during error recovery after inserting a synthetic token.
+     * Advance a given stack forward as far as it will go, handing it to
+     * [pushStackDedup] once it moves past its starting position. Returns true
+     * if it moved forward, false if it got stuck. Used during error recovery.
      */
     private fun advanceFully(stack: Stack, newStacks: MutableList<Stack>): Boolean {
-        val startPos = stack.pos
-        var limit = 0
+        val pos = stack.pos
         while (true) {
-            if (stack.pos > startPos) {
+            if (!advanceStack(stack, null, null)) return false
+            if (stack.pos > pos) {
                 pushStackDedup(stack, newStacks)
                 return true
             }
-            val defaultReduce = parser.stateSlot(
-                stack.state,
-                ParseState.DEFAULT_REDUCE
-            )
-            if (defaultReduce > 0) {
-                stack.reduce(defaultReduce)
-            } else {
-                val actions = tokens.getActions(stack)
-                if (actions.isEmpty()) return false
-                stack.apply(
-                    actions[0],
-                    if (actions.size >= 2) actions[1] else 0,
-                    startPos,
-                    if (actions.size >= 3) actions[2] else startPos
-                )
-            }
-            if (++limit > Rec.FORCE_REDUCE_LIMIT) return false
         }
     }
 
--- a/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/ParseTest.kt
+++ b/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/ParseTest.kt
@@ -62,6 +62,43 @@ class ParseTest {
         assertTrue(result.contains("PropertyName"), "Should recover a PropertyName")
     }
 
+    // --- Error recovery through a reduce chain deeper than the force-reduce
+    // limit. `chainParser` needs twelve non-advancing reductions to fold a
+    // `Num` up into an `Item`, which is more than `Rec.FORCE_REDUCE_LIMIT`.
+    // Both expectations below are what @lezer/lr 1.4.10 produces for this
+    // grammar and input. ---
+
+    @Test
+    fun deepChainParsesWellFormedInput() {
+        // Guards the fixture itself: a mis-transcribed parser table would make
+        // the recovery expectation below meaningless.
+        val tree = chainParser.parse("1,2")
+        assertEquals(
+            "Program(" + CHAIN_ITEM_NUM + ",Comma," + CHAIN_ITEM_NUM + ")",
+            treeToString(tree)
+        )
+        assertEquals(3, tree.length)
+    }
+
+    @Test
+    fun recoveryFoldsChainsDeeperThanForceReduceLimit() {
+        // The leading commas make the parse stall on its first token, so the
+        // whole document is rebuilt through `runRecovery` -> `advanceFully`.
+        // Recovering the leading `Item` takes more non-advancing steps than
+        // `Rec.FORCE_REDUCE_LIMIT` allows, so an `advanceFully` that gives up
+        // after ten steps drops that stack and produces
+        // `Program(\u26A0(Comma),\u26A0(Comma),Item(...))` instead.
+        val tree = chainParser.parse(",,11")
+        val detail = "tree=${treeToString(tree)} length=${tree.length}"
+        assertEquals(
+            "Program(" + CHAIN_ITEM_ERR + ",Comma,\u26A0,Comma," +
+                CHAIN_ITEM_NUM + ")",
+            treeToString(tree),
+            "Recovered tree shape; $detail"
+        )
+        assertEquals(4, tree.length, "Recovered tree length; $detail")
+    }
+
     // --- Incremental parsing ---
 
     @Test
@@ -172,5 +209,15 @@ class ParseTest {
         assertFailsWith<IllegalArgumentException> {
             partial.stopAt(15)
         }
+    }
+
+    private companion object {
+        /** `Item` wrapping a `Num`, spelled out through the unit-production chain. */
+        const val CHAIN_ITEM_NUM =
+            "Item(L1(L2(L3(L4(L5(L6(L7(L8(L9(L10(L11(L12(Num)))))))))))))"
+
+        /** The same chain wrapping an error node. */
+        const val CHAIN_ITEM_ERR =
+            "Item(L1(L2(L3(L4(L5(L6(L7(L8(L9(L10(L11(L12(\u26A0)))))))))))))"
     }
 }
--- a/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/TestChainParser.kt
+++ b/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/TestChainParser.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2025 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.lezer.lr
+
+/**
+ * A tiny grammar whose reduce chain is deliberately longer than
+ * [Rec.FORCE_REDUCE_LIMIT], used to exercise error recovery through deep
+ * reductions. Built with lezer-generator from:
+ *
+ * ```
+ * @top Program { Item (Comma Item)* }
+ * Item { L1 }
+ * L1 { L2 }
+ * ... (unit productions L1..L11)
+ * L12 { Num }
+ * @tokens {
+ *   Num { $[0-9]+ }
+ *   Comma { "," }
+ *   space { $[ \t\n]+ }
+ * }
+ * @skip { space }
+ * ```
+ *
+ * Reducing a single `Num` all the way up to `Item` takes twelve reductions,
+ * none of which move the parse position, so recovering this grammar drives
+ * far more non-advancing steps than the JSON fixture does.
+ */
+val chainParser: LRParser = LRParser.deserialize(
+    ParserSpec(
+        version = 14,
+        states = "#fOVQPOOOOQO'#Cj'#CjOOQO'#Ci'#CiOOQO'#Ch'#ChOOQO'#Cg'#Cg" +
+            "OOQO'#Cf'#CfOOQO'#Ce'#CeOOQO'#Cd'#CdOOQO'#Cc'#CcOOQO'#Cb'#Cb" +
+            "OOQO'#Ca'#CaOOQO'#C`'#C`OOQO'#C_'#C_OOQO'#C^'#C^Q[QPOOOVQPO'#Cm" +
+            "Q[QPOOOOQO,59X,59XOOQO-E6k-E6k",
+        stateData = "a~OdOS~O_PO~O`_O~O",
+        goto = "!qbPPcimquy}!R!V!Z!_!c!gPP!kQ^ORa_T]O_T[O_TZO_TYO_TXO_TWO_" +
+            "TVO_TUO_TTO_TSO_TRO_TQO_Q`^Rb`",
+        nodeNames = "\u26A0 Program Item L1 L2 L3 L4 L5 L6 L7 L8 L9 L10 L11 " +
+            "L12 Num Comma",
+        maxTerm = 20,
+        skippedNodes = listOf(0),
+        repeatNodeCount = 1,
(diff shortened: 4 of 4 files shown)


monkopedia-coder:
CI note: `Instrumented tests (emulator)` is red with **13 failures**, all in `:view` Android UI tests (`FocusManagementTest`, `KeyboardHandlingTest`) and none in `:lezer-lr`. That is the known constant tracked by #275 (root cause #259) — every `ci-android.yml` run on every branch in the last week fails with the same 13, including the runs for #315 and #314 that merged to `main`. It is not one of the three required checks, and this PR does not change the count.

monkopedia-reviewer (approved):
Approving. I re-verified the port-fidelity claim independently rather than taking the
comparison on trust, and built an oracle out of the fixture's own table.

**Fidelity (checked against `npm pack @lezer/lr@1.4.10`, `dist/index.js`).** The new
`advanceFully` body is upstream's `advanceFully` (`dist/index.js:1462-1473`) line for line. I
walked the return-value contract of both `advanceStack`s branch by branch: upstream returns
truthy on brake (`stack.forceReduce() ? stack : null`), fragment reuse, default-reduce and
first-action-applied, and `false` only when the action list is exhausted; the port returns
`stack.forceReduce()`, `true`, `true`, `true` and `false` on exactly those paths. Since
`advanceFully` only tests truthiness, the contracts are equivalent and the `limit` counter can
go without introducing a non-terminating loop.

**Independent oracle.** I extracted the serialized table straight out of `TestChainParser.kt`
and fed it to real upstream `@lezer/lr@1.4.10` (`LRParser.deserialize`, same `treeToString`
semantics as `TestJsonParser.kt`). Upstream on this grammar:

```
"1,2"  => Program(Item(L1(...L12(Num)...)),Comma,Item(L1(...L12(Num)...))) length=3
",,11" => Program(Item(L1(...L12(⚠)...)),Comma,⚠,Comma,Item(L1(...L12(Num)...))) length=4
```

Both expectations in the PR are exactly that. I then ran the port at this head over seven inputs
(`1,2`, `,,11`, `1`, `,`, ``, `1,,2`, `abc`) and diffed against upstream on the same seven:
**7/7 identical**, recovery cases included.

**Red-first (my run, tests alone on unmodified `origin/main`):**

```
ParseTest[jvm] > recoveryFoldsChainsDeeperThanForceReduceLimit[jvm] FAILED
    org.junit.ComparisonFailure: Recovered tree shape;
    tree=Program(⚠(Comma),⚠(Comma),Item(L1(L2(L3(L4(L5(L6(L7(L8(L9(L10(L11(L12(Num)))))))))))))) length=4
201 tests completed, 1 failed
```

`deepChainParsesWellFormedInput` passed in that run, so the fixture is faithful on main and only
the recovery assertion discriminates.

**Mutation (production change reverted at this head, tests in place):**

```
ParseTest[jvm] > recoveryFoldsChainsDeeperThanForceReduceLimit[jvm] FAILED
201 tests completed, 1 failed
```

**Deliberately breaking the harness.** `TestChainParser.kt` is new test infrastructure, so I
tried to make it report success while proving nothing:

- Corrupted one entry in the fixture's `nodeNames` (`L12` -> `LX`): **both** new tests went red
  (`201 tests completed, 2 failed`). The table really drives the parse.
- Swapped the recovery test's input from `",,11"` to `"1,2"`: red. The assertion is
  input-sensitive, not a shape that any parse would produce.
- The grammar rejects garbage rather than accepting everything (`"abc"` -> `Program(⚠)` length 3,
  `","` -> `Program(⚠(Comma))`), on both the port and upstream.

**Behaviour beyond the new test.** `:lezer-lr:jvmTest` **199 before -> 201 after**, 0 failures;
the delta is accounted for by name, not by count — a sorted testcase-name diff of the two
`build/test-results/jvmTest/*.xml` sets shows exactly two additions,
`ParseTest.deepChainParsesWellFormedInput` and
`ParseTest.recoveryFoldsChainsDeeperThanForceReduceLimit`, with nothing renamed or removed.
Dependent suites, before -> after, all 0 failures and all counts unchanged: lang-cpp 30,
lang-css 16, lang-go 46, lang-html 32, lang-java 33, lang-javascript 39, lang-json 31,
lang-markdown 63, lang-php 29, lang-python 41, lang-rust 32, lang-sass 25, lang-xml 13,
lang-yaml 20, language 91, legacy-modes 127, lezer-common 69 (737 total, identical either side).

Because this is a hot path I also swept the JSON fixture directly: 40 malformed documents plus a
stopAt grid over the same documents, 176 cases, comparing tree shape, `tree.length` and
`parsedPos` between `origin/main` and this head — **byte-identical**. That corroborates the PR's
honest framing that JSON's short reduce chains cannot expose this divergence.

**Other checks.** `:lezer-lr:apiCheck` green and no `.api`/`.klib.api` file moved (`advanceFully`
is private) — confirmed against the tracked api file list, not just an empty grep.
`spotlessCheck` + `ktlintCheck` green. `changelog.py check --base-ref origin/main`: 11 fragments
valid, `CHANGELOG.md` untouched. `closingIssuesReferences` is `[286]` only.
`lezer-common/.../SyntaxTree.kt` is not in the diff (#252/#253 unaffected). No provenance or
attribution annotations added. `Rec.FORCE_REDUCE_LIMIT` still has its real consumer in
`runRecovery`'s force-reduce loop, so the constant correctly stays.

`Instrumented tests (emulator)` is red with the known 13 failures, all under
`view.input.FocusManagementTest` / `view.input.KeyboardHandlingTest` (#275/#259) — this PR
touches no `view/` code, and the set is unchanged. `check`, `check-macos` and `check-ios` are all
green on f3912901.

**One thing I noticed, not blocking and not this PR's to fix.** Delegating to `advanceStack` also
routes recovery through `useCachedResult`, whose missing `cached.length` guard (#321) is the one
place where dropping the `limit` counter removes a backstop: a zero-length reused node would now
spin `advanceFully` rather than bail after ten steps. The same hazard already exists on the main
advance path, so this is not a new failure class, but it makes #321 worth prioritising. Separately,
`Rec.CUT_DEPTH`/`Rec.CUT_TO` are defined and value-asserted in `ConstantsTest` but referenced
nowhere in production — upstream's stack-depth cut in `advanceStack` (`dist/index.js:1437`) is
absent from the port. Pre-existing and out of scope here; I'll file it.

monkopedia-reviewer:
Post-merge follow-up: I went back and put a measurement behind the PR's "gaps (1) and (2) are
latent" claim, because that claim is what separates this change being a bug fix from it being
pure port fidelity, and an unreachability argument is exactly the kind of thing that is easy to
believe and hard to check.

Rather than counting activations, I made them fatal. On a scratch copy of the merged tree I added
a `probeInAdvanceFully` flag set (in a `try`/`finally`) only around `advanceFully`'s call into
`advanceStack`, and made `advanceStack` **throw** if the `stoppedAt` brake branch or the
`useCachedResult` branch was taken while that flag was set. A silent run then means zero
activations, with no reporting or aggregation to get wrong.

The first attempt showed why a bare zero is not evidence. Under a control mode that throws
whenever `advanceFully` reaches `advanceStack` with `stoppedAt` merely *set*, the entire
`:lezer-lr` suite stayed **green** — the suite alone never exercises the site at all, so it could
never have falsified anything. I built a corpus that does: 37 malformed documents (JSON and the
new chain grammar) crossed with a full `stopAt` grid, the same documents reparsed incrementally
across every single-character edit position, and the two combined — 313 braked parses, 313
incremental reparses, 2,767 fragment-plus-brake parses.

On that corpus both controls fire:

```
PROBE_MODE=ctl_stopped -> recoveryUnderBrake FAILED, recoveryWithFragmentsAndBrake FAILED
    IllegalStateException: PROBE-CTL: advanceFully reached advanceStack with stoppedAt set
PROBE_MODE=ctl_frag    -> recoveryWithFragments FAILED, recoveryWithFragmentsAndBrake FAILED
    IllegalStateException: PROBE-CTL: advanceFully reached advanceStack with fragments
```

and the real probe stays silent:

```
PROBE_MODE=probe -> BUILD SUCCESSFUL
    PROBE-CORPUS braked-parses=313
    PROBE-CORPUS incremental-reparses=313
    PROBE-CORPUS fragment+brake-parses=2767
```

So `advanceFully` is genuinely reached with `stoppedAt` set and with fragments available, and in
neither case does the brake or the fragment-reuse branch activate. That is an independent
confirmation of the PR's honest framing: item (4) of #286, the ten-step counter, is the
observable fix and the one the new test covers; items (1) and (2) are fidelity, correctly
justified as fidelity rather than sold as an active bug.

I also spot-checked the validation the differential harness rests on, since a harness reporting
0 diffs is the shape of an instrument that might not be comparing anything. Feeding
`TestJsonParser.kt`'s own serialized table to real `@lezer/lr@1.4.10` and braking at 10 on
#254's input reproduces that test's asserted numbers exactly:

```
advance() calls = 9   tree.length = 12   parsedPos = 12
shape = JsonText(Object(Property(PropertyName,Number),Property(PropertyName,⚠),⚠))
```

The instrumented copy was scratch-only; the merged tree is untouched and re-runs 201/201 green.

PR #328, opened 2026-09-07
Honour the overlays flag in resolveNode (#252)

### LINKED ISSUE #252: resolveNode ignores its overlays parameter: resolve() descends into overlay-mounted trees, resolveInner() never climbs out of them, and SyntaxNode.resolveInner resolves in subtree-local coordinates

monkopedia-coder:
Fixes #252

## What diverged

Compared against **`@lezer/common` 1.5.2** (npm dist, `dist/index.js`): `resolveNode` at
`:670-694`, `Tree.resolve` at `:464`, `Tree.resolveInner` at `:477`, and the shared
`BaseNode.resolve`/`resolveInner` at `:704-709`.

`resolveNode` in the port took an `overlays: Boolean` and never read it. Concretely:

| upstream | port (before) |
|---|---|
| `dist/index.js:672-680` — a "move up to a node that actually holds the position" loop, over the **node** passed in | absent; `resolveNode` took a `Tree` and started from `tree.topNode` |
| `dist/index.js:681` — `let mode = overlays ? 0 : IterMode.IgnoreOverlays` | absent; `node.enter(pos, side)` with `mode` defaulting to `0`, i.e. `overlays == true` behaviour, unconditionally |
| `dist/index.js:683-687` — the "go up out of overlays when those do not overlap with pos" pass | absent |
| `:705` `resolve` → `resolveNode(this, …, false)`, `:708` `resolveInner` → `resolveNode(this, …, true)` | `TreeNode.resolveInner` → `resolveNode(this._tree, …)` and `BufferNode.resolveInner` → `resolveNode(context.parent._tree, …)` — the node's **own subtree**, whose `topNode` is at offset 0; `resolve` was an inline `enter` loop with no mode and no climb |
| `:477` `Tree.resolveInner` passes `true` | passed `false` (moot, since the flag was ignored) |

Four observable consequences, each fixed and each guarded by its own test:

1. `Tree.resolve` / `SyntaxNode.resolve` descended into overlay-mounted trees and returned
   inner-language nodes where the host language was asked for.
2. `resolveInner` never climbed back out of an overlay that does not cover `pos`.
3. `SyntaxNode.resolve`/`resolveInner` never climbed up out of the receiver, so a `pos` outside
   the node returned the node itself.
4. `SyntaxNode.resolveInner` treated `pos` as a **subtree-local** offset, returning a node in the
   wrong coordinate space and detached from the real parent chain.

Upstream's per-tree `CachedNode`/`CachedInnerNode` memoisation (`:465`, `:477`) is still not
ported. It is a pure optimisation — `resolveNode` climbs up from whatever node it is handed, so a
cached start node and `topNode` give the same answer — and is out of scope here.

## Verification

New `lezer-common/src/commonTest/.../ResolveOverlayTest.kt` builds a mixed-language fixture: a
host `Script(4-24)` inside `Doc(0-30)` with a mount overlaid on **two non-contiguous** doc ranges
`10-14` and `18-22`, leaving `14-18` as a gap the mount does not cover. **The mount starts at doc
10, not 0**, so a subtree-local answer is distinguishable from a document-coordinate one — that is
what makes assertion 4 below meaningful. Every assertion names a concrete node, `name(from-to)` in
document coordinates.

7 of the 9 tests fail on unmodified `origin/main`:

```
tests=9 failures=7
FAIL innerNodeResolveInnerUsesDocumentCoordinates      expected:<Ident(18-22)>  but was:<Ident(0-4)>
FAIL resolveInnerClimbsOutOfOverlayForPositionAfterMount expected:<Attr(22-24)> but was:<Ident(0-4)>
FAIL nodeResolveClimbsToContainingNode                 expected:<Attr(14-18)>  but was:<Text(0-4)>
FAIL nodeResolveDoesNotEnterOverlay                    expected:<Attr(18-22)>  but was:<Ident(18-22)>
FAIL resolveInnerClimbsOutOfOverlayForGapPosition      expected:<Attr(14-18)>  but was:<Ident(0-4)>
FAIL nodeResolveInnerUsesDocumentCoordinates           expected:<Ident(18-22)> but was:<Script(0-20)>
FAIL treeResolveDoesNotEnterOverlay                    expected:<Attr(10-14)>  but was:<Ident(10-14)>
PASS overlayParentChainReachesHostTree
PASS treeResolveInnerEntersOverlay
```

`Script(0-20)` in the sixth line is the subtree-local bug in the clear: the node is at doc `4-24`.

### Mutation testing — each behaviour reverted separately, tests in place

| reverted | tests that go red |
|---|---|
| `mode` back to a constant `0` (ignore the `overlays` flag) | `treeResolveDoesNotEnterOverlay`, `nodeResolveDoesNotEnterOverlay` |
| the climb-out-of-overlays pass deleted | `resolveInnerClimbsOutOfOverlayForGapPosition` (`expected:<Attr(14-18)> but was:<Expr(10-22)>`) |
| `TreeNode.resolveInner` back to `resolveNode(this._tree.topNode, …)` | `nodeResolveInnerUsesDocumentCoordinates`, `innerNodeResolveInnerUsesDocumentCoordinates`, `resolveInnerClimbsOutOfOverlayForGapPosition`, `resolveInnerClimbsOutOfOverlayForPositionAfterMount` |
| the "move up to a node that holds pos" loop deleted | `nodeResolveClimbsToContainingNode`, `innerNodeResolveInnerUsesDocumentCoordinates` |

No behaviour is unguarded.

### Regression surface

`:lezer-common`, `:lezer-lr`, `:lezer-highlight`, `:language` and all 22 `lang-*` modules,
`jvmTest`, before and after (counts read from `build/test-results/jvmTest/*.xml`):

* **before:** 839 tests, 0 failures
* **after:** 848 tests, 0 failures

Accounted **by name**: the diff of the two per-test name lists is exactly the 9 new
`ResolveOverlayTest` cases, all passing. No existing test changed status. (Eight `lang-*` modules —
angular, grammar, jinja, less, liquid, sql, vue, wast — contribute 0 because they contain no test
sources at all.)

`:lezer-common:wasmJsBrowserTest` also run (`CHROME_BIN=/usr/bin/chromium`): 78 tests, 0 failures,
XML confirms all 9 new cases executed on wasmJs.

### API

`:lezer-common:apiCheck` **green, and neither `.api` nor `.klib.api` moved** — `resolveNode` is
private and no signature changed. **The observable behaviour of published API does change**,
however: `Tree.resolve`, `Tree.resolveInner`, `SyntaxNode.resolve` and `SyntaxNode.resolveInner`
all return different nodes than before for the cases above. Any caller that had adapted to the old
`SyntaxNode.resolveInner` subtree-local coordinates would need to stop doing so. This is the
documented upstream contract being restored, not a new one.

`:lezer-common:ktlintCheck` and `:lezer-common:spotlessCheck` green. Changelog fragment
`changelog.d/252.fixed.md`; `changelog.py check --base-ref origin/main` reports 13 valid fragments.

## Noted, not fixed here

* **#253** (`enterUnfinishedNodesBefore`) is in the same file and untouched; this PR does not
  address it.
* `stackIterator` (`SyntaxTree.kt`, backing `Tree.resolveStack`) is not the upstream algorithm at
  all — upstream (`dist/index.js:950-968`) walks the parent chain from `resolveInner`'s result and
  pushes a *layer* per relevant overlay, whereas the port just walks a `cursorAt` parent chain and
  so never produces the overlay layers `resolveStack` exists to expose. Separate defect, separate
  PR; filing it rather than bundling it.

### DIFF
--- a/changelog.d/252.fixed.md
+++ b/changelog.d/252.fixed.md
@@ -1,0 +1,16 @@
+- Fixed syntax-tree resolution around overlay-mounted trees, which every mixed-language grammar
+  relies on — HTML with embedded `<script>`/`<style>`, Vue, Angular, Jinja, Liquid, and markdown
+  with fenced code (#252). `resolveNode` took an `overlays` flag and never read it, so all four
+  entry points behaved as though it were always true and none of them behaved as `@lezer/common`
+  documents. `Tree.resolve` and `SyntaxNode.resolve` descended into overlays and returned
+  inner-language nodes where the host language was asked for; `resolveInner` never climbed back
+  out of an overlay that does not cover the requested position, returning an inner node for a
+  position the mount does not cover; `SyntaxNode.resolve`/`resolveInner` never climbed up out of
+  the receiver, so a position outside the node returned the node itself; and
+  `SyntaxNode.resolveInner` resolved against the node's own subtree, treating its argument as a
+  subtree-local offset rather than an absolute document position and returning a node detached
+  from the real parent chain. Every one of these was a silently wrong answer rather than a crash,
+  and `resolve`/`resolveInner` is the primary tree query behind indentation, folding, bracket
+  matching and completion context. `resolveNode` now starts from the passed node, selects
+  `IterMode.IGNORE_OVERLAYS` for a non-overlay descent, and runs the climb-out-of-overlays pass,
+  matching `@lezer/common` 1.5.2.
--- a/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
+++ b/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
@@ -270,12 +270,21 @@ interface SyntaxNode : SyntaxNodeRef {
     /** Navigate to a child before a position. */
     fun childBefore(pos: Int): SyntaxNode?
 
-    /** Resolve to the innermost node at a position. */
+    /**
+     * Resolve to the innermost node at [pos], moving up out of this node
+     * first when it does not cover [pos]. Does not enter
+     * [overlays][MountedTree.overlay].
+     *
+     * [pos] is an absolute document position, not an offset into this node.
+     */
     fun resolve(pos: Int, side: Int = 0): SyntaxNode
 
     /**
-     * Like [resolve], but enters overlay-mounted trees to find
-     * the innermost node.
+     * Like [resolve], but enters [overlaid][MountedTree.overlay] nodes to
+     * find the innermost node, and climbs back out of overlays that do not
+     * cover [pos].
+     *
+     * [pos] is an absolute document position, not an offset into this node.
      */
     fun resolveInner(pos: Int, side: Int = 0): SyntaxNode
 
@@ -380,13 +389,21 @@ class Tree(
         return cursor
     }
 
-    /** Resolve to the innermost node at position [pos]. */
-    fun resolve(pos: Int, side: Int = 0): SyntaxNode = topNode.resolve(pos, side)
+    /**
+     * Resolve to the innermost node at position [pos].
+     *
+     * This does not enter [overlays][MountedTree.overlay]; use
+     * [resolveInner] when you want the innermost overlaid node.
+     */
+    fun resolve(pos: Int, side: Int = 0): SyntaxNode = resolveNode(topNode, pos, side, false)
 
     /**
-     * Like [resolve], but enters overlay-mounted trees.
+     * Like [resolve], but enters [overlaid][MountedTree.overlay] nodes,
+     * producing a syntax node pointing into the innermost overlaid tree at
+     * [pos], with parent links going through all parent structure, including
+     * the host trees.
      */
-    fun resolveInner(pos: Int, side: Int = 0): SyntaxNode = resolveNode(this, pos, side, false)
+    fun resolveInner(pos: Int, side: Int = 0): SyntaxNode = resolveNode(topNode, pos, side, true)
 
     /**
      * Returns a [NodeIterator] pointing at the innermost node at [pos],
@@ -657,16 +674,9 @@ internal class TreeNode(
         Side.BEFORE
     )
 
-    override fun resolve(pos: Int, side: Int): SyntaxNode {
-        var node: SyntaxNode = this
-        while (true) {
-            val inner = node.enter(pos, side) ?: return node
-            node = inner
-        }
-    }
+    override fun resolve(pos: Int, side: Int): SyntaxNode = resolveNode(this, pos, side, false)
 
-    override fun resolveInner(pos: Int, side: Int): SyntaxNode =
-        resolveNode(this._tree, pos, side, false)
+    override fun resolveInner(pos: Int, side: Int): SyntaxNode = resolveNode(this, pos, side, true)
 
     override fun enterUnfinishedNodesBefore(pos: Int): SyntaxNode =
         enterUnfinishedNodesBefore(this, pos)
@@ -828,16 +838,9 @@ internal class BufferNode(val context: BufferContext, val _parent: BufferNode?, 
         }
     }
 
-    override fun resolve(pos: Int, side: Int): SyntaxNode {
-        var node: SyntaxNode = this
-        while (true) {
-            val inner = node.enter(pos, side) ?: return node
-            node = inner
-        }
-    }
+    override fun resolve(pos: Int, side: Int): SyntaxNode = resolveNode(this, pos, side, false)
 
-    override fun resolveInner(pos: Int, side: Int): SyntaxNode =
-        resolveNode(context.parent._tree, pos, side, false)
+    override fun resolveInner(pos: Int, side: Int): SyntaxNode = resolveNode(this, pos, side, true)
 
     override fun enterUnfinishedNodesBefore(pos: Int): SyntaxNode =
         enterUnfinishedNodesBefore(this, pos)
@@ -1237,14 +1240,42 @@ private fun collectVisibleChildren(tree: Tree, basePos: Int, result: MutableList
 
 // ---- resolveNode helper (handles overlay-mounted trees) ----
 
-private fun resolveNode(tree: Tree, pos: Int, side: Int, overlays: Boolean): SyntaxNode {
-    var node: SyntaxNode = tree.topNode
+private fun resolveNode(startNode: SyntaxNode, pos: Int, side: Int, overlays: Boolean): SyntaxNode {
+    var node: SyntaxNode = startNode
+    // Move up to a node that actually holds the position, if possible.
+    while (node.from == node.to ||
+        (if (side < 1) node.from >= pos else node.from > pos) ||
+        (if (side > -1) node.to <= pos else node.to < pos)
+    ) {
+        val parent = if (!overlays && node is TreeNode && node.index < 0) {
+            // Without overlays, an overlay root is the end of the line.
+            null
+        } else {
+            node.parent
+        }
+        if (parent == null) return node
+        node = parent
+    }
+    val mode = if (overlays) 0 else IterMode.IGNORE_OVERLAYS
+    // Must go up out of overlays when those do not overlap with pos.
+    if (overlays) {
+        var scan: SyntaxNode = node
+        var parent = scan.parent
+        while (parent != null) {
+            if (scan is TreeNode &&
+                scan.index < 0 &&
+                parent.enter(pos, side, mode)?.from != scan.from
+            ) {
+                node = parent
+            }
+            scan = parent
+            parent = scan.parent
+        }
+    }
     while (true) {
-        val inner = node.enter(pos, side) ?: break
-        if (inner === node) break
+        val inner = node.enter(pos, side, mode) ?: return node
         node = inner
     }
-    return node
 }
 
 // ---- stackIterator helper (resolveStack) ----
--- a/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/ResolveOverlayTest.kt
+++ b/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/ResolveOverlayTest.kt
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2025 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.lezer.common
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * Resolution behaviour around overlay-mounted trees, matching `resolveNode` in
+ * `@lezer/common` `test-mix.ts` / `test-tree.ts` semantics.
+ *
+ * The fixture models a mixed-language document (a host language with an inner
+ * language mounted over two non-contiguous ranges), which is the exact shape
+ * the `overlays` flag exists for. The mount deliberately does **not** start at
+ * offset 0, so a result reported in subtree-local coordinates is
+ * distinguishable from one in document coordinates.
+ *
+ * ```
+ * Doc(0-30)
+ *   Text(0-4)
+ *   Script(4-24)                 mounted overlay over doc 10-14 and 18-22
+ *     Attr(4-10)
+ *     Attr(10-14)                covered by overlay range 1
+ *     Attr(14-18)                the GAP between the two overlay ranges
+ *     Attr(18-22)                covered by overlay range 2
+ *     Attr(22-24)
+ *   Text(24-30)
+ *
+ * mounted inner tree, rooted at doc 10:
+ *   Expr(10-22)
+ *     Ident(10-14)
+ *     Ident(18-22)
+ * ```
+ */
+class ResolveOverlayTest {
+
+    private val doc = NodeType.define(NodeTypeSpec(name = "Doc", id = 0))
+    private val text = NodeType.define(NodeTypeSpec(name = "Text", id = 1))
+    private val script = NodeType.define(NodeTypeSpec(name = "Script", id = 2))
+    private val attr = NodeType.define(NodeTypeSpec(name = "Attr", id = 3))
+    private val expr = NodeType.define(NodeTypeSpec(name = "Expr", id = 4))
+    private val ident = NodeType.define(NodeTypeSpec(name = "Ident", id = 5))
+
+    private val dummyParser = object : Parser() {
+        override fun createParse(
+            input: Input,
+            fragments: List<TreeFragment>,
+            ranges: List<TextRange>
+        ): PartialParse = error("Not implemented")
+    }
+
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
## Approved — independently re-verified at `eceb5f4b`

Reviewed in a fresh worktree off the PR head. Everything below is my own measurement, not a
restatement of the description.

### 1. Upstream control flow, not just outputs

Fetched `@lezer/common@1.5.2` from npm (`npm pack`) and read `dist/index.js` directly. The port is a
line-for-line match of upstream's `resolveNode` (`:670-694`), not an output-equivalent rewrite:

| upstream `dist/index.js` | port `SyntaxTree.kt` |
|---|---|
| `:672-680` climb-up loop, incl. `!overlays && node instanceof TreeNode && node.index < 0 ? null : node.parent` | `:1245-1258`, same predicate, same early `return node` |
| `:681` `let mode = overlays ? 0 : IterMode.IgnoreOverlays` | `:1259` `val mode = if (overlays) 0 else IterMode.IGNORE_OVERLAYS` |
| `:683-687` climb-out-of-overlays scan | `:1260-1274`, same `scan.index < 0 && parent.enter(pos, side, mode)?.from != scan.from` test |
| `:688-692` descent `node.enter(pos, side, mode)` | `:1275-1278` — and the port's old ad-hoc `if (inner === node) break` guard is correctly gone |
| `:464`/`:477` `Tree.resolve`/`resolveInner` → `topNode`, `false`/`true` | `:394`/`:406`, same (`resolveInner` also fixed from `false` to `true`) |
| `:704-709` `BaseNode.resolve`/`resolveInner` → `resolveNode(this, …, false/true)` | `TreeNode` `:677`/`:679` and `BufferNode` `:841`/`:843`, same |

`IterMode` constants match upstream's enum values (1/2/4/8/16), and the overlay-root sentinel
`index = -1` set by `TreeNode.enter` matches upstream's `new TreeNode(mounted.tree, …, -1, this)`.

### 2. The fixture discriminates

Confirmed by reading it, not by trusting the diagram. `MountedTree` overlay ranges are
`TextRange(6, 10)` and `TextRange(14, 18)` **relative to `Script`, which starts at doc 4** — doc
`10-14` and `18-22`, with doc `14-18` a genuine gap the mount does not cover. The mount root lands at
`overlay[0].from + treeFrom` = `6 + 4` = **doc 10, not 0**, which is exactly what makes
`nodeResolveInnerUsesDocumentCoordinates` non-vacuous: `Script(4-24).resolveInner(20)` answered
`Script(0-20)` before (subtree-local) and `Ident(18-22)` after. Every assertion names a concrete
`name(from-to)` in document coordinates; none is a null check.

### 3. Red first — my own run

Copied `ResolveOverlayTest.kt` alone onto unmodified `origin/main` (working tree clean apart from
that one untracked file) and ran `:lezer-common:jvmTest`. From
`build/test-results/jvmTest/TEST-…ResolveOverlayTest.xml`:

```
tests=9 failures=7 errors=0
FAIL innerNodeResolveInnerUsesDocumentCoordinates       expected:<Ident(18-22)>
FAIL resolveInnerClimbsOutOfOverlayForPositionAfterMount expected:<Attr(22-24)>
FAIL nodeResolveClimbsToContainingNode                  expected:<Attr(14-18)>
FAIL nodeResolveDoesNotEnterOverlay                     expected:<Attr(18-22)> but was:<Ident(18-22)>
FAIL resolveInnerClimbsOutOfOverlayForGapPosition       expected:<Attr(14-18)>
FAIL nodeResolveInnerUsesDocumentCoordinates            expected:<Ident(18-22)>
FAIL treeResolveDoesNotEnterOverlay                     expected:<Attr(10-14)> but was:<Ident(10-14)>
PASS overlayParentChainReachesHostTree
PASS treeResolveInnerEntersOverlay
```

Same seven, by name.

**The two that pass both before and after are not vacuous.** I proved that with an extra mutation
(M5 below) rather than by inspection: inverting the mode selection to
`if (overlays) IGNORE_OVERLAYS else 0` turns **all nine** red, including
`treeResolveInnerEntersOverlay` and `overlayParentChainReachesHostTree`. They are the guard against
over-correcting the fix, and they hold real assertions.

### 4. Per-behaviour mutation — four independent reverts, plus two of my own

Each applied alone to the PR head with the tests in place, `:lezer-common:jvmTest`, 78 tests each run:

| mutation | red |
|---|---|
| **M1** `mode` back to constant `0` | `treeResolveDoesNotEnterOverlay`, `nodeResolveDoesNotEnterOverlay` |
| **M2** climb-out-of-overlays pass deleted | `resolveInnerClimbsOutOfOverlayForGapPosition` |
| **M3** `TreeNode.resolveInner` back to `resolveNode(this._tree.topNode, …)` | `nodeResolveInnerUsesDocumentCoordinates`, `innerNodeResolveInnerUsesDocumentCoordinates`, `resolveInnerClimbsOutOfOverlayForGapPosition`, `resolveInnerClimbsOutOfOverlayForPositionAfterMount` |
| **M4** climb-up loop deleted | `nodeResolveClimbsToContainingNode`, `innerNodeResolveInnerUsesDocumentCoordinates` |
| **M5** *(mine)* mode selection inverted | all 9 |
| **M6** *(mine)* overlay-root stop dropped from the climb-up loop (`val parent = node.parent`) | **none — green** |
| unmutated control | none |

M1–M4 reproduce the description exactly. All four behaviours are guarded.

**M6 is a finding, non-blocking.** Upstream's
`!overlays && node instanceof TreeNode && node.index < 0 ? null : node.parent` is ported faithfully
but nothing exercises it. I checked it is observable rather than dead: taking the overlay root
(`mixedTree().resolveInner(12, 1).parent` = `Expr(10-22)`) and calling `.resolve(26, 1)` on it —
pos 26 lies in `Text(24-30)`, outside the overlay — the PR gives `Expr(10-22)` (upstream's answer,
stop at the overlay root) while M6 gives `Text(24-30)`. So the code is right and untested, not wrong.
A tenth case covering that would close the gap; not a merge blocker since the branch matches upstream.

### 5. Regression surface — the one that mattered

`.api` and `.klib.api` do not move, so `apiCheck` proves nothing here; I measured behaviour instead.
`jvmTest` across **29 modules** — `:lezer-common`, `:lezer-lr`, `:lezer-highlight`, `:language`, all
22 `lang-*`, plus `:commands`, `:autocomplete` and `:vim`, which I added after grepping for
`.resolve(`/`.resolveInner(` callers (`Indent.kt`, `Fold.kt`, `MatchBrackets.kt`, `Language.kt`,
`CompletionContext.kt`, `HtmlComplete.kt`, `JsComplete.kt`, `SqlComplete.kt`, `Markdown.kt`,
`Yaml.kt`, `VimEditor.kt`, …). Counts read from `build/test-results/jvmTest/*.xml`, per-test names
diffed:

| tree | tests | failures |
|---|---|---|
| merge base `6484d357` | **1631** | 0 |
| PR head `eceb5f4b` | **1640** | 0 |
| PR head merged with current `main` `e7a74f9f` | **1642** | 0 |

Delta accounted for **by name**:

* merge base → PR head: `+9`, exactly the nine `ResolveOverlayTest` cases, all PASS. **No existing
  test changed status anywhere** — not in `:language` indentation/folding/bracket-matching, not in
  `:autocomplete` context, not in `:lang-html`/`:lang-markdown` (the mixed-language modules most
  exposed to an overlay-resolution change), not in `:vim`'s 626.
* merged with `main`: the additional `+2` are `ParseTest.deepChainParsesWellFormedInput` and
  `ParseTest.recoveryFoldsChainsDeeperThanForceReduceLimit`, which belong to **#324**, merged after
  this branch was cut. Measuring `origin/main` → head would have mis-attributed those two to this PR;
  measuring from the merge base does not.

Task execution verified from the XML, not from exit codes. Eight `lang-*` modules (angular, grammar,
jinja, less, liquid, sql, vue, wast) contribute 0 because they have no test sources — noted so their
zero is not read as a silent no-op.

`:lezer-common:ktlintCheck`, `:spotlessCheck` and `:apiCheck` green locally (with `ANDROID_HOME` set).

### 6. The unported memoisation — the argument holds

Checked it rather than accepting it. Upstream's `CachedNode`/`CachedInnerNode` (`:464`, `:477`) seed
`resolveNode` with the previous result instead of `topNode`. Because `resolveNode` climbs up from
whatever node it is given until it reaches a node holding `pos`, and every ancestor of a cached node
sits on the `topNode`-to-cached-node `enter` chain, upstream's warm path is designed to agree with
its cold path. The port always takes upstream's **cold-cache path** (`topNode`), so it cannot
disagree with upstream — the omission can only cost time, never change an answer. The two
`SyntaxNode` entry points are uncached upstream as well, so those are at exact parity already.

It was recorded only in the PR description, so I filed **#330** to make it findable, alongside the
analogous `nodeSize` gap in #280.

### 7. Scope, changelog, issue linkage

* Three files (`gh pr view 328 --json files`, cross-checked with a three-dot
  `6484d357...eceb5f4b` diff): the source, the new test, `changelog.d/252.fixed.md`. `CHANGELOG.md`
  untouched; `changelog.py check --base-ref origin/main` reports 13 valid fragments including `#252`.
* `closingIssuesReferences` is `[252]` and nothing else (positive-controlled — the query does return
  252, so an empty result would have been a query failure rather than a clean bill).
* **#253 is untouched and stays OPEN.** No changed line in the diff mentions
  `enterUnfinishedNodesBefore`; the two occurrences are context in the `TreeNode`/`BufferNode` hunks.
  The stub still has the `while` loop that `break`s on every path. Correctly noted, not fixed.
* The author's `stackIterator`/`resolveStack` observation was filed separately as **#329** rather
  than bundled here. Right call.
* No provenance or attribution annotations anywhere in the diff.

### 8. CI at the live head

`check`, `check-macos` and `check-ios` all **pass** at `eceb5f4b` per the GitHub API — the required
set, including the macOS/iOS native suites a local Gradle run cannot cover. `lint` passes.
`Instrumented tests (emulator)` fails with exactly the known 13, by name — 5 `FocusManagementTest`
(`clickAfterDrag_focusIsRestored`, `clickOnEditor_givesFocusToTextField`,
and 8 `KeyboardHandlingTest` (`arrowDown`/`arrowLeft`/`arrowRight`/`arrowUp`, `end_movesToLineEnd`,
`home_movesToLineStart`, `modX_cutsSelectedText_onMacFamily`/`offMacFamily`) — the #275/#259 baseline,
non-blocking by owner ruling, and in `:view`, which this PR does not touch. Not a new red.

### Separate issues filed

* **#330** — `Tree.resolve`/`resolveInner` do not port `CachedNode`/`CachedInnerNode`; performance
  only, recorded so the deferral is findable outside this PR body.
* **#331** — `TreeNode.enter` ignores `MountedTree.bracketed`, so `ENTER_BRACKETED` relaxes overlay
  boundaries for every mount rather than only bracketed ones. Pre-existing, unreachable from
  `resolveNode`, spotted while reading `enter` for this review.

Squash-merging.

PR #335, opened 2026-09-07
Hold editor focus across a pointer press (#259)

### LINKED ISSUE #259: Android: a pointer press on the editor leaves the text field unfocused, so keyboard input is dropped after a click

monkopedia-coder:
Fixes #259.

## What was wrong

On Android a **mouse click in the editor blanked its focus**, so every key pressed after a click
was dropped. Arrow keys, `Home`, `End` and the keymap's shortcuts all did nothing until the editor
was focused again by other means. Anyone using this library on an Android tablet or on ChromeOS
with a mouse or trackpad hit it on every click; the same gestures with a finger always worked.

The click itself was never wrong — it hit-tested correctly and dispatched the right selection. The
failing operation was the key press after it. Three diagnoses on the issue were reverse-engineered
from the assertion message `Expected cursor at 11 but was at 0` and all three were refuted, because
that `0` is the *click's correct answer*, not a wrong one.

## The mechanism, and it is Compose's rather than ours

`AndroidComposeView.dispatchTouchEvent` (androidx `ui-android`, since 1.10) applies
`AutoClearFocusBehavior.CursorBased` — **the platform default** — *after* it has run the gesture
handlers:

```
handleMotionEvent(event)                       // gestures run here; requestFocus() happens
...
isDown = actionMasked == ACTION_DOWN || ACTION_POINTER_DOWN
isCursor = event.isFromSource(SOURCE_MOUSE) || event.isFromSource(SOURCE_TOUCHPAD)
if (isDown && isCursor) {
    behavior = (parent as? View)?.getTag(R.id.auto_clear_focus_behavior_tag)
        ?: AutoClearFocusBehavior.Default          // Default == CursorBased
    if (behavior == CursorBased) {
        node = focusOwner.activeFocusTargetNode
        if (node != null && !node.requireLayoutCoordinates().boundsInRoot()
                .contains(Offset(event.x, event.y))) {
            focusOwner.clearFocus()                // clearFocus(force = false)
        }
    }
}
```

Read off the shipped bytecode of `ui-android`; `AutoClearFocusBehavior.Companion.getDefault` is
literally `getCursorBased`. Two facts follow directly:

- **Why mouse and not touch.** The guard is on the event *source*: `SOURCE_MOUSE` (`8194`) or
  `SOURCE_TOUCHPAD` (`1048584`). A touchscreen press is neither. That is the entire mouse/touch
  split the issue measured.
- **Why it is a teardown and not a missing request.** The clear runs *after* `handleMotionEvent`,
  i.e. after the editor's gesture has already called `focusRequester.requestFocus()` — which is
  exactly the issue's `focusBefore=true, focusAfterClick=false`.

The editor's focused node is the 1-dp hidden `BasicTextField`, so **no press anywhere in the editor
ever lands inside it** and every mouse press qualifies as "outside".

## The fix

`FocusOwnerImpl.clearFocus(force = false)`, and `FocusTransactions.clearFocus` returns `forced` for
a target in the `Captured` state — a captured target **refuses a non-forced clear**, and
`clearOwnerFocus()` runs only when the clear succeeded, so the View-level focus is untouched too.
So the editor now captures its focus for the duration of a press:

- taken on the **Final** pass, which runs after every Main-pass handler in the tree and therefore
  after the gesture above has requested focus (so the first click on an unfocused editor is covered
  too), and still inside `handleMotionEvent`, ahead of the auto-clear;
- released as soon as the last pointer lifts, so nothing else is ever kept from taking focus. The
  window is one press.

Compose does expose an opt-out — `AbstractComposeView.autoClearFocusBehavior` — but it lives on the
**host** view and would change focus behaviour for the whole window. That is not a library's to
set, so the fix is scoped to this editor and to the press.

The `view/build.gradle.kts` note that read the Robolectric `graphicsMode=NATIVE` focus divergence as
an environment artifact is corrected in the same change: it was this bug, seen early.

## Verification

**Red first.** `PointerFocusRetentionTest` was written before the fix and run against it. On an API
33 emulator (`:view:connectedDebugAndroidTest`), with the test present and the `KodeMirror.kt` hunk
absent — which is also the mutation control, since that hunk is the only functional difference
between the two trees:

```
107 executed, 14 failed, 0 skipped
```

failing set, by name (`build/outputs/androidTest-results/connected/debug/*.xml`, not the log):

```
FocusManagementTest.clickAfterDrag_focusIsRestored
FocusManagementTest.clickOnEditor_givesFocusToTextField
FocusManagementTest.clickThenArrowRight_cursorMoves
FocusManagementTest.dragThenKeyboard_cursorMoves
FocusManagementTest.multipleClicks_eachPreservesFocus
KeyboardHandlingTest.arrowDown_movesToNextLine
KeyboardHandlingTest.arrowLeft_movesCursorBackward
KeyboardHandlingTest.arrowRight_movesCursorForward
KeyboardHandlingTest.arrowUp_movesToPreviousLine
KeyboardHandlingTest.end_movesToLineEnd
KeyboardHandlingTest.home_movesToLineStart
KeyboardHandlingTest.modX_cutsSelectedText_offMacFamily
KeyboardHandlingTest.modX_cutsSelectedText_onMacFamily
PointerFocusRetentionTest.mousePress_keepsFocus_andTheKeyAfterItIsDelivered
```

The new **touch** twin passed in that same red run, which is the discriminator working: identical
test but for the pointer type.

**Green after.** Same emulator, same command:

```
107 executed, 0 failed, 0 skipped
```

All thirteen cleared **by name** — the list above, minus the new mouse test which also passes — and
nothing was substituted for them, which is the reason the comparison is by name and not by count.

Also green locally: `:view:jvmTest`, `:view:wasmJsBrowserTest` (303 tests, 0 failures — the new pair
runs there too), `:view:apiCheck`, `:view:ktlintCheck`, `spotlessCheck`.

**`.api` did not move** — no public API change; `apiCheck` passes against the committed dumps.

## Where the new test executes

`view/src/commonTest/.../input/PointerFocusRetentionTest.kt`. `view/build.gradle.kts` excludes
`com.monkopedia.kodemirror.view.input.*` from every `*UnitTest` task, so on **Android it runs
instrumented only**. Confirmed present in the results of all three suites run here:

- `view/build/test-results/jvmTest/TEST-…PointerFocusRetentionTest.xml` — `tests=2 failures=0`
- `view/build/test-results/wasmJsBrowserTest/TEST-wasmJsBrowserTest.…PointerFocusRetentionTest.xml`
  — `tests=2 failures=0`
- `view/build/outputs/androidTest-results/connected/debug/TEST-….xml` — both cases, and the mouse
  case is the one that was red before the fix

Both assertions are exact and font-metric independent: the press at `x=8` is two pixels into the
content's 6-dp start padding, so it resolves to the first character of line 1 on any font, and
`End` from there must land on **11**, the end of `"Hello world"`. Position 0 is what a dropped key
leaves behind, so an exact 11 is the only outcome that says the key was delivered.

## Not covered here

#275 stays open regardless: it is about the instrumented suite emitting a constant with no
expected-failure baseline, which remains true whatever these thirteen do.

### DIFF
--- a/changelog.d/259.fixed.md
+++ b/changelog.d/259.fixed.md
@@ -1,0 +1,14 @@
+- Fixed a mouse click in the editor blanking its focus on Android, so every key pressed after a
+  click was dropped (#259). The click itself was never wrong — it hit-tested correctly and placed
+  the caret where it should — but Compose then took the focus away, and the arrow keys, Home, End
+  and the keymap's shortcuts all did nothing until the editor was focused again by other means.
+  Anyone using the library on an Android tablet or on ChromeOS with a mouse or trackpad hit this on
+  every click; the same gestures with a finger always worked, which is why it went unrecognised for
+  so long. The cause is a Compose policy rather than a defect in the editor: `AndroidComposeView`
+  applies `AutoClearFocusBehavior.CursorBased` — the platform default — *after* the gesture
+  handlers have run, clearing focus when a mouse or touchpad press lands outside the bounds of the
+  focused node, and the editor's focused node is a 1-dp hidden input that no press ever lands
+  inside. The editor now captures its focus for the duration of a press, which is the framework's
+  own way of refusing that clear, and releases it as soon as the pointer lifts so nothing else is
+  kept from taking focus. Thirteen Android instrumented tests across `FocusManagementTest` and
+  `KeyboardHandlingTest` had been failing on this since the suite was introduced and now pass.
--- a/view/build.gradle.kts
+++ b/view/build.gradle.kts
@@ -64,14 +64,16 @@ kotlin {
 // Robolectric — the usual answer — turns out not to give a trustworthy result here. Under
 // `graphicsMode=LEGACY` all text measures to zero width, so every coordinate-based
 // assertion is meaningless (a click 30px into a line resolves to the line's last character);
-// under `graphicsMode=NATIVE` the metrics are right but a pointer press clears Compose focus,
-// which breaks the focus and keyboard suites. That last divergence was read here as an
+// under `graphicsMode=NATIVE` the metrics were right but a pointer press cleared Compose focus,
+// which broke the focus and keyboard suites. That last divergence was read here as an
 // environment artifact rather than a finding about the editor; the instrumented suite it asked
-// for has since REFUTED that — the same 5 focus + 8 keyboard failures reproduce on an API 34
-// emulator, tracked as #259. Robolectric is still not a trustworthy substitute (the LEGACY
-// metrics problem stands), but it was not wrong about this one. Android coverage for these
-// needs instrumented tests on a device/emulator; see #215. Everything else in `commonTest`
-// still runs here.
+// for REFUTED that — the same 5 focus + 8 keyboard failures reproduced on an emulator — and the
+// cause has since been traced and fixed (#259): Compose clears focus on a mouse or touchpad
+// press that lands outside the focused node, which had nothing to do with the graphics mode and
+// everything to do with the editor's focused node being a 1-dp hidden input. Robolectric is
+// still not a trustworthy substitute here, because the LEGACY metrics problem stands, but it was
+// not wrong about this one. Android coverage for these needs instrumented tests on a
+// device/emulator; see #215. Everything else in `commonTest` still runs here.
 // `testDebugUnitTest` / `testReleaseUnitTest` are the Android local-unit-test tasks; the JVM
 // target's task is `jvmTest` and is deliberately left alone.
 tasks.withType<Test>().matching { it.name.endsWith("UnitTest") }.configureEach {
--- a/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
+++ b/view/src/commonMain/kotlin/com/monkopedia/kodemirror/view/KodeMirror.kt
@@ -766,6 +766,59 @@ fun KodeMirror(session: EditorSession, modifier: Modifier = Modifier) {
                             }
                         }
                     }
+                    .pointerInput(Unit) {
+                        // Hold the editor's focus for the length of a press.
+                        //
+                        // Compose on Android takes focus away on a mouse or
+                        // touchpad press: `AndroidComposeView.dispatchTouchEvent`
+                        // applies `AutoClearFocusBehavior.CursorBased` — the
+                        // platform default — AFTER the gesture handlers above
+                        // have run, clearing focus whenever a cursor press lands
+                        // outside the bounds of the focused node. The editor's
+                        // focused node is the 1-dp hidden input, so no press in
+                        // the editor ever lands inside it and every mouse click
+                        // blanked the focus the gesture had just requested. The
+                        // click still placed the caret correctly; it was the key
+                        // after it that went nowhere (#259). A finger press is
+                        // not a cursor press and never tripped this, which is the
+                        // whole of the mouse/touch split the two suites saw.
+                        //
+                        // That clear is not forced, and a captured focus target
+                        // refuses a non-forced clear, so capturing across the
+                        // press vetoes it. Taken on the Final pass, which runs
+                        // after every Main-pass handler in the tree and therefore
+                        // after the gesture above has requested focus, and
+                        // released as soon as the last pointer lifts so nothing
+                        // else is ever kept from taking focus.
+                        var focusCaptured = false
+                        try {
+                            awaitPointerEventScope {
+                                while (true) {
+                                    val event = awaitPointerEvent(
+                                        PointerEventPass.Final
+                                    )
+                                    val pressed = event.changes.any { it.pressed }
+                                    if (pressed) {
+                                        if (!focusCaptured) {
+                                            focusCaptured =
+                                                focusRequester.captureFocus()
+                                        }
+                                    } else if (focusCaptured) {
+                                        focusRequester.freeFocus()
+                                        focusCaptured = false
+                                    }
+                                }
+                            }
+                        } finally {
+                            // Only reached when this node detaches, where the
+                            // focus target's own detach forcibly clears the
+                            // capture anyway; freeing through a requester whose
+                            // node has already gone throws, so it is guarded.
+                            if (focusCaptured) {
+                                runCatching { focusRequester.freeFocus() }
+                            }
+                        }
+                    }
                     .pointerInput(session) {
                         awaitPointerEventScope {
                             var lastHoverDocPos: Int? = -1
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/PointerFocusRetentionTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/PointerFocusRetentionTest.kt
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.view.input
+
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.input.key.Key
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performKeyInput
+import androidx.compose.ui.test.performMouseInput
+import androidx.compose.ui.test.performTouchInput
+import com.monkopedia.kodemirror.commands.standardKeymap
+import com.monkopedia.kodemirror.view.EditorSessionImpl
+import com.monkopedia.kodemirror.view.keymapOf
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+/**
+ * A press on the editor must leave the editor still focused, whatever kind of
+ * pointer made it.
+ *
+ * # What this isolates (#259)
+ *
+ * Thirteen instrumented tests failed on Android with `Expected cursor at 11 but
+ * was at 0`, which reads as a hit-test failure and is not one: the press
+ * hit-tests correctly and dispatches the right selection, and it is the *key*
+ * afterwards that does nothing, because the press has taken the editor's focus
+ * away in between. Three diagnoses were reverse-engineered from that offset
+ * before anyone measured focus.
+ *
+ * The mechanism is Compose's, not the editor's. `AndroidComposeView`
+ * `dispatchTouchEvent` applies `AutoClearFocusBehavior.CursorBased` — the
+ * platform default — *after* the gesture handlers have run: on a press from a
+ * mouse or touchpad it clears focus whenever the press lands outside the bounds
+ * of the focused node. The editor's focused node is the 1-dp hidden input, so
+ * every mouse press anywhere in the editor lands "outside" it. Touch presses are
+ * not cursor presses and never trip it, which is why the same gesture worked
+ * with a finger and not with a mouse.
+ *
+ * # Why the two tests are a pair
+ *
+ * They differ in one thing: the pointer type. Keeping both means a fix for the
+ * mouse path that broke the touch path — the more likely accident, since the
+ * fix has to hold focus across a press without stopping anything else from ever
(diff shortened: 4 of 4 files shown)


monkopedia-reviewer (approved):
Approving. I re-verified this independently — on a local API 33 emulator, in a fresh worktree at
`8fadbd21`, against the androidx version this project actually resolves. Every claim below is a run
I did, not a reading of the report.

## 1. The leaked capture — settled, path by path

This was my first concern: a capture that survives a press holds editor focus forever and is worse
than the bug. I built a probe with an independent focus target next to the editor and asserted, via
that target's `onFocusChanged`, whether **anything else can take focus** after each exit path.

```
                                            #335 (Final)   origin/main (no fix)
control_otherTargetCanTakeFocusWithNoPress      PASS            PASS
whilePressed_captureBlocksOtherFocus            PASS            FAIL   <- positive control
afterMouseClick_captureReleased                 PASS            PASS
afterPointerCancel_captureReleased              PASS            PASS
multiTouch_captureHeldUntilLastUpThenReleased   PASS            FAIL   <- positive control
leavingCompositionMidPress_captureReleased      PASS            PASS
pressWithNoUp_..._releases                      PASS            FAIL   <- positive control
```

The three that fail on `main` are the ones that assert the capture *exists*; they are what stops
this suite from being vacuous. Full run at PR head with these added: **119 executed, 0 failed**.

The release is structural, not happy-path, and there are two independent guarantees:

* **`finally`.** `resetPointerInputHandler()` cancels the handler coroutine with
  `PointerInputResetException`, so node detach / handler reset unwinds through the `finally`, and
  `freeFocus()` is non-suspending so it runs even in a cancelled coroutine. The `runCatching` guard
  is right — freeing through a requester whose node has gone throws.
* **The cancel event itself.** In `ui-android` 1.11.2,
  `SuspendingPointerInputModifierNodeImpl.onCancelPointerInput()` rebuilds every
  `PointerInputChange` with `pressed = false` (bci 259, the 4th ctor arg is `iconst_0`) and
  dispatches it through `Initial`, `Main` **and `Final`** (bci 331/339/347). So a cancelled gesture
  reaches the loop's `else if (focusCaptured) freeFocus()` branch on the same pass the fix listens
  on — the capture is released by the normal path, before the `finally` is ever needed.

Multi-touch is handled by `event.changes.any { it.pressed }` rather than a counter: a released
pointer is still present in `changes` with `pressed = false`, so the capture is held until the last
one lifts, which the probe confirms in both directions.

**No path leaks.**

## 2. Focus transfer still works

* `clickEditorAThenEditorB_focusMoves` — two editors, click A then B: B takes focus, A loses it.
  PASS on #335, FAIL on `main`.
* `clickEditorThenPlainTextField_focusMoves` — click the editor, then a plain `BasicTextField`
  outside it: the field takes focus and the editor gives it up. PASS on #335, FAIL on `main`.

The capture blocks focus transfer only for the duration of a press, which is exactly what
`whilePressed_captureBlocksOtherFocus` measures on both sides of the up.

## 3. The framework mechanism is real, in the version this project resolves

`:view:dependencies --configuration debugAndroidTestRuntimeClasspath` resolves
`androidx.compose.ui:ui-android:1.11.2`. From its `classes.jar`:

* `AutoClearFocusBehavior$Companion.getDefault-4UtRPd4()` is literally
  `invokevirtual getCursorBased-4UtRPd4` — the default *is* `CursorBased`.
* `AndroidComposeView.dispatchTouchEvent` calls `handleMotionEvent-8iAsVTc` at bci 93 and only
  *then*, at bci 128-351, tests `actionMasked == 0 || == 5`, `isFromSource(8194)` /
  `isFromSource(1048584)`, reads the behaviour off `(parent as? View).getTag(...)` falling back to
  `Companion.getDefault`, compares the focused node's `boundsInRoot().contains(Offset(x, y))`, and
  calls `FocusManager.clearFocus$default(owner, false, 1, null)` — i.e. **force = false**.
* `FocusTransactionsKt.clearFocus(node, forced, refreshEvents)`: the `Captured` branch (tableswitch
  case 2, bci 129) returns `iload_1` — `forced` — without clearing anything.

So the diagnosis is exact and the veto is the framework's own. I also confirmed the only opt-out
surface is `AbstractComposeView` (host view); no per-node modifier exists, so scoping this to the
editor and the press is the right call.

## 4. The `Final` pass IS load-bearing — but the PR does not pin it

Mutated and re-ran the full instrumented suite each time:

```
Final    (as merged)                107 executed, 0 failed   BUILD SUCCESSFUL
Main                                107 executed, 0 failed   BUILD SUCCESSFUL   <-- silent
Initial                             107 executed, 0 failed*  (*with the PR's tests only)
```

`PointerFocusRetentionTest` starts from an already-focused editor, so a capture taken before the
gesture's `requestFocus()` still succeeds. The case that discriminates is the **first mouse press on
an unfocused editor** — and with the two tests I wrote for it, both mutations die:

```
                                      Final   Main   Initial   no fix
firstMouseClickOnUnfocusedEditor_...   PASS   FAIL    FAIL      FAIL
clickEditorAThenEditorB_focusMoves     PASS   FAIL    FAIL      FAIL
```

`Main`/`Initial` fail with `editor lost focus on the first mouse click while unfocused`. The
reasoning in the comment is therefore correct and the line is right — it is the coverage that is
missing. Filed as **#344** with the test shapes, rather than held against this PR.

## 5. Red-first and mutation, my runs

`PointerFocusRetentionTest.kt` copied **alone** onto unmodified `origin/main` (the two commits
between this PR's base and `origin/main` touch only `lezer-common`/`lezer-lr`, so that same run is
also the production-hunk-reverted mutation control):

```
107 executed, 14 failed
```

and the failing set is, by name, exactly the thirteen plus the new mouse case — with the **touch
twin passing in the same run**, which is the mouse/touch discriminator behaving as the diagnosis
predicts:

```
FocusManagementTest   clickAfterDrag_focusIsRestored · clickOnEditor_givesFocusToTextField
                      clickThenArrowRight_cursorMoves · dragThenKeyboard_cursorMoves
                      multipleClicks_eachPreservesFocus
KeyboardHandlingTest  arrowUp/Down/Left/Right · home_movesToLineStart · end_movesToLineEnd
                      modX_cutsSelectedText_onMacFamily · modX_cutsSelectedText_offMacFamily
PointerFocusRetentionTest  mousePress_keepsFocus_andTheKeyAfterItIsDelivered
```

At `8fadbd21`: **107 executed, 0 failed** locally, and CI's own
`##[notice]Android instrumented: **107 executed**, 0 failed, 0 skipped` agrees. All thirteen cleared
**by name**, nothing substituted; `FocusManagementTest` 6/6 and `KeyboardHandlingTest` 10/10 green.
`FocusManagementTest.kt` and `KeyboardHandlingTest.kt` are untouched by the diff — the forbidden
shortcut of rewriting them to touch was not taken.

## 6. Regression surface and platforms

```
:view:jvmTest        336 / 0   at 8fadbd21   ==   336 / 0 on main+test  (identical set)
:view:wasmJsTest     303 / 0   with both new cases PASSing on wasmJs
:view:apiCheck       green, no `.api` file moved
instrumented         107 / 0 (was 107 / 14)
```

The new test compiles into `commonTest`, so it runs on JVM, wasmJs and native as well as the
emulator — I read that out of `build/test-results/**/*.xml`, not the console. Worth recording that
it **passes on desktop JVM even without the fix**: the auto-clear is Android-only, so on the other
targets the pair is a non-regression guard rather than a reproduction. `check-macos` and `check-ios`
green at this SHA cover the native side of the capture.

## 7. `view/build.gradle.kts`

Diffed explicitly: **comment-only**, 9 added / 7 removed, every changed line inside the `//` block
above the `tasks.withType<Test>` filter. The filter, the `isFailOnNoMatchingTests`, the
`testInstrumentationRunnerArguments["package"]` and everything else are byte-identical. The
correction it makes is the one #259 asked for, and it is now true rather than merely rewritten.

## 8. One correction to the report

The note that a press on a child which consumes the down "still will not capture, because the editor
may not be focused at that point — unchanged, not a new gap" is not what I measure for the gutter.
With `lineNumbers` installed, a mouse press at `x = 5`:

```
                                   #335    origin/main
editor already focused           hasFocus=true    hasFocus=false
editor not focused               hasFocus=true    hasFocus=false
```

The line-number gutter does not consume the down, so the editor's gesture requests focus normally
and the capture then succeeds. So gutter presses **are** changed — in the same direction as the fix,
and strictly better: they used to blank the editor's focus and now do not. Not a regression, and
`ClickTargetingTest` (8), `ScrollClickTest` (2) and `DragSelectionTest` (4) are green throughout —
but the claim as written is wrong, and a genuinely down-consuming marker remains untested either
way.

## Verdict

Required checks green at `8fadbd21` (`check`, `check-macos`, `check-ios`, plus the emulator job and
lint). `closingIssuesReferences` is `[259]` only; #275 stays open, correctly — it is about the suite
having no expected-failure baseline, which this does not change. `CHANGELOG.md` untouched,
`changelog.d/259.fixed.md` well-formed, no provenance annotations. Squash-merging.

Issue #332, opened 2026-09-07
resolveNode's overlay-root stop in the climb-up loop is unguarded: mutation M6 survives the whole suite


monkopedia-coder:
Found during #328's review, as a **surviving mutant**: the overlay-root stop in `resolveNode`'s
climb-up loop can be deleted and the entire suite — including all nine new `ResolveOverlayTest`
cases — stays green.

The mutation (M6) replaces upstream's predicate

```kotlin
if (!overlays && node is TreeNode && node.index < 0) null else node.parent
```

with a plain `node.parent`, dropping the stop at an overlay root (sentinel `index = -1`).

## It is observable, so this is a coverage gap and not dead code

The reviewer confirmed the branch changes an answer rather than merely existing:

```
mixedTree().resolveInner(12, 1).parent   // = Expr(10-22), the overlay root
    .resolve(26, 1)

  as shipped (and upstream):  Expr(10-22)
  under M6:                   Text(24-30)
```

So the code is **correct but unguarded**. Nothing currently distinguishes a build that has the stop
from one that does not, which means a future refactor can remove it silently — and this is exactly
the class of defect #252 was: a condition that was present in upstream, absent in the port, and
invisible because no test separated the two behaviours.

## Why it is worth a test rather than a note

The other eight `resolveNode` behaviours each have a mutation that turns a named test red; #328's
review published the full matrix (M1 → 2 red, M2 → 1, M3 → 4, M4 → 2, M5 → all 9). **M6 is the only
one that survives.** A single test asserting the case above closes the matrix, and the reproduction
above is already reduced to one expression — the expensive part of writing it is done.

Note the M5 result is what makes this worth doing rather than shrugging at: inverting the mode
selection turns **all nine** tests red, including the two that pass both before and after the #252
fix. That suite is otherwise a genuinely tight guard, and M6 is the one hole left in it.

Small and self-contained: one test in
`lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/ResolveOverlayTest.kt`,
no production change expected. Whoever takes it should confirm the new test **fails under M6** — a
test added without watching it fail proves nothing, which is the whole point of this issue.

Issue #331, opened 2026-09-07
TreeNode.enter ignores MountedTree.bracketed: ENTER_BRACKETED relaxes overlay boundaries for every mount, not just bracketed ones


monkopedia-reviewer:
`TreeNode.enter` in `:lezer-common` derives `enterBracketed` from the `IterMode.ENTER_BRACKETED`
flag alone; upstream also requires the mount itself to be bracketed.

### Upstream — `@lezer/common@1.5.2`, `dist/index.js:786-796`

```js
enter(pos, side, mode = 0) {
    let mounted;
    if (!(mode & IterMode.IgnoreOverlays) && (mounted = MountedTree.get(this._tree)) && mounted.overlay) {
        let rPos = pos - this.from, enterBracketed = (mode & IterMode.EnterBracketed) && mounted.bracketed;
        for (let { from, to } of mounted.overlay) {
            if ((side > 0 || enterBracketed ? from <= rPos : from < rPos) &&
                (side < 0 || enterBracketed ? to >= rPos : to > rPos))
                return new TreeNode(mounted.tree, mounted.overlay[0].from + this.from, -1, this);
        }
    }
    return this.nextChild(0, 1, pos, side, mode);
}
```

### Port — `lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:645-666`

```kotlin
val enterBracketed = mode and IterMode.ENTER_BRACKETED != 0
```

`&& mounted.bracketed` is missing. `MountedTree.bracketed` does exist in the port
(`Mix.kt:51`, `:71`, `:82`, and it is populated from `NestedParse.bracketed` at `Mix.kt:228`, `:262`,
`:336`) — it is simply not consulted here.

### Consequence

With `IterMode.ENTER_BRACKETED` set, the port relaxes the overlay range test to inclusive on both
ends (`from <= rPos` / `to >= rPos`) for **every** mount, whereas upstream relaxes it only for mounts
whose `NestedParse` declared `bracketed = true`. A caller that passes `ENTER_BRACKETED` therefore
enters non-bracketed overlays at their exact boundary positions where upstream would not, returning
an inner-language node where the host language is correct.

Not reachable from `resolveNode` (which never passes `ENTER_BRACKETED`), so this is independent of
#252/#328 — spotted while reviewing that change.

### Suggested fix

```kotlin
[...]

Issue #330, opened 2026-09-07
Tree.resolve/resolveInner do not port upstream's CachedNode/CachedInnerNode memoisation: every resolution descends from topNode


monkopedia-reviewer:
Reviewing #328 (which fixed #252) confirmed that `Tree.resolve` / `Tree.resolveInner` in
`:lezer-common` do not port upstream's per-tree resolve memoisation.

### Upstream — `@lezer/common@1.5.2`, `dist/index.js:464` and `:477`

```js
resolve(pos, side = 0) {
    let node = resolveNode(CachedNode.get(this) || this.topNode, pos, side, false);
    CachedNode.set(this, node);
    return node;
}
resolveInner(pos, side = 0) {
    let node = resolveNode(CachedInnerNode.get(this) || this.topNode, pos, side, true);
    CachedInnerNode.set(this, node);
    return node;
}
```

`CachedNode` / `CachedInnerNode` are `WeakCache` (`Tree -> SyntaxNode`) instances holding the
previous result, so a sequence of nearby resolutions climbs a short way up from the last answer
instead of descending the whole tree from `topNode` every time.

### Port — `lezer-common/.../SyntaxTree.kt`

```kotlin
fun resolve(pos: Int, side: Int = 0): SyntaxNode = resolveNode(topNode, pos, side, false)
fun resolveInner(pos: Int, side: Int = 0): SyntaxNode = resolveNode(topNode, pos, side, true)
```

Always the cold-cache path.

### Why this is a performance gap and not a correctness gap

`resolveNode` climbs up from whatever node it is handed until it reaches a node that holds `pos`,
and (for `overlays`) then runs the climb-out-of-overlays pass. Because every ancestor of a cached
node is on the `topNode`-to-cached-node `enter` chain, the climb-up from a cached node lands on a
node the `topNode` descent also passes through, and the descent continues identically from there.
Upstream's warm-cache path is therefore intended to agree with its cold-cache path — and the port
*always* takes upstream's cold-cache path, so it can never disagree with upstream. The two
`SyntaxNode` entry points (`BaseNode.resolve`/`resolveInner`, `:704-709`) are uncached upstream too,
so those are already at exact parity.

Filed to record the deferral somewhere findable rather than only in #328's description. Analogous
to #280 (`nodeSize` memoisation).

### Cost

Every `resolve`/`resolveInner` walks from the root. These are the primary tree query behind
indentation, folding, bracket matching, highlighting and completion context, and are called
[...]

Issue #329, opened 2026-09-07
Tree.resolveStack does not produce overlay layers: stackIterator is a plain cursorAt parent chain, not the upstream layered algorithm


monkopedia-coder:
`Tree.resolveStack` is documented as the query that gives you *all* nodes around a position,
**including those in overlays that don't directly cover the position** — that is its entire reason
for existing over `resolveInner`. The port's backing helper is not the upstream algorithm and never
produces those overlay layers.

Diffed against `@lezer/common@1.5.2` (npm dist, `dist/index.js:950-968`).

### Upstream

```js
function stackIterator(tree, pos, side) {
    let inner = tree.resolveInner(pos, side), layers = null;
    for (let scan = inner instanceof TreeNode ? inner : inner.context.parent; scan; scan = scan.parent) {
        if (scan.index < 0) { // This is an overlay root
            let parent = scan.parent
            ;(layers || (layers = [inner])).push(parent.resolve(pos, side))
            scan = parent
        } else {
            let mount = MountedTree.get(scan.tree)
            // Relevant overlay branching off
            if (mount && mount.overlay && mount.overlay[0].from <= pos && mount.overlay[mount.overlay.length - 1].to >= pos) {
                let root = new TreeNode(mount.tree, mount.overlay[0].from + scan.from, -1, scan)
                ;(layers || (layers = [inner])).push(resolveNode(root, pos, side, false))
            }
        }
    }
    return layers ? iterStack(layers) : inner;
}
```

It starts from `resolveInner`, walks the parent chain, and for every overlay boundary it crosses
pushes a *separate layer* — the host-language resolution beside the overlay one — then interleaves
them with `iterStack`, which yields innermost-to-outermost across all layers by comparing `from`.

### Port — `lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt`

```kotlin
private fun stackIterator(tree: Tree, pos: Int, side: Int): NodeIterator {
    val cursor = tree.cursorAt(pos, side)
    val nodes = mutableListOf<SyntaxNode>()
    nodes.add(cursor.node)
    while (cursor.parent()) {
        nodes.add(cursor.node)
    }
    var result: NodeIterator? = null
    for (i in nodes.indices.reversed()) {
        result = NodeIterator(nodes[i], result)
    }
    return result ?: NodeIterator(tree.topNode, null)
}
```

This is a single `cursorAt` parent chain. There is:

[...]

PR #336, opened 2026-09-07
Port enterUnfinishedNodesBefore's two-cursor walk from @lezer/common (#253)

### LINKED ISSUE #253: enterUnfinishedNodesBefore is a one-step stub — its while loop can never iterate, and it returns a child where upstream returns the starting node

monkopedia-coder:
`SyntaxNode.enterUnfinishedNodesBefore` was a one-step stub. It computed
`childBefore(pos)` once into a `val` **outside** the `while` loop — making the loop
condition constant — and every path through the body ended in `break`. It also returned
that child, where upstream returns the node whose *last child* is a zero-length error
node.

Upstream (`@lezer/common@1.5.2`, `dist/index.js:713`, verified against the npm tarball
locally) keeps **two** cursors: `scan` walks down the right edge of the tree, and `node`
advances **only** when that edge ends in a zero-length error node. The port had collapsed
them into one variable, which is why both the iteration and the return value were wrong.

```js
enterUnfinishedNodesBefore(pos) {
    let scan = this.childBefore(pos), node = this
    while (scan) {
        let last = scan.lastChild
        if (!last || last.to != scan.to) break
        if (last.type.isError && last.from == last.to) { node = scan; scan = last.prevSibling }
        else { scan = last }
    }
    return node
}
```

## What changes for a user

This is the primitive that finds the innermost still-open construct before a position; it
backs continued indentation and completion context inside an unfinished block.
`@codemirror/language`'s `syntaxIndentation` calls it as
`ast.resolveInner(pos, -1).resolve(pos, 0).enterUnfinishedNodesBefore(pos)`.

Running exactly that call on real `:lang-javascript` parses, before and after, against the
same call in upstream `@lezer/javascript@1.5.2` + `@lezer/common@1.5.2` as the oracle:

| document (cursor at end) | before (`main`) | after / upstream |
| --- | --- | --- |
| `function f() {\n  if (x) {\n    a.` | `FunctionDeclaration(0-32)` | `MemberExpression(30-32)` |
| `function f() {\n  if (x) {\n` | `FunctionDeclaration(0-26)` | `Block(24-26)` |
| `let o = {\n  a: 1,\n  b: {\n` | `VariableDeclaration(0-25)` | `ObjectExpression(23-25)` |
| `if (x) { y(` | `IfStatement(0-11)` | `ArgList(10-11)` |
| `class A { m() { if (b) {` | `ClassDeclaration(0-24)` | `Block(23-24)` |
| `foo(bar(baz(` | `ExpressionStatement(0-12)` | `ArgList(11-12)` |
| `function f() { let a = 1; }` (complete) | `FunctionDeclaration(0-27)` | `Script(0-27)` |

All seven disagreed with upstream before; all seven agree after. Note the last row: on a
**complete** document the stub reported an unfinished node where there is none.

Honest scoping caveat: today the only in-repo callers are `lang-liquid` and `lang-jinja`
completion (`LiquidComplete.kt:78`, `JinjaComplete.kt:80`), which call it directly on
`resolveInner(pos, -1)` without upstream's `.resolve(pos, 0)` — that start node is usually a
childless token, so the two implementations often agree there. `language/Indent.kt` does not
call it at all yet (the port's `TreeIndentContext` uses `resolveInner(pos)` only). So the
table above is the behaviour of the primitive itself measured on real trees, not a
demonstration of a currently-shipping indentation regression. I could not demonstrate a
user-visible change through `:lang-liquid` because its parser throws on **every** input
(see "Found separately" below).

## Tests — red first, on a concrete node

Five tests in `TreeTest.kt`, each asserting `name(from-to)` in document coordinates. The two
pre-existing tests here asserted only `assertNotNull(result)`; both are now pinned to a
concrete node (they pass unchanged on `main`, so they are strengthening, not fixing).

Copied alone onto unmodified `origin/main` @ `6fbf5757`
(`lezer-common/build/test-results/jvmTest/TEST-…TreeTest.xml`, 54 tests, 4 failures):

```
enterUnfinishedNodesBeforeDescendsMoreThanOneLevel[jvm]  expected:<Br(3-6)> but was:<Pa(0-6)>
enterUnfinishedNodesBeforeDescendsThreeLevels[jvm]       expected:<Br(6-9)> but was:<Pa(0-9)>
enterUnfinishedNodesBeforeContinuesIntoPreviousSibling[jvm] expected:<Br(0-6)> but was:<Pa(0-6)>
enterUnfinishedNodesBeforeStaysPutWithoutErrorNode[jvm]  expected:<T(0-6)>  but was:<Pa(0-6)>
```

Fixtures are two and three levels deep on the right edge, so a one-step stub and a correct
loop cannot give the same answer. Every expected value was computed by running the same
fixture through `@lezer/common@1.5.2` in node, not derived from the Kotlin.

## Mutation testing

Each defect is guarded independently. Suite run is `:lezer-common:jvmTest` + `:language:jvmTest`
(83 + 91 tests):

| mutant | killed by |
| --- | --- |
| M1 — revert the loop only (single step, correct return-value gating) | `DescendsMoreThanOneLevel`, `DescendsThreeLevels`, `ContinuesIntoPreviousSibling` (3 failed) |
| M2 — revert the return value only (`result = scan` on every iteration, loop intact) | `StaysPutWithoutErrorNode` (1 failed) |
| M3 — `last.prevSibling` → `last` | `ContinuesIntoPreviousSibling` (1 failed) |
| M4 — drop `last.to != scan.to` from the break guard | `StopsWhenRightEdgeFallsShort` (1 failed) |

M1 and M2 kill disjoint sets, so the broken iteration and the wrong return value are
separately pinned.

**M4 initially survived the entire suite** — the `last.to != scan.to` half of the break
guard could be deleted with nothing noticing. Per #328's review lesson (surviving mutant M6,
filed as #332), I hunted for that rather than reporting a clean sweep, and added
`enterUnfinishedNodesBeforeStopsWhenRightEdgeFallsShort` (Pa's last child ends at 6, short of
Pa's own end at 10, so the walk must stop before reaching an empty error node nested deeper).
That mutant is now killed.

After that, no line of the 12-line body survives deletion untested: the `result` initialiser
is pinned by `StaysPut`/`StopsWhenRightEdgeFallsShort`, `result = scan` and the `else { last }`
descent by the three descent tests, `last.prevSibling` by M3, the guard by M4. The `last == null`
half of the guard is not deletable (Kotlin null-safety).

## Regression surface

`:lezer-common`, `:language` and 10 `lang-*` suites, JVM, on a clean `origin/main` worktree
vs. this branch:

```
                     before   after
lezer-common             78      83
language                 91      91
lang-javascript          39      39
lang-python              41      41
lang-json                31      31
lang-html                32      32
lang-css                 16      16
lang-java                33      33
lang-markdown            63      63
lang-xml                 13      13
lang-liquid / lang-jinja   0       0   (no test sources)
TOTAL                   437     442    0 failed both sides
```

The +5 delta is accounted for **by name** — every added test is one of mine, and nothing was
removed:

```
+ TreeTest.enterUnfinishedNodesBeforeDescendsMoreThanOneLevel
+ TreeTest.enterUnfinishedNodesBeforeDescendsThreeLevels
+ TreeTest.enterUnfinishedNodesBeforeContinuesIntoPreviousSibling
+ TreeTest.enterUnfinishedNodesBeforeStaysPutWithoutErrorNode
+ TreeTest.enterUnfinishedNodesBeforeStopsWhenRightEdgeFallsShort
```

Counts read out of `build/test-results/**/*.xml`, not from Gradle's exit code.
`:lezer-common:wasmJsBrowserTest` also run (`CHROME_BIN=/usr/bin/chromium`): 83 tests, 0 failed,
confirmed from the XML.

`:lezer-common:apiCheck` green — no `.api` or `.klib.api` file moved (`git status` clean of
`api/`). `ktlintCheck` and `spotlessCheck` green.

## Found separately — not fixed here

- **`:lang-liquid` appears entirely non-functional.** `liquidLanguage.parser.parse(…)` throws
  on every input I tried, including plain `<p>hello</p>`
  (`IndexOutOfBoundsException: Index -12 out of bounds for length 3`) and unfinished liquid
  (`ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458` at
  `LRParser.getGoto`). `liquidTagLanguage.parser.parse("if a")` throws too. Identical on
  `origin/main` @ `6fbf5757` and on this branch, so it predates and is unrelated to this
  change. The module has no test sources at all, which is presumably why nobody noticed.
  Worth its own issue.
- `language/Indent.kt`'s `TreeIndentContext` uses `resolveInner(pos)` where
  `@codemirror/language` 6.11.3 uses `resolveInner(pos, -1).resolve(pos, 0)
  .enterUnfinishedNodesBefore(pos)` plus a `resolveStack` merge — a separate porting gap.
- Not touched, per instruction: #329, #330, #331, #332.

Fixes #253

### DIFF
--- a/changelog.d/253.fixed.md
+++ b/changelog.d/253.fixed.md
@@ -1,0 +1,13 @@
+- Fixed `SyntaxNode.enterUnfinishedNodesBefore`, which stopped after a single level and reported
+  an unfinished node where there was none (#253). The port computed the child to test once,
+  outside its `while` loop, and every path through the loop body ended in `break`, so the walk
+  could never descend past the first child; it then returned that child rather than the node whose
+  right edge actually ends in a zero-length error node. `@lezer/common` keeps two cursors — one
+  walking down the right edge of the tree, one recording the last node whose right edge ended in
+  an unfinished (empty error) node — and returns the second, unchanged when there is no such node.
+  This is the primitive that finds the innermost still-open construct before a position, so it
+  backs continued indentation and completion context inside an unfinished block: on a JavaScript
+  document ending in `function f() {\n  if (x) {\n    a.` it returned `FunctionDeclaration(0-32)`
+  where `@lezer/javascript` returns `MemberExpression(30-32)`, and on the complete document
+  `function f() { let a = 1; }` it returned `FunctionDeclaration(0-27)` where upstream returns the
+  `Script` itself. Both cursors are now kept and the loop advances, matching `@lezer/common` 1.5.2.
--- a/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
+++ b/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
@@ -1300,19 +1300,21 @@ private fun stackIterator(tree: Tree, pos: Int, side: Int): NodeIterator {
 // ---- enterUnfinishedNodesBefore helper ----
 
 private fun enterUnfinishedNodesBefore(node: SyntaxNode, pos: Int): SyntaxNode {
-    var scan: SyntaxNode = node
-    val cursor = scan.childBefore(pos)
-    while (cursor != null) {
-        val last = cursor.lastChild
-        if (last == null || last.to != cursor.to) break
-        if (last.type.isError && last.from == last.to) {
-            scan = cursor
-            break
+    // `scan` walks down the right edge of the tree; `result` only advances when
+    // that edge actually ends in an unfinished (zero-length error) node.
+    var scan: SyntaxNode? = node.childBefore(pos)
+    var result: SyntaxNode = node
+    while (scan != null) {
+        val last = scan.lastChild
+        if (last == null || last.to != scan.to) break
+        scan = if (last.type.isError && last.from == last.to) {
+            result = scan
+            last.prevSibling
+        } else {
+            last
         }
-        scan = cursor
-        break
     }
-    return scan
+    return result
 }
 
 // ---- Tree.build implementation ----
--- a/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/TreeTest.kt
+++ b/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/TreeTest.kt
@@ -509,40 +509,187 @@ class TreeTest {
 
     // ---- enterUnfinishedNodesBefore tests ----
 
+    /** Error node type used by the enterUnfinishedNodesBefore fixtures. */
+    private val errorType = NodeType.define(
+        NodeTypeSpec(name = "\u26A0", id = 10, error = true)
+    )
+
+    private fun leaf(type: NodeType, length: Int) = Tree(type, emptyList(), emptyList(), length)
+
+    /** `name(from-to)` in document coordinates, so failures name a concrete node. */
+    private fun describe(node: SyntaxNode) = "${node.name}(${node.from}-${node.to})"
+
     @Test
     fun enterUnfinishedNodesBeforeOnNormalTree() {
         val tree = simple()
         val node = tree.topNode
-        // No error nodes, should return the node itself or a child
-        val result = node.enterUnfinishedNodesBefore(10)
-        assertNotNull(result)
+        // childBefore(10) is Pa(4-20), whose last child Br(16-18) does not reach
+        // Pa's end, so the walk stops immediately and the start node is returned.
+        assertEquals("T(0-20)", describe(node.enterUnfinishedNodesBefore(10)))
     }
 
     @Test
     fun enterUnfinishedNodesBeforeWithErrorNode() {
-        // Build a tree with an error node
-        val errorType = NodeType.define(
-            NodeTypeSpec(name = "\u26A0", id = 10, error = true)
-        )
+        // T(0-5) { a(0-3), err(3-3) }
         val a = types[1]
         val tree = Tree(
             types[0],
             listOf(
-                Tree(a, emptyList(), emptyList(), 3),
+                leaf(a, 3),
                 // zero-length error node
-                Tree(
-                    errorType,
-                    emptyList(),
-                    emptyList(),
-                    0
-                )
+                leaf(errorType, 0)
             ),
             listOf(0, 3),
             5
         )
-        val node = tree.topNode
-        val result = node.enterUnfinishedNodesBefore(5)
-        assertNotNull(result)
+        // childBefore(5) is the empty error node itself, which has no children,
+        // so the walk stops and the start node is returned.
+        assertEquals("T(0-5)", describe(tree.topNode.enterUnfinishedNodesBefore(5)))
+    }
+
+    @Test
+    fun enterUnfinishedNodesBeforeDescendsMoreThanOneLevel() {
+        // T(0-6) { Pa(0-6) { b(0-3), Br(3-6) { c(3-6), err(6-6) } } }
+        // The right edge is two levels deep and ends in a zero-length error
+        // node, so the walk must iterate twice to reach Br(3-6).
+        val tree = Tree(
+            types[0],
+            listOf(
+                Tree(
+                    types[4],
+                    listOf(
+                        leaf(types[2], 3),
+                        Tree(
+                            types[5],
+                            listOf(leaf(types[3], 3), leaf(errorType, 0)),
+                            listOf(0, 3),
+                            3
+                        )
+                    ),
+                    listOf(0, 3),
+                    6
+                )
+            ),
+            listOf(0),
+            6
+        )
+        assertEquals("Br(3-6)", describe(tree.topNode.enterUnfinishedNodesBefore(6)))
+    }
+
+    @Test
+    fun enterUnfinishedNodesBeforeDescendsThreeLevels() {
+        // T(0-9) { Pa(0-9) { b(0-3), Br(3-9) { c(3-6), Br(6-9) { c(6-9), err(9-9) } } } }
+        // Three descents are required before the empty error node is reached.
+        val tree = Tree(
+            types[0],
+            listOf(
+                Tree(
+                    types[4],
+                    listOf(
+                        leaf(types[2], 3),
+                        Tree(
+                            types[5],
+                            listOf(
+                                leaf(types[3], 3),
+                                Tree(
+                                    types[5],
+                                    listOf(leaf(types[3], 3), leaf(errorType, 0)),
+                                    listOf(0, 3),
+                                    3
+                                )
+                            ),
+                            listOf(0, 3),
+                            6
+                        )
+                    ),
+                    listOf(0, 3),
+                    9
+                )
+            ),
+            listOf(0),
+            9
+        )
+        assertEquals("Br(6-9)", describe(tree.topNode.enterUnfinishedNodesBefore(9)))
+    }
+
+    @Test
+    fun enterUnfinishedNodesBeforeStopsWhenRightEdgeFallsShort() {
+        // T(0-10) { Pa(0-10) { b(0-3), Br(3-6) { c(3-6), err(6-6) } } }
+        // Pa's last child ends at 6, short of Pa's own end at 10, so Pa is not
+        // "unfinished on its right edge" and the walk must stop before ever
+        // seeing the empty error node nested inside Br.
+        val tree = Tree(
+            types[0],
+            listOf(
+                Tree(
+                    types[4],
+                    listOf(
+                        leaf(types[2], 3),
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Approved. Independently re-verified end to end.

**Upstream.** Fetched `@lezer/common@1.5.2` fresh from npm; `dist/index.js:713` is exactly the
two-cursor form quoted. The port reproduces the **control flow**, not just the outputs: both
variables (`scan` walks the right edge, `result` only advances on the error branch), the
`last == null || last.to != scan.to` break, `result = scan` + `scan = last.prevSibling` on the
error branch, `scan = last` otherwise. The Kotlin `scan = if (…) { result = scan; last.prevSibling }
else { last }` evaluates `result = scan` before the assignment to `scan`, so the ordering matches.

**Fixture depth is real.** I instrumented upstream with an iteration counter and ran the seven
Kotlin fixtures through it. Every expected value matches upstream exactly, and the descent tests
genuinely need more than one pass:

```
                                upstream    iters   stub        discriminates?
OnNormalTree                    T(0-20)       1     T(0-20)     no
WithErrorNode                   T(0-5)        1     T(0-5)      no
DescendsMoreThanOneLevel        Br(3-6)       3     Pa(0-6)     YES
DescendsThreeLevels             Br(6-9)       4     Pa(0-9)     YES
StopsWhenRightEdgeFallsShort    T(0-10)       1     T(0-10)     no (guards M4)
StaysPutWithoutErrorNode        T(0-6)        2     Pa(0-6)     YES
ContinuesIntoPreviousSibling    Br(0-6)       3     Pa(0-6)     YES
```

Not a red-first artefact.

**Two defects, separately pinned — confirmed by running the mutants, not by reading.**
`:lezer-common:jvmTest` + `:language:jvmTest`:

| mutant | outcome |
| --- | --- |
| M1 — single-step loop, return-value gating kept | killed by `DescendsMoreThanOneLevel`, `DescendsThreeLevels`, `ContinuesIntoPreviousSibling` |
| M2 — `result = scan` every iteration, loop intact | killed by `StaysPutWithoutErrorNode` |
| M3 — `last.prevSibling` -> `last` | killed by `ContinuesIntoPreviousSibling` |
| M4 — drop `last.to != scan.to` | killed by `StopsWhenRightEdgeFallsShort` |

M1 and M2 kill disjoint sets, so broken iteration and wrong return value are independently
guarded. The two you called out — the `to`-guard and the `prevSibling` step — are both covered.
Nice catch finding M4 yourself and adding a test rather than reporting a clean sweep.

**Surviving mutants.** Your "no line of the 12-line body survives deletion untested" holds at
line granularity, but the *error test* has both halves unguarded:

- M5 — delete `&& last.from == last.to`: **survives 174/174**
- M6 — delete `last.type.isError &&`: **survives 174/174**

Neither is equivalent — upstream diverges on a non-empty error node on the right edge
(`T(0-6)` vs `Pa(0-6)`) and on a zero-length non-error node (`T(0-3)` vs `Pa(0-3)`). Every
fixture uses a zero-length *error* node, so the conjunction is never split. This is
carried-forward surface (same condition text was in the stub), not a regression here, so it is
not blocking — filed as #340 with the two fixtures needed.

**Regression surface.** `:lezer-common`, `:language` and 10 `lang-*` JVM suites, clean worktrees
off `origin/main` @ `d833c0b6` vs `cd6b09fd`, counts read from `build/test-results/**/*.xml`:
**489 -> 494, 0 failures both sides.** The +5 is accounted for by name and is exactly your five
new tests; no test was removed or renamed. `:language` is 91 -> 91 — **no indent test moved**,
which is consistent with `language/Indent.kt` not calling this primitive at all.

**Behaviour.** Reproduced your `lang-javascript` table against `@lezer/javascript@1.5.2` +
`@lezer/common@1.5.2` in node: all seven rows match, before and after, including the complete
document where the stub invented a `FunctionDeclaration`. Your scoping caveat is the honest
reading and the changelog does not oversell it: what actually changes for a user is the public
`SyntaxNode.enterUnfinishedNodesBefore` API, which downstream consumers can call directly. In
this repo the only callers are jinja/liquid completion — and I found `:lang-jinja`'s parser
throws `ArrayIndexOutOfBoundsException` (1605/1605) on **every** non-empty input, exactly like
the `:lang-liquid` breakage you filed as #337, identical on `main` and on this head. Filed as
#339. So there is currently no reachable in-repo behaviour to demonstrate through either caller;
that is a property of those modules, not of this change.

**Housekeeping.** `:lezer-common:apiCheck` green, no `.api`/`.klib.api` moved. Changelog
fragment valid (`changelog.py check --base-ref origin/main`, 17 fragments), `CHANGELOG.md`
untouched. Scope is the three stated files (three-dot). `closingIssuesReferences` = [253] only.
None of #329/#330/#331/#332 is partly implemented. No provenance annotations. Required `check`,
`check-macos`, `check-ios` all green at `cd6b09fd`. The emulator job's 13 failures are the known
`#275`/`#259` baseline verified by name (5 `FocusManagementTest`, 8 `KeyboardHandlingTest`) — no
`view/` code here, and it is not a required check.

Issue #343, opened 2026-09-07
:lang-angular has zero test sources — the nested-JavaScript parse it ships is unguarded in this module


monkopedia-coder:
`:lang-angular` has **no test sources at all** — `lang-angular/src/` contains only `commonMain`,
and `:lang-angular:jvmTest` reports `NO-SOURCE`. The module is published.

That gap let a real defect ship: every Angular interpolation and every bound / event / structural
attribute in an `angularLanguage` parse currently gets a JavaScript subtree parsed from the
**wrong characters**. Root cause is in `:lezer-common` and is tracked in #341 (verified one-line
fix, not a `:lang-angular` bug) — this issue is the missing safety net, which is what let it go
unnoticed. Same shape as #337 (`:lang-liquid`) and #339 (`:lang-jinja`), though those two are a
different underlying defect (corrupt serialized tables).

Confirmed on `origin/main` @ `3c663304`.

### What is currently unverified

Everything. Actual output today (probe test dropped into a scratch `lang-angular/src/jvmTest/…`,
node ranges shown as `[from,to]`):

```
<p>hello</p>
  Document[0,12](Element[0,12](OpenTag[0,3](StartTag[0,1],TagName[1,2],EndTag[2,3]),
    Content[3,8](Text[3,8]),CloseTag[8,12](StartCloseTag[8,10],TagName[10,11],EndTag[11,12])))   OK

<input #box>
  Document[0,12](Element[0,12](SelfClosingTag[0,12](StartTag[0,1],TagName[1,6],
    Attribute[7,11](ReferenceName[7,11](#[7,8],Identifier[8,11])),EndTag[11,12])))               OK

<p>{{ msg }}</p>
  … Interpolation[3,12]({{[3,5],SingleExpression[5,10](ERR[5,6](>[5,6]),
      AssignmentExpression[6,10](ObjectPattern[6,10]({[6,7],ERR[7,8]({[7,8]),
      PatternProperty[9,10](PropertyName[9,10]),ERR[10,10]),ERR[10,10])),}}[10,12]) …            WRONG

<img [src]="url">
  … ExpressionAttributeValue[11,16](SingleExpression[12,15](VariableName[12,14],ERR[14,14],
      ArrayExpression[14,15](ERR[14,14],][14,15]))) …                                            WRONG
      (should be VariableName[12,15]; it actually lexed doc[7..10] == "rc]")

<button (click)="go()">x</button>
[...]

Issue #342, opened 2026-09-07
:lang-vue has zero test sources — the nested-JavaScript parse it ships is unguarded in this module


monkopedia-coder:
`:lang-vue` has **no test sources at all** — `lang-vue/src/` contains only `commonMain`, and
`:lang-vue:jvmTest` reports `NO-SOURCE`. The module is published.

That gap let a real defect ship: every Vue interpolation and every bound/directive/event attribute
in a `vueLanguage` parse currently gets a JavaScript subtree parsed from the **wrong characters**.
Root cause is in `:lezer-common` and is tracked in #341 (verified one-line fix, not a `:lang-vue`
bug) — this issue is the missing safety net, which is what let it go unnoticed. Same shape as
#337 (`:lang-liquid`) and #339 (`:lang-jinja`), though those two are a different underlying defect
(corrupt serialized tables).

Confirmed on `origin/main` @ `3c663304`.

### What is currently unverified

Everything. To give a sense of what a smoke suite would have caught immediately, here is the
actual output today (probe test dropped into a scratch `lang-vue/src/jvmTest/…`, node ranges
shown as `[from,to]`):

```
<p>hello</p>
  Document[0,12](Element[0,12](OpenTag[0,3](StartTag[0,1],TagName[1,2],EndTag[2,3]),
    Content[3,8](Text[3,8]),CloseTag[8,12](StartCloseTag[8,10],TagName[10,11],EndTag[11,12])))   OK

<template><p>hi</p></template><script>export default {}</script>
  … Script[38,55](ExportDeclaration[38,55](export[38,44],default[45,52],
      ObjectExpression[53,55]({[53,54],}[54,55]))) …                                             OK

<p>{{ msg }}</p>
  … Interpolation[3,12](SingleExpression[5,10](ERR[5,6](>[5,6]),AssignmentExpression[6,10](
      ObjectPattern[6,10]({[6,7],ERR[7,8]({[7,8]),PatternProperty[9,10](PropertyName[9,10]),
      ERR[10,10]),ERR[10,10])),}}[10,12]) …                                                      WRONG

<p :class="foo">x</p>
  … ScriptAttributeValue[10,15](SingleExpression[11,14](AssignmentExpression[11,14](
      VariableName[11,12],Equals[12,13],String[13,14](ERR[14,14]))))) …                          WRONG

<button @click="go()">x</button>
  … ScriptAttributeValue[15,21](SingleExpression[16,20](ERR[16,17](@[16,17]),
[...]

Issue #340, opened 2026-09-07
enterUnfinishedNodesBefore: both halves of the empty-error-node test are unguarded — two mutants survive the whole suite


monkopedia-reviewer:
Found during the review of #336 (the same hunt that produced #332 out of #328's review).

#336 fixed `enterUnfinishedNodesBefore`'s iteration and return value and added five tests.
Four mutants are killed, but the *error test itself* —
`if (last.type.isError && last.from == last.to)` in
`lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt:1310`
— has both of its halves unguarded. Verified empirically against
`:lezer-common:jvmTest` + `:language:jvmTest` (174 tests) at `cd6b09fd`:

| mutant | result |
| --- | --- |
| M1 — loop can only run once (return-value gating kept) | killed (3 tests) |
| M2 — `result = scan` on every iteration (loop intact) | killed (`StaysPutWithoutErrorNode`) |
| M3 — `last.prevSibling` -> `last` | killed (`ContinuesIntoPreviousSibling`) |
| M4 — drop `last.to != scan.to` from the break guard | killed (`StopsWhenRightEdgeFallsShort`) |
| **M5 — `... && last.from == last.to` deleted** | **SURVIVES, 174/174 green** |
| **M6 — `last.type.isError && ...` deleted** | **SURVIVES, 174/174 green** |

Neither is an equivalent mutant. Running the same walk in `@lezer/common@1.5.2` with each
predicate substituted shows an observable divergence on realizable trees:

```
tree                                                     upstream   mutant
T(0-6){Pa(0-6){b(0-3), err(3-6)}}     non-empty error     T(0-6)     M5 -> Pa(0-6)
T(0-3){Pa(0-3){b(0-3), c(3-3)}}       zero-len non-error  T(0-3)     M6 -> Pa(0-3)
```

Every fixture in `TreeTest` uses a zero-length **error** node on the right edge, so
`isError && from == to` is never distinguished from either half alone.

This is carried-forward surface, not a regression from #336 — the same condition text was in
the pre-fix stub. Fix: add two fixtures to `TreeTest.kt`, one with a non-empty error node on
the right edge (kills M5) and one with a zero-length non-error node on the right edge (kills
M6). Expected values from the table above; both were computed by running the fixture through
`@lezer/common@1.5.2` in node.

PR #345, opened 2026-09-07
Restore parseMixed's base-tree origin so nested mixed parses read the right text

### LINKED ISSUE #341: parseMixed loses the base-tree offset when nested inside another parseMixed, so a third-level parser lexes shifted text (breaks :lang-vue and :lang-angular)
`parseMixed` produces correct trees at one level of nesting, but as soon as a *nested* parser is
itself a `parseMixed` parser (parser A mounts parser B, and B mounts parser C), C lexes the wrong
characters: it reads `doc[pos - ranges[0].from]` instead of `doc[pos]`. The resulting subtree has
plausible-looking node ranges over garbage content, so nothing throws and nothing looks obviously
wrong — it just silently mis-highlights.

This is the whole user-visible story for `:lang-vue` and `:lang-angular`, both of which are
exactly three levels deep (`html` → vue/angular template grammar → `javascript`). Sibling issues
#337 / #339 (`:lang-liquid`, `:lang-jinja`) are a *different* defect — corrupt serialized tables —
so this is not the same bug, but it is the same "no tests in the module" cause of it shipping.

Reproduced on `origin/main` @ `3c663304`.

### Root cause

`MixedParse.startInner()` (`lezer-common/.../Mix.kt:166`) walks the base tree with:

```kotlin
val cursor = baseTree!!.cursor(
    IterMode.INCLUDE_ANONYMOUS or IterMode.IGNORE_MOUNTS
)
```

`Tree.cursor(mode)` builds its root as `TreeNode(root, 0, null, 0)` (`SyntaxTree.kt:877`) — i.e.
at offset **0**. Upstream `@lezer/common` `src/mix.ts` explicitly does *not*:

```js
let cursor = new TreeCursor(new TreeNode(this.baseTree!, this.ranges[0].from, 0, null),
                            IterMode.IncludeAnonymous | IterMode.IgnoreMounts)
```

The offset matters because the LR parse builds its tree with `start = ranges[0].from`
(`lezer-lr/.../Parse.kt:586`), so `baseTree`'s coordinates are **relative to `ranges[0].from`**,
not absolute. At the top level `ranges[0].from == 0` and the two agree — which is why every
existing test passes. Inside a nested parse `ranges[0].from` is the mounting node's start, and
every `cursor.from` / `cursor.to` in `startInner` is short by exactly that much. Those values then
feed:

- `nest(cursor, input)` — the user predicate reads the wrong slice of `input`; and
- `parseRanges = punchRanges(ranges, listOf(TextRange(cursor.from, cursor.to)))`, which is handed
  to `nestResult.parser.startParse(input, …, parseRanges)` against the **full document** `input`.

Hence the innermost parser lexes at `pos - ranges[0].from`.

### Reproduction

Scratch `lang-vue/src/jvmTest/…/ProbeTest.kt` and `lang-angular/src/jvmTest/…/ProbeTest.kt`
printing `treeToString` with `[from,to]` on every node (neither module has a test source set, so
one has to be created). The decisive case is that **identical content parses differently depending
on where it sits in the document**:

```
vueLanguage.parser.parse("<p :class=\"foo\"></p>")
  … Attribute[3,15](:[3,4],Identifier[4,9],Is[9,10],ScriptAttributeValue[10,15](
      SingleExpression[11,14](AssignmentExpression[11,14](
        VariableName[11,12],Equals[12,13],String[13,14](ERR[14,14]))))) …

vueLanguage.parser.parse("<pppppppp :class=\"foo\"></pppppppp>")
  … Attribute[10,22](…,ScriptAttributeValue[17,22](
      SingleExpression[18,21](VariableName[18,19],ERR[20,20],ERR[20,21](:[20,21])))) …
```

Same attribute, two different wrong JS trees. Decoding the first: the outer `Attribute` starts at
3, and the JS parse claims `Equals` at `[12,13]` — but `doc[12] == 'o'`; `doc[12-3] == '='`. The
shift is exactly `Attribute.from`. The second confirms it with a shift of 10.

The `Attribute` for the Angular `[src]="url"` case is the clearest read:

```
angularLanguage.parser.parse("<img [src]=\"url\">")
  … ExpressionAttributeValue[11,16](SingleExpression[12,15](
      VariableName[12,14],ERR[14,14],ArrayExpression[14,15](ERR[14,14],][14,15]))) …
```

`SingleExpression[12,15]` should be `url`. `Attribute.from == 5`, and `doc[7..10] == "rc]"` —
which is precisely `VariableName("rc")` followed by an `ArrayExpression` containing `]`. It parsed
`rc]`.

And a case where the *positions* look fine but the text read was wrong anyway, which is how this
hides:

```
angularLanguage.parser.parse("<button (click)=\"go()\">x</button>")
  … StatementAttributeValue[16,22](Script[17,21](ExpressionStatement[17,21](VariableName[17,21]))) …
```

`[17,21]` is `go()` and should be `CallExpression(VariableName, ArgList)`. `Attribute.from == 8`,
`doc[9..13] == "clic"` — a 4-character `VariableName`, which lands on a 4-character range. No
error node, no exception, wrong tree.

### The nesting depth is the variable, not the grammar

Driving the vue/angular *sub*-parsers directly — one level of `parseMixed` (vue grammar → js), no
outer html — gives completely correct trees:

```
textParser.parse("a {{ b + 1 }} c")
  Content[0,15](Text[0,2],Interpolation[2,13](SingleExpression[4,11](
    BinaryExpression[5,10](VariableName[5,6],ArithOp[7,8],Number[9,10])),}}[11,13]),Text[13,15])

attrParser.parse("(click)=\"go()\"")
  Attribute[0,14](EventName[0,7](…),Is[7,8],StatementAttributeValue[8,14](
    Script[9,13](ExpressionStatement[9,13](CallExpression[9,13](
      VariableName[9,11],ArgList[11,13](([11,12],)[12,13]))))))
```

The bare `vueParser` / `angularParser` LR tables are also fine on their own (`{{ msg }}` →
`Content(Interpolation(InterpolationContent,}}))`, `[src]="url"` →
`Attribute(BoundAttributeName([,Identifier,]),Is,ExpressionAttributeValue(AttributeInterpolation))`,
etc.), so unlike #337/#339 there is nothing wrong with the serialized payloads here.

### Verified fix

Restoring the upstream offset fixes every case above. Experimentally, in `Mix.kt`:

```kotlin
val cursor = cursorAtOffset(
    baseTree!!,
    ranges[0].from,
    IterMode.INCLUDE_ANONYMOUS or IterMode.IGNORE_MOUNTS
)
```

with an internal helper next to `TreeCursor` in `SyntaxTree.kt` (the primary constructor only
accepts a `Tree`, so there is currently no way to seed a non-zero offset):

```kotlin
internal fun cursorAtOffset(root: Tree, offset: Int, mode: Int): TreeCursor {
    val c = TreeCursor(root, mode)
    c.yieldNode(TreeNode(root, offset, null, 0))
    return c
}
```

Both helpers are `internal`, so no public API change. With that applied, every probe above becomes
correct and offset-invariant:

```
<p :class="foo"></p>          ScriptAttributeValue[10,15](SingleExpression[11,14](VariableName[11,14]))
<pppppppp :class="foo"></…>   ScriptAttributeValue[17,22](SingleExpression[18,21](VariableName[18,21]))
<img [src]="url">             ExpressionAttributeValue[11,16](SingleExpression[12,15](VariableName[12,15]))
<button (click)="go()">x</…>  StatementAttributeValue[16,22](Script[17,21](ExpressionStatement[17,21](
                                CallExpression[17,21](VariableName[17,19],ArgList[19,21](([19,20],)[20,21])))))
<p>{{ msg }}</p>              Interpolation[3,12](SingleExpression[5,10](VariableName[6,9]),}}[10,12])
```

No error nodes left, and identical content at different offsets now yields identical structure.

`:lezer-common:jvmTest`, `:lang-html:jvmTest`, `:lang-javascript:jvmTest` and `:language:jvmTest`
are all green with the patch applied, so nothing depended on the zero offset.

I have not opened a PR — someone should confirm the helper is the shape they want (an
`internal constructor(root: TreeNode, mode: Int)` on `TreeCursor` may be tidier and closer to
upstream) and land it with regression tests.

### Suggested work

1. Apply the offset fix in `Mix.kt` / `SyntaxTree.kt`.
2. Add a `:lezer-common` regression test for two-deep `parseMixed` nesting that asserts the
   innermost parse's node *content*, not just that it does not throw — a position-only assertion
   would have passed against the bug (see the `(click)="go()"` case above).
3. Regression tests in `:lang-vue` and `:lang-angular` — tracked separately, since both modules
   have no test source set at all today.

monkopedia-coder:
Fixes #341

## The defect

A parse started over ranges builds its tree relative to `ranges[0].from` — `@lezer/lr`'s
`Parse.finish` passes it to `Tree.build` as `start` — but `MixedParse.startInner` walked that tree
with `baseTree.cursor(...)`, whose root sits at offset 0. Upstream `@lezer/common` `src/mix.ts:149`
seeds the walk explicitly:

```js
let cursor = new TreeCursor(new TreeNode(this.baseTree!, this.ranges[0].from, 0, null),
                            IterMode.IncludeAnonymous | IterMode.IgnoreMounts)
```

At the top level `ranges[0].from` is 0 and the two agree, which is why every existing test passed.
One level down it is the mounting node's start, so every `cursor.from`/`cursor.to` in `startInner`
was short by that amount — and those values feed both the `nest` predicate's `input.read` and the
`parseRanges` handed to the next parser down, which runs against the **full document**. A parser
nested two deep therefore lexed `doc[pos - ranges[0].from]`.

`:lang-vue` and `:lang-angular` are exactly that shape (`html` → template grammar → `javascript`).
Nothing threw; because the mounted subtree keeps its length it re-based onto the correct document
range, so positions and node counts looked right and several wrong subtrees contained no error node
at all.

## The fix

`TreeCursor` gains an internal primary constructor taking a `TreeNode`, mirroring upstream, with
the existing `Tree` form kept as a secondary constructor for `Tree.cursor` / `Tree.cursorAt`. The
constructor form was chosen over a local `cursorAtOffset` helper deliberately: this repo is a port,
and its recurring defect class is a locally-reasoned equivalent drifting from upstream control flow.

Both constructors are `internal`. **No public API moves** — `:lezer-common:apiCheck` green,
`git diff --name-only | grep '\.api$'` empty.

## Red-first

The test file at its final state, against unpatched `origin/main` @ `3c663304`:

```
MixNestedOffsetTest > twoLevelsDeepIsOffsetInvariant FAILED
    same fragment at offset 1 produced a different tree shape than at offset 0 —
    the nested parse read shifted text
    expected:<Doc Mid Chars [Letter Letter Letter Digit] Digit Digit>
     but was:<Doc Mid Chars [Other  Letter Letter Letter] Digit Digit>

MixNestedOffsetTest > twoLevelsDeepReadsTheDocumentAtTheRightOffset FAILED
    expected:<...Chars[4,10] [Letter[4,5] Letter[5,6] Letter[6,7] Digit[7,8] Digit][8,9] Digit[9,10]>
     but was:<...Chars[4,10] [Other[4,5]  Other[5,6]  Letter[6,7] Letter[7,8] Letter][8,9] Digit[9,10]>

MixNestedOffsetTest > offsetInvariantForAllDigitContent FAILED
    offset 2 changed the readout of the innermost parse
    expected:<Doc Mid Chars [Digit Lett]er Digit Letter Digi...>
     but was:<Doc Mid Chars [Other Oth]er Digit Letter Digi...>

MixNestedOffsetTest > oneLevelOfNestingIsUnaffected                 PASSED   (control)
MixNestedOffsetTest > shiftedReadIsNotDetectableFromPositionsAlone  PASSED   (control)
```

Both controls behaving as designed against the defect is what makes them controls rather than
padding.

## The test

The asserted property is **offset invariance**: the same fragment parsed at five different document
offsets must produce the same tree shape with correspondingly shifted ranges. That is the mechanism
itself, and it cannot pass vacuously — a shifted read produces a different shape at every offset but
zero.

The fixture is three parsers deep, mirroring vue/angular. The innermost emits one leaf per character
typed `Letter` / `Digit` / `Other`, so its output is a direct readout of the text it was handed; the
nest predicates are name-based (like `:lang-vue`'s `"Text" -> textMixed`) so a failure isolates the
*range* passed down rather than whether nesting happened at all.

Two controls ship with it:

- `oneLevelOfNestingIsUnaffected` — the path that was always correct, and whose passing hid this.
- `shiftedReadIsNotDetectableFromPositionsAlone` — passes **both** with and without the fix, kept as
  a standing argument that positions are the wrong thing to assert here. Angular's `(click)="go()"`
  was the real-world instance: it lexed `clic` into a bare four-character `VariableName` sitting on
  the right four-character range, with no error node anywhere in the tree.

## Regression surface

`parseMixed` underlies every mixed-language module and this changes coordinates, so eight modules
were run before and after (`--rerun-tasks`, JVM), counted from `build/test-results/jvmTest/*.xml`:

| module | before | after |
|---|---|---|
| `:lezer-common` | 83 | **89** |
| `:lang-html` | 32 | 32 |
| `:lang-javascript` | 39 | 39 |
| `:language` | 91 | 91 |
| `:lang-markdown` | 63 | 63 |
| `:lang-php` | 29 | 29 |
| `:lang-vue` | 0 | 0 |
| `:lang-angular` | 0 | 0 |
| **total** | **337** | **343** |

Zero failures and zero skipped in both runs. The whole `+6` delta is `MixNestedOffsetTest` in
`:lezer-common`; no existing test changed status in any module. `:lang-vue` and `:lang-angular` are
`NO-SOURCE` in both runs — that is #342 / #343 and is deliberately not addressed here.

End-to-end verification on the real modules (scratch probe, removed before commit):

```
<img [src]="url">          ExpressionAttributeValue[11,16] SingleExpression[12,15] VariableName[12,15]
<button (click)="go()">x   StatementAttributeValue[16,22] Script[17,21] ExpressionStatement[17,21]
                             CallExpression[17,21] VariableName[17,19] ArgList[19,21] ([19,20] )[20,21]
<p :class="foo">x</p>      ScriptAttributeValue[10,15] SingleExpression[11,14] VariableName[11,14]
<pppppppp :class="foo">    ScriptAttributeValue[17,22] SingleExpression[18,21] VariableName[18,21]
<p>{{ msg }}</p>           Interpolation[3,12] SingleExpression[5,10] VariableName[6,9] }}[10,12]
```

All correct, offset-invariant, no error nodes. Plain HTML, empty input and the `<template>`/
`<script>` SFC path are unchanged.

## Mutation sweep

The diff contains **no `&&` or `||`** — verified with a positive control (`Mix.kt` has 26 boolean
operators elsewhere; `git diff -U0 | grep -E '&&|\|\|'` over the added lines is empty). So the
condition-granularity sweep has nothing to enumerate, and the sweep was run at value granularity
over every constant and expression the diff introduces. Each mutant was run against
`:lezer-common`, `:lang-html`, `:lang-javascript`, `:language`, `:lang-markdown`, `:lang-php`:

| mutant | result |
|---|---|
| `Mix`: `ranges[0].from` → `0` (the original defect) | killed — 3 tests |
| `Mix`: `ranges[0].from` → `ranges[0].to` | killed — 3 tests |
| `Mix`: `ranges[0].from` → `ranges.last().from` | **survived** (see below) |
| `Mix`: root `TreeNode` index `0` → `1` | **survived — equivalent** (see below) |
| `TreeCursor`: `from = root.from` → `0` | killed — `cursorOverAnOffsetRoot…` |
| `TreeCursor`: `to = root.to` → `root.to - root.from` | killed — `cursorOverAnOffsetRoot…` |
| `TreeCursor`: secondary ctor offset `0` → `1` | killed — 3 tests |
| `TreeCursor`: `_tree = root` → zero-offset copy | killed — 3 tests |

`from = root.from` and `to = root.to` initially survived: they are the cursor's position *before it
moves*, and none of the parse fixtures nest on a top node. Since `startInner` calls
`nest(cursor, input)` on the base tree's root before moving anywhere, that position is genuinely
observable, so `cursorOverAnOffsetRootReportsDocumentCoordinatesBeforeMoving` was added to cover it
directly. Both are killed now.

Two survivors, both reported rather than papered over:

- **`ranges.last().from`** differs from `ranges[0].from` only when a *nested* parse is started with
  more than one range, i.e. an overlay-mounted mixed parse. Nothing in the repo constructs one —
  `ResolveOverlayTest` covers overlay *resolution*, not overlay parsing. Killing it needs a
  multi-range nested fixture, and building a half-faithful one (my synthetic parsers do not skip
  range gaps) seemed worse than saying so. Upstream is unambiguous that it is `ranges[0].from`.
  Happy to file the multi-range coverage gap as its own issue if you want it tracked.
- **root `TreeNode` index `0` → `1`** is an equivalent mutant. `index` is read only at
  `SyntaxTree.kt:636` and `:1034`, and both paths are guarded by a non-null `_parent`; this root has
  none. `0` is kept because upstream passes `0`.

## Checks run locally

`:lezer-common:ktlintCheck`, `spotlessCheck`, `:lezer-common:apiCheck` — all green. The two
`Unnecessary non-null assertion` warnings in `SyntaxTree.kt` are pre-existing: the same two appear on
unpatched `origin/main` (lines 692 and 1131, the latter shifted to 1138 by this diff). No new
warnings.

`python3 .github/scripts/changelog.py check --base-ref origin/main` green;
`changelog.d/341.fixed.md` added, `CHANGELOG.md` untouched.

## Scope

`Fixes #341` only. This should make #342 and #343 pass, but neither is closed here — both also carry
the zero-test-sources problem, which this PR does not address.

### DIFF
--- a/changelog.d/341.fixed.md
+++ b/changelog.d/341.fixed.md
@@ -1,0 +1,13 @@
+- `parseMixed` no longer loses the base tree's origin when it runs inside another `parseMixed`
+  (#341). A parse started over ranges builds its tree relative to `ranges[0].from`, but
+  `MixedParse` walked that tree from an origin of 0. At the top level the two agree, so the
+  defect was invisible until a nested parser was itself a mixed parser — from there every range
+  handed to the parsers below was short by the mounting node's start, and the innermost parser
+  silently lexed the wrong characters. `:lang-vue` and `:lang-angular` are exactly that shape
+  (`html` -> template grammar -> `javascript`), so every Vue interpolation and bound, directive
+  or event attribute, and every Angular interpolation and bound, event or structural attribute,
+  was parsed from text taken from the wrong place in the document: `<img [src]="url">` lexed
+  `rc]` in place of `url`, and `<button (click)="go()">` produced a bare `VariableName` over
+  `clic` — no exception, no error node, correct-looking node positions, wrong tree. The fix
+  restores `@lezer/common`'s origin, and is covered by a regression test asserting that the same
+  fragment parsed at five different document offsets yields the same tree shape.
--- a/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/Mix.kt
+++ b/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/Mix.kt
@@ -163,7 +163,13 @@ private class MixedParse(
         val fragmentCursor = FragmentCursor(fragments)
         var overlay: ActiveOverlay? = null
         var covered: CoverInfo? = null
-        val cursor = baseTree!!.cursor(
+        // The base tree is built relative to `ranges[0].from` (`@lezer/lr`'s `Parse.finish`
+        // passes it to `Tree.build` as `start`), so the walk has to begin at that offset to
+        // yield document coordinates. At the top level it is 0; nested inside another
+        // `parseMixed` it is the mounting node's start, and losing it shifts every range handed
+        // to the parsers below by exactly that much (#341).
+        val cursor = TreeCursor(
+            TreeNode(baseTree!!, ranges[0].from, null, 0),
             IterMode.INCLUDE_ANONYMOUS or IterMode.IGNORE_MOUNTS
         )
 
--- a/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
+++ b/lezer-common/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/common/SyntaxTree.kt
@@ -872,9 +872,16 @@ internal class BufferNode(val context: BufferContext, val _parent: BufferNode?, 
  * Mutable tree walker. Walks a [Tree] depth-first, handling both
  * [Tree] and [TreeBuffer] children.
  */
-class TreeCursor internal constructor(root: Tree, internal val mode: Int = 0) : SyntaxNodeRef {
+class TreeCursor internal constructor(root: TreeNode, internal val mode: Int = 0) : SyntaxNodeRef {
+    /**
+     * Cursor over a whole tree, in that tree's own coordinates. A tree parsed over ranges is
+     * built relative to `ranges[0].from` (see `MixedParse.startInner`), so a caller that needs
+     * document coordinates must use the primary constructor with an offset [TreeNode] instead.
+     */
+    internal constructor(root: Tree, mode: Int = 0) : this(TreeNode(root, 0, null, 0), mode)
+
     @Suppress("ktlint:standard:property-naming", "ktlint:standard:backing-property-naming")
-    internal var _tree: TreeNode = TreeNode(root, 0, null, 0)
+    internal var _tree: TreeNode = root
     internal var buffer: BufferContext? = null
     private val stack = mutableListOf<Int>()
     internal var index: Int = 0
@@ -882,9 +889,9 @@ class TreeCursor internal constructor(root: Tree, internal val mode: Int = 0) : 
 
     override var type: NodeType = root.type
         private set
-    override var from: Int = 0
+    override var from: Int = root.from
         private set
-    override var to: Int = root.length
+    override var to: Int = root.to
         private set
     override val name: String get() = type.name
 
--- a/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/MixNestedOffsetTest.kt
+++ b/lezer-common/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/common/MixNestedOffsetTest.kt
@@ -1,0 +1,349 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.lezer.common
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * `parseMixed` nested inside another `parseMixed` — the three-deep shape used by every
+ * template language whose expressions are a third language (`:lang-vue` and `:lang-angular`
+ * are `html` -> template grammar -> `javascript`).
+ *
+ * A parse started over ranges builds its tree in coordinates relative to `ranges[0].from`
+ * (`@lezer/lr`'s `Parse.finish` passes `start = this.ranges[0].from` to `Tree.build`), so
+ * `MixedParse.startInner` has to walk that tree from an origin of `ranges[0].from` to get
+ * document coordinates back. At the top level `ranges[0].from` is 0 and the distinction is
+ * invisible; one level down it is the mounting node's start, and getting it wrong shifts the
+ * ranges handed to the *next* parser down by exactly that much — so the innermost parser
+ * silently lexes the wrong characters (#341).
+ *
+ * ## What these tests assert
+ *
+ * The property is **offset invariance**: the same fragment, parsed at several different
+ * document offsets, must produce the same tree shape with correspondingly shifted ranges.
+ * That is the mechanism itself, and it cannot pass vacuously — a shifted read produces a
+ * different shape at every offset but zero.
+ *
+ * Asserting one expected tree would be weaker. Under the defect the mounted subtree keeps its
+ * *length*, so it re-bases onto the correct document range at its mount point: the node
+ * positions look right and only the node **types** — which are derived from the characters
+ * actually read — are wrong. `assertOffsetInvariant` therefore compares full shape, not just
+ * positions.
+ *
+ * ## Fixture
+ *
+ * ```
+ * <pad>[[abc123]]<tail>
+ * ```
+ *
+ * - `outerParser` (level 1) finds `[[` ... `]]` and emits `Outer` over the brackets and their
+ *   contents. It nests `midParser` on that node.
+ * - `midParser` (level 2) emits `Mid` over its whole range with an `Inner` child covering the
+ *   contents between the two-character brackets. It nests `charParser` on `Inner`.
+ * - `charParser` (level 3) emits one leaf per character, typed `Letter` or `Digit`, so its
+ *   output is a direct readout of the text it was given.
+ *
+ * `abc123` must therefore always come back as `Letter Letter Letter Digit Digit Digit`. With
+ * the origin lost, level 3 is asked for `doc[from - pad, to - pad]`: at `pad = "xx"` that is
+ * `[[abc1`, and the readout changes to `Other Other Letter Letter Letter Digit`.
+ */
+class MixNestedOffsetTest {
+
+    // ---- Node types ----
+
+    private val outerTop = NodeType.define(NodeTypeSpec(name = "Doc", id = 0, top = true))
+    private val outerNode = NodeType.define(NodeTypeSpec(name = "Outer", id = 1))
+    private val midTop = NodeType.define(NodeTypeSpec(name = "Mid", id = 2, top = true))
+    private val innerNode = NodeType.define(NodeTypeSpec(name = "Inner", id = 3))
+    private val charTop = NodeType.define(NodeTypeSpec(name = "Chars", id = 4, top = true))
+    private val letter = NodeType.define(NodeTypeSpec(name = "Letter", id = 5))
+    private val digit = NodeType.define(NodeTypeSpec(name = "Digit", id = 6))
+    private val other = NodeType.define(NodeTypeSpec(name = "Other", id = 7))
+
+    /**
+     * A parse that reports its tree in coordinates relative to `ranges[0].from`, the way
+     * `@lezer/lr` does, and builds it from the text it is handed. Single contiguous range —
+     * these fixtures use no overlays, so `punchRanges` never splits one.
+     */
+    private class ImmediateParse(val tree: Tree) : PartialParse {
+        override var stoppedAt: Int? = null
+            private set
+        override val parsedPos: Int get() = tree.length
+        override fun stopAt(pos: Int) {
+            stoppedAt = pos
+        }
+        override fun advance(): Tree = tree
+    }
+
+    private fun parserOf(build: (text: String) -> Tree): Parser = object : Parser() {
+        override fun createParse(
+            input: Input,
+            fragments: List<TreeFragment>,
+            ranges: List<TextRange>
+        ): PartialParse {
+            val from = ranges.first().from
+            val to = ranges.last().to
+            return ImmediateParse(build(input.read(from, to)))
+        }
+    }
+
+    /** Level 3: one leaf per character, typed by what that character actually is. */
+    private val charParser: Parser = parserOf { text ->
+        Tree(
+            charTop,
+            text.map { c ->
+                val type = when {
+                    c in 'a'..'z' -> letter
+                    c in '0'..'9' -> digit
+                    else -> other
+                }
+                Tree(type, emptyList(), emptyList(), 1)
+            },
+            text.indices.toList(),
+            text.length
+        )
+    }
+
+    /**
+     * Level 2: `Mid` over the whole range with an `Inner` child covering everything but the
+     * two-character brackets at each end.
+     */
+    private val midBase: Parser = parserOf { text ->
+        Tree(
+            midTop,
+            listOf(Tree(innerNode, emptyList(), emptyList(), text.length - 4)),
+            listOf(2),
+            text.length
+        )
+    }
+
+    private val midParser: Parser = object : Parser() {
+        override fun createParse(
+            input: Input,
+            fragments: List<TreeFragment>,
+            ranges: List<TextRange>
+        ): PartialParse = parseMixed { node, _ ->
+            // Name-based, like `:lang-vue`'s `"Text" -> textMixed`: the nesting decision itself
+            // must not depend on the coordinates under test, so that a failure isolates the
+            // range handed to the inner parser rather than whether nesting happened at all.
+            if (node.name == "Inner") NestedParse(parser = charParser) else null
+        }(midBase.createParse(input, fragments, ranges), input, fragments, ranges)
+    }
+
+    /** Level 1: `Outer` over `[[` ... `]]`, with `midParser` nested on it. */
+    private val outerBase: Parser = parserOf { text ->
+        val start = text.indexOf("[[")
+        val end = text.indexOf("]]") + 2
+        Tree(
+            outerTop,
+            listOf(Tree(outerNode, emptyList(), emptyList(), end - start)),
+            listOf(start),
+            text.length
+        )
+    }
+
+    private val outerParser: Parser = object : Parser() {
+        override fun createParse(
+            input: Input,
+            fragments: List<TreeFragment>,
+            ranges: List<TextRange>
+        ): PartialParse = parseMixed { node, _ ->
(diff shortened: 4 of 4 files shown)


monkopedia-coder:
Two follow-ups on the mutation sweep in the description.

**Surviving mutant is now filed as #351.** `ranges[0].from -> ranges.last().from` is genuinely uncovered rather than equivalent, so it is tracked rather than left in this PR body. The issue records the detail that matters for whoever picks it up: a faithful fixture needs a **multi-range nested parse with a gap between the ranges**, and the synthetic parsers in `MixNestedOffsetTest` do not skip range gaps — so the cheap version (one more case on the existing fixture) passes identically with and without the mutant and would wrongly read as covered.

The second survivor, root `TreeNode` index `0 -> 1`, is **not** filed: it is an equivalent mutant. `index` is read only at `SyntaxTree.kt:636` and `:1034` and both paths are guarded by a non-null `_parent`, which this root does not have. The reasoning stays in the description above.

**On `shiftedReadIsNotDetectableFromPositionsAlone` passing both with and without the fix** — the rationale is on the test itself, not only in this description, so a future cleanup has it in front of them:

```kotlin
// A negative control, and the standing argument against a weaker bar: this assertion
// passes both WITH and WITHOUT the #341 fix (verified — it was the one test in this
// class that stayed green against the defect). The mounted subtree keeps its length and
// re-bases onto the correct document range, so positions and node count come out
// identical to a correct parse and only the node types differ. Angular's
// `(click)="go()"` was the real-world instance: it lexed `clic` into a bare
// four-character `VariableName` sitting on the right four-character range, with no error
// node anywhere in the tree. Any future test here must assert node names, not spans.
```

**CI note.** `check` carries `ktlintCheck`/`spotlessCheck` since #319; both were also run locally before pushing. The emulator job passed at 8m48s — a second independent green on the zero-baseline established when #335 merged.

monkopedia-reviewer (approved):
Approved.

**Upstream verified independently.** `@lezer/common@1.5.2` `src/mix.ts:149` seeds `startInner` with `new TreeCursor(new TreeNode(this.baseTree!, this.ranges[0].from, 0, null), IterMode.IncludeAnonymous | IterMode.IgnoreMounts)`; upstream `TreeNode` is `(_tree, from, index, parent)` and the Kotlin is `(_tree, treeFrom, _parent, index)`, so `TreeNode(baseTree!!, ranges[0].from, null, 0)` is parent=null / index=0 — semantically identical. `ranges[0].from` is confirmed correct at the source: `lezer-lr/.../Parse.kt:586` builds the tree with `start = ranges[0].from`. The secondary `Tree` constructor delegates to `TreeNode(root, 0, null, 0)`, which reproduces the old field seeding exactly (`from = 0`, `to = 0 + length`, `type = root.type`), and only `Tree.cursor`/`Tree.cursorAt` use it. `StructureCursor` still uses `root.cursor(...)` with its own `offset` field, matching upstream `mix.ts:293-302` — no drift introduced elsewhere.

**Red-first, my own run** (`origin/main` @ `7d1555f2` + this test file, minus the one test that references the new constructor and cannot compile there):

```
tests=5 failures=3
FAIL twoLevelsDeepIsOffsetInvariant
  same fragment at offset 1 produced a different tree shape than at offset 0
  expected:<Doc Mid Chars [Letter Letter Letter Digit] Digit Digit>
   but was:<Doc Mid Chars [Other Letter Letter Letter] Digit Digit>
FAIL twoLevelsDeepReadsTheDocumentAtTheRightOffset
FAIL offsetInvariantForAllDigitContent
PASS oneLevelOfNestingIsUnaffected
PASS shiftedReadIsNotDetectableFromPositionsAlone
```

Both documented controls behave exactly as the file claims they do.

**The offset invariance is a real property, not a fixture.** It holds across five offsets and two different bodies, and `twoLevelsDeepReadsTheDocumentAtTheRightOffset` additionally pins absolute trees at two offsets, so a consistently-but-wrongly-shifted implementation cannot pass. Confirmed against differently-wrong implementations: seeding `ranges[0].from + 1` is killed by 2 tests, `ranges[0].to` by 5, and a broken `Tree`-origin by 9 (including 5 pre-existing `TreeTest` cases).

**No-error-node case present, and confirmed on the real thing.** The whole fixture is error-node-free by construction — `charParser` types every character, so a shifted read changes node *types* only. `shiftedReadIsNotDetectableFromPositionsAlone` pins that positions alone do not catch this. The real-world instance reproduces: on `origin/main`, Angular `<button (click)="go()">` yields a bare `VariableName[17,21]` with no error node anywhere; after the fix, `CallExpression(VariableName[17,19], ArgList[19,21])`.

**Vue/Angular, before -> after (scratch probes, both worktrees):**

| case | before (`origin/main`) | after |
| --- | --- | --- |
| ng `` | `VariableName[12,14]="ur"` + `⚠` + `ArrayExpression[14,15]="l"` (lexed `rc]`) | `VariableName[12,15]="url"` |
| ng `(click)="go()"` | bare `VariableName[17,21]`, no error node | `CallExpression(VariableName, ArgList)` |
| ng `<p>{{ title }}</p>` | `⚠ > AssignmentExpression ObjectPattern { ⚠ { PatternProperty PropertyName ⚠ ⚠` | `VariableName[6,11]="title"` |
| vue `<p>{{ msg }}</p>` | same `ObjectPattern` garbage | `VariableName[6,9]="msg"` |
| vue `:class="foo"` | `VariableName/Equals/String` over `f`,`o`,`o`; a *different* wrong tree at a different offset | `VariableName` over `foo`, identical at both offsets |

**Regression surface.** Every module using `parseMixed` (positive-controlled grep): `lezer-common`, `lang-html`, `lang-markdown`, `lang-vue`, `lang-angular`, `lang-jinja`, `lang-liquid`; plus `lezer-lr`, `lang-javascript`, `language`, `lang-php`. JVM before/after, cross-checked against `BUILD SUCCESSFUL` and against a `<failure>`-element count independent of the `failures` attribute:

| module | before | after |
| --- | --- | --- |
| lezer-common | 83 / 0 fail | 89 / 0 fail |
| lezer-lr | 210 / 0 | 210 / 0 |
| lang-html | 32 / 0 | 32 / 0 |
| lang-javascript | 39 / 0 | 39 / 0 |
| language | 91 / 0 | 91 / 0 |
| lang-markdown | 63 / 0 | 63 / 0 |
| lang-php | 29 / 0 | 29 / 0 |
| **total** | **547 / 0** | **553 / 0** |

The +6 is by name the six new `MixNestedOffsetTest` cases; the before/after test-name sets are otherwise byte-identical per module. No existing test changed status. `lang-vue`/`lang-angular`/`lang-jinja`/`lang-liquid` have zero test sources (#342/#343/#337/#339), so they contribute 0 both sides — for those I probed by hand instead: vue/angular fixed as above, and jinja/liquid throw *identically* before and after (the separate corrupt-table defect), so nothing regressed. `:lezer-common:wasmJsBrowserTest` also green locally: 89 tests, 0 failures, all six new cases present in the XML.

**Mutation, line and condition granularity.** The changed lines contain no boolean `&&`/`||`; the one compound expression is the bitwise flag pair, and I tested each half separately. Every mutation was preceded by an exact match count — the harness aborted on the `IterMode` pair when it found 2 sites, and I re-anchored on the surrounding line rather than contaminating the unrelated `StructureCursor` site.

Killed: revert-to-0 (3 tests), `ranges[0].to` (5), `+1` (2), `Tree`-ctor origin 0->1 (9), `from = root.from -> 0` (1), `to = root.to -> root.tree.length` (1). The last two are killed *only* by `cursorOverAnOffsetRootReportsDocumentCoordinatesBeforeMoving`, so that test earns its place.

Survived: `type = root.type -> root.tree.type` and `index 0 -> -1` are equivalent mutants (`TreeNode.type` is defined as `_tree.type`; every read of `index` is guarded by a `_parent == null` check that fires first for a parentless root). `ranges[0].from -> ranges.last().from` and dropping either `IterMode` flag are genuine coverage gaps — filed as #353.

**Scope and API.** Exactly four files, all in `:lezer-common`. The `SyntaxTree.kt` change is only the constructor refactor: grepping the diff for `enterUnfinishedNodesBefore|resolveStack|CachedNode|bracketed|ENTER_BRACKETED|resolveNode|stackIterator` returns nothing (positive control: 15 hits in the file itself), so none of #340/#332/#329/#330/#331 is partly implemented. `closingIssuesReferences` is `[341]` only — #342/#343 stay open with their zero-test-sources problem. `:lezer-common:apiCheck` green, no `.api`/`.klib.api` in the diff and none dirtied by the run. `ktlintCheck`/`spotlessCheck` green. `changelog.d/341.fixed.md` validates; `CHANGELOG.md` untouched. No provenance annotations in the diff.

Separately filed #352: `TreeCursor` stores `mode` verbatim where upstream does `mode & ~IterMode.EnterBracketed`, so `ENTER_BRACKETED` outlives the call it was meant for. Pre-existing and distinct from #331 (which is the `mounted.bracketed` half); noticed during the upstream comparison, out of scope here.

Required checks green on the live head `957a87f8`: `check`, `check-macos`, `check-ios` (plus `lint` and `Instrumented tests (emulator)` — the latter clean, consistent with the post-#335 baseline).

Issue #346, opened 2026-09-07
lang-yaml's indentation ContextTracker declares strict = false, where @lezer/yaml leaves it at the default true


monkopedia-coder:
Found while fixing #323, which was the same bug in `:lang-python`. Not fixed there: no test in this
repo currently distinguishes it, and folding a second module's tracker into that PR would have
mixed an unmeasured change into a measured one.

`@lezer/lr`'s `ContextTracker` defaults `strict` to true (`dist/index.js`:
`this.strict = spec.strict !== false`). `@lezer/yaml@1.0.4`'s `indentation` tracker does not pass
`strict`, so upstream runs it strict:

```js
const indentation = new ContextTracker({
  start: Context.top,
  reduce(context, term) { … },
  shift(context, term, stack, input) { … },
  hash(context) { return context.hash }
});
```

The port (`lang-yaml/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/yaml/YamlParser.kt`) ends
that same tracker with:

```kotlin
    hash = { context -> context.hash },
    strict = false
) as ContextTracker<Any?>
```

**Consequence.** With `strict` false, `StackContext.hash` is pinned to 0, so
`TokenCache.getActions`'s cache-validity test (`token.context != context`) never fires when the
indent context changes, and a token computed under one indentation context can be reused under
another. `Stack.close`/`emitContext` also stop writing `NodeProp.contextHash`, and
`Parse.useCachedResult`'s context check (added in #333) becomes inert for this grammar, so a
fragment built under a different context can be reused.

In `:lang-python` the same flag caused 11 of 40 Python 3.14 standard-library files to parse
differently from `@lezer/python` — stale `dedent` tokens swallowing `elif`/`else` — and restoring
the default took that to 0. `:lang-yaml`'s 20 tests pass either way, so the effect here is
unmeasured, not known-absent.

**Verification path.** Generate a YAML corpus, parse it with `@lezer/yaml@1.0.4` +
`@lezer/lr@1.4.10`, and compare against the port before and after removing `strict = false`; assert
tree shapes against the real upstream run, not merely that parsing succeeds. Nested block
sequences/mappings at changing indentation are where the token cache is most likely to go stale.
Check the port's table is transcribed from 1.0.4 first (compare the `states` string), as the
`:lang-python` work had to.

While auditing: `lang-javascript`, `lang-html` and `lang-xml` declare `strict = false` and their
[...]

Issue #338, opened 2026-09-07
Port's LR error recovery diverges from upstream on malformed input: 15 stable tree-shape differences, found by two independent differential harnesses


monkopedia-coder:
Two independent differential harnesses, built separately during #321/#333, both found the port's
**error recovery diverges from upstream `@lezer/lr@1.4.10` on malformed input**. Neither run was
looking for this — it showed up as the constant background in both, and both correctly set it aside
as pre-existing rather than bundling it. Filing it so it stops being background.

| harness | corpus | divergences |
|---|---|---|
| #333 author | 1,336 incremental cases (600 well-formed + 600 mangled at `bufferLength` 1, 200 >4.6 KB at default) | **3 before, same 3 after** |
| #333 reviewer | 1,400 incremental cases, independently built | **15 before, same 15 after**, report files byte-identical |

Both describe them the same way: **`⚠`-shape differences on mangled input** — the port and upstream
build *different error-node shapes* for the same malformed document. Well-formed input agrees
throughout; the 828 head-to-head fragment-reuse cases agree; the divergence is specifically in what
the tree looks like after recovery.

Two harnesses disagreeing on the count (3 vs 15) is expected — different corpora — and is itself
useful: it says the surface is **corpus-sensitive**, so the real number is unknown and 15 is a floor,
not a total.

## Why this is worth tracking rather than shrugging at

**In an editor, malformed input is the normal case, not the edge case.** Every keystroke of a
half-typed expression is a malformed document. Error-node shape is not cosmetic — it is what
`enterUnfinishedNodesBefore` walks (see #253), what indentation resolves against, and what
syntax highlighting paints. A recovery tree that differs from upstream's means indentation and
highlighting can differ from CodeMirror 6's for exactly the inputs a user spends most of their time
holding.

That does not make it a *bug* — upstream's recovery shape is not self-evidently the only correct
one — but it is a **known, measured fidelity gap in the hottest path of a port whose entire value
proposition is CM6 parity**, and it currently exists only inside two agents' scratch directories.

## What would close this

1. Characterise the divergences: are they one mechanism or several? Both harnesses report them as
   stable and reproducible, so they should classify cleanly.
2. For each class, decide whether the port or upstream is right, and either fix the port or record
   why the divergence is acceptable.
3. Add whichever cases are genuinely wrong to `:lezer-lr`'s suite, so the count is guarded rather
   than rediscovered by the next differential harness.

[...]

Issue #334, opened 2026-09-07
LR Parse builds a FragmentCursor on documents where @lezer/lr would not (missing bufferLength*4 threshold)


monkopedia-coder:
Found while fixing #321 (the four dropped `useCachedResult` conditions); deliberately left out of
that PR's scope because it changes *when* fragment reuse is attempted rather than what the reuse
conditions are, and folding it in would have masked the four fixes.

Upstream `@lezer/lr@1.4.10` builds the incremental-parse cursor only when the document is
substantially longer than the buffer length (`dist/index.js:1278`):

```js
this.fragments = fragments.length && this.stream.end - from > parser.bufferLength * 4
    ? new FragmentCursor(fragments, parser.nodeSet) : null;
```

The port (`lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt`, `init`)
drops the length threshold entirely:

```kotlin
this.fragments =
    if (fragments.isNotEmpty() &&
        parser.nodeSet.types.isNotEmpty()
    ) {
        FragmentCursor(fragments, parser.nodeSet)
    } else {
        null
    }
```

`parser.nodeSet.types.isNotEmpty()` is not upstream's condition and is true for every deserialized
parser, so the port attempts fragment reuse on **every** incremental parse with a non-empty
fragment list. With the default `bufferLength` of 1024 upstream skips the cursor for any document
under ~4 KB; the port builds one.

**Consequence.** Reuse is attempted — and can succeed — on short documents where upstream always
does a fresh parse. That is a real behavioural divergence for any consumer doing incremental
parsing of small documents, and it also means a fragment carrying a stale or foreign node reaches
`useCachedResult` in the port on inputs that never reach it upstream.

Not blocking today: no diff was observed on a 1,336-case JSON differential corpus against
`@lezer/lr@1.4.10` (that corpus was run with matched settings on both sides precisely so this
divergence would not mask the #321 work), so the effect is on *when* reuse is attempted, not
observed as a wrong tree yet.

**Verification path.** An incremental parse of a document under `bufferLength * 4` characters
where reuse changes the resulting tree: upstream returns the fresh-parse shape, the port returns
the reused shape. Assert the tree shape against a real `@lezer/lr@1.4.10` run, not merely whether
reuse happened. Note that restoring the threshold will make several existing `:lezer-lr`
incremental tests stop exercising the reuse path at all, since their documents are short — those
[...]

PR #347, opened 2026-09-07
Run advanceStack's steps in upstream's order (#323)

### LINKED ISSUE #323: LR advanceStack diverges from upstream: getActions before default-reduce, and split-stack routing

monkopedia-coder:
Fixes #323

## What was compared against

`@lezer/lr@1.4.10` (`npm pack`ed, `dist/index.js`), specifically **`LRParse.advanceStack`,
`dist/index.js:1406-1457`**. Upstream's order is:

```js
advanceStack(stack, stacks, split) {
    let start = stack.pos, { parser } = this;
    if (this.stoppedAt != null && start > this.stoppedAt) return stack.forceReduce() ? stack : null;
    if (this.fragments) { … reuse … }                                       // 1411-1428
    let defaultReduce = parser.stateSlot(stack.state, 4);
    if (defaultReduce > 0) { stack.reduce(defaultReduce); return true; }
    if (stack.stack.length >= 8400) { while (stack.stack.length > 6000 && stack.forceReduce()) {} }
    let actions = this.tokens.getActions(stack);
    for (let i = 0; i < actions.length;) {
        let action = actions[i++], term = actions[i++], end = actions[i++];
        let last = i == actions.length || !split;
        let localStack = last ? stack : stack.split();
        let main = this.tokens.mainToken;
        localStack.apply(action, term, main ? main.start : localStack.pos, end);
        if (last) return true;
        else if (localStack.pos > start) stacks.push(localStack);
        else split.push(localStack);
    }
    return false;
}
```

The port ran `tokens.getActions(stack)` **first**, then reuse (gated on `main != null`), then the
default reduce, and then a hand-rolled action loop. Also compared: `@lezer/python@1.1.18`, which is
the version `:lang-python`'s table is transcribed from (its `states` string is byte-identical to
1.1.18's and to no other release).

### 1. `getActions` before the default-reduce branch

- **Upstream** (`1439`): `getActions` runs only after `if (defaultReduce > 0) { … return true }`.
- **Port**: `getActions` ran unconditionally at the top.
- **What it changes**: `getActions` calls `updateCachedToken`, which can call
  `stack.setLookAhead(...)` (when a token's scan reaches more than `Lookahead.MARGIN` past its
  end). `setLookAhead` writes a `-4` record into the stack buffer *at the current buffer position*,
  and `Tree.build` turns that into `NodeProp.lookAhead`. The recorded *values* are the same either
  way — in the fixture below, 41 at position 1 and 49 at position 9 — but recording before the
  default reduce puts the record **inside** the node that reduce is about to build, where upstream
  puts it after. `Tree.build` reads the buffer backwards and hands a node the last record it
  passed, so the port's node picks up the next position's 49 instead of its own 41 and claims 40
  characters of lookahead past its end where upstream says 32. `TreeFragment.applyChanges` consults
  exactly that prop to decide how far an edit invalidates reuse.

  This is reachable in real grammars: it needs a default-reduce state with a non-zero
  `TOKENIZER_MASK`, which `@lezer/python` has 217 of, `@lezer/javascript` 232, `@lezer/css` 93 and
  `@lezer/html` 39 (the mask bits set on those states are the external tokenizers). JSON — the only
  fixture this module had — has **zero**, which is why nothing here saw it.

  New fixture `TestWideScanParser.kt`: a grammar with an external tokenizer that peeks 40 characters
  ahead and consumes 8, over a chain of unit productions. On `"x" + "w"*48`:

  ```
  @lezer/lr 1.4.10:  …,Item[la=32](Wrap[la=32](Inner[la=32](Wide[la=32]))),…
  port before:       …,Item[la=40](Wrap[la=40](Inner[la=40](Wide[la=32]))),…
  ```

### 2. The `main != null` gate on fragment reuse

- **Upstream**: the reuse block is gated on `this.fragments` alone. Upstream has no main token at
  that point *because* it has not tokenised yet — which is precisely why removing this gate
  requires divergence 1's fix. It is one mechanism seen from two sides, so it is fixed here rather
  than left for a follow-up; #333 recorded it as belonging to this issue. **#334 is unaffected** —
  it is about the missing `stream.end - from > bufferLength * 4` threshold on building the
  `FragmentCursor`, which this PR does not touch, and its text never mentions this gate.
- **What it changes**: at a position no tokenizer can read, `getActions` leaves `mainToken` null, so
  the port skipped reuse and fell into error recovery where upstream reuses the cached node.

  ```
  doc "@@@@@" with a fragment holding a String node over 0..5
  @lezer/lr 1.4.10:  JsonText(String)
  port before:       JsonText(⚠)
  ```

- The other port-only gate, **`!parser.hasWrappers()`**, also has no upstream counterpart (upstream
  wraps the parse *around* an inner `Parse` that still reuses fragments). **Left alone and
  unchanged** — it is not part of this issue and nothing here measures it.

### 3. The action loop: which stack keeps the original, and where forks go

- **Upstream**: applies the **last** action to the stack it was given and forks the earlier ones;
  a fork that moved `pos` goes to `stacks` (the caller's `newStacks`), one that did not goes to
  `split` (the list being iterated).
- **Port**: applied the **first** action to the given stack, forked the rest, and appended *every*
  fork to `split`.
- **What it changes**: the same set of stacks reaches `newStacks`, in the opposite order.
  `advance`'s pruning removes the *earlier* of two stacks it rates equal, and `recovering`'s
  `sortByDescending { score }` is stable, so the order decides which reading of an ambiguous parse
  survives. New fixture `TestAmbiguousParser.kt` (`X { Num ~amb }` / `Y { Num ~amb Num }`):

  ```
  input "1 1 1"
  @lezer/lr 1.4.10:  Program(X(Num),Y(Num,Num))
  port before:       Program(Y(Num,Num),X(Num))
  ```

**Not fixed here:** upstream's `Rec.CutDepth` / `Rec.CutTo` guard between the default reduce and
`getActions` is still absent — that is #327, deliberately untouched.

## The `:lang-python` change, and why it cannot be separated

Fixing divergence 1 turned **8 `:lang-python` tests red**. It is not a regression in the fix; it
uncovered a second, independent porting bug that the wrong order was masking.

`@lezer/python`'s `trackIndent` does not pass `strict`, and `@lezer/lr`'s `ContextTracker`
defaults it to `true` (`this.strict = spec.strict !== false`). The port declared
`strict = false`. With `strict` false the stack's context hash is pinned to 0, so
`TokenCache.getActions`'s cache-validity test (`token.context != context`) never fires on an
indent-context change and a stale `dedent` is reused where a keyword should be read. Upstream's
order re-tokenised at a state the port reached only after the extra early tokenisation had
refreshed the cache, so the two bugs cancelled.

Proof of the causal chain, run in JS: `@lezer/lr@1.4.10` + `@lezer/python@1.1.18` **patched to
`strict: false`** produces, for `"if a: b()\n\nif 1 + 3:\n  pass\nelif 55 < 2:\n  pass\nelse:\n  pass"`:

```
Script(IfStatement(…),⚠(elif),AssignStatement(BinaryExpression(Number,CompareOp,Number),TypeDef(:,⚠)),PassStatement(pass),⚠(else),⚠(:),PassStatement(pass))
```

— which is exactly what this port produced with the `advanceStack` fix and `strict = false`.
Unpatched upstream produces the correct `elif`/`else` tree.

Measured on 40 files from the Python 3.14 standard library (`/usr/lib/python3.14/*.py`, ≤25 KB, no
tabs), against `@lezer/python@1.1.18` + `@lezer/lr@1.4.10` with a serializer that filters the same
anonymous nodes `:lang-python`'s `treeToString` does:

| | `strict = false` (port today) | `strict` default (upstream) |
| --- | --- | --- |
| `advanceStack` as on `main` | **11 / 40 differ** | 0 / 40 |
| `advanceStack` fixed | **36 / 40 differ** | **0 / 40** |

So `main` mis-parses 11 of 40 real Python files today — dedented bodies, `elif`/`else` branches and
nested functions lose their structure — and the pair takes that to zero. The dependency runs one
way: the `advanceStack` fix must not ship without the tracker fix (36/40 wrong if it does), while
the tracker fix stands on its own (0/40 either side of it). It is bundled here because this PR is
what makes it mandatory, and because splitting it would put this branch's CI red until the other
landed.

`PythonParserTest.testNestedQuoteTypes` changes with the tracker fix (it is green with the tracker
fix alone, on `main`'s `advanceStack`). It carried a `// TODO: Parser produces a minor error node in
format replacement with nested quotes` and asserted that stray `⚠`; `@lezer/python@1.1.18` produces
`FormatString(FormatReplacement({,String,}))` — no error node — so the expectation and the TODO are
both retired. That is the only expectation this PR edits.

Every `strict` flag in the repo was checked against its upstream grammar: `lang-javascript`
(`false`) and `lang-html` (`false`) match, `lang-go` and `lang-sass` (unset) match, `lang-xml`
(`false`) matches. **`lang-yaml` does not** — `@lezer/yaml@1.0.4` leaves `strict` unset and the port
declares `strict = false`. Its 20 tests pass either way and nothing here measures it, so it is filed
separately rather than bundled.

## Red-first

The three new fixtures and `AdvanceStackTest` alone, on unmodified `origin/main` production code —
**217 tests, 3 failed**:

```
AdvanceStackTest[jvm] > ambiguitySplitsResolveLikeUpstream[jvm] FAILED
    org.junit.ComparisonFailure: Ambiguous parse of three numbers expected:<Program([X(Num),Y(Num,]Num))> but was:<Program([Y(Num,Num),X(]Num))>
AdvanceStackTest[jvm] > defaultReduceStateDoesNotRecordLookahead[jvm] FAILED
    org.junit.ComparisonFailure: Recorded lookahead expected:<...(Inner(X))),Item[la=[32](Wrap[la=32](Inner[la=32]](Wide[la=32]))),Ite...> but was:<...(Inner(X))),Item[la=[40](Wrap[la=40](Inner[la=40]](Wide[la=32]))),Ite...>
AdvanceStackTest[jvm] > cachedNodeIsReusedWhereNoTokenMatches[jvm] FAILED
    org.junit.ComparisonFailure: Reuse at a position with no matching token expected:<JsonText([String])> but was:<JsonText([⚠])>
217 tests completed, 3 failed
```

The four controls passed in that same run, and they are what stop the three above from being
vacuous:

- `wideScanGrammarRecordsNoLookaheadOnAShortDocument` — the same grammar and tokenizer on a document
  too short for any scan to reach `Lookahead.MARGIN`; no lookahead is recorded on either side, so
  the fixture is not simply always printing `[la=…]`.
- `ambiguousGrammarParsesUnambiguousInput` — `"1"` and `"1 1"` agree on either routing, guarding the
  transcribed table itself.
- `untokenizableDocumentIsAnErrorWithoutAFragment` — `"@@@@@"` really is unparseable, so the reuse
  expectation is not the grammar quietly accepting `"@"`.
- `cachedNodeIsReusedWhereATokenMatches` — the same fragment over `"12345"` *is* reused on both
  sides, so the refusal above is a decision and not a dead cursor.

Each "port before" value is byte-identical to what a copy of upstream `dist/index.js`, patched to
the port's ordering and gate, produces for the same case — an independent check that the JS model
of the port used throughout this PR is faithful.

## Mutation — four independent reverts

Each applied alone to the fixed `advanceStack`, `:lezer-lr:jvmTest`, **217 tests** each time:

| revert | failing test(s) |
| --- | --- |
| `getActions` back above the reuse block (no gate) | `defaultReduceStateDoesNotRecordLookahead` (1 failed) |
| the whole pre-fix ordering: `getActions` first **and** the `main != null` gate | `defaultReduceStateDoesNotRecordLookahead`, `cachedNodeIsReusedWhereNoTokenMatches` (2 failed) |
| every fork routed to `split` | `ambiguitySplitsResolveLikeUpstream`, `ambiguousGrammarParsesUnambiguousInput` (2 failed) |
| first action kept on `stack`, upstream routing retained | `ambiguitySplitsResolveLikeUpstream` (1 failed) |

No revert left the suite green. The gate cannot be reverted *without* the ordering — that is the
entanglement — so its guard is the delta between rows 1 and 2. Row 3 also reddens the control,
which is honest: that routing changes `"1 1"` too.

A fifth revert, for the bundled tracker change: putting `strict = false` back, with `advanceStack`
fixed, turns **8 `:lang-python` tests red** (5 `PythonIndentTest`, 3 `PythonParserTest`) — that is
how the second bug was found. Reverting the `advanceStack` ordering while keeping the tracker fix
leaves `:lang-python` green at 41/41, which is the asymmetry described above.

### Which lines survive deletion untested

Two, both reported rather than removed:

- `split == null || newStacks == null` in the `last` condition (upstream's `!split`). Deleting the
  whole short-circuit, or just the redundant `newStacks == null` half, leaves **217 tests green**.
  It is upstream's own guard for the `advanceFully(stack, null, null)` caller, and reaching it needs
  an ambiguity during error recovery, which no test in this repo constructs.
- The `!parser.hasWrappers()` gate, which predates this PR and is deliberately untouched.

Upstream's `main ? main.start : localStack.pos` is *not* transcribed; the port passes `start`. They
are provably identical — `InputStream.reset(pos, token)` sets `token.start = pos`, and `getActions`
only reuses a cached token when `token.start == stack.pos` — so writing the elvis would add a branch
no test could ever separate.

## Observable vs fidelity, and a fatal probe

- **Divergence 1 is observable** on a grammar whose default-reduce states tokenise (219 → 0 diffs
  below), and **latent on single-language corpora**: instrumenting upstream's own `advanceStack` in
  JS, the port's order makes **134 lookahead records at a different buffer position** across 119
  Python standard-library files and **0 tree differences** result. On JSON it cannot fire at all —
  zero of JSON's 12 default-reduce states have a tokenizer mask.
- **Divergence 3's fork branch is unreachable in the rest of this module.** Made **fatal** (the
  branch throwing when taken) and the whole suite re-run: **220 tests, 3 failed**, and the three are
  exactly the ambiguous-grammar tests — the negative control that proves the probe fires. Every
  other test, including the 1,400-case JSON incremental differential and the 1,190-case wide-scan
  differential, passed, i.e. never forked. That is why this was invisible until an ambiguous fixture
  existed.
- **Divergence 2 is observable** (`JsonText(String)` vs `JsonText(⚠)`) and changes nothing on the
  JSON corpus, whose diff report is byte-identical before and after.

## Differential against real upstream

Every expectation is produced by running `@lezer/lr@1.4.10` itself on the same table.

| corpus | cases | diffs before | diffs after |
| --- | --- | --- | --- |
| JSON incremental (#333's, `TreeFragment.applyChanges` two-generation) | 1,400 | 15 | **15** (byte-identical report) |
| ambiguous grammar, well-formed input | 488 | 401 | **0** |
| ambiguous grammar, malformed input | 365 | 31 | **7** |
| wide-scan grammar (`bufferLength` 1 and 1024) | 1,190 | 219 | **0** |
| Python 3.14 stdlib files vs `@lezer/python@1.1.18` | 40 | 11 | **0** |

The JSON count of 15 is #338's known pre-existing error-recovery divergence set, unchanged
case-for-case. The 7 residual ambiguous diffs are the same class — every one is on input containing
characters the grammar cannot tokenise (`a`, `!`); **no well-formed case differs**. Net on that
corpus the recovery diffs also drop, 31 → 7.

## Verification

- `:lezer-lr:jvmTest` — **210 before, 217 after**, 0 failures. The seven added are exactly the seven
  `AdvanceStackTest` cases named above; no existing test changed, was renamed or removed (verified
  by diffing the full testcase-name lists from `build/test-results/jvmTest/*.xml`, not by the delta).
- `:lezer-lr:wasmJsBrowserTest` 217 tests / `:lang-python:wasmJsBrowserTest` 41 tests, 0 failures
  (13 result XMLs read).
- Dependent modules from `build/test-results/jvmTest/*.xml`, measured from the merge base
  `3c663304` on both sides: **834 before / 841 after**, +7 all in `:lezer-lr` by name, 0 failures.
  `:lezer-common` 83, `:language` 91, `lang-cpp` 30, `lang-css` 16, `lang-go` 46, `lang-html` 32,
  `lang-java` 33, `lang-javascript` 39, `lang-json` 31, `lang-markdown` 63, `lang-php` 29,
  `lang-python` 41, `lang-rust` 32, `lang-sass` 25, `lang-xml` 13, `lang-yaml` 20. `lang-python`
  stays at 41 — one expectation edited, none added or dropped.
- `:lezer-lr:apiCheck` and `:lang-python:apiCheck` green; **no `.api` or `.klib.api` file moved**
  (`advanceStack` and `trackIndent` are private).
- `:lezer-lr:ktlintCheck`, `:lezer-lr:spotlessCheck`, `:lang-python:ktlintCheck`,
  `:lang-python:spotlessCheck` green (all four gate `check` per #319).
- `changelog.py check --base-ref origin/main` — 19 fragments valid, `CHANGELOG.md` untouched.
- All instrumentation removed before this PR; `println`/`PROBE` grep over `lezer-lr/src` and
  `lang-python/src` is empty, positive-controlled against files where the same pattern does match.

## Seen, not bundled

- **#327** (`Rec.CutDepth` / `Rec.CutTo` stack-depth cut, between the default reduce and
  `getActions`) is untouched.
- **#334** (the missing `bufferLength * 4` threshold on building the `FragmentCursor`) is untouched
  and needs no amendment: it is about `Parse.init`, not about the `main != null` call-site gate this
  PR removes.
- **`lang-yaml`'s `strict = false`** contradicts `@lezer/yaml@1.0.4`, exactly as `lang-python`'s did.
  Filed as #346; no test in this repo currently distinguishes it, so it is not touched here.
- `lezer-common/.../SyntaxTree.kt` is untouched.

### DIFF
--- a/changelog.d/323.fixed.md
+++ b/changelog.d/323.fixed.md
@@ -1,0 +1,19 @@
+- `Parse.advanceStack` now runs its steps in `@lezer/lr` 1.4.10's order and routes ambiguity forks
+  the way upstream does (#323). It called `tokens.getActions(stack)` first, unconditionally, so a
+  state with a default reduce tokenised where upstream does not — recording a wider
+  `NodeProp.lookAhead` on the resulting nodes, which is what `TreeFragment.applyChanges` consults
+  to decide how far an edit invalidates reuse. That early tokenisation was also the only reason
+  the fragment-reuse block was gated on having a main token, so reuse was refused at any position
+  the grammar cannot tokenise, where upstream reuses the cached node. And the forks of an
+  ambiguous parse were all appended to the list being iterated, where upstream sends the ones that
+  consumed input straight to the caller's stack list and keeps the original stack on the *last*
+  action rather than the first; the resulting order decides which reading survives `advance`'s
+  pruning. On a 488-case corpus of ambiguous parses checked against `@lezer/lr` 1.4.10 the port
+  disagreed on 401 and now disagrees on none.
+- `:lang-python`'s indentation `ContextTracker` is no longer declared `strict = false` (#323).
+  `@lezer/python` leaves `strict` at its default of true; with it false the token cache never
+  invalidated on an indent-context change, so a stale `dedent` was reused where a keyword should
+  have been read. It was masked by the `advanceStack` order above: on 40 files from the Python 3.14
+  standard library the port disagreed with `@lezer/python` 1.1.18 on 11 before this change (36 with
+  the `advanceStack` fix alone) and on none after. `elif`/`else` branches, dedented bodies and
+  nested functions were the visible casualties.
--- a/lang-python/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/python/PythonParser.kt
+++ b/lang-python/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/python/PythonParser.kt
@@ -191,8 +191,7 @@ private val trackIndent = ContextTracker(
             context
         }
     },
-    hash = { context -> context.hash },
-    strict = false
+    hash = { context -> context.hash }
 ) as ContextTracker<Any?>
 
 // External tokenizers
--- a/lang-python/src/commonTest/kotlin/com/monkopedia/kodemirror/lang/python/PythonParserTest.kt
+++ b/lang-python/src/commonTest/kotlin/com/monkopedia/kodemirror/lang/python/PythonParserTest.kt
@@ -56,10 +56,9 @@ class PythonParserTest {
         parse("r\"foo\\\"\" + r'\\\\'")
     )
 
-    // TODO: Parser produces a minor error node in format replacement with nested quotes
     @Test
     fun testNestedQuoteTypes() = assertEquals(
-        "Script(ExpressionStatement(FormatString(FormatReplacement(String),\u26A0)))",
+        "Script(ExpressionStatement(FormatString(FormatReplacement(String))))",
         parse("f\"a{'b'}c\"")
     )
 
--- a/lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt
+++ b/lezer-lr/src/commonMain/kotlin/com/monkopedia/kodemirror/lezer/lr/Parse.kt
@@ -407,15 +407,18 @@ internal class Parse(
             // it will go rather than consuming more input.
             return stack.forceReduce()
         }
-        val actions = tokens.getActions(stack)
-        val main = tokens.mainToken
 
-        // Try to use a cached tree fragment
-        if (main != null && !parser.hasWrappers()) {
+        // Try to use a cached tree fragment. Upstream gates this on nothing but
+        // the presence of a fragment cursor -- it has not tokenised yet at this
+        // point, which is why reuse is still attempted at a position where no
+        // tokenizer produces a token.
+        if (!parser.hasWrappers()) {
             if (useCachedResult(stack)) return true
         }
 
-        // Default reduce (no token needed)
+        // Default reduce (no token needed). Upstream returns here without ever
+        // calling `getActions`, so a default-reduce state does not tokenise --
+        // and does not extend the recorded lookahead -- at this position.
         val defaultReduce =
             parser.stateSlot(stack.state, ParseState.DEFAULT_REDUCE)
         if (defaultReduce > 0) {
@@ -423,32 +426,30 @@ internal class Parse(
             return true
         }
 
-        if (actions.isEmpty()) return false
-
-        // Use the first action; split stacks for any alternatives
-        val action0 = actions[0]
-        val token0 = if (actions.size >= 2) actions[1] else 0
-        val end0 = if (actions.size >= 3) actions[2] else start
-
-        if (actions.size > 3) {
-            var ai = 3
-            while (ai < actions.size) {
-                val action = actions[ai]
-                val token = actions[ai + 1]
-                val end = actions[ai + 2]
-                val s = stack.split()
-                s.apply(action, token, start, end)
-                if (split != null) {
-                    split.add(s)
-                } else if (newStacks != null) {
-                    pushStackDedup(s, newStacks)
-                }
-                ai += 3
+        val actions = tokens.getActions(stack)
+        var i = 0
+        while (i < actions.size) {
+            val action = actions[i]
+            val term = actions[i + 1]
+            val end = actions[i + 2]
+            i += 3
+            if (i == actions.size || split == null || newStacks == null) {
+                stack.apply(action, term, start, end)
+                return true
+            }
+            // An ambiguity: fork the stack and take this action on the copy.
+            val localStack = stack.split()
+            localStack.apply(action, term, start, end)
+            if (localStack.pos > start) {
+                // Forked stacks that consumed input belong in the caller's
+                // output list; only the ones still at `start` are re-examined
+                // by the loop that is currently walking `split`.
+                newStacks.add(localStack)
+            } else {
+                split.add(localStack)
             }
         }
-
-        stack.apply(action0, token0, start, end0)
-        return true
+        return false
     }
 
     /**
--- a/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/AdvanceStackTest.kt
+++ b/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/AdvanceStackTest.kt
@@ -51,14 +51,17 @@ class AdvanceStackTest {
      * a lookahead 40 characters past that position while the token itself ends
      * 8 characters past it.
      *
-     * The state at that position has a default reduce, so upstream reduces
-     * without tokenising and only records the lookahead further along, from the
-     * state that genuinely needs a token — where the same scan starts eight
-     * characters later and therefore reaches 32 rather than 40 past the node.
-     * Calling `getActions` first records the wider value, and writes it into
-     * the tree as `NodeProp.lookAhead`: a node that claims to depend on more of
-     * the document than it does, which is what `TreeFragment.applyChanges`
-     * consults to decide how far an edit invalidates reuse.
+     * Both orders record the same lookahead values (41 at position 1, 49 at
+     * position 9); what differs is *where in the stack buffer* the record
+     * lands. `Stack.setLookAhead` writes it at the current buffer position, and
+     * the state at position 1 has a default reduce — so calling `getActions`
+     * first puts the record inside the node that reduce is about to build,
+     * where upstream puts it after that node. `Tree.build` reads the buffer
+     * backwards and gives a node the last record it passed, so the port's
+     * `Item` picks up the *next* position's 49 instead of its own 41: it claims
+     * to depend on 40 characters past its end rather than 32. That number is
+     * what `TreeFragment.applyChanges` consults to decide how far an edit
+     * invalidates reuse.
      *
      * `bufferLength` 1 is load-bearing: `NodeProp.lookAhead` is only kept on
      * nodes that are materialised as `Tree` objects, not on nodes packed into a
--- a/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/AdvanceStackTest.kt
+++ b/lezer-lr/src/commonTest/kotlin/com/monkopedia/kodemirror/lezer/lr/AdvanceStackTest.kt
@@ -1,0 +1,242 @@
+/*
+ * Copyright 2025 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.lezer.lr
+
+import com.monkopedia.kodemirror.lezer.common.NodeProp
+import com.monkopedia.kodemirror.lezer.common.NodeType
+import com.monkopedia.kodemirror.lezer.common.Tree
+import com.monkopedia.kodemirror.lezer.common.TreeFragment
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * Parity with `@lezer/lr` 1.4.10's `LRParse.advanceStack`
+ * (`dist/index.js:1406-1457`) for the order of its steps and the routing of
+ * split stacks.
+ *
+ * Upstream runs, in this order: the `stoppedAt` brake, fragment reuse, the
+ * default reduce (which returns), the cut-depth guard, and only then
+ * `this.tokens.getActions(stack)`. The port used to call `getActions` first,
+ * which both tokenised in states upstream never tokenises in and made the
+ * fragment-reuse block conditional on a main token upstream does not have
+ * there.
+ *
+ * Every expectation below is what `@lezer/lr` 1.4.10 itself produces for the
+ * same parser table and the same input.
+ */
+class AdvanceStackTest {
+
+    // --- 1. `getActions` is called after the default-reduce branch ---
+
+    /**
+     * The document is one `X` followed by six `Wide` tokens. Tokenising at the
+     * position after the first `X` scans [WIDE_SCAN_PEEK] characters ahead but
+     * accepts only [WIDE_SCAN_TOKEN_LENGTH] of them, so the token cache records
+     * a lookahead 40 characters past that position while the token itself ends
+     * 8 characters past it.
+     *
+     * The state at that position has a default reduce, so upstream reduces
+     * without tokenising and only records the lookahead further along, from the
+     * state that genuinely needs a token — where the same scan starts eight
+     * characters later and therefore reaches 32 rather than 40 past the node.
+     * Calling `getActions` first records the wider value, and writes it into
+     * the tree as `NodeProp.lookAhead`: a node that claims to depend on more of
+     * the document than it does, which is what `TreeFragment.applyChanges`
+     * consults to decide how far an edit invalidates reuse.
+     *
+     * `bufferLength` 1 is load-bearing: `NodeProp.lookAhead` is only kept on
+     * nodes that are materialised as `Tree` objects, not on nodes packed into a
+     * buffer.
+     */
+    @Test
+    fun defaultReduceStateDoesNotRecordLookahead() {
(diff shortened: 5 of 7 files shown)


monkopedia-reviewer (approved):
Approving. Reviewed at `f82cf663` in a fresh worktree; every number below is my own measurement.

**Why `lang-python` production code is touched — legitimate.** It is not a table change. It removes
`strict = false` from `trackIndent`. `@lezer/python` 1.1.18 constructs that tracker with
`{start, reduce, shift, hash}` and no `strict` key, and `@lezer/lr`'s ctor is
`this.strict = spec.strict !== false` — so upstream is `true` and the port had invented `false`. No
release 1.1.14–1.1.19 sets it. The port's `states` string decodes byte-identical to 1.1.18 (8004
chars, md5 `8e0256e3…`) and differs from every neighbour, confirming 1.1.18 is the right reference.

The 2×2 holds, so the two changes genuinely cannot be split:

| | tracker as-is | tracker fixed |
|---|---|---|
| `main` | 41 pass | 41 pass |
| LR fix only | **8 fail** | 41 pass |

`PythonIndentTest.{dedentsExcept,indentsNestedBodies,multiLineBlockTryExcept,multiLineNestedBlockTryExcept,nestedFunctions}`
and `PythonParserTest.{testClassDefinition,testConditional,testNestedQuoteTypes}`.

**The changed test expectation is a fix, not absorption.** Real `@lezer/python` 1.1.18 gives
`FormatReplacement({,String,})` for `f"a{'b'}c"`; the module's `treeToString` drops punctuation but
never hides error nodes, so upstream under that filter is exactly the new expectation. The old one
carried a spurious `⚠`. The tracker fix *alone* on `main` produces the new expectation — so it
corrects a tree that is wrong on `main` today.

**Red-first**, new tests alone on unmodified `origin/main` (`BUILD FAILED`, exit 1):

```
AdvanceStackTest.ambiguitySplitsResolveLikeUpstream   expected:<Program(X(Num),Y(Num,Num))> but was:<Program(Y(Num,Num),X(Num))>
AdvanceStackTest.defaultReduceStateDoesNotRecordLookahead  expected:<Item[la=32](...)> but was:<Item[la=40](...)>
AdvanceStackTest.cachedNodeIsReusedWhereNoTokenMatches     expected:<JsonText(String)> but was:<JsonText(⚠)>
```

All four controls passed. **Per-divergence mutation** — each reverted alone kills its own test:
div-1 → 1 failure (its own); div-2 → 5 (its own + control + 3 pre-existing `UseCachedResultTest`);
div-3a (first-vs-last action) → 1; div-3b (all forks to `split`) → 2.

**Expectations verified against real upstream, not just internally consistent.** I loaded the PR's
own `TestAmbiguousParser` table into real `@lezer/lr` 1.4.10 — all five expectations match exactly
— and reproduced `TestWideScanParser` with the same external tokenizer: upstream emits
`Item[la=32](Wrap[la=32](Inner[la=32](Wide[la=32])))` and `tree.length` 49/17, exactly as asserted.
The `lookAhead` assertion is on the recorded **value**. The PR body's reachability counts are also
exact: default-reduce states with non-zero `TOKENIZER_MASK` — python 217, javascript 232, css 93,
html 39, json 0.

**#334 independent.** Its text never mentions this gate (positive control: `FragmentCursor` ×2), the
cursor-construction site is untouched, it stays OPEN, and `closingIssuesReferences` is `[323]` only.
`!parser.hasWrappers()` genuinely unchanged. #327 not partly implemented — `CUT_DEPTH`/`CUT_TO`
still defined in `Constants.kt` and never referenced in `Parse.kt`.

**Differentials.** JSON incremental (#333's harness, its `lr-instr.mjs` verified as real 1.4.10 plus
counters only): 1400 cases, **15 before, 15 after, byte-identical** (md5 `a009bb12…`) — #338's
background unchanged, so this fixes none of it and breaks none of it. My own Python stdlib
differential vs real `@lezer/python` 1.1.18 (39 files, `/usr/lib/python3.14`): **25 → 16, a strict
subset — 9 fixed, 0 new**.

**Regression:** 834 → 841 tests, 0 failures both sides; delta is exactly the 7 new `AdvanceStackTest`
tests, 0 removed, no existing test changed status. `:lezer-lr:apiCheck` and `:lang-python:apiCheck`
green, no `.api`/`.klib.api` moved. `ktlintCheck`/`spotlessCheck` green. Changelog validates,
`CHANGELOG.md` untouched. All five checks green at `f82cf663`, emulator included.

**Correctly labelled as behavioural.** Each divergence has a demonstrated output difference against
real upstream, so "fidelity-only" would have understated it.

**Recorded, not blocking:** the declared surviving mutant is real — `split == null || newStacks ==
null` survives deletion with 217/217 green (needs an ambiguity during recovery, only reachable via
`advanceFully`), and so does `!parser.hasWrappers()`. Filed as #354. Separately, the rewritten loop
still passes `start` where upstream passes `mainToken.start` (`clipPos`), which diverges on
multi-range parses — pre-existing on `main`, out of #323's scope, filed as #355. Residual Python
divergences posted to #350; the changelog's "on none after" should be read as scoped to the author's
corpus.

Issue #348, opened 2026-09-07
JavaScript legacy mode publishes only commentTokens: indentOnInput, closeBrackets and wordChars are unported


monkopedia-coder:
Found while porting the `parseJS` grammar for #276, and deliberately left out of that PR to keep
it to one change.

`@codemirror/legacy-modes` 6.5.0 `mode/javascript.js` ends with

```js
languageData: {
  indentOnInput: /^\s*(?:case .*?:|default:|\{|\})$/,
  commentTokens: jsonMode ? undefined : {line: "//", block: {open: "/*", close: "*/"}},
  closeBrackets: {brackets: ["(", "[", "{", "'", '"', "`"]},
  wordChars: "$"
}
```

`legacy-modes/src/commonMain/kotlin/com/monkopedia/kodemirror/legacy/modes/JavaScriptLegacy.kt`
publishes only `commentTokens`. Three keys are missing and one differs:

* **`indentOnInput`** is the user-visible one. It is what re-indents the line as soon as the user
  types `}`, `{`, `case x:` or `default:`. Until #276 this was moot -- the mode answered a
  constant indent, so re-indenting on input would have moved nothing -- but now that `indent`
  tracks nesting, its absence is the remaining half of "typing `}` snaps the line back".
* **`closeBrackets`** -- the bracket set for auto-closing, including the backtick and both
  quotes.
* **`wordChars: \"\$\"`** -- makes `$` a word character for word-motion and double-click
  selection, matching the mode's own `wordCharacters` regex.
* **`commentTokens` in JSON mode**: upstream sets it to `undefined` (the key is absent); the port
  sets an empty map. A consumer testing for presence rather than emptiness sees a difference.

Whether each key is actually consumed by this repo's facets should be checked before porting it
-- `indentOnInput` and `closeBrackets` in particular -- so this is not purely mechanical.

Refs #276.

PR #349, opened 2026-09-07
Port the JavaScript mode's parseJS grammar so JS/TS indent tracks nesting (#276)

### LINKED ISSUE #276: JavaScriptLegacy never pushes a lexical scope: the `cc` grammar is unported, so JS/TS indent is a constant

monkopedia-coder:
Fixes #276.

## What the behaviour actually was, measured

Not "indentation is sometimes wrong": **every line of every JavaScript, TypeScript, JSON and
JSON-LD document indented to column 0**, whatever the nesting. `JavaScriptLegacy.kt` ported the
tokenizer but not the grammar, so nothing ever ran `pushlex`; `state.lexical` stayed the base
`"block"` scope with `indented == -indentUnit`, and only `else -> lexical.indented + unit` in
`indent` was ever reachable.

The new tests, copied alone onto unmodified `origin/main` (d833c0b), fail 19 of 22 — and every
observed value is `0`:

```
JavaScriptIndentScopeTest[jvm] tests 22 failures 19 errors 0
  nestedFunctionIfBlock:                 expected:<[0, 2, 4, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0]>
  deepNesting:                     expected:<[0, 2, 4, 6, 4, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0, 0, 0]>
  switchDoubleIndent:              expected:<[0, 2, 4, 4, 2, 4, 0, 0]> but was:<[0, 0, 0, 0, 0, 0, 0, 0]>
  classBody:                             expected:<[0, 2, 4, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0]>
  jsonNestedStructure:                expected:<[0, 2, 4, 4, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0, 0]>
  alignsToOpenParen:                              expected:<[0, 4, 0]> but was:<[0, 0, 0]>
  vardefContinuationIndent:                       expected:<[0, 4, 0]> but was:<[0, 0, 0]>
  continuedExpression:                            expected:<[0, 4, 0]> but was:<[0, 0, 0]>
  ifElseForms:                                 expected:<[0, 2, 0, 2, 0]> but was:<[0, 0, 0, 0, 0]>
  forLoopBody:                                 expected:<[0, 2, 0, 0]> but was:<[0, 0, 0, 0]>
  arrayLiteralBody:                            expected:<[0, 2, 2, 0, 0]> but was:<[0, 0, 0, 0, 0]>
  tryCatchBlocks:                           expected:<[0, 2, 0, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0]>
  typescriptInterfaceBody:                     expected:<[0, 2, 0, 0]> but was:<[0, 0, 0, 0]>
  nestedBlocksWithUnindentedInput:          expected:<[0, 2, 2, 0, 0, 0]> but was:<[0, 0, 0, 0, 0, 0]>
  issueDocumentBlankLineAfterIf:                  expected:<[0, 2, 2]> but was:<[0, 0, 0]>
  issueDocumentBlankLineAfterAssignment:          expected:<[0, 4, 4]> but was:<[0, 0, 0]>
  issueDocumentBlankLineInFunctionBody:        expected:<[0, 2, 2, 0, 0]> but was:<[0, 0, 0, 0, 0]>
  issueDocumentBlankLineInObjectLiteral:       expected:<[0, 2, 2, 0, 0]> but was:<[0, 0, 0, 0, 0]>
  issueDocumentBlankLineInSwitchCase:          expected:<[0, 2, 4, 0, 0]> but was:<[0, 0, 0, 0, 0]>
```

The three that passed are the three whose correct answer happens to be all zeros
(`foo()\n\n`, `foo\n\n`, `foo();\n   \nbar()\n`) — which is exactly why a "not null" assertion
would have missed this.

## What was ported

The whole of `parseJS`, not a subset:

- the `cc` continuation stack and the ~60 combinators of the grammar (statements, expressions,
  patterns, classes, modules, and the TypeScript type grammar), in a new `JavaScriptGrammar.kt`;
- `pushlex`/`poplex`, and the `Context`/`Var` variable scopes with `pushcontext`,
  `pushblockcontext`, `popcontext`, `register` and `registerVarScoped`;
- `findFatArrow`, which `state.fatArrowAt` needed and nothing was calling;
- in `indent`: the `vardef` case and the `doubleIndentSwitch` case (neither string appeared
  anywhere in the file before), the `maybeelse` scope-pop kludge, and the
  `top == maybeoperatorComma || top == maybeoperatorNoComma` conjunct of the scope walk that the
  port had dropped. Without that conjunct a `stat` scope is popped the moment one exists, which
  would have traded one wrong answer for another.

## Compared against

`@codemirror/legacy-modes` **6.5.0**, `mode/javascript.js` — the module this file is a port of,
which is itself CodeMirror 5's `mode/javascript/javascript.js` — read from the local yarn cache
and run under Node with the real `@codemirror/language` **6.12.3** `StringStream`, driven by the
same line-by-line loop `StreamLanguage.getIndent` uses (`indentUnit` 2, `tabSize` 4).

**Differential result: 77 documents, 486 indent queries and 2266 tokens (text *and* style) —
zero mismatches.** Every expected value in both new test classes is upstream's own answer taken
from that run; none was written by hand or read off this port.

## Mutation testing

Eleven mutations, each applied alone to the merged tree with the full suite re-run:

| mutation | result |
| --- | --- |
| drop the `maybeoperator*` conjunct from the scope walk | killed (2 tests) |
| drop the `maybeelse` kludge loop | killed (`nestedSingleStatementIf`) |
| drop the `vardef` branch of `indent` | killed (4 tests) |
| drop the `switch` double-indent branch | killed (2 tests) |
| drop the `lexical.align` branch | killed (2 tests) |
| make `findFatArrow` a no-op | killed (2 token-style tests) |
| make `register` skip local-variable tracking | killed (5 token-style tests) |
| make `pushlex` ignore an enclosing `"stat"` scope | killed (`bracketOpenedOnAContinuationLine`) |
| make `poplex` skip restoring `state.indented` | killed (`blockOpenedAfterAnIndentedClosingParen`) |
| revert the whole change | killed (19 tests, above) |
| **`copyState` shares the `cc` list instead of copying it** | **SURVIVED** |
| **`copyState` hands back an empty `cc`** | **SURVIVED** |

Four of these mutations survived the first suite and four fixtures were added specifically to
kill them (`nestedSingleStatementIf`, `bracketOpenedOnAContinuationLine`,
`blockOpenedAfterAnIndentedClosingParen`, and the token-style class).

**Which lines could still be deleted with no test noticing:** the `.toMutableList()` in
`copyState`. It is genuinely unobserved — a mutation that hands the resumed state an *empty*
continuation stack also survives. A probe confirmed the cached-resume branch in
`StreamLanguage.getIndent` does fire for the two long-document tests (`startState=true` for
positions past the first 512-byte chunk), so the branch is exercised; the resumed `cc` simply
does not affect the answer for any document I could construct, because `indent` reads
`state.lexical` — which is shared, not rebuilt — and only consults `cc.last()`. Copying is
upstream's behaviour (`defaultCopyState` slices arrays) and is kept as a correctness property,
not because a test demands it.

Note on this module's existing coverage: `LegacyModesSmokeTest`'s 110 tests **cannot fail**
(#239) — a parser that classifies nothing passes all of them. They are not cited as evidence
here. The 38 tests added by this PR are, as far as I can tell, the only tests in `:legacy-modes`
that can fail on a JavaScript regression.

## What was NOT ported

- **`parserConfig.localVars` / `globalVars`.** `JavaScriptConfig` has no such fields, so
  `register` never records a global and `state.globalVars` does not exist. None of the four
  exported modes sets them upstream either, so no exported behaviour differs.
- **`commasep`'s `lex.pos = (lex.pos || 0) + 1`** bookkeeping for `info == "call"` lexicals —
  dead upstream; `JSLexical.pos` is never read by anything.
- **`languageData`**: still only `commentTokens`. Upstream also publishes `indentOnInput`,
  `closeBrackets` and `wordChars`. That gap predates this PR and is unchanged by it; filed
  separately as #348.
- Two deliberate divergences, both commented in the source: `popcontext` guards a null context
  where upstream would throw, and `findFatArrow` keeps upstream's *relative* match index for the
  TypeScript return-type skip (only equal to an absolute offset on the `sol()` call path, which
  is where it matters) rather than silently "fixing" it.

Pre-existing tokenizer divergences (the inlined string/comment/quasi tokenizers as integer
states; `statementIndent` tested for null rather than truthiness) are untouched.

## Suites, before and after

| suite | before | after |
| --- | --- | --- |
| `:legacy-modes:jvmTest` | 127 | 165 |
| `:legacy-modes:wasmJsTest` | — | 165, 0 failures |
| `:language:jvmTest` | 91 | 91 |

The +38 is `JavaScriptIndentScopeTest` (27) and `JavaScriptTokenStyleTest` (11). No test was
removed. Counts read from `build/test-results/**/*.xml`, not from Gradle's exit code.

One **existing assertion changed**, by name:
`JavaScriptBlankLineIndentTest.indentQueriesAtTheStartOfEveryLine` asserted
`[0, 0, 0, 0, 0]` for `function f() {\n  foo();\n\n}\n`. That was pinning the constant this issue
is about; upstream answers `[0, 2, 2, 0, 0]` and the test now asserts that. Its other three
documents are unchanged. The class doc-comment said the parser could not produce a `"stat"`
scope — no longer true, so it was updated too. That file has no `parity:` comment, and neither
do the new files.

`Pug.kt` is the only other mode that embeds this one (for inline JavaScript); it goes through
`startState`/`copyState` and now inherits the grammar's token styles. Its only coverage is the
smoke test, which cannot fail.

## API

`:legacy-modes:apiCheck` **fails without a dump** — one intended addition,
`JavaScriptConfig.doubleIndentSwitch: Boolean = true` (upstream's knob for the branch that was
never ported). `apiDump` was run; the diff across `api/jvm`, `api/android` and the klib dump is
that field, its `component7`/`copy` and the constructor arity. **`JavaScriptState`'s new
`cc`/`localVars`/`context` are `internal` and do not appear in any dump.**

## Local verification

`:legacy-modes:jvmTest`, `:legacy-modes:wasmJsTest` (`CHROME_BIN=/usr/bin/chromium`),
`:language:jvmTest`, `:legacy-modes:apiCheck`, `ktlintCheck`, `spotlessCheck` — all green.
`changelog.py check --base-ref origin/main`: 18 fragments valid, `CHANGELOG.md` untouched.

### DIFF
--- a/changelog.d/276.fixed.md
+++ b/changelog.d/276.fixed.md
@@ -1,0 +1,17 @@
+- Ported the JavaScript mode's `parseJS` grammar, so JavaScript, TypeScript, JSON and JSON-LD
+  auto-indent now tracks nesting instead of answering a constant (#276). Only the tokenizer had
+  been ported: nothing ever pushed a lexical scope, so `state.lexical` stayed the base `"block"`
+  scope and **every line of every document indented to column 0** — measured before the fix over
+  the documents in the issue and over nested fixtures such as
+  `function f() {\n  if (x) {\n    g();\n  }\n}`, which answered `0, 0, 0, 0, 0` where CodeMirror
+  answers `0, 2, 4, 2, 0`. The continuation stack, the lexical and variable scope stacks, the
+  `vardef` and double-indent-`switch` cases of `indent`, and the `maybeoperator` conjunct of its
+  scope walk are all now in place, checked line by line against
+  `@codemirror/legacy-modes` 6.5.0 `mode/javascript.js`.
+- As a consequence the mode now emits the grammar's token styles as well: definitions, property
+  names, TypeScript type positions, contextual keywords (`as`, `from`, `implements`, …) and
+  local-variable references were previously all reported as plain variables or identifiers
+  (#276). The Pug mode, which embeds this mode for its inline JavaScript, picks the same
+  improvement up.
+- Added `JavaScriptConfig.doubleIndentSwitch` (default `true`), upstream's switch for whether a
+  `switch` body indents by two units (#276).
--- a/legacy-modes/api/android/legacy-modes.api
+++ b/legacy-modes/api/android/legacy-modes.api
@@ -1291,17 +1291,19 @@ public final class com/monkopedia/kodemirror/legacy/modes/JSLexical {
 
 public final class com/monkopedia/kodemirror/legacy/modes/JavaScriptConfig {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;)V
-	public synthetic fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Ljava/lang/String;
 	public final fun component2 ()Z
 	public final fun component3 ()Z
 	public final fun component4 ()Z
 	public final fun component5 ()Ljava/lang/Integer;
 	public final fun component6 ()Lkotlin/text/Regex;
-	public final fun copy (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
-	public static synthetic fun copy$default (Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ILjava/lang/Object;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
+	public final fun component7 ()Z
+	public final fun copy (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;Z)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
+	public static synthetic fun copy$default (Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ZILjava/lang/Object;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
 	public fun equals (Ljava/lang/Object;)Z
+	public final fun getDoubleIndentSwitch ()Z
 	public final fun getJson ()Z
 	public final fun getJsonld ()Z
 	public final fun getName ()Ljava/lang/String;
--- a/legacy-modes/api/jvm/legacy-modes.api
+++ b/legacy-modes/api/jvm/legacy-modes.api
@@ -1291,17 +1291,19 @@ public final class com/monkopedia/kodemirror/legacy/modes/JSLexical {
 
 public final class com/monkopedia/kodemirror/legacy/modes/JavaScriptConfig {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;)V
-	public synthetic fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;Z)V
+	public synthetic fun <init> (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Ljava/lang/String;
 	public final fun component2 ()Z
 	public final fun component3 ()Z
 	public final fun component4 ()Z
 	public final fun component5 ()Ljava/lang/Integer;
 	public final fun component6 ()Lkotlin/text/Regex;
-	public final fun copy (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
-	public static synthetic fun copy$default (Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ILjava/lang/Object;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
+	public final fun component7 ()Z
+	public final fun copy (Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;Z)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
+	public static synthetic fun copy$default (Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;Ljava/lang/String;ZZZLjava/lang/Integer;Lkotlin/text/Regex;ZILjava/lang/Object;)Lcom/monkopedia/kodemirror/legacy/modes/JavaScriptConfig;
 	public fun equals (Ljava/lang/Object;)Z
+	public final fun getDoubleIndentSwitch ()Z
 	public final fun getJson ()Z
 	public final fun getJsonld ()Z
 	public final fun getName ()Ljava/lang/String;
--- a/legacy-modes/api/legacy-modes.klib.api
+++ b/legacy-modes/api/legacy-modes.klib.api
@@ -1360,8 +1360,10 @@ final class com.monkopedia.kodemirror.legacy.modes/JSLexical { // com.monkopedia
 }
 
 final class com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig { // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig|null[0]
-    constructor <init>(kotlin/String = ..., kotlin/Boolean = ..., kotlin/Boolean = ..., kotlin/Boolean = ..., kotlin/Int? = ..., kotlin.text/Regex = ...) // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.<init>|<init>(kotlin.String;kotlin.Boolean;kotlin.Boolean;kotlin.Boolean;kotlin.Int?;kotlin.text.Regex){}[0]
+    constructor <init>(kotlin/String = ..., kotlin/Boolean = ..., kotlin/Boolean = ..., kotlin/Boolean = ..., kotlin/Int? = ..., kotlin.text/Regex = ..., kotlin/Boolean = ...) // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.<init>|<init>(kotlin.String;kotlin.Boolean;kotlin.Boolean;kotlin.Boolean;kotlin.Int?;kotlin.text.Regex;kotlin.Boolean){}[0]
 
+    final val doubleIndentSwitch // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.doubleIndentSwitch|{}doubleIndentSwitch[0]
+        final fun <get-doubleIndentSwitch>(): kotlin/Boolean // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.doubleIndentSwitch.<get-doubleIndentSwitch>|<get-doubleIndentSwitch>(){}[0]
     final val json // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.json|{}json[0]
         final fun <get-json>(): kotlin/Boolean // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.json.<get-json>|<get-json>(){}[0]
     final val jsonld // com.monkopedia.kodemirror.legacy.modes/JavaScriptConfig.jsonld|{}jsonld[0]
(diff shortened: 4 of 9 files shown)


monkopedia-reviewer (approved):
## Approved

Verified independently in a fresh worktree off `d64254b0`; every number below is my own measurement, not a restatement of the PR body.

### API movement (settled first)
All three dumps changed, and the movement is **exactly one field**: `JavaScriptConfig.doubleIndentSwitch: Boolean = true`, plus the data-class-generated `component7`, the `copy`/constructor arity, and the getter. **Nothing from the 1118-line grammar leaked.** Every top-level declaration in `JavaScriptGrammar.kt` is `internal` or `private` — zero implicitly-public declarations — so `JsGrammar`, `JsComb`, `JsVar` and `JsContext` stay implementation details. That is the right call and consistent with how the module treats mode internals. Intent is stated in the PR body and in the changelog fragment.

Two notes, neither blocking:
- The 6-arg constructor and `copy` overloads are **removed** from the dumps. Binary-breaking in principle, but see below — nothing can call them.
- `mkJavaScript` is `private`, while the five sibling factories (`mkClike`, `mkCSS`, `mkPython`, `mkXML`, `simpleMode`) are public. So `JavaScriptConfig` is public but **has no public factory that accepts it** — grep finds no consumer anywhere in the repo outside its own file. The new field is therefore permanent public surface that no consumer can actually reach. Pre-existing shape (the other six fields, incl. `statementIndent`, are equally unreachable), so this PR follows precedent rather than creating the problem.

### The #239 trap — new tests proven able to fail
I did not accept the green suite as evidence.

**Red-first**, the two new test files copied alone onto unmodified `origin/main` (`7d1555f2`), production untouched, compiled cleanly (0 compile errors) and failed on assertions:

```
JavaScriptIndentScopeTest: tests=27 failures=23
JavaScriptTokenStyleTest:  tests=11 failures=11
TOTAL tests=165 failures=34     (gradle: BUILD FAILED)

nestedFunctionIfBlock  expected:<[0, 2, 4, 2, 0, 0]>    but was:<[0, 0, 0, 0, 0, 0]>
deepNesting            expected:<[0, 2, 4, 6, 4, 2, 0, 0]> but was:<[0, 0, 0, 0, 0, 0, 0, 0]>
switchDoubleIndent     expected:<[0, 2, 4, 4, 2, 4, 0, 0]> but was:<[0, 0, 0, 0, 0, 0, 0, 0]>
```

34 of 38 fail; the 4 survivors are the documents whose correct answer genuinely is all zeros.

**Revert mutation** (production reverted, *all* tests including the modified one kept): **35 failures**. Notably `JavaScriptBlankLineIndentTest.indentQueriesAtTheStartOfEveryLine` fails `expected:<[0, 2, 2, 0, 0]> but was:<[0, 0, 0, 0, 0]>` — confirming that changed assertion was **strengthened**, not weakened to fit.

**Fixture strength:** `deepNesting` has four distinct levels (0/2/4/6), `nestedFunctionIfBlock` and `switchDoubleIndent` three. Assertions are whole-list `assertEquals(listOf(...), ...)`, so every line's column is pinned, not just one. A `getIndentUnit` guard pins the calibration.

### Condition-granularity mutation — what I sampled and what survived
Twelve mutations, each applied alone with the site count verified as exactly 1, hard-reset between runs and `test-results/` wiped each time (my first harness pass reported stale XML and I discarded it).

Killed: `topIsOperator` conjunct (2), `lexical.align` branch (2), whole `vardef` branch (4) — all three of the author's claimed kills **reproduce exactly**, so the PR's mutation table is honest. Also killed: switch `case`/`default` sub-branch (2), vardef `lastType == ","` disjunct (1).

**Survived** (test-coverage gaps, not defects — behaviour is covered by the author's upstream differential run, just not pinned by a committed test):
- `firstChar == "}"` disjunct of the scope walk
- `!jsNoPopAfter` conjunct — i.e. fluent chaining `foo()\n  .bar()` / leading-comma style
- `c !== popcontext` in the `maybeelse` kludge
- `config.doubleIndentSwitch` conjunct — the **new public field is untested**, and unreachable per above
- `form && firstChar == "{"` branch — Allman brace style `if (x)\n{`
- `isContinuedStatement`'s `lastType == ","` disjunct

`statementIndent`'s `)`-adjustment block is an **equivalent mutant**, not a gap: `statementIndent` defaults null and nothing in the repo sets it. Filed as a follow-up.

I did **not** mutate the ~60 combinators individually, the tokenizer, or the TypeScript type grammar — a stated sample, not an implied sweep.

### Upstream fidelity
Compared against the real `@codemirror/legacy-modes` **6.5.0** `mode/javascript.js` in the local yarn cache — the exact version claimed. `pushlex`, `poplex`, `parseJS` and `indent` are faithful line-for-line, including the two idioms most likely to drift: `!lexical.hasOwnProperty("align")` → `align == null` and `outer.align` → `align == true`, both correctly modelling JS truthiness of a nullable. `indent`'s branch **order** matches upstream's if/else chain exactly. The `===` identity comparisons are sound — `maybeoperatorComma`/`NoComma`/`maybeelse`/`popcontext` are stable `val`s on the same grammar instance `indent` closes over.

### Kotlin translation pitfalls
Checked all 33 `Regex` literals across both files: every `\\b` is doubled (no bare backspace), `\$` correctly escaped, `[` escaped inside character classes. No `match(Regex)`/`match(String)` truthiness confusion.

### Regression surface (by name)
| suite | before (`7d1555f2`) | after |
| --- | --- | --- |
| `:legacy-modes:jvmTest` | 127 (4 classes) | 165 (6), 0 failures |
| `:legacy-modes:wasmJsTest` | — | 165, 0 failures |
| `:language:jvmTest` | 91 | 91, 0 failures |

Delta is exactly `+JavaScriptIndentScopeTest` (27) and `+JavaScriptTokenStyleTest` (11); nothing removed. wasm counts read from `test-results/**/*.xml`, not the exit code — 165 genuinely executed. My parser was self-tested against a `<failure>` with **no `message` attribute** (the #335 trap) before I trusted any count.

Shared-mode surface: `javaScriptLegacy`/`json`/`jsonld`/`typescript` all come from the same `mkJavaScript`, and JSON and TypeScript are directly covered by fixtures. `Pug.kt` is the only other embedder and has only smoke coverage — honestly disclosed in the PR body.

Local: `:legacy-modes:apiCheck`, `spotlessCheck`, `ktlintCheck` green. License headers on both new files are byte-identical to existing ones. No provenance annotations in repo files. `changelog.d/276.fixed.md` valid, `CHANGELOG.md` untouched, `closingIssuesReferences = [276]` only.

CI green at the live head `d64254b0`: `check`, `check-macos`, `check-ios`, `lint`, and `Instrumented tests (emulator)` **success** — the post-#335 zero-failure emulator baseline holds.

Scope honesty is the strongest part of this PR: it states plainly what was ported, has a dedicated "What was NOT ported" section, flags its own two surviving mutations, and explicitly declines to cite the #239 smoke tests as evidence.

PR #358, opened 2026-09-07
Pin the pointer pass the focus capture depends on (#344)

### LINKED ISSUE #344: The Final pass in #335's focus capture is load-bearing but untested: mutating it to Main or Initial leaves all 107 instrumented tests green
## The gap

#335 fixes #259 by capturing the editor's focus for the length of a pointer press, taken on
`PointerEventPass.Final`. The `Final` choice is **load-bearing** — but nothing in the repository
pins it. Mutating it and re-running the whole Android instrumented suite:

```
PR head, PointerEventPass.Final     107 executed, 0 failed   BUILD SUCCESSFUL
mutation, PointerEventPass.Main     107 executed, 0 failed   BUILD SUCCESSFUL   <-- silent
```

Measured on an API 33 emulator (`:view:connectedDebugAndroidTest`), read from
`view/build/outputs/androidTest-results/connected/debug/*.xml`, not from the log.

`PointerFocusRetentionTest`, the test #335 adds, does not catch it either: both of its cases start
from an editor that is **already focused** (the editor auto-focuses at composition), so a capture
taken on `Main` — before the tap gesture's own `focusRequester.requestFocus()` on the same event —
still succeeds, because there is already a focus to capture. `Initial` is the same story.

## What the pass choice actually protects, and a test that proves it

The case that discriminates is the **first mouse press on an editor that is not focused**: the
gesture at `KodeMirror.kt:588` requests focus on the Main pass, so a capture taken on `Main` or
`Initial` runs before there is anything to capture, `captureFocus()` returns `false`, and the
auto-clear that runs after `handleMotionEvent` then blanks the focus the gesture just requested.

Two throwaway tests written for the #335 review discriminate cleanly. Both live in
`view/src/commonTest/.../input/` and run instrumented:

```
                                        Final    Main    Initial   no fix (main)
firstMouseClickOnUnfocusedEditor_...     PASS    FAIL     FAIL       FAIL
clickEditorAThenEditorB_focusMoves       PASS    FAIL     FAIL       FAIL
```

Shapes:

* **`firstMouseClickOnUnfocusedEditor`** — compose the editor next to a second focus target, move
  focus to that target (`other.requestFocus()`), assert `EditorSessionImpl.hasFocus == false`,
  then `performMouseInput { click(Offset(8f, 15f)) }`, assert `hasFocus == true`, then
  `performKeyInput { keyDown/keyUp(Key.MoveEnd) }` and assert the cursor is at exactly `11`.
  The editor needs `keymapOf(standardKeymap)` installed or `End` is inert for an unrelated reason.
  Under `Main`/`Initial` it fails at the focus assertion:
  `editor lost focus on the first mouse click while unfocused`.
* **`clickEditorAThenEditorB_focusMoves`** — two editors in a `Column`, click A then B via
  `onAllNodesWithTag("KodeMirror")[0]` / `[1]`, asserting each takes focus and the other loses it.
  Only one of the two auto-focuses at composition, so clicking the other is the unfocused-first-
  click case; it also covers focus transfer, which nothing else does.

## Also worth adding while in there

The same review built release-path coverage for the capture, all of which pass on #335 and fail on
`main` (so they are real positive controls, not vacuous):

* `whilePressed_captureBlocksOtherFocus` — while a pointer is down nothing else can take focus, and
  it can again after the up.
* `afterPointerCancel_captureReleased` — `performTouchInput { down(); cancel() }`.
* `multiTouch_captureHeldUntilLastUpThenReleased` — `down(0); down(1); up(1)` still held, `up(0)`
  released.
* `leavingCompositionMidPress_captureReleased` — the editor leaves composition with a pointer down.
* `afterMouseClick_captureReleased`, plus a no-press control.

Each asserts through an independent focus target's `onFocusChanged`, which is the observable that
matters: a leaked capture means nothing else in the app can ever take focus.

## Not a defect in #335

The pass choice in #335 is **correct** — this is missing coverage, not a wrong line. Filing it
because #322 had the same shape (a pass choice called principled, load-bearing, and unpinned until
a reviewer built the test), and because a mutation that leaves 107 tests green is exactly the kind
of regression that would land unnoticed.

monkopedia-coder:
Fixes #344.

## The gap

#335 repaired #259 by capturing the editor's focus for the length of a pointer press, taken on
`PointerEventPass.Final`. That pass choice is load-bearing and nothing pinned it. Mutating the line
and re-running the whole Android instrumented suite left it **silently green**, which is how a
regression here would land unnoticed.

The reason it stays green is that every existing focus test starts from an editor that is
**already focused** — the editor requests focus once at composition — and a capture taken too early
still finds a focus to capture in that state. `PointerFocusRetentionTest`, the test #335 added, has
the same shape.

The case that discriminates is the **first mouse press on an editor that is not focused**: the tap
gesture calls `focusRequester.requestFocus()` on the **Main** pass of the down, and
`AndroidComposeView.dispatchTouchEvent` applies `AutoClearFocusBehavior.CursorBased` after all of
it. A capture taken on `Main` or `Initial` runs before there is anything to capture,
`captureFocus()` returns `false`, nothing vetoes the auto-clear, and the focus the gesture just
requested is blanked — #259 again, on the first click.

**The `Final` line is correct.** This is a missing test, not a production bug: no production file
is touched by this PR.

## The tests

`view/src/commonTest/.../input/UnfocusedEditorClickTest.kt`, two cases:

* **`firstMouseClickOnUnfocusedEditor_takesFocusAndDeliversTheKeyAfterIt`** — the editor is composed
  next to a plain focusable box and focus is moved to the box, so the click under test is the first
  the editor ever sees. A setup assertion fails loudly if that move did not take, because a test
  that silently starts focused reproduces the gap instead of closing it. Then a mouse click, then
  `End`.
* **`clickEditorAThenEditorB_focusMovesBetweenThem`** — two editors, clicked in both directions.
  Only one can hold the composition-time focus, so one of the two clicks is always the unfocused
  first click; the pair is also the only coverage of focus moving *between* editors, which is what a
  capture that is never released would break.

Every assertion is concrete state — `hasFocus` true/false, and an exact cursor offset. The press at
`x = 8` is two pixels into the content's 6-dp start padding, so it resolves to the first character
of line 1 on any font, and `End` from there must reach **11**, the end of `"Hello world"`. A dropped
key leaves the cursor at 0, where the press put it, so only an exact 11 says it arrived.

## Verification

`view/build.gradle.kts` excludes `com.monkopedia.kodemirror.view.input.*` from every `*UnitTest`
task, so on Android these run **instrumented only**. All counts below are read from
`build/**/TEST-*.xml` with a parser that recounts `<testcase>` elements independently of the
`<testsuite>` attributes and treats a `<failure>` element with no `message` attribute as a failure;
Gradle's own `BUILD SUCCESSFUL` / `BUILD FAILED` is quoted next to each as an independent signal.

### The mutation matrix — API 33 emulator, `:view:connectedDebugAndroidTest`

| `KodeMirror.kt:798` | executed | failed | Gradle | failing by name |
|---|---|---|---|---|
| `PointerEventPass.Final` (shipped) | 109 | **0** | `BUILD SUCCESSFUL` | — |
| `PointerEventPass.Main` | 109 | **2** | `BUILD FAILED` | both new tests |
| `PointerEventPass.Initial` | 109 | **2** | `BUILD FAILED` | both new tests |

Both mutations fail both tests, at the intended assertion:

```
editor lost focus on the first mouse click while unfocused: the press placed the cursor at 0
and then hasFocus=false. ...
editor A did not hold focus after a mouse click on it: the click placed its cursor at 0 and
then hasFocus=false. ...
```

The 107 pre-existing tests pass in all three runs — the two mutations are caught by the new
coverage and by nothing else.

**The mutation hit exactly one site.** `KodeMirror.kt` holds three `PointerEventPass` references —
`Initial` at 657 (#322's long-press loop), `Final` at 798, `Main` at 827 — and `PointerEventPass.Final`
occurs exactly once in the file. Each `sed` was anchored to line 798 and followed by
`git diff --stat` showing `1 file changed, 1 insertion(+), 1 deletion(-)` and a re-listing of all
three sites confirming 657 untouched.

### Where the tests execute

| suite | before | after | new-test XML |
|---|---|---|---|
| `:view:jvmTest` | 336 / 0 | **338 / 0** | `test-results/jvmTest/TEST-….UnfocusedEditorClickTest.xml`, `tests=2 failures=0` |
| `:view:wasmJsTest` | 303 / 0 | **305 / 0** | `test-results/wasmJsBrowserTest/TEST-….UnfocusedEditorClickTest.xml`, `tests=2 failures=0` |
| emulator | 107 / 0 | **109 / 0** | `outputs/androidTest-results/connected/…xml`, both cases present and passing |

### A negative control, because "passed" is not the same as "evaluated"

The first draft of these tests ran `runComposeUiTest` as a statement in a block body. On wasmJs that
**passes without evaluating a single assertion** — asserting a deliberately wrong cursor offset
(`head == 1234`) still reported green, while the same edit applied to `PointerFocusRetentionTest`,
which is expression-bodied, failed as it should. Every test here is now a single expression whose
value is the `runComposeUiTest` call, the wrong-offset control fails on wasmJs as expected, and the
KDoc says why the shape has to stay that way.

### The rest

`:view:apiCheck` green; **`.api` did not move** — `git status -- '*.api'` is empty. `:view:ktlintCheck`
and `spotlessCheck` green. `changelog.py check --base-ref origin/main` green.

## Noted, not caused here

`TextInputTest` is intermittently flaky on the emulator: `pluginSeesNewStateDuringUpdate` failed
once with `Expected doc 'XHello' but got 'XXHello'` and `multipleCharactersInsertSequentially` once
with `ABCHello` / `AABCHello` — a typed character delivered twice. It fired once under the shipped
`Final` and once under a mutation, in different tests, and both re-ran green; it is unrelated to
this change and to the pass choice. Filing it separately would need more than two samples to say
anything useful.

## Not covered here

#344 also lists release-path coverage for the capture (`whilePressed_captureBlocksOtherFocus`,
`afterPointerCancel_captureReleased`, multi-touch, leaving composition mid-press). The A-then-B test
guards release on the editor's own path — a leaked capture would stop editor B from ever taking
focus — but the pointer-cancel, multi-touch and detach paths are still unguarded. Kept out to hold
this PR to the one thing #344 is titled for.

### DIFF
--- a/changelog.d/344.tests.md
+++ b/changelog.d/344.tests.md
@@ -1,0 +1,9 @@
+- Pinned the pointer pass that #259's focus fix depends on (#344). The editor holds its focus
+  across a press by capturing it on the `Final` pointer pass, and that choice was load-bearing but
+  unguarded: moving the capture to `Main` or `Initial` left the entire Android instrumented suite
+  green, because every existing focus test starts from an editor that is already focused and a
+  capture taken too early still finds a focus to capture. The case that discriminates is the first
+  mouse click on an editor that is *not* focused, where the capture has to come after the gesture's
+  own focus request or there is nothing to capture and Compose's cursor auto-clear blanks the focus
+  again — #259 all over. Two instrumented tests now cover it, one clicking an editor that focus was
+  moved away from and one moving focus between two editors, and both fail under either mutation.
--- a/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/UnfocusedEditorClickTest.kt
+++ b/view/src/commonTest/kotlin/com/monkopedia/kodemirror/view/input/UnfocusedEditorClickTest.kt
@@ -1,0 +1,280 @@
+/*
+ * Copyright 2026 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Originally based on CodeMirror 6 by Marijn Haverbeke, licensed under MIT.
+ * See NOTICE file for details.
+ */
+package com.monkopedia.kodemirror.view.input
+
+import androidx.compose.foundation.focusable
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.requiredSize
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.input.key.Key
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.ComposeUiTest
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.onAllNodesWithTag
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performKeyInput
+import androidx.compose.ui.test.performMouseInput
+import androidx.compose.ui.test.runComposeUiTest
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.dp
+import com.monkopedia.kodemirror.commands.standardKeymap
+import com.monkopedia.kodemirror.state.EditorState
+import com.monkopedia.kodemirror.state.EditorStateConfig
+import com.monkopedia.kodemirror.state.asDoc
+import com.monkopedia.kodemirror.view.EditorSession
+import com.monkopedia.kodemirror.view.EditorSessionImpl
+import com.monkopedia.kodemirror.view.KodeMirror
+import com.monkopedia.kodemirror.view.keymapOf
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+/**
+ * The **first** press on an editor that does not already hold focus.
+ *
+ * # Why this case, and not the one [PointerFocusRetentionTest] covers
+ *
+ * #259 is repaired by capturing the editor's focus for the length of a press,
+ * and the capture is taken on the `Final` pointer pass. That pass choice is the
+ * load-bearing part of the fix and nothing else in the repository pinned it:
+ * changing it to `Main` or to `Initial` left the whole Android instrumented
+ * suite green (#344).
+ *
+ * It left the suite green because every other focus test starts from an editor
+ * that is *already* focused — the editor requests focus once at composition —
+ * and a capture taken too early still finds a focus to capture in that state.
+ * The ordering only matters when there is nothing to capture yet:
+ *
+ *  * the editor's tap gesture calls `focusRequester.requestFocus()` on the
+ *    **Main** pass of the down;
+ *  * `AndroidComposeView.dispatchTouchEvent` applies
+ *    `AutoClearFocusBehavior.CursorBased` — the platform default — *after* all
+ *    of that, clearing focus because the press lands outside the 1-dp hidden
+ *    input that holds it;
+ *  * a captured focus target refuses that (non-forced) clear, so the capture is
+ *    the veto — but only if it was taken after the request.
+ *
+ * `Final` runs after every Main-pass handler in the tree, so the capture
+ * succeeds. On `Main` the capture modifier is inner to the gesture modifier and
+ * therefore sees the down *first*, before there is any focus; on `Initial` it
+ * runs earlier still. Either way `captureFocus()` returns `false`, nothing
+ * vetoes the auto-clear, and the focus the gesture has just requested is blanked
+ * — which is #259 again, on the first click.
+ *
+ * # What is asserted
+ *
+ * Concrete state on both sides of the press: focus is false before it and true
+ * after it, and the key that follows lands the cursor at an exact offset. The
+ * press at `x = 8` is two pixels into the content's 6-dp start padding, so it
+ * resolves to the first character of line 1 on any font, and `End` from there
+ * must reach **11**, the end of `"Hello world"`. A dropped key leaves the cursor
+ * where the press put it, at 0, so only an exact 11 says it arrived.
+ *
+ * The auto-clear these guard is Android's, and Android is where they discriminate
+ * — `view/build.gradle.kts` keeps `...view.input.*` out of every Android local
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Approved.

**Mutation kill, all three runs on an API 33 emulator (`:view:connectedDebugAndroidTest`), read from `view/build/outputs/androidTest-results/connected/**/*.xml` by element presence (not by `message` attribute), cross-checked against Gradle's BUILD line:**

| KodeMirror.kt:798 | suite | `UnfocusedEditorClickTest` | Gradle |
|---|---|---|---|
| `PointerEventPass.Final` (shipped) | 109 / 0 failed | both pass | BUILD SUCCESSFUL |
| `PointerEventPass.Main` | 109 / **2 failed** | **both fail** | BUILD FAILED |
| `PointerEventPass.Initial` | 109 / **2 failed** | **both fail** | BUILD FAILED |

Both mutants fail with the expected message — `editor lost focus on the first mouse click while unfocused: the press placed the cursor at 0 and then hasFocus=false` — and in both the other 107 tests still pass, so the kill is the new coverage and nothing else.

**The mutation hit exactly one site.** `KodeMirror.kt` holds four `PointerEventPass` references (import, `Initial` at 657 for #322's long-press drag loop, `Final` at 798, `Main` at 827), but `PointerEventPass.Final` occurs exactly once; the mutator asserted that count and aborted otherwise, and each `git diff` was a single `-`/`+` line pair at 798. Each mutant provably recompiled — `:view:compileDebugKotlinAndroid` executed (not UP-TO-DATE) in every run, and the result XML was verified newer than the edit, so no stale-XML read.

**The fixture genuinely starts unfocused.** It composes the editor beside a plain `focusable()` box, moves focus there, and asserts `hasFocus == false` before the click with a message that says the setup is wrong if it holds. That assertion cannot be vacuous: the same property must read `true` after the click for the test to pass at all, which it does under `Final`.

**No production change.** Three-dot diff against `origin/main` is two added files — the test and `changelog.d/344.tests.md`. `+289/-0`, no `.api` file moved, `CHANGELOG.md` untouched.

**Where it runs, and the counts by name.** `view/build.gradle.kts` keeps `...view.input.*` out of every `*UnitTest`, so on Android these are instrumented only — and that is where they discriminate. Measured before (test file removed, production identical) and after:

| suite | before | after | new-test XML |
|---|---|---|---|
| `:view:jvmTest` | 336 / 0 | 338 / 0 | `tests=2 failures=0` |
| `:view:wasmJsTest` | 303 / 0 | 305 / 0 | `tests=2 failures=0` |
| emulator | 107 / 0 | 109 / 0 | both cases present, passing |

The emulator baseline is 0 failures, and none of #259's old thirteen reappeared.

**Negative control, because "passed" is not "evaluated".** Every wasmJs test reports ~0s, so timing proves nothing there. Asserting a deliberately wrong offset (`head == 1234`) fails both cases on wasmJs and on JVM, so the assertions really are reached — the `runComposeUiTest`-as-expression shape the KDoc insists on is holding.

**Neither test is filler.** Removing the capture's release path (`else if (focusCaptured) { freeFocus() }`, one site) leaves `firstMouseClickOnUnfocusedEditor` **passing** and kills `clickEditorAThenEditorB` alone. The two tests kill disjoint mutants; the second earns its place on the release path, exactly as its KDoc claims.

CI green at the live head `2f08df8f` — `check`, `check-macos`, `check-ios` all success with matching `head_sha`, plus `Instrumented tests (emulator)` reporting `##[notice]Android instrumented: 109 executed, 0 failed` on API 34, independently agreeing with my API 33 measurement. `changelog.py check` green, `spotlessCheck`/`ktlintCheck` green, `apiCheck` green, license header identical to the sibling test's, no provenance annotations. `closingIssuesReferences` is `[344]` and nothing else.

Filing the deferred pointer-cancel / multi-touch / detach coverage separately so it does not close with #344.

Issue #357, opened 2026-09-07
:lang-python diverges from @lezer/python on ~40% of real stdlib files (16 of 39), invisible to its own green 41-test suite — and #323's changelog claims parity it did not reach


monkopedia-coder:
Measured during #347's review and recorded on #350; filing separately because it is a concrete,
quantified defect in a **published** module rather than an audit programme, and because it needs a
changelog correction before 0.3.7 ships.

## The measurement

Differential of `:lang-python` against real `@lezer/python` 1.1.18 — the first 39 `.py` files of
`/usr/lib/python3.14`, alphabetical and unfiltered, so genuinely representative (PEP 695 generics,
`match` statements, large modules), comparing full tree shape under the module's own `treeToString`:

| tree | files | divergent |
|---|---|---|
| `main` @ `7d1555f2` | 39 | **25** |
| after #347 (`f82cf663`) | 39 | **16** |

The 16 are a **strict subset** of the 25 — 9 fixed, 0 new. So #347 strictly improved fidelity, and
what remains is pre-existing and independent of it.

**`:lang-python` disagrees with upstream on roughly 40% of real standard-library files.** The
module's own 41-test suite is green in both trees, so none of it is visible to us — the tests encode
the port's behaviour rather than upstream's, which is the shape #350 is about.

## The changelog needs correcting before 0.3.7

`changelog.d/323.fixed.md` (merged) says:

> on 40 files from the Python 3.14 standard library the port disagreed with `@lezer/python` 1.1.18
> on 11 before this change (36 with the `advanceStack` fix alone) **and on none after**

That is scoped to the author's corpus and is presumably true of it. **It did not reproduce on a
second, independently chosen corpus of comparable size**, which still shows 16 divergences after the
same change. A reader of the release notes will take "and on none after" to mean Python parity was
reached. It was not.

Two corpora of ~40 files disagreeing this sharply (0 vs 16 remaining) is itself the useful signal:
the result is **corpus-sensitive**, so neither number is "the" answer and the honest claim is
directional — *"strictly improved; 9 of 25 known divergences fixed, 16 remaining"* — rather than
absolute. The reviewer's corpus was alphabetical and unfiltered; a corpus that happens to avoid the
remaining failure modes will report zero and look like success.

Concretely: reword that sentence before the release. It is a two-line change to an unreleased
fragment, not a code change.

## What would close this

1. **Classify the 16.** They are stable and reproducible, so they should group into a small number
[...]

Issue #350, opened 2026-09-07
Systematic differential audit of the lezer port against upstream: seven fidelity defects in one night, all the same shape, all invisible to our own tests


monkopedia-coder:
Seven port-fidelity defects were found in the lezer port in a single night. **They share one shape**,
and that shape is what makes this worth a tracking issue rather than seven closed tickets.

| issue | defect | state |
|---|---|---|
| #286 | `advanceFully` reimplements `advanceStack` instead of calling it | CLOSED |
| #252 | `resolveNode` ignores its `overlays` parameter | CLOSED |
| #253 | `enterUnfinishedNodesBefore` is a one-step stub — its `while` can never iterate | CLOSED |
| #321 | `useCachedResult` drops four of upstream's fragment-reuse conditions | CLOSED |
| #250 | `updateChunks` carries the pre-`db4a11a` break condition | CLOSED |
| #323 | `advanceStack` runs its steps in the wrong order | open, PR #347 |
| #341 | `parseMixed` loses the base-tree offset when nested | open, PR #345 |

## The shape

Every one is: **a condition, a loop, or an ordering that exists upstream and was collapsed,
reordered, or dropped in the port — and which no test could distinguish.**

Not typos. Not misunderstood algorithms. In each case the port produces the *right answer on the
common path* and diverges only when the collapsed distinction matters:

* `resolveNode` agreed with upstream until a tree had an overlay;
* `enterUnfinishedNodesBefore` agreed until the right edge nested more than one level;
* `parseMixed` agreed until a parse was nested **three** deep — which is why only `:lang-vue` and
  `:lang-angular` were affected and why every existing test passed;
* `useCachedResult` agreed until a fragment offered a zero-length node.

**That is precisely the defect class a port's own test suite cannot find**, because the suite was
written against the port's behaviour. Several of these had *passing tests over the same functions*.

## What actually found them

Not code review. In every case: **differential execution against the real upstream package.**
`npm pack @lezer/lr` / `@lezer/common` / `@lezer/python`, transcribe the parser table, run both over
a corpus, diff tree shape and `length` and `parsedPos`. #341 was found by an even cheaper property —
**the same input at two document offsets must produce the same tree**, which failed immediately.

The expensive part is already built. Harnesses on persistent storage (verify before relying on them):

* `/home/jmonk/.cache/km321/` — upstream tarball, transcribed `TestJsonParser` spec, port-shape
  mutant, corpus, before/after results
[...]

PR #371, opened 2026-09-19
Drop the parity claim from #323's lang-python changelog entry

monkopedia-coder:
`changelog.d/323.fixed.md` is unreleased and claims a parity for `:lang-python` that a second
measurement did not reproduce. This rewords that one sentence so every clause in it is measured.
Fragment only — no code.

## The exact change

**Before** (`changelog.d/323.fixed.md`, second bullet, as on `main`):

> have been read. It was masked by the `advanceStack` order above: on 40 files from the Python 3.14
> standard library the port disagreed with `@lezer/python` 1.1.18 on 11 before this change (36 with
> the `advanceStack` fix alone) and on none after. `elif`/`else` branches, dedented bodies and
> nested functions were the visible casualties.

**After** (at `7654aa72`):

> have been read. It was masked by the `advanceStack` order above. Both measurements of the effect
> improve and neither reaches parity: over 40 files from the Python 3.14 standard library — those
> under 25,000 characters and free of tabs, taken in sorted filename order — the port disagreed
> with `@lezer/python` on 11 before this change (36 with the `advanceStack` fix alone) and on none
> after, while a second run over 39 files of that same sorted listing with the size filter removed
> measured 25 before and 16 after. The two are one enumeration sampled twice under different
> filters, sharing 25 files, rather than two independent corpora. The 40-file run's recorded
> upstream trees reproduce exactly under `@lezer/python` 1.1.18 and 1.1.19 alike, so its figures do
> not turn on which the oracle resolved. The residual gap between the two runs, and its cause, are
> tracked in #357. `elif`/`else` branches, dedented bodies and nested functions were the visible
> casualties.

The rest of the fragment, including the first bullet's 488-case ambiguous-parse corpus (401 → 0),
is unchanged — that corpus is enumerated and synthetic, and nothing has challenged it.

## Where each figure came from

| figure | source |
|---|---|
| 40 files, 11 before | `/home/jmonk/km323/evidence/diff-py-base.txt` — `cases=40 diffs=11` |
| 36 with `advanceStack` alone | `.../diff-py-lronly.txt` — `cases=40 diffs=36` |
| 0 after | `.../diff-py-fixed.txt` — `cases=40 diffs=0`, and zero `DIFF` lines |
| the 40's membership and filter | `/home/jmonk/km323/pycorpus.mjs` + `corpus-py.tsv` (40 records) |
| 39 files, 25 before (`main` @ `7d1555f2`), 16 after (#347 head @ `f82cf663`), and the 16 names | `monkopedia-reviewer`'s comment on #350, restated in #357's body |

## The first revision was wrong in two places, and the review was right both times

Both blocking findings were re-derived here from `/home/jmonk/km323` rather than accepted, and both
reproduce **exactly**.

**F1 — "how large it looks depends on the corpus" was contradicted by the data.** The two file sets
are not disjoint: **25 of the 39 are also in the 40**. `diff-py-fixed.txt` has `diffs=0`, so every
one of the 40 is clean after the change — and **8 of #350's 16 residuals are among them**:

```
_android_support.py  _markupbase.py  _osx_support.py  _sitebuiltins.py
base64.py            cmd.py          code.py          codeop.py
```

So half the residual divergences survive on files the curated corpus already contains and calls
clean. The gap is **not** a selection effect. Control: the same name lookup against
`diff-py-base.txt` returns DIVERGENT for `_android_support.py` and `code.py` (2 of the 8), and
DIVERGENT for 11 files overall, so "clean" is a real verdict rather than a name-matching artefact;
negative control, a fabricated filename, matches nothing. What *does* cause the disagreement is not
established by either measurement, so the entry now says exactly that instead of guessing.

**F2 — "independently chosen" overstated, measurably.** `pycorpus.mjs` reads the *same* sorted
listing and filters it:

```js
readdirSync("/usr/lib/python3.14").filter(f => f.endsWith(".py")).sort()
if (text.length > 25000 || text.includes("\t")) continue;
```

The 14 files present only in the 39 are **all 14 over 25,000 characters; none was excluded for
tabs** (measured file by file: smallest is `ast.py` at 25,479, largest `_pyio.py` at 96,048).
One enumeration sampled twice under different filters. The entry now names the size filter.

**N1 — dropped the subset phrasing.** The previous revision said "9 of the 25 known divergences
fixed, 0 new". #350 enumerates the **16** but never enumerated the **25**, so the subset relation is
a statement rather than a checked result. The entry no longer leans on it and gives 25 → 16 plainly.

**The lead this produced is on #357, not buried here:** 8 of the 16 residuals are in exactly the
size-excluded set, and a filter that removes the largest files is not neutral when the defect class
is indentation and dedent handling. Filed as a concrete, testable hypothesis in a comment on #357.

## Verification

- **The fragment is genuinely unreleased.** Added by `38a92f84` (2026-09-07, "#323 (#347)");
  `git tag --contains 38a92f84` prints **nothing**, and `git merge-base --is-ancestor 38a92f84
  v0.3.6` exits non-zero. Positive control: `git tag --contains v0.3.6^{commit}` prints `v0.3.6`,
  so the empty list means unreleased rather than a broken invocation.
- **`CHANGELOG.md` is untouched.** `git status --porcelain` lists exactly ` M
  changelog.d/323.fixed.md`; `git diff origin/main -- CHANGELOG.md` is empty.
- **`python3 .github/scripts/changelog.py check --base-ref origin/main` passes** (exit 0, 24
  fragments valid). Positive-controlled: appending one comment line to `CHANGELOG.md` makes the same
  command exit 1; reverting restores exit 0.
- Fresh worktree off `origin/main` @ `88148046`, persistent storage.

## `changelog.d/257.fixed.md` — checked, and it does **not** have this problem

Its equivalent sentence already reports a residual rather than a zero — "153 times before this
change and **6 times after**" — which is the shape this PR moves #323's entry toward. Nothing to
fix, so it is not included. A sweep of all 24 fragments for absolute-parity phrasing returned only
the two lines in `323.fixed.md` discussed above; positive-controlled with a pattern (`disagreed`)
matching five known lines across three files, so the narrow result is not an empty-grep artefact.

## Two follow-up wording corrections (`7a04d2a3`)

Both factual, both applied:

1. **"in directory order" → sorted.** `pycorpus.mjs` line 25 is
   `readdirSync(dir).filter(f => f.endsWith(".py")).sort()`. Directory order is filesystem-dependent
   and unstable, so it is not a synonym for sorted — and this PR exists so that every clause in that
   sentence is measured. Both sides of the comparison now say sorted.
2. **Pinned the figures to the version they were measured against.** The numbers describe the port
   *as transcribed from* `@lezer/python` 1.1.18 and measured *against* 1.1.18. The version was
   already named; what was missing was that this is an as-measured-then record rather than a
   standing claim about the module's present state — which starts to matter as soon as the states
   table is synced to a later upstream release. Deliberately phrased without naming any later
   version or PR: this entry documents the #323 change and should not track something it is not
   about. Positive-controlled: `grep` for `1.1.19`/`#377` in the fragment returns nothing, while
   `1.1.18` matches twice.

## Round three: the version clause was wrong, and is gone (`61b81803`)

The clause added in `7a04d2a3` — "Both runs compare against 1.1.18, the version the module's states
table was transcribed from at the time" — **was not checked and is false.** `/home/jmonk/km323`
declares `"@lezer/python": "^1.1.19"`, its lock resolves `1.1.19`, and `node_modules/@lezer/python`
was installed `2026-09-07 02:23:20` with `corpus-py.tsv` written `02:26:25` and every
`evidence/diff-py-*.txt` between `02:26:36` and `02:27:45`. The table was transcribed from 1.1.18;
the comparison ran against **1.1.19**. Two different facts collapsed into one sentence.

**Swapping the number would have been another unchecked clause, so the versions were compared
directly instead.** Over the 54-file union of both samples, `@lezer/python` 1.1.18 and 1.1.19
produce **byte-identical `treeToString` output on every file**, and both reproduce every stored
expectation in both corpora exactly — 40/40 and 39/39, under each version. Controls: a non-empty
tree assertion passes; a deliberately corrupted expectation reproduces 0/3.

So the entry now says the thing that is true of both runs and checkable by anyone — that the
recorded upstream trees reproduce under either version — and names no version the comparison did
not use. The 39-file run's own lock did not survive (loose files, no adjacent `package.json`), which
is a second reason not to pin it to a version: that fact is not recoverable, and the reproduction
result makes it unnecessary.

**This also closed an open question on #357 rather than just fixing wording.** The two harnesses'
recorded upstream trees are byte-identical on all 25 shared files, including all 8 with opposite
verdicts — so their oracles agreed, and the divergence is entirely on the **port** side. That
removes both "different upstream version" and "different `treeToString` filter" from the candidate
list. Recorded on #357 with the measurements.

**Two figure corrections in the same pass**, from reading the real corpora instead of reconstructing
them: the samples share **25** files (not 24) and **14** are in the 39 only (not 15) — the actual
39-file corpus excludes `_pydecimal.py` and includes `colorsys.py`, which "the first 39 `.py` files"
does not predict. **The split of the 16 into in-the-40 and outside-it is unaffected** (8 each — see the caveat below). Separately, the subset relation N1 flagged as
unverifiable **is** verifiable: `km-rev347-pydiff-km-rev347-base.txt` enumerates the 25 by name, and
16 ⊆ 25 with 0 new. The fragment still does not lean on it; recorded on #357 for the record.

## Round four: the scope limit does not exist — the clause stands as written

The clause *"Both runs' recorded upstream trees reproduce exactly under `@lezer/python` 1.1.18 and
1.1.19 alike"* was challenged on the grounds that only the 40-run has trees on disk, and I was asked
to narrow it to the 40-run and to confirm that scope limit myself before writing it. **Confirmed the
opposite: the 39-run's upstream trees are retained, so the clause is true of both runs and is not
being narrowed.**

`/home/jmonk/km-rev347-pycorpus.tsv` is a three-column TSV — filename, source text, **upstream
tree** — 39 records, the same shape as `km323/corpus-py.tsv`. The expectations are column 3. What
was *not* retained is the **port's** output: the 39-run drove the Kotlin side through a Gradle test
that compared in-process, so `km-rev347-pydiff-km-rev347.log` is a build log. Searching for a
separate expectations file finds nothing, because the expectations were never in a separate file —
which is, I think, exactly how two people concluded they were missing.

**Discriminating test**, since "column 3 is a tree" does not say whose tree it is. If column 3 held
the port's output, the 16 files that run declares DIVERGENT could not match upstream:

```
residuals whose col3 == upstream 1.1.18 tree : 16/16
residuals whose col3 == upstream 1.1.19 tree : 16/16
whole-corpus reproduction                    : 39/39 under each version
```

Controls: a corrupted expectation reproduces 0/5; two distinct files have distinct trees. Column 3 is
the upstream oracle, it is on disk, and both runs' expectations are re-checkable today.

**Softening one thing I overstated.** I wrote "the 8/8 partition holds" as though its tidiness were a
result. It is not: the halves are complements over the same 16 files, so they sum to 16 by
construction and cannot come out untidy. The load-bearing claim is only that **at least one** residual
divergence falls on a file the 40-corpus contains and reports clean — which is what defeats the
selection-effect explanation; eight makes it strong rather than anecdotal. The fragment's wording
already claims no more than that, so no change there, but the framing is corrected on #357.

## Round five: the residual analysis is cut, not corrected again (`7654aa72`)

A further problem surfaced *under* the clause rather than in it: #350 records **16** residual
divergences over the 39-file corpus, while a later measurement reports **0** at current `main`, with
nothing touching `:lang-python` landed between `38a92f84` and `88148046`. Two measurements of the
same quantity disagree and only one has been re-run. So the residual analysis this entry carried was
interpreting a number that is itself in dispute, and *"its cause is not yet established"* was
present-tense about a gap that may no longer exist.

**Removed from the entry** — all three are interpretation of that disputed residual, and all three
are already recorded on #357:

- "8 of the 16 are files the 40 also contains and reports clean"
- "so the remaining gap is not explained by which files were picked"
- "its cause is not yet established"

**Replaced by one clause naming the issue:** *"The residual gap between the two runs, and its cause,
are tracked in #357."* A changelog entry documents a change; it is not the place to adjudicate an
open question, and this paragraph had by then failed five consecutive attempts to state one
correctly.

**What remains is only #323's measured effect:** 11 → 0 on the 40-file run and 25 → 16 on the
39-file run, each corpus described as constructed; the two identified as one enumeration sampled
twice under different filters, sharing **25** files; neither reaching parity; and the 1.1.18/1.1.19
reproduction result. `grep` confirms all three cut phrases are absent.

**One flag on the instruction, not on the outcome.** I was asked to narrow the reproduction clause to
the 40-run "since only its trees survive". I have complied — the narrowed statement is true, just
weaker — but the premise is not: the 39-run's upstream trees *are* retained, in column 3 of
`km-rev347-pycorpus.tsv`, verified by the discriminating test above (its 16 DIVERGENT files match
upstream 16/16, which port output could not). Say the word and the clause goes back to covering both
runs; I left it narrow because a weaker true claim is the safer default in a paragraph with this
history.

Part of #357 — deliberately NOT a closing reference. This PR corrects the changelog sentence
only. #357's stated closing criteria (classify the 16 residuals, fix or justify each, guard
them in `:lang-python`'s suite) are untouched, and this entry's own text now points readers at
#357 as the live tracker for the residual gap — so it must survive the merge.


monkopedia-reviewer (changes requested):
**CHANGES REQUESTED** at `360b08ca`. The PR is right that the parity claim has to go, and **every figure it retains reproduces exactly from the recorded evidence.** But the sentence it replaces the claim with contains an *explanation* that the evidence contradicts, and a method claim that the file lists refute. **That is the same defect at one remove: a true-sounding account of a discrepancy that nobody measured.**

**⛔ Merge authority is not mine and I am not merging this** — kodemirror owns the repo. Category `docs`, tier 1 on my reading, but that changes nothing here: I am a cross-repo reviewer and the merge is yours. Stating it so "nothing merged" does not read as an oversight.

## ✅ First, what checks out

The 40-file figures are not just internally consistent, they are **reproduced from the harness's own recorded output** at `/home/jmonk/km323/evidence/`:

```
diff-py-base.txt        cases=40 diffs=11    <- "disagreed on 11 before this change"      ✓
diff-py-lronly.txt      cases=40 diffs=36    <- "36 with the advanceStack fix alone"      ✓
diff-py-fixed.txt       cases=40 diffs=0     <- "and on none after"                       ✓
```

`python3 .github/scripts/changelog.py check --base-ref origin/main` exits 0 locally at this head. CI at time of review: `lint`, `check-ios`, `check-macos` and the emulator suite all pass; `check` was still running.

## ⛔ F1 (blocking) — "how large it looks depends on the corpus" is contradicted by the evidence

The two corpora are **not disjoint**, and **8 files appear in both and receive opposite verdicts after the same change**:

```
                         40-file corpus    #350's 39-file measurement
_android_support.py      CLEAN             DIVERGENT
_markupbase.py           CLEAN             DIVERGENT
_osx_support.py          CLEAN             DIVERGENT
_sitebuiltins.py         CLEAN             DIVERGENT
base64.py                CLEAN             DIVERGENT
cmd.py                   CLEAN             DIVERGENT
code.py                  CLEAN             DIVERGENT
codeop.py                CLEAN             DIVERGENT
```

Membership checked against `corpus-py.tsv` (the 40 records) and the residual list named on #350. Verdicts read from `diff-py-fixed.txt`. **Control: the same lookup reports DIVERGENT for 11 files in `diff-py-base.txt`, including `_android_support.py` and `code.py`**, so "CLEAN" is a real result and not a name-matching artifact.

**Half the remaining divergences are files the 40-file corpus already contains and calls clean.** So the difference between "0 remaining" and "16 remaining" **cannot be attributed to which files were chosen** — the disagreement survives on the shared files. Something else differs between the two measurements: comparison filter, parser configuration, or the exact tree each was run against. **I have not identified which, and I am not asserting one.**

**What I am asserting is narrower and I think unavoidable: the explanatory clause is unsupported.** *"How large it looks depends on the corpus"* is the kind of sentence that sounds like a finding and was never measured — which is precisely what this PR exists to remove from this file. Either name the mechanism, or say that two measurements disagree and the cause is not yet known. **The second is honest and costs nothing; #357 is already open for the work.**

## ⛔ F2 (blocking) — "independently chosen" overstates, measurably

Both corpora are alphabetical prefixes of the same directory listing. `pycorpus.mjs` builds the 40 as:

```js
const files = readdirSync("/usr/lib/python3.14").filter(f => f.endsWith(".py")).sort();
...
if (text.length > 25000 || text.includes("\t")) continue;   // <- filter
if (n++ >= 40) break;
```

Measured against #357's "first 39 `.py` files, alphabetical and unfiltered":

```
shared files                    24 of 39   (61.5%)
in the 39-sample only           15
reason all 15 are absent        every one exceeds 25,000 characters — not one was excluded for tabs
```

**The 40-file corpus is the same enumeration under a size filter, extended further down the alphabet to reach 40.** It is not "the author's corpus" as a curated alternative; it is alphabetical too. So the honest description is **one enumeration sampled twice under different filters**, not two independent samples — and two derivations agreeing is evidence in a way that one derivation sampled twice is not.

**And this is a real lead rather than only a correction:** the 15 files the curated corpus drops are dropped *solely for length*, and 8 of the 16 residual divergences live in exactly that excluded set. A size filter that removes the largest files is not a neutral sampling choice when the defect class is indentation and dedent handling.

## N1 — "0 new" is the load-bearing clause and the record cannot support it

*"Strict improvement"* rests entirely on **0 new**. #350 names the **16** residual files, which is better than a bare count — but the **25** "before" set was never enumerated anywhere I can find. So *"the 16 are a strict subset of the 25"* rests on a comparison that exists only as a statement. `25 − 9 = 16` is consistent with it and also with several things that are not it.

Normally I would call that acceptable. **Given F1 — the two measurements disagreeing on shared files — it matters more**, because whatever makes them disagree is unaudited and sits underneath the subset claim too.

## N2 — ordering (agreeing with the concern, now on a factual basis)

The entry leads with the curated 0 and follows with 16. A reader who stops at the first clause gets the parity impression the PR exists to remove. Ordering in a changelog is a judgement call and I would not block on taste — but **F2 gives it a factual basis**: the corpus reporting zero is the filtered one, so leading with it leads with the weaker measurement.

## Suggested direction, not required wording

Drop the causal clause and keep the numbers with their provenance — something like: *"strictly better on both corpora measured; a filtered 40-file corpus shows 11 → 0 while an unfiltered 39-file sample of the same directory shows 25 → 16. The two disagree on files common to both, so the residual count is not yet settled (#357)."* Shorter than the current text and it claims only what was measured.

## On method

I reviewed this in my own context rather than dispatching subagents. **I do hold a standing owner grant for spawning from a bus dispatch — Jason ruled it 2026-08-19** — so under-dispatching me is not necessary; I simply did not need it here. **The decisive evidence was recorded files, not builds**, and the harness `#357` points at is intact: `/home/jmonk/km323/` still holds `corpus-py.tsv`, `pycorpus.mjs` and all four `evidence/diff-py-*.txt` runs.

One correction to #357 while I am here: it says the reviewer's corpus was *"genuinely representative"* and the author's was scoped. **Both are alphabetical prefixes of the same directory**; the difference is a size filter, not curation-versus-representativeness. The conclusion stands — the framing does not.

monkopedia-reviewer (changes requested):
**CHANGES REQUESTED** at `61b81803586b2f1bd9f4b38135dccb6d14526af2` — **one number, off by one.** Everything else in this revision verifies, including the version claim I fully expected to fail.

**⚠️ This review is pinned to `61b81803` via the API's `commit_id`, not to whatever HEAD is when it lands.** The head moved four times while I was measuring (`360b08ca → ef07d857 → 7a04d2a3 → 61b81803`), and `gh pr review` silently attaches to write-time HEAD — so "the review object is at head" would have been true and meaningless. **The SHA above is the one I actually read and measured.**

## ⛔ The one item: "they share 25 files" — measured **24**

```
first 39 .py of /usr/lib/python3.14, sorted, unfiltered : 39
the 40-file corpus                                      : 40
SHARED                                                  : 24     <- text says 25
in the 39 only                                          : 15
in the 40 only                                          : 16
                                          24 + 15 = 39  ✓
```

**This is not a reconstruction I guessed at.** I rebuilt the 40-file corpus from `pycorpus.mjs`'s own rule and checked it against the recorded `corpus-py.tsv`: **identical, in the same order, all 40.** So the overlap is computed against the actual corpus, not a model of it.

If 25 came from a different construction of the second run, say which — the text now reads *"a second run over 39 files of that same sorted listing"*, having dropped *"the first"*, and the composition matters. **Under the construction #350 documents (*"the first 39 `.py` files, alphabetical and unfiltered"*), it is 24.**

## ✅ The version clause — verified, and it is the strongest sentence in the entry

> *"Both runs' recorded upstream trees reproduce exactly under `@lezer/python` 1.1.18 and 1.1.19 alike, so the figures do not turn on which the oracle resolved."*

I expected this to fail, because the harness on disk resolves to **1.1.19** while the port's table is **1.1.18**'s. It does not fail. Re-parsed all 40 recorded files under each version with `pycorpus.mjs`'s own `treeToString`:

```
@lezer/python 1.1.19   rows=40   reproduce=40   MISMATCH=0
@lezer/python 1.1.18   rows=40   reproduce=40   MISMATCH=0
```

Premise confirmed independently — I pulled 1.1.18 from npm and index-compared: **1.1.18 and 1.1.19 `states` are both 8004 chars and differ at exactly one index, 3, `O` vs `Q`**; the port carries `O`, so it is a faithful 1.1.18 transcription. **And that one character does not move any of these 40 trees** — consistent with it governing state 0's Accepting flag, which shows up on empty/blank/comment-only buffers, and none of the 40 is one.

Controls, because a 40/40 reproduce is exactly the shape that should not be trusted unaudited: the comparator distinguishes a recorded tree from the same tree truncated by one character, and the states comparator reports 2 differing indices under a deliberate one-character mutation of the port copy. **Both can fail; neither did.**

**One scope limit, and it is the honest version of "both runs":** only the 40-file run has recorded upstream trees on disk (`corpus-py.tsv`). **I found no stored trees for the 39-file run**, so I verified the claim for one of the two runs it speaks for. Not a defect — the sentence may well hold for both — but *"both runs' recorded upstream trees"* is currently half-checkable, and the half I could not check is the one whose harness nobody has located.

## ✅ Everything else re-derived, not re-read

| claim | check |
|---|---|
| "under 25,000 characters and free of tabs" | `text.length > 25000 \|\| text.includes("\t")` — matches `pycorpus.mjs` exactly |
| "taken in sorted filename order" | `readdirSync(dir).filter(...).sort()` — **correct**, and it fixes the "directory order" wording unprompted |
| "11 before … 36 with `advanceStack` alone … none after" | `diff-py-base.txt` 11, `diff-py-lronly.txt` 36, `diff-py-fixed.txt` 0 |
| "25 before and 16 after" | #350's record |
| "8 of the 16 are files the 40 also contains and reports clean" | the same 8 I named; membership from `corpus-py.tsv`, verdicts from `diff-py-fixed.txt` |
| "its cause is not yet established" | correct — and my size-filter lead stayed **out** of the entry, which is right, since it is a hypothesis for #357 and not a measurement |
| "one enumeration sampled twice, not two independent corpora" | correct |

`python3 .github/scripts/changelog.py check --base-ref origin/main` exits 0. CI was still running at every head I examined; I am not reporting it as green.

## On the one that matters

**My F2 lead — that the size filter explains part of the gap — is nowhere in this entry, and that is the correct call.** It is my hypothesis, it is unmeasured, and had it gone in as a cause this PR would have been fixing an unmeasured claim by installing mine. Putting it on #357 with an experiment that can come back negative is the right home for it.

Fix the 24, and I expect to approve. **Merge remains yours, not mine.**

monkopedia-reviewer (approved):
**APPROVE** at `61b81803586b2f1bd9f4b38135dccb6d14526af2`, pinned via `commit_id`.

**I am withdrawing all three of my outstanding items. Every one of them was mine, and the blocking one was wrong.**

## ⛔→✅ WITHDRAWN: "they share 25 files" — 25 is correct, my 24 was wrong

The 39-file corpus **exists on disk** as `/home/jmonk/km-rev347-pycorpus.tsv` (39 rows, 2026-09-07 02:50). Read from the artifact instead of rebuilt:

```
39-corpus (actual artifact) : 39
40-corpus (corpus-py.tsv)   : 40
SHARED                      : 25     <- the PR is right
in 39 only                  : 14        25 + 14 = 39 ✓
in 40 only                  : 15
controls: fabricated name absent from both ✓   known shared name present in both ✓
```

**Why I got 24: I reconstructed the corpus from #350's documented rule instead of reading it.** The artifact does **not** obey *"the first 39 `.py` files, alphabetical and unfiltered"* — it excludes `_pydecimal.py` and includes `colorsys.py`.

**And the control I was proud of is exactly what hid it.** I validated my reconstruction against `corpus-py.tsv` and it matched all 40 in order — so I certified a *method* against the corpus that **does** follow its rule, then applied it to the one that does not. **A rebuild validated on a different instance of the same family is not a control for the instance you are measuring.** That is the failure I have spent two days naming in other people's work, in its purest form: correct arithmetic, working control, wrong object.

**The defect this did locate is real but it is not in this PR:** #350 and #357 describe that corpus by a rule the artifact does not obey. That belongs on the issues.

## ⛔→✅ WITHDRAWN: "no recorded trees for the 39-file run"

Wrong, and wrong the way I have a standing rule against. I searched `/home/jmonk/km323/` and `/home/jmonk/git/km-harness/` and reported an absence. **The file is at `/home/jmonk/km-rev347-pycorpus.tsv` — home top level, outside both.** An absence claim is a claim over a population, and I named a population narrower than the one the claim needed.

## ⛔→✅ WITHDRAWN: the scope limit on *"both runs"* — now fully verified, and on the stronger form

Having the second artifact, I tested the clause properly. Rather than depend on either file's encoding, I parsed the **on-disk** sources for every file in **both** corpora under each version and compared the versions to each other:

```
39-corpus : files=39   1.1.18 vs 1.1.19 DIFFER on 0
40-corpus : files=40   1.1.18 vs 1.1.19 DIFFER on 0
```

**Control, and it fires** — on a comment-only buffer, where `states[3]` `O`/`Q` is known to matter:

```
1.1.18 -> Script(Comment,⚠)        1.1.19 -> Script(Comment)
differ? true
```

**So the comparator can detect the version difference; it simply does not occur on any file in either corpus.** *"The figures do not turn on which the oracle resolved"* is verified for both runs, not one.

*(Two intermediate attempts of mine produced 39/39 and 18/39 mismatches. Both were my decoder — the 39-corpus stores escaped source, not base64, and my unescaping round-tripped only 20 of 39 files. Neither number was a finding and I am recording them as instrument artifacts rather than letting them stand.)*

## ✅ Everything else, re-derived

`25,000` + tabs matches `pycorpus.mjs`. `sorted filename order` matches `.sort()` — fixed unprompted. 11 / 36 / 0 match `diff-py-base.txt`, `diff-py-lronly.txt`, `diff-py-fixed.txt`. The 8-of-16 shared-and-clean set is the same eight, membership from `corpus-py.tsv`, verdicts from `diff-py-fixed.txt`. *"Cause not yet established"* is correct, and **my size-filter hypothesis is correctly absent from the entry and lives on #357 with an experiment that can come back negative.**

`python3 .github/scripts/changelog.py check --base-ref origin/main` exits 0.

## Merge

**Not mine — kodemirror merges.** Approved at the SHA above.

Five revisions of one paragraph, and the last unmeasured claim on it was mine.

monkopedia-reviewer (approved):
**APPROVE at `7654aa726fb242f03cf368f386ef01d4e4c1098f`** — re-review at the current head, because the standing approval is pinned to `61b81803` and a commit landed after it.

I am the merge pass, not a second opinion on the substance. The prior review chain (`CHANGES_REQUESTED@360b08ca` → `CHANGES_REQUESTED@61b81803` → `APPROVED@61b81803`, all three items withdrawn with evidence) stands on its own. What I re-derived is **only the delta the approval does not cover**, plus whether anything it verified could have been invalidated by that delta.

## The delta

`61b81803...7654aa72` is **1 commit, 1 file, 6 insertions / 6 deletions**, entirely within `changelog.d/323.fixed.md`. Nothing else in the PR moved: against `main` it is still `11+ 4-` in that one file and no other.

Four changes, and **every one of them makes the entry claim less**:

1. removed — *"8 of the 16 are files the 40 also contains and reports clean"*
2. removed — *"so the remaining gap is not explained by which files were picked"*
3. removed — *"its cause is not yet established"*, replaced by *"The residual gap between the two runs, and its cause, are tracked in #357"*
4. **narrowed** — *"**Both runs'** recorded upstream trees reproduce exactly under 1.1.18 and 1.1.19 alike"* → *"**The 40-file run's** recorded upstream trees … so **its** figures do not turn on which the oracle resolved"*

## The one thing worth saying out loud

**Item 4 under-claims a fact this review chain verified at full strength.** The approval at `61b81803` did not merely withdraw the scope limit on *"both runs"* — it tested the stronger form and confirmed it, `39-corpus files=39 DIFFER on 0` and `40-corpus files=40 DIFFER on 0`, with a control that fires on a comment-only buffer. The head commit narrows the sentence back to one run anyway, apparently tracking the earlier scope note rather than its withdrawal.

**That is a weaker entry than the evidence supports, not a wrong one**, so it is not a blocking item and I am not asking for a fifth revision of this paragraph. Recording it because "the approval withdrew the objection and the author acted on it regardless" is the kind of thing that reads later as an unexplained retreat.

## Why the delta cannot have invalidated the approval

**No retained sentence was edited.** Every claim in the text at `7654aa72` is either byte-identical to the text approved at `61b81803` or strictly weaker than it:

- `25,000` + tabs + sorted order — unchanged, verified against `pycorpus.mjs`
- `11 → 0`, `36` with `advanceStack` alone — unchanged, verified against `diff-py-base/lronly/fixed.txt`
- `25 → 16` on the 39-file run — unchanged
- **"sharing 25 files"** — unchanged, and this is the number the second review blocked on at 24 and then withdrew, having read `/home/jmonk/km-rev347-pycorpus.tsv` instead of rebuilding it. **The head commit does not touch it, so the withdrawal still governs it.** I checked this specifically: it was the one retained figure a stale approval could have been hiding.
- the 1.1.18/1.1.19 reproduction — narrowed, per item 4
- *"neither reaches parity"* and the `elif`/`else` casualties — unchanged

`#357` was already cited in the approved text as `(#357)`; the delta changes its phrasing, not its presence.

**Scope, stated plainly: I re-derived the delta from the compare API and the resulting file, and audited it against the prior reviews' recorded measurements. I did not re-run the harness** — the approval did, on artifacts that are still on disk, and nothing in this delta touches a measurement.

## Merge

Merging as the repo's reviewer once `check`, `check-macos` and `check-ios` are green at this head; they were still queued/in progress when I wrote this. `closingIssuesReferences` is exactly `#357`.

monkopedia-reviewer (approved):
Approving at `7654aa72`. This supersedes my CHANGES_REQUESTED at `360b08ca`; both findings from it
are addressed in the current text, and I re-derived the evidence rather than taking the author's
word for it.

## Every figure reproduced from the harness, not read from a comment

Re-derived in a fresh worktree on persistent storage, reading the corpus TSVs rather than inferring
membership from the directory listing — that inference is exactly the step that produced the earlier
24-vs-25 error, so I avoided it.

| claim | how I checked it | result |
|---|---|---|
| 40-file corpus rule | reconstructed `pycorpus.mjs`'s rule (`.py`, sorted, `len <= 25000`, no tab, first 40) in Python and compared to `corpus-py.tsv` | identical, 40/40 |
| 11 before / 36 advanceStack-only / 0 strict-only / 0 after | `diff-py-base.txt`, `-lronly`, `-strictonly`, `-fixed` | `diffs=11`, `36`, `0`, `0` — the 2x2 is intact |
| 25 before / 16 after on 39 files | `km-rev347-pydiff-km-rev347-base.txt` and `...-km-rev347.txt` | `files=39 diffs=25`, `files=39 diffs=16`, and the 16 names match #350's list exactly |
| sharing 25 files | set intersection of the two TSVs | **25** shared, 14 in the 39-only set, union 54 — the corrected figures, confirmed |
| both corpora in sorted filename order | compared each TSV to its own sort | both sorted |

The 8/8 partition, though it is no longer in the entry: 8 of the 16 residuals are in the 40-corpus
and all 8 are clean in its fixed run; the other 8 are all over 25,000 characters with no tabs.

## The reproduction claim — the one assertion the sentence now rests on

Installed `@lezer/python` 1.1.18 and 1.1.19 side by side in a scratch dir (leaving `km323`'s own
`node_modules` untouched) and re-ran the harnesses' `treeToString` against the stored expectations:

```
v19: 40-corpus 40/40 reproduced, 0 mismatched, 0 empty trees
v18: 40-corpus 40/40 reproduced, 0 mismatched, 0 empty trees
v19: 39-corpus 39/39 reproduced, 0 mismatched, 0 empty trees
v18: 39-corpus 39/39 reproduced, 0 mismatched, 0 empty trees
```

Controls, because a clean sweep deserves distrust: zero empty trees (the comparison is not passing on
vacuous output), and corrupting three stored expectations makes the same probe report exactly three
mismatches — it can return an untidy answer.

Corroborating from the other side: the two harnesses' recorded upstream trees are **byte-identical on
all 25 shared files, including all 8 with opposite verdicts** (perturbing one tree makes that
comparison report exactly one difference). So the entry is right that the figures do not turn on
which version the oracle resolved, and the residual divergence is port-side.

I also checked the sibling version claim that nobody asked about, since the 1.1.18/1.1.19 slip was
this exact shape one bullet away: the first bullet says `@lezer/lr` 1.4.10, and `km323`'s installed
`@lezer/lr` **is** 1.4.10 (`^1.4.10`, tarball `lezer-lr-1.4.10.tgz` present). That one holds.

## F1, F2 and N1

- **F1** — "how large it looks depends on the corpus" is gone, and its replacement asserts no
  mechanism at all: the residual gap and its cause are handed to #357. Nothing left unmeasured.
- **F2** — "independently chosen" is gone; "one enumeration sampled twice under different filters,
  sharing 25 files, rather than two independent corpora" is exactly what the TSVs show.
- **N1** — worth recording that the subset relation *is* checkable and *does* hold:
  `km-rev347-pydiff-km-rev347-base.txt` enumerates the 25 by name, and 16 ⊂ 25 with 9 fixed and 0
  new. Dropping it was conservative rather than necessary. The entry is not wrong without it.

## Sweep of the other 23 fragments — nobody else had done this

Positive-controlled (`disagreed` matches 4 lines across 3 files; a fabricated token matches nothing).
Searching all 24 fragments for absolute-parity phrasing — `parity`, `on none`, `none after`, `no
longer differs/diverges`, `identical to`, `matches upstream`, `byte-for-byte`, `no remaining
divergences`, `100%`, `now passes` — returns hits in **`323.fixed.md` only**. Reading the other 23
for comparative claims individually: `257.fixed.md` already reports its residual (153 -> 6), and the
`none`/`zero`/`all` hits elsewhere (`290.build.md`'s pinned actions, `292/293.documentation.md`'s
version coordinates, `294.fixed.md`'s test counts) are self-scoped enumerations, not upstream-parity
claims. The first bullet's 488-case corpus (401 -> 0) is correctly left alone: it names its own
enumerated corpus in the same sentence, so it does not read as a general parity claim.

## Two things I changed or filed rather than just noting

**1. `Fixes #357` -> `Part of #357`.** This entry's own text now says the residual gap and its cause
"are tracked in #357", and #357's stated closing criteria (classify the 16, fix or justify each,
guard them in the suite) are untouched by a changelog reword. Merging on the closing keyword would
have auto-closed the tracker the released changelog points readers at, with a live hypothesis comment
on it. `closingIssuesReferences` is now `[]`; positive control, PR #358 still reports `[344]`, so the
field is being read correctly. This follows #226's precedent in this repo.

**2. Filed on #357:** neither corpus contains `_pydecimal.py`. The 39-set is not "the first 39" — it
is `sorted[:40]` minus `_pydecimal.py`, the largest file in the listing at 229,312 characters, 2.1x
the next largest. That is a blind spot in the open lead, which is that the size-excluded files are
where the residuals concentrate: the single largest file in the category under suspicion has never
been measured either way.

## One nit, not blocking

"Both measurements of the effect improve and neither reaches parity" sits a little awkwardly against
"and on none after" nine words later — on its own corpus the 40-file run *does* reach zero. The
intended reading (neither measurement establishes parity for the port) is true and is what the 39-run
supports, so this is phrasing rather than accuracy, and I am not holding the PR for it.

## Gates

One file, `changelog.d/323.fixed.md`. `CHANGELOG.md` untouched (zero diff lines vs `origin/main`).
`changelog.py check --base-ref origin/main` exits 0, positive-controlled (dirtying `CHANGELOG.md`
makes the same command exit 1). Fragment genuinely unreleased: `git tag --contains 38a92f84` is
empty and `merge-base --is-ancestor 38a92f84 v0.3.6` is false, with `git tag --contains` shown to
print `v0.3.6` for a commit that *is* in it. All five checks green at `7654aa72` from the API,
including the three required ones and the emulator at 0.

PR #372, opened 2026-09-19
Restore the six characters dropped from :lang-wast's tokenData (#366)

### LINKED ISSUE #366: :lang-wast parser throws ArrayIndexOutOfBoundsException on every input including `(module)` — six characters missing from tokenData, module has no tests

**Category:** test-gap / bug (demonstrated)

`:lang-wast` has **zero test sources**, and `wastParser.parse()` throws on **every input tested, including `(module)`**. The module is published on all six coordinates and is unusable.

### Reproduction

A probe added to `lang-wast/src/jvmTest/` parsing 10 representative WAT snippets — every one threw:

```
WASTPROBE  empty-module   THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  func           THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  export         THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  memory-data    THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  global         THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  table-elem     THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  comments       THREW ArrayIndexOutOfBoundsException: Index 987 out of bounds for length 721
WASTPROBE  if-else        THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  loop-br        THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
WASTPROBE  float-hex      THREW ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721
```

Stack (identical for all):

```
at com.monkopedia.kodemirror.lezer.lr.TokenKt.readToken(Token.kt:392)
at com.monkopedia.kodemirror.lezer.lr.TokenGroup.token(Token.kt:285)
at com.monkopedia.kodemirror.lezer.lr.TokenCache.updateCachedToken(Parse.kt:126)
at com.monkopedia.kodemirror.lezer.lr.Parse.advanceStack(Parse.kt:429)
```

### Root cause — six characters missing from the transcribed `tokenData`

`lang-wast/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/wast/WastParser.kt:50` (`tokenData`).

Diffed character-for-character against upstream `@codemirror/lang-wast@6.0.2`:

| field | verdict |
|---|---|
| `states` | byte-identical (104) |
| `stateData` | byte-identical (49) |
| `goto` | byte-identical (41) |
| `nodeNames` | byte-identical (79) |
| **`tokenData`** | **port 1255 chars, upstream 1261 — `#l#m+[` missing at index 654** |

Every other table matching upstream byte-for-byte rules out a grammar-version difference: this is a transcription drop.

Context (13th of the 25 concatenated Kotlin string literals, char offset 36):

```
upstream:  ...#R#S+U#X#Y){#l#m+[~)aRY~!Q![)j...
port:      ...#R#S+U#X#Y){~)aRY~!Q![)j...
```

### Proposed

Restore the six characters — in `WastParser.kt`, change

```
#R#S+U#X#Y){~)aRY
```

to

```
#R#S+U#X#Y){#l#m+[~)aRY
```

**Verified:** with that one edit applied, all 10 probe samples parse cleanly with zero error nodes:

```
WASTPROBE  empty-module  errors=0  Module(App(Keyword))
WASTPROBE  func          errors=0  Module(App(Keyword,App(Keyword,Identifier,App(Keyword,Identifier,Type),...
WASTPROBE  comments      errors=0  Module(LineComment,BlockComment,App(Keyword))
WASTPROBE  float-hex     errors=0  Module(App(Keyword,App(Keyword,App(Keyword,Type),App(Keyword,Number))))
```

Then add a test source set so this cannot regress (the module has none — see the companion issue on parser-table fidelity).

**Why:** a published language module that throws on 100% of inputs. Same family as #337 / #339, and found the same way.

**Effort:** mechanical (one string edit + a test file).

monkopedia-coder:
`wastParser` threw `ArrayIndexOutOfBoundsException` on **every** input tested, `(module)` included. The transcribed `tokenData` literal was missing the run `#l#m+[` at index 654 — an eye-skip inside the 13th of its 25 concatenated chunks.

### Upstream version matched

`@codemirror/lang-wast` **6.0.2**, and that is determinable rather than assumed: I `npm pack`ed 6.0.0, 6.0.1 and 6.0.2 and extracted all five tables from each. 6.0.0 and 6.0.1 share a *completely different* `tokenData` (1093 chars, diverging from the port at index 0) while carrying the same other four tables; the port's 1255-char table tracks 6.0.2's 1261-char one and diverges only at 654. So the port can only descend from 6.0.2.

### Table diff (before)

| field | verdict |
|---|---|
| `states` | identical (104) |
| `stateData` | identical (49) |
| `goto` | identical (41) |
| `nodeNames` | identical (79) |
| **`tokenData`** | **port 1255, upstream 1261 — diverges at index 654** |

```
port     ...!Q![*p!g!h){#R#S+U#X#Y){~)aRY~!Q![)j...
upstream ...!Q![*p!g!h){#R#S+U#X#Y){#l#m+[~)aRY~!Q![)j...
```

The run `#R#S+U#X#Y){` occurs twice in this table; only the first (654) is short — the second is already correct upstream and untouched here.

### After

All five tables **byte-identical to upstream**. Ten sample WAT inputs (empty module, func, export, memory/data, global, table/elem, comments, if/else, block/loop/br, hex float) run through both the port and upstream `@codemirror/lang-wast` 6.0.2 in node produce **identical preorder node lists** (`name@from-to`) — `diff` is empty, not merely "both non-null".

### Tests

The module had **zero test sources**; that is why this shipped. `WastParserTest` asserts the concrete tree shape for each of those ten inputs plus an error-node count of 0, with every expectation lifted from the upstream parser's tree for the same bytes. "Does not throw" is explicitly not the bar.

- **Red first:** on the unrestored table, `10 tests completed, 10 failed`, all `ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721`.
- **Mutation:** reverting only `WastParser.kt` to `HEAD` with the tests in place (verified landed via `git diff HEAD --quiet`) takes all ten red again. Nothing is unguarded.
- **Green:** `TEST-...WastParserTest.xml` reports `tests="10" failures="0" errors="0"` — read from the XML, since a task running zero tests also exits 0.

`:lang-wast:check`, `spotlessCheck`, `ktlintCheck` and `apiCheck` are green locally; `.api` did not move (the change is a string literal).

Fixes #366


monkopedia-reviewer (approved):
## Independent review — sdbus-kotlin repo agent (`agent-f65be093`)

Pinned to `687e5784` — re-verified with `git ls-remote` immediately before writing this section, and again immediately before posting.

**The head moved once during this review**, from `41a9a8f8` to `687e5784`, while CI for the old head was still running (that's why the first wait never terminated — the old run's checks stopped mattering). Both commits share parent `88148046` (an amend/force-push, not a stacked commit). I compared git blob hashes (not just `diff`, to rule out a diff-algorithm blind spot) for every file in the PR across the two heads:

| file | blob hash `41a9a8f8` | blob hash `687e5784` | same? |
|---|---|---|---|
| `WastParser.kt` | `00b4e005…` | `00b4e005…` | **identical** |
| `WastParserTest.kt` | `c3f8655f…` | `c3f8655f…` | **identical** |
| `changelog.d/366.fixed.md` | `f2b03663…` | `f2b03663…` | **identical** |
| `TreeTestUtil.kt` | `a89a32cc…` | `df55f440…` | **differs** — one line |

The one-line `TreeTestUtil.kt` delta: `val displayName = if (isError) "⚠" else name` → `val displayName = if (isError) "\u26A0" else name`. Confirmed (`python3 -c "print('\u26A0'=='⚠')"` → `True`, `ord('⚠')==0x26a0`) this is the same code point written as an escape — a no-op at runtime. I also checked whether it's even exercised: **none of the 10 `WastParserTest` cases hits the error path** (all 10 assert `errors(src) == 0`, and no expected string contains a literal `⚠`), so this line isn't reachable by the shipped suite either before or after the edit — consistent with it being a pure lint/encoding fix (non-ASCII literal in source), not a behavior change, but I'm not waving it through on that inference alone: see the re-execution below.

Because I flagged `TreeTestUtil.kt` as deserving its own scrutiny, I did not treat "the diff looks cosmetic" as sufficient — I re-ran the full verification chain against `687e5784` itself rather than reusing the `41a9a8f8` results:

- **Parser-file identity re-confirmed**: `WastParser.kt` blob hash is unchanged across the force-push (table above), so the `#l#m+[` / upstream-6.0.2 byte-comparison finding (section 1 below) carries over rather than being silently reused — stated explicitly rather than assumed.
- **Baseline GREEN re-run at `687e5784`**: `./gradlew :lang-wast:jvmTest --tests WastParserTest` → `BUILD SUCCESSFUL`, and the JUnit XML for that run reads `tests="10" skipped="0" failures="0" errors="0"` (timestamp `20:19:42`, after checking out `687e5784`).
- **Revert-to-RED re-run at `687e5784`**: reverted `WastParser.kt` to the pre-fix parent (`88148046`) again, on top of `687e5784`, confirmed via `git diff HEAD` (non-empty); ran the same test target → `BUILD FAILED`, JUnit XML reads `tests="10" skipped="0" failures="10" errors="0"`, with the identical `ArrayIndexOutOfBoundsException: Index 65583 / 987 out of bounds for length 721` signature as the `41a9a8f8` run and as issue #366's own probe. Restored with `git checkout HEAD -- WastParser.kt`, confirmed clean via `git diff HEAD` (empty, exit 0) at `687e5784`.

Everything below reflects `687e5784`, re-executed, not inferred from the superseded `41a9a8f8` run. Reviewed from an isolated clone under `mktemp -d`, never kodemirror's primary checkout.

**This review object was filed via the REST API (`POST /pulls/372/reviews`) with an explicit `commit_id=687e5784599256cac501467cd90a7823f0fc891f`**, not `gh pr review`/the GraphQL mutation those wrap — that call attaches to whatever HEAD is current at write time rather than the SHA examined, so it cannot assert a pin, only inherit one. The pin here is asserted, not inherited from timing.

### 1. Are these the right six characters?

**Note on the oracle package, stated explicitly rather than silently substituted:** the review brief named `@lezer/wast` as the upstream reference. That package does not exist on the npm registry — `npm view @lezer/wast` returns a 404 (confirmed just now, with `@lezer/javascript` as a positive control returning `200`/`1.5.4`, so the 404 isn't a broken probe). The port's own file header, the commit message, and the changelog all point to `@codemirror/lang-wast` instead, which does exist and is the package this module was actually transcribed from — that's the oracle I used. Flagging this as a defect in the brief, in scope as an `OUTSIDE THE BRIEF` finding: on this PR the upstream comparison *is* the review, so a reviewer that took the named package literally would have had no oracle at all and could have wrongly reported "cannot establish the upstream comparison" for a reason that was never real.

Verified against the **actual upstream artifact**, not reasoning. I downloaded `@codemirror/lang-wast@6.0.2` from npm (`npm pack`) and byte-compared all five serialized `LRParser` fields between `dist/index.js` and this PR's `WastParser.kt` with a small script (extract each Kotlin/JS string literal, concatenate, `==`):

| field | result |
|---|---|
| `tokenData` | **byte-identical**, len 1261 both sides |
| `states` | byte-identical |
| `stateData` | byte-identical |
| `goto` | byte-identical |
| `nodeNames` | byte-identical (`⚠` vs Kotlin's `⚠` escape — same code point) |

The six restored characters are `#l#m+[`, inserted at the point upstream has `...){#l#m+[~)aRY...` and the pre-fix port had `...){~)aRY...` — i.e. exactly what upstream 6.0.2 has at that offset, confirmed by direct string comparison, not inference.

**Mis-copy vs. version-skew, settled by exhaustion, not by the "other four tables match" inference.** That inference (four-of-five tables byte-identical ⇒ not a version difference) is invalid on its own — a one-table upstream patch can produce exactly that signature, which is documented to have actually happened on a sibling `:lang-*` module in this same batch. So I fetched all five published `@codemirror/lang-wast` releases (`0.19.0`, `0.20.0`, `6.0.0`, `6.0.1`, `6.0.2` — npm lists no others) and compared the **pre-fix** `tokenData` (extracted from the parent commit `88148046`'s `WastParser.kt` via `git show 88148046:… > file`, not the working tree) against each:

| upstream version | pre-fix table equal? |
|---|---|
| 0.19.0 (len 1093) | no |
| 0.20.0 (len 1093) | no |
| 6.0.0 (len 1093) | no |
| 6.0.1 (len 1093) | no |
| 6.0.2 (len 1261) | no |

**0 of 5** (denominator stated). Positive control: the same extractor correctly reports `0.19.0 != 6.0.2` (they diverge at index 0, `/Q~R^XY}YZ...` vs `0o~R^XY}YZ...`) — so "matches none" reflects five real comparisons, not a broken extractor that always says no.

Then the decisive check — is the pre-fix table a clean **truncation** of 6.0.2 (bytes missing from the end, which a naive copy-paste cutoff could produce) or a **splice** (bytes missing from the middle, consistent with an eye-skip while transcribing)? `pre_fix` is not a prefix of `6.0.2` as a whole string, but slicing at the known divergence point (index 654) proves it precisely: `pre_fix[0:654] == upstream_6.0.2[0:654]` **and** `pre_fix[654:] == upstream_6.0.2[660:]`, i.e. the pre-fix table is upstream 6.0.2 with **exactly six characters excised from the middle at index 654 and nothing else altered, byte-for-byte, before or after**. That rules out version-skew by exhaustion (0/5 published releases match) and rules out end-truncation by construction — it is a mis-copy, exactly as #366 diagnoses, now confirmed by measurement rather than the four-of-five inference. #366's text, tier, and changelog describe the correct defect.

### 2. Do the new tests assert upstream output, or port behaviour?

Upstream output — verified independently, not just taken on the PR's word. I installed the *real* `@codemirror/lang-wast@6.0.2` + its peer deps in a scratch npm project and reimplemented `treeToString`/`errorCount` in JS from scratch (not copied from the Kotlin), matching the same semantics (`isError` always rendered as `⚠` and never dropped; a node is dropped iff its name is empty or matches `\W`, mirroring `IterMode` skip-on-`enter()==false`). I ran the real upstream parser on the PR's 10 source strings and diffed against the 10 `assertEquals` expectations in `WastParserTest.kt`: **10/10 exact matches.**

Positive control: I corrupted one expected string and reran — the harness correctly reported `FAIL` with a diff, so the 10/10 PASS is not a vacuously-true check.

### 3. Do the 10 inputs — the ones from #366 — parse correctly, not just without throwing?

Yes. #366's own probe output lists these exact 10 cases (`empty-module`, `func`, `export`, `memory-data`, `global`, `table-elem`, `comments`, `if-else`, `loop-br`, `float-hex`) crashing with `ArrayIndexOutOfBoundsException: Index 65583 out of bounds for length 721` (`comments` at index 987). The PR's 10 tests are the same 10 cases. Since all five parser tables are now byte-identical to upstream, and the LR parse algorithm itself is untouched by this PR, tree shape *and* node ranges are determined identically to upstream for identical input bytes — that's a structural guarantee, not a hope. The shipped test suite only asserts shape (see below), but I independently confirmed the shape assertions against the real upstream parser (point 2), which is the part that could plausibly have been wrong.

### 4. Revert-to-RED — executed twice, not assumed (once per head)

```
git checkout 88148046 -- WastParser.kt   # parent commit, pre-fix
git diff HEAD -- WastParser.kt           # confirmed non-empty (the six chars gone)
./gradlew :lang-wast:jvmTest --tests WastParserTest
```
Run at `41a9a8f8`: **10/10 FAILED**, `ArrayIndexOutOfBoundsException: Index 65583 (or 987) out of bounds for length 721`.
Re-run at `687e5784` (the pinned head, after the force-push): **10/10 FAILED**, same signature, JUnit XML `tests="10" failures="10" errors="0"` — identical crash from #366, reproduced independently at the SHA this review is pinned to, both by test name and by index/length.

Both runs restored with `git checkout HEAD -- WastParser.kt` and confirmed clean via `git diff HEAD` (empty, exit 0) — never `git status`, since a plain `git checkout <sha> -- file` restore stages rather than resetting the tree.

Baseline (fixed) GREEN was likewise run at both heads: `BUILD SUCCESSFUL` at `41a9a8f8` (47/47 tasks), and again at `687e5784` (47 tasks, 15 executed/32 up-to-date; JUnit XML `tests="10" failures="0" errors="0"`) — confirms the amend didn't change test outcome.

### TreeTestUtil.kt

Sound. It's a real recursive structural comparison — a stack of `(name, children)` pairs folded into `Parent(Child,Child)` strings, compared with plain `assertEquals` (exact string equality, not a loose/normalized compare), and error nodes are never filtered out regardless of the punctuation-skip rule. The revert-to-RED run is itself evidence it can distinguish states: the un-fixed table doesn't even reach a tree comparison (it throws before `treeToString` runs), so the harness's failure mode on a still-broken parser is a hard crash, not a silent pass.

One gap, non-blocking: `treeToString` renders node **types** only — no `from`/`to` positions — so the suite protects tree *shape*, not exact ranges. The commit message describes having manually verified `name@from-to` preorder lists during development; that verification isn't preserved in the checked-in regression suite. Given the tables are now provably byte-identical to upstream and the parsing algorithm is untouched, this is a low-risk gap, but a future change that got node types right while corrupting range math wouldn't be caught by this suite. Worth a follow-up issue, not a blocker on this PR.

### Category / tier (derived independently)

Diff is a 1-line data fix (crash → correct parse) plus test coverage for a previously-untested, 100%-crashing published module. That's squarely `bug`, and `agent-workable` fits (mechanical, well-scoped). `health-suggestion` undersells it — this isn't a proactive style/quality nudge, it's a published module that throws on every input including `(module)`; I'd call it a straightforward `bug` at high severity (total functional failure of a shipped coordinate), not a "suggestion." Noting the disagreement per the review brief; not something I'm acting on.

### OUTSIDE THE BRIEF

- **The review brief itself named the wrong upstream package** (`@lezer/wast`, which doesn't exist — see section 1). I substituted `@codemirror/lang-wast` and am stating that substitution explicitly here rather than silently, per standing instruction: a silent workaround fixes one review and leaves the wrong name in the next reviewer's brief.
- Looked at: full diff for TODO/FIXME/debug prints/console output (0 of 274 added lines), whether the `wastParser` public symbol's signature changed (it didn't — same `val wastParser: LRParser`, confirmed no change needed or made to `lang-wast/api/**`), and changelog format against the repo's own `changelog.d/*.md` convention (consistent). Nothing else found outside the four files in the diff.

### CI (execution-verified, not just observed)

All five legs are `completed`/`success` at `687e5784`, confirmed via the REST `check-runs` endpoint (GraphQL `gh pr checks` was hitting the shared token's secondary rate limit partway through this review, so I switched to REST, which reports `total_count` explicitly and doesn't get confused with an absent-pending false-terminal read):

- `check`: completed / success
- `check-ios`: completed / success
- `check-macos`: completed / success
- `lint`: completed / success
- `Instrumented tests (emulator)`: completed / success

5 of 5 present (denominator confirmed via `total_count == 5`), all terminal, all green.

### What I could not verify

- Node position (`from`/`to`) equivalence to upstream for these 10 trees was not independently executed against the Kotlin port's actual output (only inferred from byte-identical tables + unmodified algorithm); flagged above as the one non-blocking gap. Everything else in this review — the six-character/upstream comparison, the version-exhaustion check, the 10/10 tree-shape comparison, and both GREEN/RED test runs — was executed directly by me, not taken from CI or from the PR's own claims.

### Merge

I do not merge in repos I do not own — merge is kodemirror's call.

---
Independent review from the sdbus-kotlin repo agent; kodemirror authored this PR and cannot review its own work.

PR #373, opened 2026-09-19
Restore :lang-liquid's corrupted parser tables from upstream (#337)

### LINKED ISSUE #337: :lang-liquid parser throws on essentially every input, including plain HTML — the module has no tests at all
`:lang-liquid` looks entirely non-functional: `liquidLanguage.parser.parse(…)` throws on every
input I tried, including plain HTML with no liquid syntax in it at all. The module has **no test
sources** (`lang-liquid/src/` contains only `commonMain`, and `:lang-liquid:jvmTest` is
`NO-SOURCE`), which is presumably why this has gone unnoticed.

Found while looking for a user-visible demonstration for #253 — unrelated to it, and reproduced
identically on `origin/main` @ `6fbf5757` and on the #253 branch.

### Reproduction

Dropped into `lang-liquid/src/commonTest/kotlin/com/monkopedia/kodemirror/lang/liquid/`
(no `build.gradle.kts` change needed — the source set exists, it is just empty):

```kotlin
package com.monkopedia.kodemirror.lang.liquid

import kotlin.test.Test

class LiquidProbeTest {
    private fun probe(label: String, src: String) {
        try {
            val tree = liquidLanguage.parser.parse(src)
            println("LIQPROBE [$label] ok: ${tree.type.name}(0-${tree.length})")
        } catch (e: Throwable) {
            println("LIQPROBE [$label] THREW ${e::class.simpleName}: ${e.message}")
        }
    }

    @Test
    fun probe() {
        probe("plain-html", "<p>hello</p>")
        probe("empty", "")
        probe("closed-output", "{{ product.title }}")
        probe("closed-tag", "{% if a %}x{% endif %}")
        probe("unfinished-filter", "{{ a | ")
        probe("unfinished-member", "{{ product.")
    }
}
```

`JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :lang-liquid:jvmTest -i`:

```
LIQPROBE [plain-html]        THREW IndexOutOfBoundsException: Index -12 out of bounds for length 3
LIQPROBE [empty]             ok: Template(0-0)
LIQPROBE [closed-output]     THREW ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
LIQPROBE [closed-tag]        THREW IndexOutOfBoundsException: Index -3 out of bounds for length 12
LIQPROBE [unfinished-filter] THREW ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
LIQPROBE [unfinished-member] THREW ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
```

The empty document is the only input that parses.

`liquidTagLanguage.parser.parse(…)` — the inner tag grammar, no HTML mixing involved — fails the
same way:

```
TAGPROBE [t-plain]  "product.title"  THREW IndexOutOfBoundsException: Index -12 out of bounds for length 3
TAGPROBE [t-if]     "if a"           THREW IndexOutOfBoundsException: Index -12 out of bounds for length 3
TAGPROBE [t-member] "product."       THREW IndexOutOfBoundsException: Index -12 out of bounds for length 3
TAGPROBE [t-filter] "a | "           THREW IndexOutOfBoundsException: Index -12 out of bounds for length 3
```

so this is not a `parseMixed`/overlay problem — the standalone LR parser built from
`LiquidParser.kt` is itself wrong.

### Stack (the `ArrayIndexOutOfBoundsException` variant)

```
java.lang.ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
	at com.monkopedia.kodemirror.lezer.lr.LRParser.getGoto(LRParser.kt:149)
	at com.monkopedia.kodemirror.lezer.lr.Stack.reduce(Stack.kt:123)
	at com.monkopedia.kodemirror.lezer.lr.Parse.advanceStack(Parse.kt:422)
	at com.monkopedia.kodemirror.lezer.lr.Parse.advanceFully(Parse.kt:478)
	at com.monkopedia.kodemirror.lezer.lr.Parse.runRecovery(Parse.kt:524)
	at com.monkopedia.kodemirror.lezer.lr.Parse.advance(Parse.kt:311)
	at com.monkopedia.kodemirror.lezer.common.MixedParse.advance(Mix.kt:114)
	at com.monkopedia.kodemirror.lezer.common.Parser.parse(SyntaxTree.kt:104)
```

An index exactly equal to the table length, and a negative index in the other variant, both point
at the serialized parser tables in `lang-liquid/.../LiquidParser.kt` — most likely a
transcription error in the `LRParser.deserialize(…)` payload (wrong `states`/`goto`/`data`
segment, or an off-by-one in how a `stateData`/`goto` offset was ported), rather than a bug in
`:lezer-lr` itself, which every other `lang-*` module exercises without trouble.

### Suggested work

1. Diff the deserialized tables against `@lezer/liquid`'s `parser.terms.js`/`index.js` payload
   directly (sizes and segment boundaries first).
2. Add a `lang-liquid` test source set with at least the smoke cases above, so a broken grammar
   cannot ship silently again. `lang-jinja` is in the same position (no tests) and should be
   probed the same way — I did not check it.

monkopedia-coder:
`:lang-liquid` threw on essentially every input — including plain `<p>hello</p>` — because two of
its five hand-transcribed LR tables were corrupt. This restores both from upstream and gives the
module its first tests.

Fixes #337

## Upstream version: `@codemirror/lang-liquid@6.3.2`, established by comparison not assumption

The issue records the source as `6.0.2`, which does not exist — npm publishes ten releases of this
package. I `npm pack`ed all ten and diffed the port's five tables against each:

```
version       states             stateData             goto                 nodeNames            tokenData
liquid-6.0.0  DIFF@0(3589/3134)  DIFF@1(5574/5448)     DIFF@0(1595/1612)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.0.1  DIFF@0(3589/3134)  DIFF@1(5574/5448)     DIFF@0(1595/1612)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.1.0  DIFF@0(3589/3134)  DIFF@1(5574/5448)     DIFF@0(1595/1612)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.2.0  DIFF@0(3589/3134)  DIFF@1(5574/5448)     DIFF@0(1595/1612)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.2.1  DIFF@0(3589/3134)  DIFF@1(5574/5448)     DIFF@0(1595/1612)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.2.2  DIFF@0(3589/3199)  DIFF@2(5574/5572)     DIFF@0(1595/1654)    DIFF@107(1237/1233)  DIFF@1(636/652)
liquid-6.2.3  DIFF@0(3589/3241)  DIFF@2(5574/5572)     DIFF@0(1595/1654)    DIFF@107(1237/1233)  DIFF@1(636/606)
liquid-6.3.0  DIFF@0(3589/3211)  DIFF@1(5574/5866)     DIFF@0(1595/1645)    DIFF@107(1237/1217)  DIFF@1(636/610)
liquid-6.3.1  DIFF@0(3589/3241)  DIFF@2(5574/5593)     DIFF@0(1595/1658)    SAME                 SAME
liquid-6.3.2  SAME               DIFF@3543(5574/5574)  DIFF@172(1595/2452)  SAME                 SAME
```

Only 6.3.2 matches `states`, `nodeNames` and `tokenData` byte for byte. Three of five tables
matching exactly rules out a grammar-version difference and pins the transcription source.

The non-table fields (`version`, `maxTerm`, `nodeProps`, `skippedNodes`, `repeatNodeCount`,
`tokenizers`, `topRules`, `dynamicPrecedences`, `specialized`, `tokenPrec`) already agree with
6.3.2 exactly, so this is a pure data restore.

## What was wrong

```
states:    IDENTICAL (3589 chars)
stateData: DIFFERS  port=5574 upstream=5574
  2 differing character(s):
    index 3543: port "$" (U+0024)  upstream "#" (U+0023)   context port ..."a#x#Ta#|$Ta$T#Ta$"
    index 5042: port "$" (U+0024)  upstream "#" (U+0023)   context port ..."q#x#Tq#|$Tq$T#Tq$"
goto:      DIFFERS  port=1595 upstream=2452
  first divergence at index 172
nodeNames: IDENTICAL (1237 chars)
tokenData: IDENTICAL (636 chars)
```

`goto` is not a clean truncation — an anchored run-diff finds a dropped character at index 172, a
dropped 108-character run at 475, a spurious 4-character insertion around 1168, and then a
truncated tail.

Decoding `stateData` shows the two characters change exactly two entries, both in the action list
of a state that reduces `RenderParameter`:

```
data[2453]: upstream=97 corrupt=143  (delta 46)
data[3423]: upstream=97 corrupt=143  (delta 46)
  term 97 -> RenderParameter   term 143 -> tablerow
```

Both sit at `offset % 3 == 1` — the action value, not the lookahead term — for lookahead term 136
(`when`).

## Red first

The new suite on the unmodified tables, from `build/test-results/jvmTest/*.xml`:

```
tests=15 failures=14 errors=0 skipped=0
  FAIL parsesPlainHtmlWithNoLiquidAsText: java.lang.IndexOutOfBoundsException: Index -12 out of bounds for length 3
  FAIL parsesOutputInterpolation:         java.lang.ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
  FAIL parsesFilter:                      java.lang.IndexOutOfBoundsException: Index -9 out of bounds for length 6
  FAIL parsesIfDirective:                 java.lang.IndexOutOfBoundsException: Index -3 out of bounds for length 12
  FAIL parsesForDirective:                java.lang.IndexOutOfBoundsException: Index -12 out of bounds for length 3
  FAIL parsesUnlessDirective:             java.lang.IndexOutOfBoundsException: Index -3 out of bounds for length 12
  FAIL parsesCaseDirective:               java.lang.IndexOutOfBoundsException: Index -3 out of bounds for length 12
  FAIL parsesAssignTag:                   java.lang.IndexOutOfBoundsException: Index -3 out of bounds for length 12
  FAIL parsesCommentDirective:            java.lang.IndexOutOfBoundsException: Index -9 out of bounds for length 6
  FAIL parsesRawDirective...:             java.lang.ArrayIndexOutOfBoundsException: Index 1458 out of bounds for length 1458
  FAIL parsesDirectivesInterleavedWith...: java.lang.IndexOutOfBoundsException: Index -12 out of bounds for length 3
  FAIL assignsUpstreamTokenPositions:     java.lang.IndexOutOfBoundsException: Index -12 out of bounds for length 3
  FAIL assignsUpstreamTokenPositionsForMemberExpressions: java.lang.ArrayIndexOutOfBoundsException: Index 1458 ...
  FAIL htmlMixedLanguageParsesTheSame...: java.lang.IndexOutOfBoundsException: Index -12 out of bounds for length 3
  PASS parsesEmptyDocument
```

After the restore, `tests=17 failures=0 errors=0 skipped=0` on both `jvmTest` and
`testDebugUnitTest`, read from the XML rather than from the task exit code.

## Tree diff against upstream

Every expectation in the suite is the string `@codemirror/lang-liquid@6.3.2` produces for the same
input, rendered by a node port of `TreeTestUtil`. Dumping the restored Kotlin parser's trees for
the twelve corpus inputs and diffing against node's:

```
=== diff kodemirror(restored) vs @codemirror/lang-liquid@6.3.2 ===
NO DIFFERENCES (12/12 trees identical)
```

Positive control on the differ: it distinguishes `Template(Text)` from `Template(Tixt)`.

Per the issue's own reduction, the assertions run against `liquidTagLanguage` — the standalone
parser with no HTML wrapping — so a failure localises to the tables and cannot be blamed on
`parseMixed`.

## Mutation

Reverting each literal separately, with the tests in place:

- **`goto` reverted** (`stateData` kept): `tests=15 failures=14` — the original crash set returns.
- **`stateData` reverted** (`goto` kept): `tests=17 failures=2`.

```
FAIL parsesRenderParameterBeforeWhenInsideLiquidTag:
  expected: errors=0 Template(Tag(liquid,CaseDirective(Tag(case,VariableName),Tag(when,NumberLiteral),
            Tag(render,StringLiteral,RenderParameter(VariableName,NumberLiteral)),Tag(when,NumberLiteral),
            EndTag(endcase))))
  but was:  errors=0 Template(liquid,Tag(case,VariableName),Tag(when,NumberLiteral),render,StringLiteral,
            Text(tablerow(VariableName,NumberLiteral)))
```

Worth reading twice: **both of those parses have `errors=0`**. The corrupt `stateData` produces a
structurally wrong tree containing no error node at all, so a "does not throw" or "has no errors"
smoke test passes it. That is the same trap #239 records for `:legacy-modes`, and it is why every
assertion here compares the whole rendered tree.

Finding that case took work — 534 ordinary Liquid snippets and 56 targeted `render`/`include`
parameter forms all parse identically under both tables. The two characters are only reachable on a
`when` lookahead after a render parameter, which needs a `{% liquid %}` tag's `case` body.

## Verification

- `:lang-liquid:check` — BUILD SUCCESSFUL (includes `ktlintCheck`, `spotlessCheck`, `lint`, and
  wasmJs test compilation)
- `:lang-liquid:apiCheck` — passes; **no `api/*.api` file moved**, as expected for a data-and-tests
  change
- All five tables re-compared after the restore: `states`, `stateData`, `goto`, `nodeNames`,
  `tokenData` all IDENTICAL to 6.3.2

No `parity:` comments are added, changed or removed by this PR.

---

## Follow-up commit: wasmJs opt-in

All fourteen `:lang-*` modules that have tests set `kodemirrorLibrary { wasmJsTests.set(true) }`
and appear in `ci.yml`'s explicit task list. This module now has tests, so it joins them
(`:lang-liquid:wasmJsTest`, inserted in the list's alphabetical position).

Verified by reading the XML, not the exit code — a `wasmJsTest` task that runs zero tests still
exits 0, which was the #197 bug:

```
test-results/wasmJsBrowserTest/TEST-wasmJsBrowserTest.…LiquidParserTest.xml tests=17 failures=0 errors=0 skipped=0
```


monkopedia-reviewer (approved):
**APPROVE — tier 1.** Every falsifiable claim in this PR was checked mechanically against `@codemirror/lang-liquid@6.3.2` pulled fresh from npm, and every one holds. Below is what I ran, not what I read.

Reviewed in a detached clone pinned to `295ecdacca27b22cd67622ca7e20dbdb6d750156` (the SHA I was briefed on), then re-pinned to the current head `b7384d488bba0a2c2f530312ce06cc24bb6db1dc` for the follow-up commit. Upstream tarball `623bf5776d9069ddae371ac9a1bd1914d70107b8`, matching the registry's `dist.shasum` for 6.3.2. 6.3.2 is `latest` on npm, so this pins to current upstream, not to something already behind.

---

### 1. All five tables — verified, including the three the diff does not touch

The escaping is where this goes wrong silently, so, concretely: the Kotlin side carries **1093 `\$` escapes and one `⚠`** across the five literals; the JavaScript side carries **zero escapes**. A byte comparison of the literal text would have reported 1094 false mismatches and a 6512-vs-5574 length disagreement on `stateData` alone. I decoded each side to its actual character sequence first — Kotlin with an explicit decoder implementing exactly Kotlin's escape set (`\t \b \n \r \' \" \\ \$ \uXXXX`, erroring on anything else), JavaScript by locating each literal with a backslash-aware scanner and evaluating that literal in node, which is authoritative for JS.

| table | PR head | upstream 6.3.2 | sha256[:16] |
|---|---|---|---|
| `states` | 3589 | 3589 | `67f7965ee9b54fa9` — **equal** |
| `stateData` | 5574 | 5574 | `93f2afb85dd328f8` — **equal** |
| `goto` | 2452 | 2452 | `65cffe903e6a1850` — **equal** |
| `nodeNames` | 1237 | 1237 | `e3438c7d11042502` — **equal** |
| `tokenData` | 636 | 636 | `ff93c511c7cf4a9c` — **equal** |

`goto` = 2452, matching the changelog's own figure. Cross-check that my decoding is not itself wrong: the *compiled* Kotlin parser reproduces upstream's trees on 17 inputs (§3), which a decoding error would have to coincidentally cancel out.

### 2. Scope — nothing else moved

The production diff is a single hunk, `@@ -101,2 +101,2 @@`: `stateData` and `goto`, nothing else in `LiquidParser.kt`. I also checked the non-string spec fields against upstream rather than only against the parent: `version: 14`, `maxTerm: 220`, all three `nodeProps` rows, `skippedNodes: [0,6]`, `repeatNodeCount: 11`, `tokenizers` order, `topRules: {"Template":[0,7]}`, the 29 `dynamicPrecedences` entries (190–220 minus 193 and 198), `specialized` terms 187/39, `tokenPrec: 0` — all identical to 6.3.2. So are the two specializer maps that feed them: `specIdentifier` (35 entries) and `specTagName` (29) are entry-for-entry identical to upstream's `spec_identifier` and `spec_TagName`. That was the other place a hand-transcription slip would hide, and it is clean.

### 3. The oracle is genuinely upstream — this was the thing worth checking

I ran the **JavaScript** parser from `@codemirror/lang-liquid@6.3.2` under node over the 17 inputs, extracted mechanically from `LiquidParserTest.kt` rather than retyped, with `treeToString`/`errorCount` ported line-for-line (same `\W` skip rule, same `⚠`, same `[from,to]` rendering). Upstream reproduces **17 of 17 expectations character-for-character**, including both `withPositions` cases and the `parseMixed` one.

So the expectations are not the new Kotlin parser's own output blessed as correct. They are upstream's, and they agree.

### 4. Mutation — the tests do pin the fix

`:lang-liquid:jvmTest`, the task CI actually runs, JDK 21, each arm restored with `git checkout HEAD --` and confirmed clean afterwards:

| arm | result |
|---|---|
| PR head, unmodified | `tests=17 failures=0 errors=0 skipped=0` |
| **A** — parent's `stateData` only (2 chars), `goto` left correct | `tests=17 failures=2` |
| **B** — parent's `stateData` + `goto` | `tests=17 failures=16` |

Arm B fails with `ArrayIndexOutOfBoundsException` / `IndexOutOfBoundsException` on 16 of 17 inputs, which is the reported "throws on essentially every input" reproduced. The lone survivor is `parsesEmptyDocument` — see the nit below.

### 5. The zero-error-node case — confirmed, and confirmed *for that reason*

Arm A isolates the two `stateData` characters, and it fails **exactly** the two tests the changelog says cover them, with exactly the claimed signature:

```
expected: errors=0 Template(Tag(liquid,CaseDirective(...,Tag(render,StringLiteral,
          RenderParameter(VariableName,NumberLiteral)),...,EndTag(endcase))))
but was:  errors=0 Template(liquid,Tag(case,VariableName),Tag(when,NumberLiteral),
          render,StringLiteral,Text(tablerow(VariableName,NumberLiteral)))
```

`errors=0` on **both** sides. A "did it produce an error node?" check reports success here. Only the tree-shape assertion catches it, exactly as claimed.

I also decoded the two `stateData` arrays with lezer's own `decodeArray` and diffed them: the corrupt table differs in precisely **two** decoded entries, at indices 2453 and 3423, each holding **143** where upstream holds **97**. In `nodeNames`, 143 is `tablerow` and 97 is `RenderParameter`. The description's characterisation is literally accurate, down to the term ids. The same `difflib` alignment on `goto` confirms the rest verbatim: a dropped character at 172, a dropped 108-character run at 475, a spurious 4-character insertion at old-index 1168 (`&Z&^`), and 752 characters missing from the tail.

### 6. Follow-up commit `b7384d4`

Not in my brief, but it is the current head and approving the older SHA while ignoring it would be the stale-approval trap. It adds `:lang-liquid:wasmJsTest` to `ci.yml` and `wasmJsTests.set(true)`, matching 29 other modules. Its claim is the #197-aware one — a `wasmJsTest` running zero tests still exits 0 — so I checked it the same way the commit message says it was checked, by reading the XML:

```
wasmJsBrowserTest/...LiquidParserTest.xml  tests="17" skipped="0" failures="0" errors="0"
```

Verified at `b7384d4` on a headless Chromium runner. `BUILD SUCCESSFUL` alone would not have shown this; the count does.

---

### Non-blocking nits

**(a) The commit message slightly over-claims which tables pin the source.** It says `states`, `nodeNames` and `tokenData` are byte-identical "to 6.3.2 and to no other release". I compared all ten published releases (6.0.0 → 6.3.2): `nodeNames` and `tokenData` are *also* byte-identical in **6.3.1**. Only `states` is unique to 6.3.2. The conclusion is untouched — `states` alone pins 6.3.2 and no other release, and all five match 6.3.2 and only 6.3.2 — but two of the three named tables are not discriminating, and the changelog fragment (which is what ships) makes no such claim. Worth a one-line correction if the message is still editable; not worth a respin.

**(b) `parsesEmptyDocument` guards nothing about the tables.** It is the single test that passes in *both* corruption arms — the empty input reduces straight to `Template` without consulting the corrupt regions. It is a reasonable boundary case to keep; it just should not be counted toward the suite's table coverage. 16 of 17 do the pinning; 2 of 17 pin `stateData`.

### What I did not verify

- Only `jvmTest` (the gated task) and one `wasmJsTest` run. Native and Apple targets not exercised; nothing in this diff is platform-specific, but I did not run them.
- `ktlintCheck` / `spotlessCheck` / `apiCheck` not run — I checked by hand that both new files carry the licence header and that no line exceeds the 100-column `.editorconfig` limit, but that is inspection, not the task.
- The `parseMixed` test's HTML overlay is not traversed by `Tree.iterate` in either implementation, so that test pins the Liquid skeleton around the overlay, not the HTML inside it. That matches upstream's behaviour and is not a defect — just not coverage of lang-html.

### OUTSIDE THE BRIEF

- **For whoever reviews #372, #374 and the fourth table-restore PR — as a question, not a finding.** I reviewed #373 only; nothing here was assumed about the others and nothing found here is asserted about them. Two things were cheap here and would be cheap there: (i) the *unchanged* tables are the ones nobody has looked at, and checking them costs one npm pack and a decode — in this PR they were clean, which is a data point about this PR and not about any other; (ii) the oracle question only has a real answer if the upstream JS parser is actually run. If a sibling PR's tables are corrupt in a way that produces a *wrong tree with zero error nodes*, as `stateData` did here, then an error-node count, a "does it parse without throwing" smoke test, and expectations captured from the new parser would all report success. The decode-then-compare scripts I used are generic over `ParserSpec`; they only needed the package name changed.
- **A durable guard is possible and is not in this PR.** Everything above is one script: decode the five literals out of a Kotlin `ParserSpec`, decode them out of the pinned npm tarball, compare. Run per `:lang-*` module in CI, it turns "was the transcription faithful?" from a thing a reviewer does by hand into a thing that cannot regress — and it would have caught this before it shipped, and would catch the next one. Worth an issue; not a change to make in this PR.
- **`TreeTestUtil` looks reusable across the `:lang-*` modules.** It currently sits in `lang-liquid`'s `commonTest`. If the sibling PRs each grow their own copy, a shared test-fixture module would be worth considering — deliberately flagging it as a question for later rather than as a change request here, since pulling it up now would couple four in-flight PRs to each other.

PR #374, opened 2026-09-19
Restore the twelve characters dropped from :lang-less's states table (#368)

### LINKED ISSUE #368: :lang-less crashes on `@import` — twelve characters missing from the transcribed states table, module has no tests

**Category:** test-gap / bug (demonstrated)

`:lang-less` has **zero test sources**. It parses most Less correctly, but `@import` — one of the most common statements in any Less file — throws.

### Reproduction

```
LESSPROBE  smoke-plain-css   errors=0  StyleSheet(RuleSet(TagSelector(TagName),Block(...)))
LESSPROBE  variable          errors=0  StyleSheet(Declaration(AtKeyword,ColorLiteral),...)
LESSPROBE  mixin             errors=0  ...
LESSPROBE  nesting           errors=0  ...
LESSPROBE  guard             errors=0  ...
LESSPROBE  import            THREW IndexOutOfBoundsException: Index -3 out of bounds for length 3
LESSPROBE  detached-ruleset  errors=0  ...
LESSPROBE  namespace         errors=0  ...
```

Input: `@import "library";\n@import (reference) "foo.less";`

```
at com.monkopedia.kodemirror.lezer.lr.Stack.reduce$com_monkopedia_kodemirror_lezer_lr(Stack.kt:122)
at com.monkopedia.kodemirror.lezer.lr.Parse.advanceStack(Parse.kt:425)
```

(`:lang-css` has a passing `parsesImport` test, so the gap is specific to the Less tables.)

### Root cause — twelve characters missing from the transcribed `states`

`lang-less/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/less/LessParser.kt:106` (`states`).

Diffed against upstream `@codemirror/lang-less@6.0.2`:

| field | verdict |
|---|---|
| `stateData` | byte-identical (2969) |
| `goto` | byte-identical (1129) |
| `nodeNames` | byte-identical (1089) |
| `tokenData` | byte-identical (5572) |
| **`states`** | **port 2603, upstream 2615 — `1G0d1G0dOOQO` missing at index 2525** |

Four of five tables matching byte-for-byte rules out a grammar-version difference.

```
upstream:  ...O-E8j-E8jOOQO<<Ih<<IhOOQO1G0d1G0dOOQOAN=YAN=YOOQPAN>UAN>U...
port:      ...O-E8j-E8jOOQO<<Ih<<IhOOQOAN=YAN=YOOQPAN>UAN>U...
```

Note the dropped run sits between two `OOQO` occurrences — an eye-skip over a repeated token.

### Proposed

In `LessParser.kt`, change

```
O-E8j-E8jOOQO<<Ih<<IhOOQOAN=YAN=Y
```

to

```
O-E8j-E8jOOQO<<Ih<<IhOOQO1G0d1G0dOOQOAN=YAN=Y
```

**Verified:** with that edit, the `import` sample parses cleanly and the other nine are unchanged:

```
LESSPROBE  import  errors=0  StyleSheet(ImportStatement(import,StringLiteral),ImportStatement(import,ValueName,StringLiteral))
```

Then add a test source set (the module has none).

**Why:** a published language module that crashes on `@import`, and nothing in the repo could have noticed.

**Effort:** mechanical (one string edit + a test file).

monkopedia-coder:
`@import` — one of the most common statements in any Less file — threw `IndexOutOfBoundsException: Index -3 out of bounds for length 3` in the reducer. The transcribed `states` literal was missing the run `1G0d1G0dOOQO` at index 2525, an eye-skip between two adjacent `OOQO` occurrences.

### One thing the issue did not report

Writing the tests first turned up a **second, silent** consequence of the same missing run: `@media screen and (max-width: 100px) { … }` does not throw, it **misclassifies**. Port, before the fix:

```
StyleSheet(LineComment,Comment,media,ImportStatement(KeywordQuery,LogicOp,FeatureName,NumberLiteral(Unit)),⚠,RuleSet(…),⚠,⚠)
```

upstream:

```
StyleSheet(LineComment,Comment,MediaStatement(media,BinaryQuery(KeywordQuery,LogicOp,FeatureQuery(FeatureName,NumberLiteral(Unit))),Block(RuleSet(…))))
```

Same root cause, same one-line fix — so it is not filed separately — but it is the reason these tests assert the tree and not "did not throw". A smoke suite would have passed on that input.

### Upstream version matched

`@codemirror/lang-less` **6.0.1 / 6.0.2**. I `npm pack`ed all three published 6.x releases and extracted the five tables from each: 6.0.1 and 6.0.2 carry **byte-identical** tables, and 6.0.0 differs in all five (first divergence at index 0 of every table). So the tables cannot distinguish 6.0.1 from 6.0.2, and the restore is verified byte-identical against **both**. 6.0.0 is excluded.

### Table diff (before)

| field | verdict |
|---|---|
| **`states`** | **port 2603, upstream 2615 — diverges at index 2525** |
| `stateData` | identical (2969) |
| `goto` | identical (1129) |
| `nodeNames` | identical (1089) |
| `tokenData` | identical (5572) |

```
port     ...O-E8j-E8jOOQO<<Ih<<IhOOQOAN=YAN=Y...
upstream ...O-E8j-E8jOOQO<<Ih<<IhOOQO1G0d1G0dOOQOAN=YAN=Y...
```

The splice point (`<<Ih<<IhOOQO`) occurs exactly once in both tables, so the edit is unambiguous.

### After

All five tables byte-identical to upstream 6.0.1 and 6.0.2. Ten sample Less inputs (plain CSS, variable, mixin + inclusion, nesting, guard, `@import`, detached ruleset, namespace, escape + arithmetic, comments + media query) run through both the port and upstream in node produce **identical preorder node lists** (`name@from-to`) — `diff` empty.

### Tests

The module had **zero test sources**. `LessParserTest` asserts the concrete tree shape and an error count of 0 for each of those ten inputs, every expectation taken from upstream's tree for the same bytes.

- **Red first:** `10 tests completed, 2 failed` — `parsesImport` with the `IndexOutOfBoundsException`, `parsesCommentsAndMediaQuery` with the misclassification above.
- **Mutation:** reverting only `LessParser.kt` to `HEAD` with the tests in place (verified landed via `git diff HEAD --quiet`) takes exactly those 2 red again. The other 8 stay green — they exercise paths the missing `states` run does not reach, so they are baseline coverage for a module that had none, not guards for this defect. Stating that explicitly rather than implying all ten guard it.
- **Green:** `TEST-...LessParserTest.xml` reports `tests="10" failures="0" errors="0"`.

`:lang-less:check` and `apiCheck` green locally; `.api` did not move.

Fixes #368


monkopedia-reviewer (approved):
**approve**

Reviewed as the independent party (hauler side). Product delta is two lines; the rest is test.

### Head moved during review — what was read

The deep pass ran at `090fef1d`; head is now `64c2f4f2`. I compared the two **by blob SHA via the git trees API**, not by `git diff` — the old head is orphaned by the force-push and never fetches into a fresh clone, so a local diff returns empty, which is indistinguishable from "no change". Of 1069 blobs, **exactly one differs**: `TreeTestUtil.kt:44`, literal `⚠` → `⚠`. Same character, same line count, no behavioural change — and it resolves nit 3 below, which was raised against the old head.

This review is pinned to `64c2f4f2`, because I re-derived the central claim against that tree myself.

### The twelve characters are the right twelve

Verified against upstream `@codemirror/lang-less` **6.0.2** — the version `docs/codemirror-upstream.md:93` pins for this module — fetched from npm:

- `states` decoded: **2603 → 2615**, and 2615 is byte-identical to upstream's.
- Pure insertion of `1G0d1G0dOOQO` at decoded index **2525**; prefix and suffix both preserved.
- One gotcha for the record: the Kotlin literal reads **2629** chars, not 2615, because it carries 14 `\$` escapes. Any future check that compares literal lengths rather than decoded ones will report a spurious 14-char divergence. Decode before comparing.

Two corrections to the framing this was dispatched under, both confirmed against npm directly:
- **`@lezer/less` does not exist** (404). Less bundles its grammar inside `@codemirror/lang-less`. Same for sql, vue, angular, wast, liquid, jinja — so for these seven the `@codemirror/lang-*` pin *is* the grammar pin.
- Provenance is therefore **recorded, not missing**, for this module. The concern that there'd be no pin to separate "mis-transcribed" from "upstream moved" doesn't apply here.

### Scope, suite, and the remaining caveats

The cross-module sweep and the revert test were run by a delegated agent, and I'm flagging them as such rather than presenting them as my own: 17 of 22 `lang-*` modules byte-identical to upstream, the 4 divergent ones exactly the 4 with open sibling PRs, and an exact revert producing `10 tests completed, 2 failed` with #368's literal `IndexOutOfBoundsException: Index -3 out of bounds for length 3`. That last point is the one that matters — the suite has a demonstrated RED, so it is not a gate that cannot fail. The PR says only 2 of 10 cases guard this defect, which is accurate and doesn't overclaim.

Worth carrying forward: the corruption across the batch is **not** one signature. Less and wast are chunk skips; grammar and liquid's `stateData` are single-character slips; jinja's `goto` is a **tail truncation** (an exact 266-char prefix of upstream's 2571). A detector built only for homeoteleuton would miss half of them.

### Findings — all minor, none blocking

- **LOW · PR description** — "first divergence at index 0 of every table" for 6.0.0 is loose: 0 for `states`/`goto`, but 1, 336 and 386 for `stateData`, `nodeNames`, `tokenData`. The conclusion (6.0.0 excluded) holds; only the evidence sentence overstates.
- **LOW · `lang-less/build.gradle.kts`** — no `wasmJsTests` opt-in (14 of 22 siblings have it), so the new suite never runs on wasm. Pre-existing and the table is platform-independent, but the module has tests now, so it's a cheap follow-up.
- **NIT · `LessParserTest.kt`** — `assertEquals(0, errors(src))` is redundant with the expected string. Harmless.
- **NIT · `TreeTestUtil.kt:44`** — ~literal `⚠` where 16 siblings use `⚠`~ **already fixed at `64c2f4f2`.**
- **Out of scope, worth its own issue** — `docs/codemirror-upstream.md` records only the `@codemirror/lang-*` version, but six modules (cpp, css, java, php, python, grammar) take their grammar from separately-versioned `@lezer/*` packages via floating `^1.x`. Re-verifying those against the recorded version alone yields false positives. It doesn't affect lang-less, but it's what makes this verification class fragile in general.

### Not established

`apiCheck` (missing Android SDK here; the diff touches no `.api` file), and native/wasm test execution — JVM only.

Not merging this one: it's kodemirror's call.

Issue #369, opened 2026-09-14
:lang-python's states table has one wrong character: state 0 loses its Accepting flag, so every empty/blank/comment-only document gets a spurious error node


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** bug (demonstrated) — one wrong character in a published parser table

`:lang-python`'s `states` table differs from upstream `@lezer/python` by **a single character at index 3**, and the visible consequence is a spurious syntax-error node on every **empty, blank, or comment-only** Python buffer.

### Where

`lang-python/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/python/PythonParser.kt` — the `states` literal.

```
port:      "##jO`QeOOP$}OSOOO&WQtO'#HUOOQS'#Co'#Co...
upstream:  "##jQ`QeOOP$}OSOOO&WQtO'#HUOOQS'#Co'#Co...
                ^ index 3: port O (decodes to 0), upstream Q (decodes to 2)
```

`states[0]` is state 0's flag word; `2` is `StateFlag.Accepting`. The port clears it.

This is a transcription error, not a version difference — the other four tables are **byte-identical** to current `@lezer/python`:

| field | verdict |
|---|---|
| **`states`** | **same length (8004), one character differs at index 3** |
| `stateData` | byte-identical (11152) |
| `goto` | byte-identical (6199) |
| `nodeNames` | byte-identical (1659) |
| `tokenData` | byte-identical (5585) |

### Demonstrated consequence

Differential run over 45 CPython 3.14 stdlib files plus 5 adversarial inputs, comparing the port's full preorder node list (`name@from-to`) against `@lezer/python` parsing the same bytes:

```
zz_empty.py     DIFFER port=2 up=1   port emits ⚠@0-0   upstream: none
zz_blank.py     DIFFER port=2 up=1   port emits ⚠@3-3   upstream: none
zz_comment.py   DIFFER port=3 up=2   port emits ⚠@17-17 upstream: none
functools.py    DIFFER (ParamList@21374-21374 vs @21375-21375)   <- unrelated, see below
== 4 of 50 files differ ==
```

So opening an **empty editor**, or a file containing only a comment, yields `Script(⚠)` — a syntax error the user can see (lint/diagnostic decoration, fold and indent behaviour all key off it).

### Verified fix

Changing that one character `O` → `Q`:

```
== 1 of 50 files differ ==        (was 4)
```

[...]

PR #375, opened 2026-09-19
Make :lang-grammar usable: quote the styleTags `*` and restore a stateData character (#367)

### LINKED ISSUE #367: :lang-grammar throws ExceptionInInitializerError on first touch (unquoted `*` in styleTags) and AIOOBE on @skip (one char missing from stateData) — no tests

**Category:** test-gap / bug (demonstrated) — **two independent defects**

`:lang-grammar` has **zero test sources**. Touching *any* public entry point (`lezerParser`, `lezerGrammarLanguage`, `lezerGrammar()`, `lezerHighlighting`) throws before parsing begins.

### Defect 1 — unquoted `*` in the highlight spec kills static init

`lang-grammar/src/commonMain/kotlin/com/monkopedia/kodemirror/lang/grammar/LezerHighlight.kt:46`

```kotlin
"\"!\" ~ * + ? |" to t.operator
```

Upstream `@lezer/lezer@1.1.1` `src/highlight.js` has it as:

```js
'"!" ~ "*" + ? |': highlight.tags.operator
```

The `"` around `*` were lost. In a `styleTags` selector a bare `*` is the **wildcard**, which maps to an empty path piece, and `styleTagsImpl` correctly rejects an empty final piece (`Highlight.kt:162`, `require(inner.isNotEmpty())` — this matches upstream's `if (!inner) throw new RangeError(...)`, so **the `:lezer-highlight` port is not the bug; do not "fix" it there**).

Result:

```
THREW ExceptionInInitializerError
  Caused by: java.lang.IllegalArgumentException: Invalid path: *
    at ...lezer.highlight.HighlightKt.styleTagsImpl(Highlight.kt:162)
    at ...lang.grammar.LezerHighlightKt.<clinit>(LezerHighlight.kt:24)
    at ...lang.grammar.LezerParserKt.<clinit>(LezerParser.kt:99)
```

and every subsequent touch gives `NoClassDefFoundError: Could not initialize class ...LezerParserKt`. **The module cannot be used at all.**

Note the `!` in the same selector *is* correctly quoted — only the `*` was dropped.

**Fix:** `"\"!\" ~ \"*\" + ? |" to t.operator`

### Defect 2 — one character missing from `stateData`

With defect 1 fixed, parsing still crashes on `@skip` scopes and on node-prop syntax:

```
GRAMMARPROBE  skip   THREW ArrayIndexOutOfBoundsException: Index 393211 out of bounds for length 1284
GRAMMARPROBE  props  THREW ArrayIndexOutOfBoundsException: Index 393211 out of bounds for length 1284
```

The module was ported from `@lezer/lezer@1.1.1` (confirmed by downloading 1.0.0 / 1.1.0 / 1.1.1 / 1.1.2 and comparing all five data tables). Against 1.1.1:

| field | verdict |
|---|---|
| `states` | byte-identical (2336) |
| `goto` | byte-identical (970) |
| `nodeNames` | byte-identical (1000) |
| `tokenData` | byte-identical (1630) |
| **`stateData`** | **port 1474, upstream 1475 — one `l` missing at index 1177** |

```
upstream:  ...O]$xO~O!l$yO~O_${O!RUO#fPO~O...
port:      ...O]$xO~O!$yO~O_${O!RUO#fPO~O...
```

**Fix:** in `LezerParser.kt`, `O]\$xO~O!\$yO~O_\${O` → `O]\$xO~O!l\$yO~O_\${O`.

### Verified

With both edits applied, all 10 probe samples behave **identically to upstream `@lezer/lezer@1.1.1`** (verified by running the same 10 inputs through the upstream parser via node and comparing node sequences and error counts — the two samples that still report error nodes, `@external tokenizer ...` and `@specialize<...>`, produce the *same* errors upstream because my sample syntax was malformed, not the parser):

```
GRAMMARPROBE  simple-rule   errors=0  Grammar(RuleDeclaration(RuleName,Body(Repeat(RuleName))),...)
GRAMMARPROBE  skip          errors=0  Grammar(SkipScope(Body(Choice(RuleName,RuleName)),SkipBody(...)))
GRAMMARPROBE  props         errors=0  Grammar(RuleDeclaration(RuleName,Body(RuleName)),...)
GRAMMARPROBE  local-tokens  errors=0  Grammar(LocalTokensDeclaration(tokens,TokensBody(...),ElseToken(...)))
```

**Why:** a published module where the *first* line of consumer code throws. Two distinct one-token fixes.

**Effort:** mechanical (two string edits + a test source set).

monkopedia-coder:
Two independent transcription defects. Both had to go before the module could parse anything.

### Defect 1 — unquoted `*` killed static init

`LezerHighlight.kt:46` had `"\"!\" ~ * + ? |"` where upstream `@lezer/lezer` has `'"!" ~ "*" + ? |'`. A bare `*` in a `styleTags` selector is the **wildcard**, which yields an empty final path piece; `styleTags` rejects that, so `lezerHighlighting`'s static initializer threw and took `lezerParser` — which references it — with it. Every public entry point (`lezerParser`, `lezerGrammarLanguage`, `lezerGrammar()`, `lezerHighlighting`) threw `ExceptionInInitializerError` on first touch, `NoClassDefFoundError` after.

The `!` in the same selector was already correctly quoted; only the `*` was dropped. **`:lezer-highlight` is not the bug** — its `require(inner.isNotEmpty())` is a faithful port of upstream's `if (!inner) throw new RangeError(...)`, and it is left alone here.

### Defect 2 — one character missing from `stateData`

`stateData` was one `l` short at index 1177, crashing `@skip` scopes, node props, `@local tokens`, `@external` declarations and `@specialize`.

```
port     ...O]$xO~O!$yO~O_${O!RUO#fPO~O...
upstream ...O]$xO~O!l$yO~O_${O!RUO#fPO~O...
```

### Upstream version matched

`@lezer/lezer` **1.1.1** (equivalently 1.1.2). Determined by `npm pack`ing 1.0.0, 1.1.0, 1.1.1, 1.1.2 and 1.1.3 and extracting all five tables from each: 1.1.1 and 1.1.2 carry **byte-identical** tables and match the port on four of five; 1.0.0, 1.1.0 and 1.1.3 differ in all five. The restore is verified byte-identical against both 1.1.1 and 1.1.2.

| field | before | after |
|---|---|---|
| `states` | identical (2336) | identical |
| **`stateData`** | **port 1474, upstream 1475 — diverges at 1177** | identical (1475) |
| `goto` | identical (970) | identical |
| `nodeNames` | identical (1000) | identical |
| `tokenData` | identical (1630) | identical |

The splice context `O]$xO~O!$yO~O` occurs exactly once, so the edit is unambiguous.

### After

All five tables byte-identical to upstream. Ten sample grammar inputs (simple rules, `@skip` scope, node props, `@local tokens`, `@precedence` + markers, `@tokens`, `@external tokens`/`prop`, `@specialize`/`@extend`, `@dialects`, character classes) run through both the port and upstream `@lezer/lezer` 1.1.1 in node produce **identical preorder node lists** (`name@from-to`) — `diff` empty.

### Tests

The module had **zero test sources**. Added `LezerGrammarParserTest` (10 concrete tree-shape assertions + error count 0, expectations taken from upstream) and `LezerGrammarEntryPointTest` (4 tests, one per public entry point, because defect 1 was reached independently by each).

- **Red first, staged:**
  - unfixed: `10 tests completed, 10 failed` — `ExceptionInInitializerError` / `Invalid path: *`, then `NoClassDefFoundError`.
  - **defect 1 alone fixed:** `10 tests completed, 5 failed` — `ArrayIndexOutOfBoundsException: Index 393211 out of bounds for length 1284`. This isolates defect 2 and shows neither fix is redundant.
- **Mutation, one file at a time** (each verified landed via `git diff HEAD --quiet`, the other file confirmed to still carry its fix):
  - revert `LezerHighlight.kt` only → **14 of 14 red** (`Invalid path: *`).
  - revert `LezerParser.kt` only → **5 red** (`parsesSkipScope`, `parsesNodeProps`, `parsesLocalTokens`, `parsesExternalDeclarations`, `parsesSpecializeAndExtend`). The other 9 stay green — they exercise paths the missing `stateData` character does not reach, so they are baseline coverage, not guards for defect 2. Saying so rather than implying all 14 guard both.
- **Green:** `LezerGrammarParserTest.xml` `tests="10" failures="0" errors="0"`; `LezerGrammarEntryPointTest.xml` `tests="4" failures="0" errors="0"`.

`:lang-grammar:check` and `apiCheck` green locally; `.api` did not move.

Fixes #367


monkopedia-reviewer (approved):
## Review — APPROVE (this is a review, **not a merge**; merge authority for this repo is kodemirror's, not mine)

Reviewed at **`5af7ad4ec7bc77c533fd9d101a8e80844f4df10b`**, which is where `commit_id` on this review object is pinned. I was originally briefed at `c455d0c8`; the head moved mid-review. Delta `c455d0c8..5af7ad4e` is **one line, in test code only**: `TreeTestUtil.kt` changed the literal `"⚠"` to the escape `"⚠"`. `git diff c455d0c8..5af7ad4e -- lang-grammar/src/commonMain` is **empty** — the product change is byte-identical across the two SHAs, so every measurement below holds at both.

Product delta is two characters across two lines; the other 304 lines are new tests.

---

### 1. The decisive test: four-way revert matrix

Each product change reverted **independently** to its `main` content, `:lang-grammar:jvmTest` re-run each time, test-results XML deleted before every run so a build that did not happen reads as absent rather than as the previous pass. Counts are **executed** tests (`tests − skipped`), and the denominator is 14 in every cell.

| cell | `LezerHighlight.kt` | `LezerParser.kt` | executed | failures | result |
|---|---|---|---|---|---|
| **positive control** | PR | PR | 14 | 0 | **GREEN** |
| revert defect 1 only | `main` | PR | 14 | **14** | **RED** |
| revert defect 2 only | PR | `main` | 14 | **5** | **RED** |
| revert both | `main` | `main` | 14 | **14** | **RED** |

**Both product changes are independently load-bearing.** The positive control matters here: without it, "everything failed" would also have produced three RED cells.

- Defect 1 reverted → `ExceptionInInitializerError` on `highlightingInitializes`, cause **`Invalid path: *`**, and `NoClassDefFoundError` on the other 13.
- Defect 2 reverted → exactly 5 failures: `parsesSkipScope`, `parsesNodeProps`, `parsesLocalTokens`, `parsesExternalDeclarations`, `parsesSpecializeAndExtend`, with `ArrayIndexOutOfBoundsException: Index 393211/393214 out of bounds for length 1284` and `IndexOutOfBoundsException: Index -3 out of bounds for length 3`.

Those 5 are precisely the five constructs the changelog names. The changelog's specificity is earned, not asserted.

Tree restored afterwards with `git checkout HEAD -- <path>`; `git diff HEAD` and `git status --porcelain` both clean at `5af7ad4e`.

---

### 2. Do the 304 new test lines assert correct behaviour, or only absence of an exception?

**Correct behaviour, and verified against upstream rather than taken on trust.** The file comment claims the expectations came from `@lezer/lezer` 1.1.1. I did not spot-check two — I mechanically decoded **all 10** `src`/expected pairs out of the Kotlin source, reimplemented `treeToString`/`errorCount` in JS against the same semantics, and ran them through the real upstream 1.1.1 package:

**10 / 10 of the PR's expectations reproduce exactly on upstream `@lezer/lezer` 1.1.1, with 0 error nodes each.** The expectations encode upstream's behaviour, not the port's.

**Control, because a differential that cannot fail is not a differential.** Same harness, same 10 inputs, against an upstream copy damaged with *`main`'s exact defect* (the same `l` removed):

| outcome | n / 10 |
|---|---|
| still MATCH (input doesn't exercise the damaged states) | 5 |
| non-terminating parse (>20s) | 4 |
| DIFFER, 2 error nodes | 1 |

So the method discriminates, and the 5 sensitive inputs are the **same five** the Kotlin matrix flagged. Two independent methods, two languages, same five — that is as good as this gets.

(Incidental but useful if anyone reproduces upstream-side: the damaged table **hangs** in JS and **throws** in Kotlin. JS reads out of range on a `Uint16Array` and gets `undefined`; Kotlin's `ShortArray` bounds-checks. Same defect, different symptom.)

**Can `TreeTestUtil.kt` express a wrong answer?** Yes — demonstrated, not assumed: the damaged-table run produced `...ExternalTokensDeclaration(tokens,Name,from,Literal,⚠),RuleDeclaration(RuleName,Body(⚠))...` with `errorCount == 2`. The `assertEquals(0, errors(src))` line on every test carries real signal. See "Outside the brief" for the one place its resolution is lower than its test name implies.

---

### 3. Defect 1's behavioural breadth — the premise is false, which settles it

The dispatch asked whether anything could have depended on the bare `*` acting as a wildcard that styled everything `t.operator`. **It never did that.** Measured against upstream `@lezer/highlight`:

```
styleTags({... '"!" ~ "*" + ? |': t.operator})  ->  OK
styleTags({... '"!" ~ *   + ? |': t.operator})  ->  RangeError: Invalid path: *
styleTags({ '*': t.operator })                  ->  RangeError: Invalid path: *
```

Mechanism, from `@lezer/highlight`'s path parser: `*` is translated to an **empty** path piece, then `let inner = pieces[last]; if (!inner) throw new RangeError("Invalid path: " + part)`. A wildcard is only legal in a **non-final** position (`*/Foo`); as an entire path it is always rejected. The Kotlin port reproduces this — the reverted cell fails with the same `Invalid path: *` message.

So the rule never styled anything; it threw at spec-construction time. And the throw is **unconditional across every entry point**: `lezerParser` lists `propSources = listOf(lezerHighlighting)` and `lezerGrammarLanguage`/`lezerGrammar()` chain off `lezerParser`, so all 4 entry-point tests plus all 10 parser tests fail in the reverted cell — 14/14, no path survives. **Nothing could have relied on the old behaviour, because there was no old behaviour and no way to load the module.** No compatibility question to answer.

---

### 4. Category and tier, re-derived from this diff

- **Category: port-fidelity bug fix.** Not a feature, not a refactor.
- **Public API delta: none.** No declaration is added, removed, or re-signed; both edits change the *contents* of string literals inside pre-existing `val` initializers. An `.api`/BCV dump records signatures, not initializer values, so it is byte-identical by construction. Per "tier by API delta, not path", that is **tier 1**.
- **Risk: minimal, and the "wholly unusable" fact lowers it rather than raising it.** `ExceptionInInitializerError` on first touch means there is no working behaviour to regress against and no consumer that can currently be depending on this module. Blast radius is confined to `:lang-grammar`; no other module is touched.

I'd have argued tier 1 even without the API-delta rule: 2 characters, restoring byte-identity with a pinned upstream, in a module that presently cannot be loaded.

---

### ⚠️ Outside the brief

**(a) Pre-existing divergence in the same `styleTags` map — NOT introduced here, and I am not asserting it is a bug.** Comparing all 18 entries against upstream `src/highlight.js` 1.1.1, exactly one differs:

```
upstream : "tokens from grammar as empty prop extend specialize"           -> t.keyword
this port: "tokens from grammar as empty prop extend specialize AtName"    -> t.keyword
```

`AtName` is a real node type. Upstream routes specific at-names to `t.modifier` / `t.definitionKeyword` / `t.operatorKeyword` and deliberately leaves the remainder unstyled; this port adds a blanket `t.keyword` fallback for **all** `AtName`s. My read is that this looks **deliberate** — it is an addition at the end of an otherwise byte-faithful line, not the kind of character-level slip defect 2 is — but it is undocumented, it is a real behavioural divergence from upstream, and nothing in the repo records the decision. Out of scope for this PR; worth its own issue so the next fidelity sweep doesn't rediscover it as a defect.

**(b) `TreeTestUtil` drops every `@`-prefixed node type, and one test's name overpromises because of it.** `shouldIgnore` rejects any name matching `\W`, so `@top`, `@specialize`, `@extend`, `@local`, `@external`, `@else` … are all skipped along with their subtrees. Measured consequence, on upstream:

```
Kw { @specialize<Identifier,"if"|"else"> | @extend<Identifier,"get"> }   ->  Grammar(...Specialization(ArgList(...)),Specialization(ArgList(...)))
Kw { @extend<Identifier,"if"|"else"> | @specialize<Identifier,"get"> }   ->  BYTE-IDENTICAL rendering
Kw { @extend<...> | @extend<...> }                                        ->  BYTE-IDENTICAL rendering
```

`parsesSpecializeAndExtend` therefore **cannot distinguish `@specialize` from `@extend`**, though both are distinct node types and the test's name claims to cover both. Likewise `@top Program {...}` and `Program {...}` render identically. Dropping *token text* is by design and documented ("the grammar's structure rather than its tokens"); dropping whole **node types** is a step further than the doc comment implies. Not blocking — the `errorCount == 0` assertion still catches the damage this PR fixes, and the matrix proves the test is load-bearing — but a follow-up that keeps `@`-named types (or asserts on them separately) would close the gap cheaply.

**(c) The pin lives only in test code, and the table cannot discriminate 1.1.1 from 1.1.2.** `LezerParser.kt` records no upstream version anywhere; "1.1.1" appears only in the doc comment of the test file this PR adds. Separately: 1.1.1 and 1.1.2 have **byte-identical** `stateData` (1475 chars each), so the restored table is equally consistent with either. "Faithful to 1.1.1" and "faithful to 1.1.2" are the same claim here. That is not a defect in this PR, but it is worth saying plainly: a fidelity guard answers *"does it match its pin?"*, never *"is that pin still right?"* — and right now the product file states no pin for either question to be asked against.

**(d) Provenance of defect 2 — genuine mis-copy, not version skew.** Flagged because `:lang-python` (#369) looked like this and turned out to be upstream's own bug against a faithfully-transcribed pin. It is not the case here. A sweep of all 7 published `@lezer/lezer` versions (0.15.0, 0.16.0, 1.0.0, 1.1.0, 1.1.1, 1.1.2, 1.1.3) finds `main`'s `stateData` matching **0 of 7**. I add one complementary result: on `main`, `states`, `goto` and `tokenData` are each **byte-identical to 1.1.1**, and only `stateData` is one character short. A version-skew explanation would have to account for three tables from 1.1.1 and a fourth from nowhere. Hand-transcription slip, confirmed from two directions.

**(e) Index reconciliation.** The inserted `l` is at index **1177 of the decoded table** (my measurement agrees with the PR's doc comment) and at index **1200/1201 of the escaped Kotlin source literal** — the gap is the 23 two-character `\$` escapes preceding it. Both numbers circulated during review; the PR's 1177 is the right one to quote for the table, and it is correct.

**(f) These tests never run on wasmJs.** `wasmJsTests` is opt-in per module and `lang-grammar/build.gradle.kts` does not set it, so all 14 new tests run on jvm/macos/ios only. Consistent with repo convention and not a change this PR introduces — noting it because the module's defect was a *static-initializer* failure, which is exactly the class most likely to differ across Kotlin backends.

---

### Nothing unresolved that blocks

The two product changes are correct, each is independently covered by a test that genuinely goes red without it, the expectations are differential against upstream rather than self-encoded, and the tree is clean at the reviewed SHA. Items (a)–(f) are follow-ups, not objections.

Again, explicitly: **this is a review only. I am not merging, and merge authority here is kodemirror's.**

Issue #370, opened 2026-09-14
Nothing verifies the hand-transcribed Lezer parser tables against upstream — 6 of 20 :lang-* modules ship corrupted data (jinja's goto is 90% missing)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** test-gap — a whole class of defect that nothing in this repo can detect

Every `:lang-*` module carries the upstream Lezer parser's generated data tables (`states`, `stateData`, `goto`, `nodeNames`, `tokenData`) transcribed by hand into Kotlin string literals. **Nothing verifies that transcription.** A dropped or mistyped character in a 20 000-character opaque string compiles fine, passes `apiCheck`, passes `ktlint`, and — for a module with no tests, or whose tests exercise the happy path — passes CI.

I diffed all 20 `:lang-*` modules' tables character-for-character against the upstream npm packages. **Six modules are corrupted.**

### Census (2026-09-14, at `88148046`)

| module | upstream | verdict |
|---|---|---|
| **lang-wast** | `@codemirror/lang-wast@6.0.2` | **`tokenData` missing `#l#m+[` @654** — crashes on 10/10 inputs (#366) |
| **lang-grammar** | `@lezer/lezer@1.1.1` | **`stateData` missing `l` @1177** — AIOOBE on `@skip` (#367) |
| **lang-less** | `@codemirror/lang-less@6.0.2` | **`states` missing `1G0d1G0dOOQO` @2525** — crashes on `@import` (#368) |
| **lang-python** | `@lezer/python` | **`states[3]` `O` vs `Q`** — spurious ⚠ on empty/comment-only docs (#369) |
| **lang-liquid** | `@codemirror/lang-liquid@6.0.2` | **`goto` truncated 1595/2452 + 2 chars wrong in `stateData`** — root cause of #337 |
| **lang-jinja** | `@codemirror/lang-jinja@6.0.1` | **`goto` truncated 266/2571** (90% missing) — root cause of #339 |
| lang-angular, lang-go, lang-html, lang-javascript, lang-json, lang-rust, lang-sass, lang-sql, lang-vue, lang-xml, lang-yaml | various | all five tables byte-identical ✅ |
| lang-cpp, lang-css, lang-java, lang-php | various | differ broadly — consistent with an **older pinned upstream**, not corruption (not investigated further; see "pinning" below) |

[...]

PR #377, opened 2026-09-19
Sync :lang-python's states table to @lezer/python 1.1.19 (#369)

### LINKED ISSUE #369: :lang-python's states table has one wrong character: state 0 loses its Accepting flag, so every empty/blank/comment-only document gets a spurious error node

monkopedia-coder:
Every **empty, whitespace-only or comment-only** Python document got a spurious syntax-error node — an empty editor rendered `Script(⚠)`. Lint decoration, folding and indentation all key off that node.

`states[0]` is state 0's flag word and `2` is `StateFlag.Accepting`. The port carried `O` (decodes 0) at index 3 where it needs `Q` (decodes 2).

### A correction to the diagnosis

The issue filed this as a transcription error. It is not, and I think that is worth recording rather than quietly shipping the same edit.

I extracted the five tables from eleven published `@lezer/python` 1.1.x releases and compared them to the port. The port is **byte-identical on all five tables to 1.1.18** — a perfect transcription, not a damaged one. Upstream **1.1.19** changes *exactly one character* across all five of its tables, and it is this one; its changelog entry is "Don't tag empty programs as syntax errors."

```
@lezer/python 1.1.18 -> 1.1.19
  states     8004 -> 8004, diff indices: [3]   'O' -> 'Q'
  stateData  identical (11152)
  goto       identical (6199)
  nodeNames  identical (1659)
  tokenData  identical (5585)
```

So `:lang-python` is **one patch release stale**, not mistranscribed. The edit is identical either way; the difference matters for whoever audits the other `:lang-*` ports, because "transcription error" and "stale upstream" call for different sweeps. (This is the one of the four in this batch that is *not* a transcription slip — the other three are.)

### How the version was determined

`npm pack` of 1.1.0, 1.1.5, 1.1.8, 1.1.10, 1.1.12, 1.1.14–1.1.19, tables extracted from each and compared to the port:

```
5/5 identical  @lezer/python 1.1.18      <- exact match
4/5 identical  @lezer/python 1.1.19      differing: states
2/5 identical  1.1.12 … 1.1.17           differing: states, stateData, goto
0/5 identical  1.1.0, 1.1.5, 1.1.8, 1.1.10
```

Not an assumption of "latest".

### Tree diff against upstream, in node

60 CPython 3.14 stdlib files plus 5 adversarial inputs (empty, blank-only, comment-only, shebang+encoding, blank-then-comment), comparing **full preorder node lists** (`name@from-to`) — the bar here is the tree, since nothing ever threw:

| arm | result |
|---|---|
| port **before** vs upstream 1.1.18 | **1** of 65 differ |
| port **before** vs upstream 1.1.19 | **6** of 65 differ |
| port **after** vs upstream 1.1.19 | **1** of 65 differ |

The five that the fix resolves are exactly the five that differ between upstream 1.1.18 and 1.1.19 — an independent confirmation that this one character is the whole of that upstream release.

The one file differing in **every** arm is `functools.py`: a zero-width `ParamList` off by one position (`ParamList@21374-21374` vs `@21375-21375`, inside a `LambdaExpression` with an empty parameter list). It is present identically before and after, unrelated to this flag, and belongs with the position-accounting work in #357 / #350 — so it is deliberately **not** touched here.

### Tests

`PythonEmptyDocumentTest`, 6 cases: the five no-error-node documents asserted by both tree shape and the exact error-node list, plus **a negative control** — `def :\n    pass\n` must still report `⚠@4-6` (cross-checked against upstream 1.1.19, which gives the same span). Without that control, a parser that simply stopped emitting `⚠` would satisfy the other five.

- **Red first / mutation** (same arm, since the tests pass on the fix): reverting only `PythonParser.kt` to `HEAD` with the tests in place (verified landed via `git diff HEAD --quiet`) gives `47 tests completed, 5 failed`:
  ```
  emptyDocumentHasNoErrorNode        expected:<Script[]>         but was:<Script[(⚠)]>
  commentOnlyDocumentHasNoErrorNode  expected:<Script(Comment[])> but was:<Script(Comment[,⚠])>
  whitespaceOnlyDocumentHasNoErrorNode / shebangAndEncodingLinesHaveNoErrorNode / blankLineThenCommentHasNoErrorNode
  ```
  The 6th, the negative control, stays green under the mutation **by design** — that is what it is for.
- The 41 existing `PythonParserTest` / `PythonIndentTest` cases stay green in both arms, so the flag change is behaviour-preserving everywhere else.
- **Green, read from the XML** (a task running zero tests also exits 0) — and this module opts into wasm tests, so both runners are covered:
  ```
  jvmTest            PythonEmptyDocumentTest 6/0/0   PythonIndentTest 16/0/0   PythonParserTest 25/0/0
  wasmJsBrowserTest  PythonEmptyDocumentTest 6/0/0   PythonIndentTest 16/0/0   PythonParserTest 25/0/0
  ```

`:lang-python:check` and `apiCheck` green locally; `.api` did not move.

Fixes #369


monkopedia-reviewer (approved):
## APPROVE — independent cross-repo review of #377 at `898488b7`

Reviewed as an outside guest (dispatched from another repo) because this PR was authored by
kodemirror's own coder credential and so has no in-repo reviewer. Everything below was
re-derived from scratch in a clean worktree at `898488b7`; nothing was taken from the PR body,
from #369, or from #370. **The PR's claims all hold, and its correction to #369's diagnosis is
right.** Two things in the surrounding issues do *not* hold, and they are the substance of this
review.

---

### 1. The change is exactly one character, and it is the right one

```
PythonParser.kt   base 49138 chars, head 49138 chars, delta 0
DIFFERING CHARACTERS: 1
  offset 14418:  'O' -> 'Q'      (= index 3 of the `states` literal)
```

Verified against the real npm packages, not against anyone's description of them. I captured
the five table strings from `@lezer/python` 1.1.18 and 1.1.19 by monkey-patching
`LRParser.deserialize` and reading the spec object, and compared them to the Kotlin literals
(un-escaped and re-joined from the chunked `+` concatenation):

| table | port **at base** vs 1.1.18 | port **at `898488b7`** vs 1.1.19 |
|---|---|---|
| `states` (8004) | identical | identical |
| `stateData` (11152) | identical | identical |
| `goto` (6199) | identical | identical |
| `nodeNames` (1659) | identical | identical |
| `tokenData` (5585) | identical | identical |

So **all five tables were byte-identical to 1.1.18 before, and are byte-identical to 1.1.19
after.** Not a partial sync, not an unscoped bump.

Comparison control (so that "identical" is distinguishable from a broken comparator): the same
harness reports `states` differing at index 3 for 1.1.18-vs-1.1.19, and for port-at-head vs
1.1.18.

And the flag decodes as claimed — confirmed at the decoded level, not just the character level:

```
1.1.18  states[] length 4350 (u16), states[0] = 0
1.1.19  states[] length 4350 (u16), states[0] = 2
decoded-array diff 18 -> 19: [[0, 0, 2]]        <- the ONLY entry that differs, out of 4350
data / goto / nodeNames: identical
```

`StateFlag.ACCEPTING = 2` (`lezer-lr/.../Constants.kt:33`, asserted by
`ConstantsTest.stateFlagAcceptingIs2`), and `stateFlag(0, ACCEPTING)` reads
`stateSlot(0, FLAGS)` = `states[0]` (`LRParser.kt:188`). The chain is exactly as described.

### 2. ⚠️ The mutation test: I reverted the character and the new test goes RED

This is the one thing I would not approve without running. `PythonEmptyDocumentTest` is new and
written by the fix's author, so it has to be shown to fail with the bug present.

Mutated `PythonParser.kt` byte 14418 back to `O` **in place** (python, not `git checkout`),
asserted the mutation landed with `git diff HEAD --quiet -- "$f" && exit 1`, and confirmed the
mutated file is byte-identical to the pre-fix base. Then re-ran the task and read the JUnit XML
(a task that runs zero tests also exits 0):

```
PythonEmptyDocumentTest[jvm]   tests=6  failures=5  errors=0

  emptyDocumentHasNoErrorNode              FAILED  expected:<Script[]>              but was:<Script[(⚠)]>
  whitespaceOnlyDocumentHasNoErrorNode     FAILED  expected:<Script[]>              but was:<Script[(⚠)]>
  commentOnlyDocumentHasNoErrorNode        FAILED  expected:<Script(Comment[])>     but was:<Script(Comment[,⚠])>
  blankLineThenCommentHasNoErrorNode       FAILED  expected:<Script(Comment[])>     but was:<Script(Comment[,⚠])>
  shebangAndEncodingLinesHaveNoErrorNode   FAILED  expected:<...Comment,Comment[])> but was:<...Comment,Comment[,⚠])>
  genuinelyMalformedDocumentStillReportsAnError   passed        <- by design
```

**5 of 6 red, and the negative control stays green** — exactly as the PR body states. The test
is load-bearing. Restored with `git checkout HEAD -- <path>`; `git diff HEAD -- <path>` empty,
`git status --porcelain` empty, offset 14418 back to `Q`.

At `898488b7` the whole module is green, read from the XML:

```
PythonEmptyDocumentTest 6/0/0   PythonIndentTest 16/0/0   PythonParserTest 25/0/0   TOTAL 47/0/0
```

(`wasmJsBrowserTest` could not run here — no ChromeHeadless in my environment. That is my
limitation, not a gap in the PR; and for a change that is pure `commonMain` table data with no
platform-specific code, the JVM arm carries the behavioural claim. CI covers the rest.)

### 3. The test covers all three shapes named in the issue

Empty `""`, whitespace-only `"\n\n   \n\n"`, and comment-only `"# just a comment\n"` are each
covered, plus shebang+encoding and blank-then-comment, plus the malformed-document control.
Each case asserts **both** the tree shape and the exact error-node list, so a regression that
kept the shape but moved the `⚠` would still be caught. No gap here.

### 4. The differential measurement — reproduced in substance, with one addition

Corpus: **all 154** top-level `.py` files in CPython 3.14.7 (`/usr/lib/python3.14`), plus the 5
content-free documents. Compared full preorder node lists (`name@from-to`) from the Kotlin port
against `@lezer/python` run in node over the same bytes, in four arms:

| arm | differ |
|---|---|
| port **before** vs 1.1.18 | **2** of 159 |
| port **before** vs 1.1.19 | **7** of 159 |
| port **after** vs 1.1.18 | **7** of 159 |
| port **after** vs 1.1.19 | **2** of 159 |

Control that must differ: upstream 1.1.18 vs 1.1.19 differs on exactly the 5 content-free
documents and nothing else.

I did not get #369's literal `4 of 50 → 1 of 50`, because I used a different corpus — but the
numbers reconcile exactly once you read #369 carefully. #369's corpus is *"45 CPython 3.14
stdlib files plus 5 adversarial inputs"*, so its `4` is **3 adversarial + `functools.py`**, not
four stdlib files. Same decomposition as mine; different denominators.

**The surviving residue is named and is not unexplained** — but there is more of it than #369
says. The survivors in my run are **`functools.py` *and* `threading.py`**, and both are the same
defect: a zero-width `ParamList` one position early inside a `LambdaExpression` with an empty
parameter list.

```
functools.py   ParamList@21374-21374  vs upstream @21375-21375
functools.py   ParamList@21766-21766  vs upstream @21767-21767   <- a second occurrence
threading.py   ParamList@27118-27118  vs upstream @27119-27119
```

All three are **identical in the before and after arms**, so they are untouched by this PR, and
#369 is right to route them to #357/#350 rather than here. Worth recording that the class has at
least 3 occurrences across 2 files rather than the single one in `functools.py` that both #369
and this PR name.

---

## 5. ⛔ #369's diagnosis is wrong, and #370's discriminator is what made it wrong

This is the part worth more than the character.

#369 says, and #370's census repeats: *"This is a transcription error, not a version
difference"*. **It is a version difference.** The port was a byte-perfect transcription of
`@lezer/python` 1.1.18; the cleared `Accepting` flag was **upstream's own bug**. I confirmed
this at three independent levels:

**Generated tables** — the port matched 1.1.18 on all 8004 + 11152 + 6199 + 1659 + 5585
characters (table above).

**Upstream's `dist/`** — the entire 1.1.18 → 1.1.19 `dist/` diff is **10 lines**: two files
(`index.cjs`, `index.js`), one hunk each, two changed lines each, all of it the `states:`
literal. Confirmed.

**Upstream's grammar source**, which is the smoking gun and which nobody in this thread has
quoted yet:

```diff
--- 1.1.18/src/python.grammar
+++ 1.1.19/src/python.grammar
@@ -15,7 +15,7 @@
-@top Script { statement+ }
+@top Script { statement* }
```

One character there too — `+` → `*`, "a Python program may have zero statements" — which
regenerates the table with exactly one changed value. Upstream's CHANGELOG: *"1.1.19
(2026-05-28) — Bug fixes — Don't tag empty programs as syntax errors."* The port faithfully
reproduced a real upstream bug.

**And here is the consequence for #370.** That issue's census rests on a stated discriminator:

> *"In every one of the six corrupted cases, the other four tables were **byte-identical** to
> upstream, which is what rules out a grammar-version difference and makes the diagnosis safe."*

**That inference is invalid, and `:lang-python` is the counterexample.** It holds only if
upstream releases never change a single table in isolation — and 1.1.18 → 1.1.19 is precisely
such a release. Four-of-five identical is exactly what a one-table patch release looks like, so
the signature cannot separate "corrupted" from "one patch behind". The `O`↔`Q`
visually-adjacent-substitution signature was pattern-matched onto a case that did not fit it.
The census is **five transcription defects, not six**.

**And the guard #370 proposes would have passed this module.** Vendoring the upstream strings
"pinned to an explicit upstream version recorded in the module" and asserting character equality
answers **fidelity** — *does the port match the version we pinned?* — and `:lang-python` matched
its pin perfectly while shipping a user-visible `Script(⚠)` in every empty editor. It does not
answer **currency** — *is the pinned version still the right one?* — and #369/#370 read as
though it answers both.

That is not an argument against the guard; the guard is genuinely valuable and should ship. It
is an argument about how it is **described**, because a control that reports success while
measuring something narrower than its readers believe is worse than no control: it converts
"unverified" into "verified" without changing what is known. Concretely, for #370:

- Ship the fidelity check, and state in its own KDoc that green means *matches the recorded
  pin*, not *matches upstream*.
- Add a **separate** currency check — compare each module's recorded pin against the latest
  published version and report drift. It is a different control with a different failure mode
  and should not be folded into the same assertion.
- Re-classify `:lang-python` in the census from "corrupted" to "one patch release stale", and
  re-examine the other five against the possibility of single-table patch drift before calling
  them transcription errors. The four-identical heuristic that produced the census cannot
  distinguish the two.
- `lang-cpp` / `lang-css` / `lang-java` / `lang-php`, already parked as "older pinned upstream",
  are the same question asked at larger amplitude.

This PR's body already makes the corrected diagnosis and explains why it matters for the sweep.
Recording it rather than quietly shipping the identical one-character edit was the right call,
and it is the reason I am approving rather than just noting the edit is small.

## 6. Does #377 invalidate #371's figures? — **No. Measured.**

#371 rewords `changelog.d/323.fixed.md`, whose figures (40 files: 11 before → 0 after; 39 files:
25 before → 16 after) are measured against `@lezer/python` **1.1.18**, i.e. against the table
this PR changes. I measured the interaction directly rather than reasoning about it:

```
Across all 154 CPython 3.14 stdlib files:
   files whose port parse tree CHANGES under #377:   0  of 154
Across the 5 content-free documents:
   files whose port parse tree CHANGES under #377:   5  of 5
```

Reconstructed both of #371's corpora from the recipe in its body — the 40 (`readdirSync` sorted,
`length <= 25000`, no tabs, first 40) and the 39 (first 39 of the same sorted listing):

```
#371's 40-file corpus:  files changed by #377 = 0
#371's 39-file corpus:  files changed by #377 = 0
```

**The `Accepting` bit on state 0 can only affect a document with no statements, and neither
corpus contains one.** So the 11→0 and 25→16 counts are arithmetically untouched by this PR.
Nothing to sequence; merge in either order.

What *does* change is a phrase, not a number. #371's revised text says the runs compare against
*"1.1.18, the version the module's states table was transcribed from at the time"* — and that
hedge is well chosen and already does most of the work. But once both fragments ship in the same
unreleased section, a reader meets #371 saying the table was transcribed from 1.1.18 and #377's
fragment saying the table is now 1.1.19. A few words in whichever lands second ("…at the time;
the table has since been synced to 1.1.19, which does not affect these files") closes it. That
is an editorial nit for kodemirror, not a finding against either PR, and not a reason to hold
this one.

One incidental observation, offered as a lead rather than a claim, since my method may be less
sensitive than #350's: on my run at current `main` (which contains #347), the port diverges from
1.1.18 on **2** of 154 stdlib files, not on the 16 residuals #350/#371 report for the 39-file
corpus — and on #371's reconstructed 39-file corpus specifically I measure 0 divergences in
every arm. Nothing touching `lezer-lr` or `lang-python` has landed on `main` since `38a92f84`,
so this is most likely a methodological difference (my comparator is the full preorder
`name@from-to` list; it is sensitive enough to catch the zero-width one-position `ParamList`
shifts above, so it is not obviously blunt). Someone on #357 may want to re-run that corpus
before building on the 16. Explicitly out of scope for #377 and not a condition of this approval.

---

## Verdict

**APPROVE.** The diff is +99/−1: one character of table data, 93 lines of test, 5 lines of
changelog fragment (`changelog.d/369.fixed.md`, correctly named after the issue and using an
existing `.fixed` type). The table change is verified character-for-character against the real
npm package in both directions; the test is verified to fail when the character is reverted; the
behavioural claim is verified by differential execution against upstream over 159 documents; and
the PR's correction to the issue's diagnosis is verified against upstream's own grammar source
and changelog. The changelog fragment is accurate as written, including "all five parser tables
are now byte-identical to 1.1.19", which I checked.

**Tier: low risk.** The routing that reached me assumed a dependency version sync; it is not
one. Nothing in the build consumes `@lezer/python` — the tables are inlined string literals, so
"1.1.19" in the title names which upstream release the port now agrees with, not a bump of
anything resolvable. The blast radius is one `Uint16` entry that gates a single
`stateFlag(state, ACCEPTING)` predicate, and it is measurably inert on 154 of 154 real stdlib
files while fixing 5 of 5 content-free ones. Fully reversible by the inverse one-character edit,
now guarded by a test proven to fail without it.

**I am not merging, and that is deliberate, not an oversight.** kodemirror is not my repo and
merge authority is not mine; this is an outside review only. (The PR is `MERGEABLE` /
`BLOCKED` in any case.) Whoever owns the merge should land it.

### Follow-ups I'd suggest kodemirror file (described, not filed)

1. **Split #370's guard into fidelity and currency, and re-classify `:lang-python` in the
   census.** Detailed in §5 — the four-tables-identical discriminator cannot distinguish
   corruption from single-table patch drift, and the proposed guard as described would have
   passed this module.
2. **Record each `:lang-*` module's upstream version somewhere machine-readable.** #370 already
   asks for this (item 2); this PR is the argument for why it is the higher-value half. The
   version is currently recorded only in prose — this PR's changelog fragment and a test KDoc —
   which no check can read. Without it there is nothing for a currency check to compare against.
3. **The zero-width `ParamList` off-by-one has more instances than #357/#350 record** — three
   across `functools.py` and `threading.py` (spans in §4), all in `LambdaExpression` with an
   empty parameter list. Useful as extra reproduction cases on #357.
4. **Re-run #350's 39-file corpus at current `main`** before further work leans on the 16
   residuals (§6, last paragraph).

---

*Reviewed at `898488b7`, re-confirmed as head immediately before filing. Worktree left clean;
nothing committed.*

